PR #3, opened 2024-02-18
add postgres support

efiShtain:
I removed most of the onConflict clauses, since we do want to fail if we try to insert an entity with the same ID

I left 2 onConflict that I think make sense, but made them work with postgres

postgres requires a uniqueIndex or primary key for onConflict clauses
gorm makes it really hard to use onConflicts with associations so I had to switch stuff to explicit transaction for mappings

also set membership_id as conflict column for env_vars, not sure if this is correct
@itayd please ack on this change


itayd (approved):
lgtm, but if i get my triggers change before you might need to change a bit (it's much simpler). we'll see who gets what first.

PR #4, opened 2024-02-19
standard dockerfile

efiShtain:
removed old docker resources (build directory)
added standard docker files (created using new docker init command)


compose.yaml
```diff
@@ -0,0 +1,49 @@
...
+# This service is built from the Dockerfile in the current directory.
+# You can add other services your application may depend on here, such as a
+# database or a cache. For examples, see the Awesome Compose repository:
+# https://github.com/docker/awesome-compose
+services:
+  server:
+    build:
+      context: .
+      target: final
+    command: ["up", "--mode=dev"]
```
daabr, line 15: Why dev mode? Even if experimental, the cloud instance should be the "full, real AK". More importantly, dev mode lacks data persistence and durability, which are critical for a non-local AK instance.

This includes: a Temporal server, DB(s) for Temporal and AK, and a Redis server for AK - all these components need to be independent and separate from day 1.

compose.yaml
```diff
@@ -0,0 +1,49 @@
...
+    ports:
+      - 9980:9980
+# The commented out section below is an example of how to define a PostgreSQL
```
daabr, line 18: "The commented out section below" - if it's not commented-out, why include it in this PR?

Also, we haven't decided yet on which DB to use, where, and how - this is scheduled for Wednesday, but if needed & possible we can discuss tomorrow too.

Lastly, all cloud configuration orchestration (i.e. meant to be used by us, not users) should be maintained in a non-public repo due to their sensitive nature. Or am I misunderstanding the purpose of this PR?

Dockerfile
```diff
@@ -0,0 +1,71 @@
...
+
+# Build the application.
+# Leverage a cache mount to /go/pkg/mod/ to speed up subsequent builds.
+# Leverage a bind mount to the current directory to avoid having to copy the
+# source code into the container.
+RUN --mount=type=cache,target=/go/pkg/mod/ \
+    --mount=type=bind,target=. \
+    CGO_ENABLED=0 go build -o /bin/ak ./cmd/ak
+
+################################################################################
```
itayd, line 26: remove example comments

compose.yaml
efiShtain: this is for developers who want to just start it and work with it, 
this is not a production compose

I will add more features to the compose file later to support a database and persistence temporal

compose.yaml
efiShtain: all cloud related stuff are in infrastructure repo, 
except for the dockerfile and this compose file which are standard in the main repo

itayd:
@daabr to approve

daabr (approved):

itayd (approved):

PR #5, opened 2024-02-19
New system test with end-to-end session execution

daabr:
The new test - `systest/testdata/workflows/builtin_funcs.txtar` - has a dual purpose:

1. (The first but not only) automated test of end-to-end execution via AK
2. Test the functionality of AK built-in functions, which are non-standard but loaded automatically in Starlark

Two commands were added to the systest:

1. `http get|post`
2. `wait` (for a specific session to have the state `COMPLETED` or `ERROR`)

Note that this test possibly uncovers flakiness in AK's HTTP connection, which needs to be investigated further. So I'm adding the test to this PR, but for now skipping it in `make test-unit` - so we can manually run it anywhere, but without suffering from CI flakiness.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ test: test-race test-cli
 
 .PHONY: test-unit
 test-unit:
-	$(GOTEST) ./...
+	$(GOTEST) ./... -skip /workflows/builtin_funcs
 
 # Subset of "test-unit", for simplicity.
 .PHONY: test-system
--- a/Makefile
+++ b/Makefile
@@ -101,6 +101,7 @@ endif
 .PHONY: test
 test: test-race test-cli
 
+# TODO(ENG-447): Fix HTTP trigger flakiness.
 .PHONY: test-unit
 test-unit:
 	$(GOTEST) ./... -skip /workflows/builtin_funcs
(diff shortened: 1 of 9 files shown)


Makefile
```diff
@@ -103,7 +103,7 @@ test: test-race test-cli
 
 .PHONY: test-unit
 test-unit:
-	$(GOTEST) ./...
+	$(GOTEST) ./... -skip /workflows/builtin_funcs
```
itayd, line 107: comment why this is disabled

itayd: with ticket

systest/ak_test.go
```diff
@@ -99,27 +101,65 @@ func runTestSteps(t *testing.T, steps []string, akPath, akAddr string) {
...
 			}
-			ak = result.(*akResult)
+
+			switch v := result.(type) {
+			case *akResult:
+				ak = v
+			case *httpRequest:
+				pendingReq = v
+			case string:
+				t.Log(v)
```
itayd, line 144: add default to not silently ignore unexpected?

itayd (approved):

Makefile
daabr: Done - see ENG-447

systest/ak_test.go
daabr: done

PR #7, opened 2024-02-20
Start and stop Temporal within the system test

daabr:
Fix for ENG-437: start and stop Temporal dev server within `systest/ak_test` instead of AK's `temporalclient`.

Verification:

```
$ ps -ef | grep temporal | grep headless | wc -l
$ gotestsum -f testname -- ./systest
$ ps -ef | grep temporal | grep headless | wc -l
```

Indirectly, had to change AK's ready log message from `ready` to `autokitteh ready` to avoid conflict with Temporal's own `DevServer ready` log message when the system test waits for the AK server to be ready.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -103,7 +103,7 @@ test: test-race test-cli
 
 .PHONY: test-unit
 test-unit:
-	$(GOTEST) ./...
+	$(GOTEST) ./... -skip /workflows/builtin_funcs
 
 # Subset of "test-unit", for simplicity.
 .PHONY: test-system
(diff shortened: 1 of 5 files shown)


daabr:
CLI test CI workflow is failing because it expects a Temporal server to be created for it. Need to decide how to proceed:

1. Install and run Temporal as a part of `tests/cli/run.sh`, until we finish migrating all tests
2. Disable/remove the old CLI test, and migrate old tests to the new systest ASAP
3. Something else?

PR #9, opened 2024-02-21
kill runtimes rpcs; modify build to receive fs.FS

itayd:
- runtimes build & run rpcs are not expected to be needed for a long while. for now kill the rpc (partial) implementation for build. 

- change the build signature to receive an `fs.FS` instead of `rootURL`. will enable to receive data from non-local-fs sources.


cmd/ak/common/runtimes.go
```diff
@@ -0,0 +1,5 @@
+package common
+
+import "go.autokitteh.dev/autokitteh/backend/runtimes"
+
+var LocalRuntimes = runtimes.New()
```
daabr: It's wasteful to create this even when it's not needed, especially when it's used only once so far. Please instantiate `runtimes.New()` directly in `cmd/builds/create.go` instead of this file.

sdk/sdkclients/sdkclients.go
```diff
@@ -54,6 +53,7 @@ func New(params sdkclient.Params) sdkservices.Services {
 		integrations: kittehs.Lazy1(sdkintegrationsclient.New, params),
 		mappings:     kittehs.Lazy1(sdkmappingclient.New, params),
 		oauth:        kittehs.Lazy1(sdkoauthclient.New, params),
+		params:       params,
```
daabr: Please add a short comment why only `params` doesn't use `kittehs.Lazy1`

daabr (changes requested):
A few minor issues to fix.

cmd/ak/common/runtimes.go
itayd: ha, done

sdk/sdkclients/sdkclients.go
itayd: done

daabr (approved):

PR #10, opened 2024-02-21
build from resources in fs or project

itayd:
build resources are taken from project storage.

this supersedes https://github.com/autokitteh/autokitteh/pull/2.


backend/internal/db/db.go
```diff
@@ -47,6 +47,7 @@ type DB interface {
 
 	ListProjects(context.Context) ([]sdktypes.Project, error)
 
+	// Returns nill, nil if no resources are set.
```
daabr, line 50:
```suggestion
	// Returns (nil, nil) if no resources are set.
```

daabr (approved):
Approved, but need to fix the failing CLI test

backend/internal/projects/config.go
```diff
@@ -0,0 +1,22 @@
...
+	AllowLocalFS bool `koanf:"allow_local_fs"`
+}
+
+type Config struct {
+	Resources ResourcesConfig `koanf:"resources"`
+}
+
+var Configs = configset.Set[Config]{
+	Default: &Config{},
+	Dev: &Config{
```
efiShtain: Maybe we should be very explicit and say this is Local mode

dev mode can still be on remote server ?
not sure, WDYT?

backend/internal/projects/resources.go
```diff
@@ -0,0 +1,73 @@
...
+
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+// If srcURL is nil, get resources from the database.
+// If srcURL scheme is empty, get resources from the database. If none found, get from filesystem, if allowed.
+// If srcURL scheme is "file", get resources from filesystem, if allowed.
+// If srcURL scheme is "project", get resources from database.
+func (p Projects) openResourcesFS(ctx context.Context, projectID sdktypes.ProjectID, srcURL *url.URL) (fs.FS, error) {
```
efiShtain: *Projects ?

backend/internal/projects/resources.go
```diff
@@ -0,0 +1,73 @@
...
+	if srcURL == nil || *srcURL == (url.URL{}) {
+		return p.openResourcesFS(ctx, projectID, &url.URL{Scheme: "project"})
```
efiShtain: do you really need the recursive call here instead of just setting the scheme to project ?

```go
if srcURL == nil || *srcURL == (url.URL{}) {
  srcURL = &url.URL{Scheme: "project"}
}
```

backend/internal/projects/resources.go
```diff
@@ -0,0 +1,73 @@
...
+		return os.DirFS(srcURL.Path), nil
+	case "project":
+		return p.openProjectResourcesFS(ctx, projectID, srcURL.Path)
+	default:
+		return nil, fmt.Errorf("%w: unhandled resources url scheme %q", sdkerrors.ErrNotImplemented, srcURL.Scheme)
+	}
+}
+
+// returns nil, nil if no resources found.
+func (p Projects) openProjectResourcesFS(ctx context.Context, projectID sdktypes.ProjectID, path string) (fs.FS, error) {
```
efiShtain: path is always "/" in this case
so its redundant

backend/internal/projects/resources.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	if files == nil {
+		return nil, nil
+	}
+
+	memfs := memfs.New()
+
+	for path, content := range files {
+		if err := memfs.MkdirAll(filepath.Dir(path), 0o755); err != nil {
```
efiShtain:
```suggestion
		if err := memfs.MkdirAll(filepath.Dir(path), 0o700); err != nil {
```

backend/internal/projects/resources.go
```diff
@@ -0,0 +1,73 @@
...
+			return nil, fmt.Errorf("failed to create directory %q in memory filesystem: %w", filepath.Dir(path), err)
+		}
+
+		if err := memfs.WriteFile(path, content, 0o644); err != nil {
```
efiShtain:
```suggestion
		if err := memfs.WriteFile(path, content, 0o600); err != nil {
```

sdk/sdktypes/project.go
```diff
@@ -19,7 +19,7 @@ var (
 )
 
 func strictValidateProject(pb *projectv1.Project) error {
-	if err := ensureNotEmpty(pb.Name, pb.ProjectId); err != nil {
+	if err := ensureNotEmpty(pb.Name, pb.ProjectId, pb.ResourcesRootUrl); err != nil {
```
efiShtain: why do we ensure this is not empty if we support not having ResourcesRootUrl ?

backend/internal/projects/resources.go
```diff
@@ -0,0 +1,73 @@
...
+// If srcURL scheme is empty, get resources from the database. If none found, get from filesystem, if allowed.
+// If srcURL scheme is "file", get resources from filesystem, if allowed.
+// If srcURL scheme is "project", get resources from database.
+func (p Projects) openResourcesFS(ctx context.Context, projectID sdktypes.ProjectID, srcURL *url.URL) (fs.FS, error) {
+	if srcURL == nil || *srcURL == (url.URL{}) {
+		return p.openResourcesFS(ctx, projectID, &url.URL{Scheme: "project"})
+	}
+
+	switch srcURL.Scheme {
+	case "":
```
efiShtain:
```suggestion
	case "", "project":
```

backend/internal/projects/resources.go
```diff
@@ -0,0 +1,73 @@
...
+			return fs, nil
+		}
+		fallthrough
+	case "file":
+		if !p.Config.Resources.AllowLocalFS {
+			return nil, errors.New("resources from local filesystem are not allowed")
+		}
+		return os.DirFS(srcURL.Path), nil
+	case "project":
+		return p.openProjectResourcesFS(ctx, projectID, srcURL.Path)
```
efiShtain:
```suggestion
```
assuming you move project to the top case

backend/internal/projects/resources.go
```diff
@@ -0,0 +1,73 @@
...
+		return p.openResourcesFS(ctx, projectID, &url.URL{Scheme: "project"})
+	}
+
+	switch srcURL.Scheme {
+	case "":
+		fs, err := p.openProjectResourcesFS(ctx, projectID, srcURL.Path)
+		if err != nil {
+			return nil, err
+		}
+		if fs != nil || !p.Config.Resources.AllowLocalFS {
```
efiShtain: maybe split this condition to two? will make it more readable

```go
if fs != nil {
  return fs, nil
}

if !p.Config.Resources.AllowLocalFS {
  return nil, nil
}
```

backend/internal/db/db.go
itayd: fixed

backend/internal/projects/config.go
itayd: need to think about it

backend/internal/projects/resources.go
itayd: no longer relevant

backend/internal/projects/resources.go
itayd: no longer relevant

backend/internal/projects/resources.go
itayd: no longer relevant

backend/internal/projects/resources.go
itayd: it's in mem, nobody cares :)

backend/internal/projects/resources.go
itayd: it's in mem, nobody cares :)

sdk/sdktypes/project.go
itayd: no longer relevant

backend/internal/projects/resources.go
itayd: no longer relevant

backend/internal/projects/resources.go
itayd: no longer relevant

backend/internal/projects/resources.go
itayd: no longer relevant

PR #11, opened 2024-02-21
feat(db): delete sessions

ig-ra:
delete sessions from DB


backend/internal/db/dbgorm/dbgorm.go
```diff
@@ -120,3 +122,17 @@ func get[T, R any](db *gorm.DB, ctx context.Context, f func(t T) (*R, error), wh
 
 	return f(r)
 }
+
+func delete[T any](db *gorm.DB, ctx context.Context, t T, where string, args ...any) error {
```
daabr, line 142: I'm not experienced enough with this code to know how this will react if you're trying to delete something which other tables reference. I assume that's why a generic `delete` didn't exist until now. Please check with @efiShtain.

backend/internal/sessions/sessions.go
```diff
@@ -100,8 +107,9 @@ func (s *sessions) Start(ctx context.Context, session sdktypes.Session) (sdktype
 				sdktypes.NewErrorSessionState(fmt.Errorf("execute workflow: %w", err), nil),
 			),
 		); uerr != nil {
-			s.z.With(zap.String("session_id", sessionID.String())).Error("update session", zap.Error(err))
+			z.Error("update session", zap.Error(err))
 		}
+		z = s.z.With(zap.String("session_id", sessionID.String()))
```
daabr: Delete this line (unnecessary and unused repetition of line 100)

backend/internal/sessionsgrpcsvc/svc.go
```diff
@@ -149,3 +149,37 @@ func (s *server) List(ctx context.Context, req *connect.Request[sessionsv1.ListR
...
+
+		return nil, connect.NewError(connect.CodeUnknown, err)
+	}
+	if session == nil { // just to ensure, Get should return an error if not found
+		return nil, connect.NewError(connect.CodeNotFound, err)
+	}
+
+	// FIXME: maybe CodeCancelled?
+	if state := sdktypes.GetSessionLatestState(session); state == sdktypes.RunningSessionStateType {
+		connect.NewError(connect.CodeFailedPrecondition, fmt.Errorf("cannot delete running session. session_id: %w", err))
```
daabr: I'd recommend splitting this like in two: create an `error` with `fmt.Errrof`, and then wrap it with `connect.NewError`.

But more importantly, you're creating an error, but not actually returning it.

backend/internal/sessionsgrpcsvc/svc.go
```diff
@@ -149,3 +149,37 @@ func (s *server) List(ctx context.Context, req *connect.Request[sessionsv1.ListR
...
+	}
+	s.sessions.Delete(ctx, sessionID)
```
daabr: Missing error/output check.

cmd/ak/cmd/sessions/delete.go
```diff
@@ -0,0 +1,58 @@
...
+)
+
+var deleteCmd = common.StandardCommand(&cobra.Command{
+	Use:     "delete [session ID] [--fail]",
+	Short:   "Delete non-running session",
+	Aliases: []string{"d"},
+	Args:    cobra.MaximumNArgs(1),
+
+	RunE: func(cmd *cobra.Command, args []string) error {
+		if len(args) == 0 {
```
daabr: A point for discussion: with other session commands, considering the latest session as a reasonable and desirable default makes sense. With `delete`, given that it's destructive, should we allow an implicit default at all, or should we force the user to specify the ID for the sake of safety? (i.e. `Args: cobra.ExactArgs(1)` above)

cmd/ak/cmd/sessions/delete.go
```diff
@@ -0,0 +1,58 @@
...
+			args = append(args, id)
+		}
+
+		r := resolver.Resolver{Client: common.Client()}
+		s, id, err := r.SessionID(args[0])
+		if err != nil {
+			return err
+		}
+
+		if s != nil {
```
daabr: Idiomatic Go style is to minimize indentation, so instead of this if/else block, do:

```
if s == nil {
    return ...err...
}

ctx, cancel := ...

File: cmd/ak/cmd/sessions/delete.go [hunk 7]
```diff
@@ -0,0 +1,58 @@
...
+			ctx, cancel := common.LimitedContext()
+			defer cancel()
+
+			common.RenderKVIfV("session", s) // FIXME: should we print deleted session?
```
daabr: If the user doesn't have to specify the sessions ID, for sure yes, but at the end, not here. If we decide to require specifying the ID (see my other comment above) it's preferable to adhere to the "silence == success" rule.

cmd/ak/cmd/sessions/delete.go
```diff
@@ -0,0 +1,58 @@
...
+			err = sessions().Delete(ctx, id)
+			if err != nil {
+				err = fmt.Errorf("delete session: %w", err)
```
daabr: Suggestion: `err = fmt.Errorf("delete session %q: %w", <session ID>, err)`

cmd/ak/cmd/sessions/delete.go
```diff
@@ -0,0 +1,58 @@
...
+			}
+		} else {
+			err = common.NewExitCodeError(common.NotFoundExitCode, fmt.Errorf("session ID %q not found", id))
+		}
+
+		if kittehs.Must1(cmd.Flags().GetBool("fail")) && err != nil {
+			return err
+		}
```
daabr: Please use `FailIfNotFound` like in the `get` command, to preserve the error message pattern that appears in all instances of the `--fail` flag across all get and list commands.

backend/internal/db/dbgorm/dbgorm.go
ig-ra: gorm by default will use soft delete, e.g. mark objects as deleted and won't fetch them as references.
@efiShtain , please correct me if I'm wrong and if you thin we should do anything else

backend/internal/sessionsgrpcsvc/svc.go
ig-ra: @daabr pay attention - I extracted re-wrapping errors as a func.

cmd/ak/cmd/sessions/delete.go
ig-ra: inviting @haimzlato to the discussion as well.

1. I think the goal of having good CLI is to facilitate user interaction with the system. Specifying sessionID (which requires copy/paste) is a bit annoying IMO. So, I suggest to allow user to delete last session without specifying explicitly the sessionID (similar to how stop CMD is implemented)
2. actually, delete command isn't that destructive. We are marking sessions as deleted in DB and allowing to do so only for non-running sessions. Moreover, session being deleted is printed to stdout, so user have a copy in case he deleted not what he  wanted
3. Although we may consider to allow user to delete implicitly only sessions resulting in ERROR and require explicit sessionID for those ended with SUCCESS

cmd/ak/cmd/sessions/delete.go
ig-ra: I feel that in case of delete CMD the silence rule isn't good enough, but the opposite. This gives the user immediate opportunity to see what was deleted and act. So if CLI is invoked by the user itself, I think we should provide clear output what was deleted. If CLI is scripted the output doesn't matter and there is --fail option that should cover the cases needed for scripting

cmd/ak/cmd/sessions/delete.go
ig-ra: My intention was to save the error in order to print at the end if --fail option is specified. That's why there is if/else, because there is no `if` that will lead to immediate return

cmd/ak/cmd/sessions/delete.go
ig-ra: hmm.. but there could be more errors. not only not found one. Please suggest the snippet?

backend/internal/db/dbgorm/dbgorm.go
```diff
@@ -99,7 +99,7 @@ func (db *gormdb) Teardown(ctx context.Context) error {
 	return nil
 }
 
-// Todo: not sure this will work with the connect method
+// TODO: not sure this will work with the connect method
```
efiShtain, line 102: ?

backend/internal/db/dbgorm/dbgorm.go
efiShtain: @ig-ra gorm will soft delete only if our model has gorm.DeletedAt type field set, which none of our models has
so if you need soft delete, please add the field (and verify the result)

second thing is about deleting stuff that other reference,
as long as we keep the correct foreign keys and cascading definition (which I'm not sure we have)
I guess this should work

this is a very dangerous function, and should probably 

we should probably test it very well in all cases if we want to keep it here

anyhow,
currently only session delete uses it, so not sure we need the generic part

and you need to check that if you delete a session, it should delete all its data (session call, attempts, what so ever or mark it as soft delete as well)

backend/internal/sessions/sessions.go
```diff
@@ -76,6 +76,11 @@ func (s *sessions) List(ctx context.Context, filter sdkservices.ListSessionsFilt
 	return s.svcs.DB.ListSessions(ctx, filter)
 }
 
+func (s *sessions) Delete(ctx context.Context, sessionID sdktypes.SessionID) error {
+	s.z.With(zap.String("session_id", sessionID.String())).Debug("delete")
```
efiShtain: I think it is better to log what happened and not what about to happen
otherwise you can see delete in log, but service can fail and you won't know if it happened or not
they deleting can also fail, so the log might indicate something other than intended

but that's my taste for logging (unless you verbose log everything searching for a bug)

backend/internal/sessionsgrpcsvc/svc.go
```diff
@@ -32,27 +32,36 @@ func Init(mux *http.ServeMux, sessions sdkservices.Sessions) {
 	mux.Handle(path, handler)
 }
 
+// re-wrap sdk as connect error
+func wrapError(err error, code connect.Code) error {
```
efiShtain: don't we have a place that already does something like that for all errors ?
and if not, this looks like a good candidate to have in a shared place

backend/internal/sessionsgrpcsvc/svc.go
```diff
@@ -32,27 +32,36 @@ func Init(mux *http.ServeMux, sessions sdkservices.Sessions) {
...
 	uid, err := s.sessions.Start(ctx, session)
 	if err != nil {
-		if errors.Is(err, sdkerrors.ErrAlreadyExists) {
-			return nil, connect.NewError(connect.CodeAlreadyExists, err)
-		} else if errors.Is(err, sdkerrors.ErrUnauthorized) {
-			return nil, connect.NewError(connect.CodePermissionDenied, err)
-		}
-
-		return nil, connect.NewError(connect.CodeUnknown, err)
+		return nil, wrapError(err, 0)
```
efiShtain: magic number ?
probably connect.CodeUnknown is better

backend/internal/sessionsgrpcsvc/svc.go
```diff
@@ -125,27 +124,55 @@ func (s *server) List(ctx context.Context, req *connect.Request[sessionsv1.ListR
...
+	if err != nil {
+		return nil, wrapError(err, 0)
+	}
+
+	if session == nil { // just to ensure, Get should return an error if not found
+		return nil, wrapError(err, connect.CodeNotFound)
+	}
+
+	// FIXME: maybe connect.CodeCancelled?
+	if state := sdktypes.GetSessionLatestState(session); state == sdktypes.RunningSessionStateType {
```
efiShtain: created is also a problematic state
I think you should check if state is not completed or failed

cmd/ak/cmd/sessions/delete.go
efiShtain: I think allowing any action without a specific id is problematic

in a real production system, session could start at any time and on many projects
delete \ get logs of a non specific value would in high probability result unintended stuff

it's not so hard to copy paste an id once

cmd/ak/cmd/sessions/delete.go
efiShtain: @daabr lots of CLIs has a --quiet option to suppress any output
maybe that's a win win options if we set the output to quiet as default or something like that

regarding delete, I tend to agree with @ig-ra 
I prefer to know what happened instead of just empty line

backend/internal/db/dbgorm/dbgorm.go
ig-ra: most of the tools and common practice is grep/search the code for common keywords/labels stated as CAP letters

cmd/ak/cmd/sessions/delete.go
ig-ra: agree with @efiShtain 
rewritten. sessionID is now required

cmd/ak/cmd/sessions/delete.go
ig-ra: done

cmd/ak/cmd/sessions/delete.go
ig-ra: so, are we agreed to print deleted session? if so, @daabr please resolve
note that now code requires sessionID, not optional anymore

backend/internal/db/dbgorm/dbgorm.go
ig-ra: I will check how cascade delete forks for both sqlite and postgress and update

backend/internal/sessionsgrpcsvc/svc.go
ig-ra: sure. I think we should open a ticket for refactoring @itayd ? @daabr ?

backend/internal/sessionsgrpcsvc/svc.go
ig-ra: rewritten

backend/internal/sessionsgrpcsvc/svc.go
ig-ra: rewritten

backend/internal/db/dbgorm/dbgorm.go
efiShtain: I though you added the comment

ig-ra: resolving then

backend/internal/sessionsgrpcsvc/svc.go
```diff
@@ -8,6 +8,7 @@ import (
 
 	"connectrpc.com/connect"
 
+	protovalidate "github.com/bufbuild/protovalidate-go"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
```
daabr:
```suggestion
	protovalidate "github.com/bufbuild/protovalidate-go"

	"go.autokitteh.dev/autokitteh/internal/kittehs"
```

cmd/ak/cmd/sessions/delete.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var deleteCmd = common.StandardCommand(&cobra.Command{
+	Use:     "delete session ID [--fail]",
```
daabr:
```suggestion
	Use:     "delete <session ID> [--fail]",
```

cmd/ak/cmd/sessions/delete.go
```diff
@@ -0,0 +1,50 @@
...
+	Short:   "Delete non-running session",
+	Aliases: []string{"d"},
+	Args:    cobra.MaximumNArgs(1),
```
efiShtain: use ExactArgs

backend/internal/db/dbgorm/dbgorm.go
```diff
@@ -120,3 +122,33 @@ func get[T, R any](db *gorm.DB, ctx context.Context, f func(t T) (*R, error), wh
 
 	return f(r)
 }
+
+// TODO: change all get functions to use this
+func get1[T any](db *gorm.DB, ctx context.Context, t T, where string, args ...any) (*T, error) {
```
efiShtain: NIT, probably getOne is better, not sure

backend/internal/db/dbgorm/dbgorm.go
```diff
@@ -120,3 +122,33 @@ func get[T, R any](db *gorm.DB, ctx context.Context, f func(t T) (*R, error), wh
 
 	return f(r)
 }
+
+// TODO: change all get functions to use this
```
efiShtain, line 126: you can't change all get to use this, since this does not support associations

backend/internal/db/dbgorm/dbgorm.go
```diff
@@ -120,3 +122,33 @@ func get[T, R any](db *gorm.DB, ctx context.Context, f func(t T) (*R, error), wh
...
+func get1[T any](db *gorm.DB, ctx context.Context, t T, where string, args ...any) (*T, error) {
+	var r T
+
+	// TODO: fetch all records and report if there is more than one record
+	result := db.WithContext(ctx).Where(where, args...).Limit(1).Find(&r)
+	if result.Error != nil {
+		return nil, translateError(result.Error)
+	}
+
+	if result.RowsAffected == 0 {
```
efiShtain, line 136: just making sure, is it RowsAffected on find ?

backend/internal/db/dbgorm/deployments.go
```diff
@@ -9,12 +9,17 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+	"gorm.io/gorm"
 )
 
+func (db *gormdb) createDeployment(ctx context.Context, deployment scheme.Deployment) error {
```
efiShtain, line 15: you should probably get a reference here instead of copy the deployment object

backend/internal/db/dbgorm/deployments.go
```diff
@@ -24,27 +29,37 @@ func (db *gormdb) CreateDeployment(ctx context.Context, deployment sdktypes.Depl
...
+	return get1(db.db, ctx, scheme.Deployment{}, "deployment_id = ?", deploymentID)
+}
+
 func (db *gormdb) GetDeployment(ctx context.Context, id sdktypes.DeploymentID) (sdktypes.Deployment, error) {
-	return get(db.db, ctx, scheme.ParseDeployment, "deployment_id = ?", id.String())
+	if d, err := db.getDeployment(ctx, id.String()); d == nil {
+		return nil, err
+	} else {
+		return scheme.ParseDeployment(*d)
+	}
```
efiShtain:
```suggestion
	d, err := db.getDeployment(ctx, id.String()); 
	if err != nil || d == nil {
		return nil, err
	} 
	return scheme.ParseDeployment(*d)
```

Not a must, I personally, really don't like using else where I can, you can self resolve this one what ever you decide

specifically about the if condition, I think it's better to be verbose and explicit

backend/internal/db/dbgorm/deployments.go
```diff
@@ -61,34 +76,57 @@ func (db *gormdb) ListDeployments(ctx context.Context, filter sdkservices.ListDe
 	}
 
 	q = q.Order("created_at desc")
+	return q
+}
+
+func (db *gormdb) listDeploymentsWithStats(ctx context.Context, filter sdkservices.ListDeploymentsFilter) ([]scheme.DeploymentWithStats, error) {
+	q := db.listDeploymentsCommonQuery(ctx, filter)
 
 	if filter.IncludeSessionStats {
```
efiShtain: you can remove the if condition, since the function name indicates you want to get the stats

backend/internal/db/dbgorm/gorm_test.go
```diff
@@ -0,0 +1,109 @@
...
+	"gorm.io/gorm/logger"
+)
+
+type dbFixture struct {
+	db     *gorm.DB
+	gormdb *gormdb
+	ctx    context.Context
+}
+
+func setupDB(dbName string) *gorm.DB {
```
efiShtain, line 26: I'm not sure we should use database for unit tests, even if it's in memory (but we should definitely consider it)


@itayd  / @daabr  what do you think?

backend/internal/db/dbgorm/deployments_test.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func TestListDeployments(t *testing.T) {
+	f := newDbFixture()
+
+	flt := sdkservices.ListDeploymentsFilter{
+		EnvID:               nil,
+		BuildID:             nil,
+		State:               sdktypes.DeploymentStateUnspecified,
+		Limit:               0,
```
efiShtain, line 53: you don't have to specify 0 if you don't want to limit
I think it makes line 69 less understandable

backend/internal/db/dbgorm/deployments_test.go
```diff
@@ -0,0 +1,126 @@
...
+	// no deployments
+	deployments, err := f.gormdb.listDeployments(f.ctx, flt)
+	assert.Nil(t, err)
+	assert.Equal(t, 0, len(deployments))
+
+	// create deployment and obtain it via list
+	deployment := makeSchemeDeployment()
+	createDeploymentWithTest(t, f.ctx, f.gormdb, deployment)
+
+	deployments, err = f.gormdb.listDeployments(f.ctx, flt)
```
efiShtain, line 66: effectively you don't check any filter capability here, fo you could pass the default filter and it work the same 
```go
sdkservices.ListDeploymentFilter{}
```

backend/internal/db/db.go
```diff
@@ -118,6 +118,7 @@ type DB interface {
 	UpdateSessionState(ctx context.Context, sessionID sdktypes.SessionID, state sdktypes.SessionState) error
 	AddSessionPrint(ctx context.Context, sessionID sdktypes.SessionID, print string) error
 	ListSessions(ctx context.Context, f sdkservices.ListSessionsFilter) ([]sdktypes.Session, int, error)
+	DeleteSession(ctx context.Context, sessionID sdktypes.SessionID) error
```
efiShtain, line 121: I think this should support hard delete with default soft

backend/internal/db/dbgorm/sessions_test.go
```diff
@@ -0,0 +1,96 @@
...
+
+func createSessionWithTest(t *testing.T, ctx context.Context, gormdb *gormdb, session scheme.Session) {
+	assert.Nil(t, gormdb.createSession(ctx, session))
+	res := gormdb.db.First(&scheme.Session{}, "session_id = ?", session.SessionID)
+	assert.Nil(t, res.Error)
+	assert.Equal(t, int64(1), res.RowsAffected)
+}
+
+func ttt() {
+}
```
efiShtain: ?

backend/internal/db/dbgorm/dbgorm.go
ig-ra: implemented with softdelete. See tests

backend/internal/sessionsgrpcsvc/svc.go
```diff
@@ -32,27 +33,43 @@ func Init(mux *http.ServeMux, sessions sdkservices.Sessions) {
 	mux.Handle(path, handler)
 }
 
+// re-wrap sdk as connect error
+func asConnectError(err error) error {
+	// in protovalidate Error() is defined on pointer type and there is no error object
+	var pValidationError *protovalidate.ValidationError
```
efiShtain: no need for p prefix here

efiShtain:
Except for the unit tests + database access issue (and some minor stuff)
LGTM

backend/internal/db/dbgorm/sessions_test.go
ig-ra: removed

backend/internal/db/db.go
ig-ra: meanwhile we won't support hard delete. let's discuss?

backend/internal/db/dbgorm/deployments_test.go
ig-ra: I would like explicitly specify that there is no limit

cmd/ak/cmd/sessions/delete.go
ig-ra: done

ig-ra: @daabr, I have a question regarding the meaning of `--fail`. In many commands we would like to fail/report failure  if the requested resource wasn't found. But it feels that in `delete` we would like to report failure also if the session was in not appropriate for deletion state (e.g. not finished/error). So, I'm in doubts regarding my decision to use this flag to report any error and not only if not found.

backend/internal/db/dbgorm/dbgorm.go
ig-ra: could u plz elaborate more? pay attention that getOneModel return Model/Schema and not SDK type

backend/internal/db/dbgorm/dbgorm.go
ig-ra: I think yes, it's the same logic that is present in original get()

backend/internal/db/dbgorm/dbgorm.go
ig-ra: I renamed both function for clarity - `getOne` and `getOneWTransform`

backend/internal/sessions/sessions.go
ig-ra: I agree with you that logs should be clear.
In this particular example it's not clear for me what's the best solution.. IMO logging before before will cover both cases - success and failure. On success you will see that session deletion was triggered and on failure you will see the trigger and error logged on nested level functions.
I could suggest (maybe) better and cleaner solution which will include call+cache err+check+log. But it feels as an overkill for our current purposes, e.g. 
```go
   if err := s.svcs.DB.DeleteSession(ctx, sessionID); err != nil {
      s.z.With(zap.String("session_id", sessionID.String())).Debug("delete")
   } else {
      s.z.With(zap.String("session_id", sessionID.String()), zap.Error(err)).Debug("delete failed")
   }
```
Should we create generic wrapper for such cases? WDYT?

backend/internal/sessionsgrpcsvc/svc.go
ig-ra: added a ticket. resolving

backend/internal/db/dbgorm/deployments.go
ig-ra: hmm.. maybe you are right. I just kept the original logic/signature for returning objects by value.
let's think together? will it be better for us/performance? Or it will change nothing? If it changes nothing, I think I will prefer to keep this by value, then.

btw, there are other functions (e.g. listDeployments/Sessions) which return list, so for consistency they should return list of references then as well and this maybe harder with Gorm, unless we will return pointer to list itself.

backend/internal/db/dbgorm/deployments.go
ig-ra: done. resolving

backend/internal/db/dbgorm/dbgorm.go
efiShtain: in gorm to use associations, you have to use .Association method, not as part of the where clause

backend/internal/db/dbgorm/gorm_test.go
```diff
@@ -0,0 +1,119 @@
...
+	logger := logger.New(
+		log.New(os.Stdout, "\r\n", log.LstdFlags), // io writer
+		logger.Config{
+			SlowThreshold: time.Second,   // Slow SQL threshold
+			LogLevel:      logger.Silent, // Log level
+			Colorful:      false,         // Disable color
+		},
+	)
+	if dbName == "" {
```
itayd, line 35: // TODO: Test will gormkitteh and for various connection strings (ie databases)

backend/internal/sessionsgrpcsvc/svc.go
```diff
@@ -125,27 +132,55 @@ func (s *server) List(ctx context.Context, req *connect.Request[sessionsv1.ListR
...
+	if err != nil {
+		return nil, asConnectError(err)
+	}
+
+	if session == nil { // just to ensure, Get should return an error if not found
+		return nil, asConnectError(err)
+	}
+
+	// FIXME: what to return? Cancelled? Aborted? FailedPrecondition?
+	if state := sdktypes.GetSessionLatestState(session); state != sdktypes.CompletedSessionStateType && state != sdktypes.ErrorSessionStateType {
```
itayd: failed precondition, move into sessions

backend/internal/sessions/sessions.go
efiShtain: in the extreme case that the process will fail after logging delete and before actually deleting it (calling the database), you will get invalid logs

in the not so uncommon case, where by a bug or mistake, the error isn't printed to log after the delete fail, you still have an invalid log

so for me, as a guideline, I try to log what happened and not what I'm about to do, at least on info level

debug logs can be as verbose as I need, but most of the times, the log level in production is info and not debug

so maybe, info should indicate what happened, and debug can do what ever required ?

ig-ra: Agree. Above log is DEBUG btw.
resolving?

backend/internal/sessionsgrpcsvc/svc.go
ig-ra: done

efiShtain (approved):

cmd/ak/cmd/sessions/delete.go
ig-ra: add FailIfError in addition to FailIfNotFound

cmd/ak/cmd/sessions/delete.go
ig-ra: used both `FailIfNotFound` and `FailedIfError`, which I introduced

cmd/ak/cmd/sessions/delete.go
ig-ra: rewritten

backend/internal/db/dbgorm/deployments.go
```diff
@@ -9,12 +9,17 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+	"gorm.io/gorm"
```
daabr, line 12: Our convention is to have (up to) 3 separate blocks inside `import()`: standard Go packages, non-AK, and AK.

backend/internal/db/dbgorm/deployments.go
```diff
@@ -24,27 +29,37 @@ func (db *gormdb) CreateDeployment(ctx context.Context, deployment sdktypes.Depl
...
 }
 
+func (db *gormdb) getDeployment(ctx context.Context, deploymentID string) (*scheme.Deployment, error) {
+	return getOne(db.db, ctx, scheme.Deployment{}, "deployment_id = ?", deploymentID)
+}
+
 func (db *gormdb) GetDeployment(ctx context.Context, id sdktypes.DeploymentID) (sdktypes.Deployment, error) {
-	return get(db.db, ctx, scheme.ParseDeployment, "deployment_id = ?", id.String())
+	d, err := db.getDeployment(ctx, id.String())
+	if d == nil || err != nil {
```
daabr, line 45: Can you add a short comment explaining when/why it's possible to get `d == nil` as well as `err == nil`?

backend/internal/db/dbgorm/deployments.go
```diff
@@ -24,27 +29,37 @@ func (db *gormdb) CreateDeployment(ctx context.Context, deployment sdktypes.Depl
...
+	}
+	return scheme.ParseDeployment(*d)
 }
 
 func (db *gormdb) DeleteDeployment(ctx context.Context, id sdktypes.DeploymentID) error {
 	var b scheme.Deployment
 	if err := db.db.WithContext(ctx).Where("deployment_id = ?", id.String()).Delete(&b).Error; err != nil {
 		return translateError(err)
 	}
+	// FIXME: delete deployment sessions as well?
```
daabr, line 56: Does this require a separate Linear ticket?

backend/internal/db/dbgorm/deployments.go
```diff
@@ -24,27 +29,37 @@ func (db *gormdb) CreateDeployment(ctx context.Context, deployment sdktypes.Depl
...
 
 	return nil
 }
 
-func (db *gormdb) ListDeployments(ctx context.Context, filter sdkservices.ListDeploymentsFilter) ([]sdktypes.Deployment, error) {
+// FIXME: fix generic in order to avoid all this
```
daabr, line 61: Does this require a separate Linear ticket?

backend/internal/db/dbgorm/deployments.go
```diff
@@ -61,34 +76,56 @@ func (db *gormdb) ListDeployments(ctx context.Context, filter sdkservices.ListDe
...
 		return nil, translateError(err)
 	}
+	return ds, nil
+}
 
-	return kittehs.TransformError(ds, scheme.ParseDeployment)
+func (db *gormdb) ListDeployments(ctx context.Context, filter sdkservices.ListDeploymentsFilter) ([]sdktypes.Deployment, error) {
+	if filter.IncludeSessionStats {
+		ds, err := db.listDeploymentsWithStats(ctx, filter)
+		if ds == nil {
```
daabr, line 118: I could be wrong, but lines 118 and 124 look weird to me, they imply a hidden connection between `ds == nil` and `err`?

backend/internal/db/dbgorm/dbgorm.go
daabr, line 126: Does this require a separate Linear ticket?

backend/internal/db/dbgorm/dbgorm.go
```diff
@@ -99,16 +99,18 @@ func (db *gormdb) Teardown(ctx context.Context) error {
...
 		z:  db.z,
 		db: db.db.Debug(),
 	}
 }
 
-func get[T, R any](db *gorm.DB, ctx context.Context, f func(t T) (*R, error), where string, args ...any) (*R, error) {
+func getOneWTransform[T, R any](db *gorm.DB, ctx context.Context, f func(t T) (*R, error), where string, args ...any) (*R, error) {
 	var r T
+
+	// TODO: fetch all records and report if there is more than one record
```
daabr, line 113: Does this require a separate Linear ticket?

backend/internal/db/dbgorm/dbgorm.go
```diff
@@ -99,16 +99,18 @@ func (db *gormdb) Teardown(ctx context.Context) error {
 	return nil
 }
 
-// Todo: not sure this will work with the connect method
+// TODO: not sure this will work with the connect method
```
daabr, line 102: Does this require a separate Linear ticket?

backend/internal/db/dbgorm/deployments_test.go
```diff
@@ -0,0 +1,122 @@
...
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"go.autokitteh.dev/autokitteh/backend/internal/db/dbgorm/scheme"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+	"gorm.io/gorm"
```
daabr: Our convention is to have (up to) 3 separate blocks inside `import()`: standard Go packages, non-AK, and AK. i.e. move gorm to the 2nd block

backend/internal/db/dbgorm/deployments.go
ig-ra: there is already

backend/internal/db/dbgorm/dbgorm.go
ig-ra: nop. just thoughts

daabr (approved):

itayd (approved):

PR #16, opened 2024-02-22
project not needed for build

itayd:
- no need for resources_paths from project - everything in a directory is built
- there is no need for project reference in build, this makes it completely decoupled from a project


cmd/ak/cmd/builds/upload.go
```diff
@@ -7,13 +7,9 @@ import (
...
 	"go.autokitteh.dev/autokitteh/cmd/ak/common"
-	"go.autokitteh.dev/autokitteh/internal/kittehs"
-	"go.autokitteh.dev/autokitteh/internal/resolver"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-var project string
-
 var uploadCmd = common.StandardCommand(&cobra.Command{
 	Use:     "upload <build file path> <--project=...>",
```
daabr:
```suggestion
	Use:     "upload <build file path>",
```

cmd/ak/cmd/builds/upload.go
```diff
@@ -26,17 +22,7 @@ var uploadCmd = common.StandardCommand(&cobra.Command{
...
-		if err != nil {
-			return err
-		}
-		if p == nil {
-			err = fmt.Errorf("project %q not found", project)
-			return common.NewExitCodeError(common.NotFoundExitCode, err)
-		}
-
-		b, err := sdktypes.BuildFromProto(&sdktypes.BuildPB{ProjectId: p.ToProto().ProjectId})
+		b, err := sdktypes.BuildFromProto(&sdktypes.BuildPB{})
```
daabr: Should we change `builds().Save()` to not require a build object as a parameter, given that it's empty by definition?

backend/internal/projects/projects.go
```diff
@@ -97,9 +91,7 @@ func (ps *Projects) Build(ctx context.Context, projectID sdktypes.ProjectID) (sd
 		return nil, err
 	}
 
-	build := kittehs.Must1(sdktypes.BuildFromProto(&sdktypes.BuildPB{
-		ProjectId: projectID.String(),
-	}))
+	build := kittehs.Must1(sdktypes.BuildFromProto(&sdktypes.BuildPB{}))
```
daabr: Doesn't it make more sense to create a new object directly, without using an empty PB and muting it's impossible translation error?

systest/testdata/builds/build.txtar
```diff
@@ -0,0 +1,217 @@
+ak builds create --path hiss.star
```
daabr, line 2:
```suggestion
# Negative test: build failure due to error in code.
ak builds create --path hiss.star
```

systest/testdata/builds/build.txtar
```diff
@@ -0,0 +1,217 @@
...
+output equals file hiss_error.txt
+return code == 1
+
+ak builds create --path single.star -o single.akb -d
```
daabr:
```suggestion
ak builds create --path single.star -o single.akb --describe
```

systest/testdata/builds/build.txtar
```diff
@@ -0,0 +1,217 @@
...
+output equals_json file single_output.json
+return code == 0
+
+ak builds create --dir multi -o multi.akb -d
```
daabr:
```suggestion
ak builds create --dir multi -o multi.akb --describe
```

daabr (commented):
Mostly nits, plus 1 small fix and a question.

cmd/ak/cmd/builds/upload.go
itayd: we might want to add stuff in the future. i think we should leave it as is.

backend/internal/projects/projects.go
itayd: yeah it's silly. going to add an sdktypes.NewBuild() function.

cmd/ak/cmd/builds/upload.go
itayd: used NewBuild

daabr (approved):

PR #17, opened 2024-02-23
systest: gracefuly stop ak + temporal

itayd:
not sure that's the prettiest thing that can be done. this is more of a poc to show that it can be stopped from a test.

made sure temporal is thoroughly killed at end of run - no lingering processes on either success or failure.

### DIFF
--- a/cmd/ak/cmd/up.go
+++ b/cmd/ak/cmd/up.go
@@ -1,9 +1,12 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
+	"os"
 
 	"github.com/spf13/cobra"
+	"go.uber.org/fx"
 
 	"go.autokitteh.dev/autokitteh/backend/basesvc"
 	"go.autokitteh.dev/autokitteh/backend/configset"
@@ -12,6 +15,8 @@ import (
 )
 
 var mode string
+
+var App *fx.App
 
 var upCmd = common.StandardCommand(&cobra.Command{
 	Use:     "up [--mode={default|dev|test}]",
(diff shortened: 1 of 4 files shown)


systest/testdata/completion.txtar
```diff
@@ -1,7 +1,7 @@
 # The completion command exists.
 ak completion bash
 return code == 0
-output contains 'bash completion for ak'
+output contains 'bash completion for axk'
```
daabr, line 4: Typo?

systest/server.go
```diff
@@ -16,11 +16,15 @@ const (
...
 	cmd.RootCmd.SetArgs([]string{"up", "--config", "http.addr=:0", "--mode", "test"})
 
-	// We don't care about execution errors here, the test will check this.
-	cmd.RootCmd.ExecuteContext(ctx) //nolint:errcheck
+	if err := cmd.RootCmd.ExecuteContext(ctx); err != nil {
+		t.Logf("ak server error: %v", err)
+	}
+}
+
+func stopAKServer(ctx context.Context, t *testing.T) {
```
daabr, line 27: `stopAKServer` is empty and unused - why does it exist?

systest/server.go
```diff
@@ -16,11 +16,15 @@ const (
...
 // Start the AK server, but in a goroutine rather than as a separate
 // subprocess: to support breakpoint debugging, and measure test coverage.
-func startAKServer(ctx context.Context) {
+func startAKServer(ctx context.Context, t *testing.T) {
 	cmd.RootCmd.SetArgs([]string{"up", "--config", "http.addr=:0", "--mode", "test"})
 
-	// We don't care about execution errors here, the test will check this.
-	cmd.RootCmd.ExecuteContext(ctx) //nolint:errcheck
+	if err := cmd.RootCmd.ExecuteContext(ctx); err != nil {
+		t.Logf("ak server error: %v", err)
```
daabr, line 23: I'm not 100% sure `t.Logf` will actually print stuff given that it's in a separate goroutine, but anyway why are you checking for an error that can be reported only after the server is shut down?

systest/ak_test.go
```diff
@@ -89,12 +91,13 @@ func setUpTest(t *testing.T) string {
...
-	t.Cleanup(cancel) // Stop the AK server's goroutine.
+	go startAKServer(ctx, t)
+
+	defer cmd.App.Stop(context.Background())
 
-	akAddr := waitForAKServer(t, combinedOutput)
+	t.Cleanup(cancel)
 
-	return akAddr
+	return waitForAKServer(t, combinedOutput)
```
daabr, line 100: Nit: I intentionally used the `akAddr` helper variable to show that `waitForAKServer` returns an address. Removing it is shorter, but less clear.

cmd/ak/cmd/up.go
```diff
@@ -13,6 +16,8 @@ import (
 
 var mode string
 
+var App *fx.App
```
daabr, line 19: Making `fx.App` a global and exported variable creates a huge risk. Now any code anywhere in the server can mess with it directly and damage the server, intentionally or not. Even if there's no damage, this goes against the separation of concerns principle.

systest/ak_test.go
```diff
@@ -89,12 +91,13 @@ func setUpTest(t *testing.T) string {
 	// Start the AK server, but in a goroutine rather than as a separate
 	// subprocess: to support breakpoint debugging, and measure test coverage.
 	ctx, cancel := context.WithCancel(context.Background())
-	go startAKServer(ctx)
-	t.Cleanup(cancel) // Stop the AK server's goroutine.
+	go startAKServer(ctx, t)
+
+	defer cmd.App.Stop(context.Background())
```
daabr, line 96: You can call `App.Stop()` here because of `App` is inappropriately exposed in `cmd/up.go` - see my comment about that in that file.

Also, this call breaks goroutine boundaries - it's called from one goroutine on an object in a different goroutine, and may cause race conditions.

Moreover, I don't think `defer` is used correctly here: it's guaranteed to run when the calling scope is closed, i.e. when `setUpTest()` is done, before the test begins, not after it ends.

Also, you're not checking if `App.Stop()` returns an errors, so you're not handling the condition where it failed to actually stop the AK server.

PR #20, opened 2024-02-24
Tiny tweaks in manifest internals

daabr:


internal/manifest/log.go
```diff
@@ -1,6 +1,8 @@
 package manifest
 
-import "fmt"
```
itayd, line 3: why

internal/manifest/log.go
```diff
@@ -11,8 +13,5 @@ type keyer interface {
 }
 
 func (l Log) For(kind string, keyer keyer) Log {
-	if kind != "" {
-		kind += " "
-	}
-	return Log(func(f string) { l.Printf("%s%v: %v", kind, keyer.GetKey(), f) })
+	return Log(func(f string) { l.Printf("%s %q: %v", kind, keyer.GetKey(), f) })
```
itayd, line 16: if `kind` is empty you'll have an extra space in the beginning

internal/manifest/manifest.go
```diff
@@ -26,8 +26,7 @@ type stringKeyer string
 func (s stringKeyer) GetKey() string { return string(s) }
 
 type Manifest struct {
-	Version string `yaml:"version,omitempty" json:"version,omitempty" jsonschema:"required"`
-
+	Version string   `yaml:"version,omitempty" json:"version,omitempty" jsonschema:"required"`
```
itayd, line 29: daniel

internal/manifest/log.go
daabr: Yes, but in AK `kind` should never be empty.

internal/manifest/log.go
daabr: standardization of import block appearance (I wouldn't have bothered to do this, but I already touched the file for a different reason)

itayd (approved):

PR #27, opened 2024-02-25
apply manifest return created project ids

efiShtain:
- Update apply proto to return project_ids list if created
- Update code to return list of project ids

this is required for the extension to use instead of having the extension parse the manifest file and understand it's structure


proto/autokitteh/apply/v1/svc.proto
```diff
@@ -11,6 +11,7 @@ message ApplyRequest {
 
 message ApplyResponse {
   repeated string logs = 1;
+  repeated string project_ids = 2;
```
itayd, line 14: why not returning all ids? do we want to return only stuff that created or also modified? or maybe just everything?

itayd (approved):

proto/autokitteh/apply/v1/svc.proto
daabr: Also, don't we support only a single project now in our manifest schema?

cmd/ak/cmd/manifest/apply.go
```diff
@@ -33,9 +33,11 @@ var applyCmd = common.StandardCommand(&cobra.Command{
 			ctx, cancel := common.LimitedContext()
 			defer cancel()
 
-			return imanifest.Execute(ctx, actions, common.Client(), func(msg string) {
+			pids, err := imanifest.Execute(ctx, actions, common.Client(), func(msg string) {
 				fmt.Fprintf(os.Stderr, "[exec] %s\n", msg)
 			})
+			fmt.Println(pids)
```
daabr: I'm OK with adding the data to the API for the extension, but do you think it's important to print it in the CLI too?

1. I'm OK with `os.Fprintf` for logs, but data printing is done in all commands with `common.Render` or a variant for lists
2. Combining stderr and stdout is usually a bad idea
3. It's unclear to the user what this line actually means, and it's not contributing new information on top of the exec logs

daabr (approved):

PR #28, opened 2024-02-26
ENG-456: Fix "config set" CLI command

daabr:
Undoing the usage of a temp directory requires reloading the XDG context, otherwise the Go package continues to use the old temp path.


itayd:
ummmm am i missing something or this is just a var name change?

efiShtain (approved):

PR #32, opened 2024-02-27
go test in single job

itayd:
will minimize resources


tebeka (approved):

.github/workflows/ci-go.yml
```diff
@@ -29,7 +29,7 @@ jobs:
       - name: Build
         run: make build bin
       - name: Test
-        run: make test-unit
+        run: make test
```
daabr, line 32: The impact of this PR is that we're running 3 slow make targets (`test-race`, `test-cli`, `test-runs`) **sequentially** in a single CI workflow, instead of running them in parallel.

As a result, we're saving 1 build (not 2, because a regular build and a build for `test-race` are not the same), but the total runtime grows from 2m+2m+3m (https://github.com/autokitteh/autokitteh/actions/runs/8120148376/job/22196966987?pr=43) to 6m (https://github.com/autokitteh/autokitteh/actions/runs/8116201065/job/22185684179?pr=32).

So yeah, in total this PR is saving 2m of compute time, but forcing the PR author to wait twice the time for a result.

I'm not sure that's a win...?

Also, didn't we agree to use the regular unit tests (i.e. without `-race`) in the GitHub CI?

itayd: @tebeka, thoughts?

daabr: btw, what if we use `make test-unit test-run` instead of `make test-race test-run test-cli`?

I wonder if the average runtime would be close enough to what we currently have for `make test-unit` - if so, that **would** be a good change.

Either way, we should keep `test-cli` separate because we're going to deprecate it soon anyway after we migrate all tests from that to `/tests/system`.

PR #42, opened 2024-03-01
feat(db): delete deployments

ig-ra:
ENG-123


backend/internal/db/dbgorm/deployments.go
```diff
@@ -43,22 +43,24 @@ func (db *gormdb) getDeployment(ctx context.Context, deploymentID string) (*sche
...
-		return translateError(err)
-	}
-	// FIXME: delete deployment sessions as well?
+func (db *gormdb) deleteDeployment(ctx context.Context, deploymentID string) error {
+	// TODO: this won't work. Consider removing generic delete function.
+	// return delete(db.db, ctx, scheme.Deployment{}, "deployment_id = ?", deploymentID)
 
-	return nil
+	var b scheme.Deployment = scheme.Deployment{DeploymentID: deploymentID}
+	return db.db.WithContext(ctx).Delete(&b).Error
```
efiShtain: did we decide on cascading deletion?
deployments should delete all of its sessions (and session all of its sub types)?

backend/internal/db/dbgorm/deployments.go
```diff
@@ -43,22 +43,24 @@ func (db *gormdb) getDeployment(ctx context.Context, deploymentID string) (*sche
...
 
-func (db *gormdb) DeleteDeployment(ctx context.Context, id sdktypes.DeploymentID) error {
-	var b scheme.Deployment
-	if err := db.db.WithContext(ctx).Where("deployment_id = ?", id.String()).Delete(&b).Error; err != nil {
-		return translateError(err)
-	}
-	// FIXME: delete deployment sessions as well?
+func (db *gormdb) deleteDeployment(ctx context.Context, deploymentID string) error {
+	// TODO: this won't work. Consider removing generic delete function.
+	// return delete(db.db, ctx, scheme.Deployment{}, "deployment_id = ?", deploymentID)
```
efiShtain: don't really this need comment

backend/internal/db/dbgorm/builds.go
```diff
@@ -10,18 +10,18 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+func (db *gormdb) saveBuild(ctx context.Context, build scheme.Build) error {
```
efiShtain: I assume you just saw this one and fixed it which is great
but if this is unrelated to this PR, please leave a comment on the PR itself or split it to a different PR
since It takes me time to understand why you did it

backend/internal/db/dbgorm/deployments.go
```diff
@@ -43,22 +43,24 @@ func (db *gormdb) getDeployment(ctx context.Context, deploymentID string) (*sche
...
+	var b scheme.Deployment = scheme.Deployment{DeploymentID: deploymentID}
+	return db.db.WithContext(ctx).Delete(&b).Error
 }
 
-// FIXME: fix generic in order to avoid all this
+func (db *gormdb) DeleteDeployment(ctx context.Context, deploymentID sdktypes.DeploymentID) error {
+	return translateError(db.deleteDeployment(ctx, deploymentID.String()))
+}
+
+// TODO: rewrite generic in order to avoid splitting to Deployment and DeploymentWithStats
```
efiShtain: TODO without a ticket would probably never be solved, 
so better to add a ticket and list it here like TODO(ENG-XYZ): ...

backend/internal/db/dbgorm/gorm_test.go
```diff
@@ -41,30 +49,23 @@ func setupDB(dbName string) *gorm.DB {
...
-			log.Printf("Could not sync logger: %v", err)
-		}
-	}()
-	return logger
-}
-
-func newDbFixture() *dbFixture {
+func newDbFixture(withoutForeignKeys bool) *dbFixture {
 	db := setupDB("") // in-memory db, specify filename to use file db
+	if withoutForeignKeys {
```
efiShtain: when would you use this ?

backend/internal/db/dbgorm/gorm_test.go
```diff
@@ -41,30 +49,23 @@ func setupDB(dbName string) *gorm.DB {
 		},
 		Logger: logger,
 	})
+	db.Exec("PRAGMA foreign_keys = ON")
```
efiShtain: you have to reimplement this logic here since you are not using our gorm wrapper
maybe you should consider using ours

backend/internal/db/dbgorm/gorm_test.go
```diff
@@ -82,39 +83,61 @@ func newDbFixture() *dbFixture {
...
+func newBuild() scheme.Build {
+	now := time.Now().UTC() // save and compare times in UTC
+
+	return scheme.Build{
+		BuildID:   testBuildID,
+		Data:      []byte{},
+		CreatedAt: now,
+	}
+}
+*/
```
efiShtain: remove comments

backend/internal/db/dbgorm/scheme/scheme.go
```diff
@@ -370,6 +370,10 @@ type Deployment struct {
 	State        int32
 	CreatedAt    time.Time
 	UpdatedAt    time.Time
+	DeletedAt    gorm.DeletedAt `gorm:"index"`
+
+	// just for foreign key constraint. Without it gorm won't enforce it
+	Sessions []*Session `gorm:"foreignKey:DeploymentID"`
```
efiShtain: could you clarify this one?

backend/internal/db/dbgorm/scheme/scheme.go
```diff
@@ -322,7 +322,7 @@ func ParseSessionCallAttemptComplete(c SessionCallAttempt) (d sdktypes.SessionCa
 
 type Session struct {
 	SessionID        string `gorm:"primaryKey"`
-	DeploymentID     string `gorm:"index"`
+	DeploymentID     string `gorm:"index;foreignKey:DeploymentID;references:DeploymentID"`
```
efiShtain: do you need both foreignKey and references here?

backend/internal/db/dbgorm/scheme/scheme.go
```diff
@@ -388,6 +392,11 @@ func ParseDeployment(d Deployment) (sdktypes.Deployment, error) {
 	return deployment, nil
 }
 
+// gorm don't cascade soft deletes. hook is transactional.
+func (d *Deployment) AfterDelete(db *gorm.DB) (err error) {
+	return db.Where("deployment_id = ?", d.DeploymentID).Delete(&Session{}).Error
```
efiShtain: what about running sessions ?
we don't allow deleting them
we should probably not allow deleting a deployment with a running session

other option is deleting but letting the running session to complete or fail

backend/internal/db/dbgorm/sessions.go
```diff
@@ -17,16 +17,24 @@ import (
...
-func (db *gormdb) GetSession(ctx context.Context, sessionID sdktypes.SessionID) (sdktypes.Session, error) {
-	return getOneWTransform(db.db, ctx, scheme.ParseSession, "session_id = ?", sessionID.String())
+func (db *gormdb) getSession(ctx context.Context, sessionID string) (*scheme.Session, error) {
+	return getOne(db.db, ctx, scheme.Session{}, "session_id = ?", sessionID)
+}
+
+func (db *gormdb) GetSession(ctx context.Context, id sdktypes.SessionID) (sdktypes.Session, error) {
+	s, err := db.getSession(ctx, id.String())
+	if s == nil || err != nil {
+		return nil, translateError(err)
```
efiShtain: what does translate error of nil return ?

backend/internal/deployments/deployments.go
```diff
@@ -150,6 +150,27 @@ func (d *deployments) Drain(ctx context.Context, id sdktypes.DeploymentID) error
...
+		}
+
+		if sdktypes.GetDeploymentState(dep) != sdktypes.DeploymentStateInactive {
+			err = drain(ctx, tx, id)
+		}
+		return err
+	})
+
+	// NOTE: deployment is inactive, so it should be safe to delete it outside of the transaction.
+	// It seems that sqlite does not support nested transactions, so we'll fail in delete hook, since DB is locked.
```
efiShtain: so... what is the solution here?

cmd/ak/cmd/deployments/delete.go
```diff
@@ -0,0 +1,40 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var deleteCmd = common.StandardCommand(&cobra.Command{
+	Use:     "delete <deployment ID> [--fail]",
+	Short:   "Delete inactive deployment",
```
efiShtain, line 12: technically the delete code above drains a deployment, so it deletes active deployments as well

cmd/ak/cmd/deployments/delete.go
```diff
@@ -0,0 +1,40 @@
...
+	RunE: func(cmd *cobra.Command, args []string) error {
+		r := resolver.Resolver{Client: common.Client()}
+		d, id, err := r.DeploymentID(args[0])
+		if err != nil {
+			return common.FailIfError(cmd, err, "deployment")
+		}
+
+		if err := common.FailIfNotFound(cmd, "deployment", d); err != nil { // test d != nil
+			return err
+		}
```
efiShtain: do we really need to check if deployments exists
and then try to delete it ?
we will get the same result by just calling delete if it exists it will be deleted (assuming it can reach inactive state)
if it is not found, we will get not found error

proto/autokitteh/sessions/v1/svc.proto
```diff
@@ -61,6 +61,12 @@ message GetLogResponse {
   SessionLog log = 1 [(buf.validate.field).required = true];
 }
 
+message DeleteRequest {
+  string session_id = 1 [(buf.validate.field).string.min_len = 1];
+}
+
+message DeleteResponse {}
+
```
efiShtain, line 69: is this part of this PR ?

sdk/sdkclients/sdkdeploymentsclient/client.go
```diff
@@ -20,25 +20,26 @@ type client struct {
...
 
-// Get implements sdkservices.Deployments.
-func (c *client) Get(ctx context.Context, id sdktypes.DeploymentID) (sdktypes.Deployment, error) {
-	resp, err := c.client.Get(ctx, connect.NewRequest(&deploymentsv1.GetRequest{DeploymentId: id.String()}))
+func New(p sdkclient.Params) sdkservices.Deployments {
+	return &client{client: internal.New(deploymentsv1connect.NewDeploymentsServiceClient, p)}
+}
+
+// Create implements sdkservices.Deployments.
+func (c *client) Create(ctx context.Context, deployment sdktypes.Deployment) (sdktypes.DeploymentID, error) {
```
efiShtain, line 28: did we change the API here intentionally to return just the id and not the whole model?

sdk/sdkclients/sdkdeploymentsclient/client.go
```diff
@@ -69,41 +70,44 @@ func (c *client) Test(ctx context.Context, id sdktypes.DeploymentID) error {
...
 
-// Create implements sdkservices.Deployments.
-func (c *client) Create(ctx context.Context, deployment sdktypes.Deployment) (sdktypes.DeploymentID, error) {
-	resp, err := c.client.Create(ctx, connect.NewRequest(&deploymentsv1.CreateRequest{Deployment: deployment.ToProto()}))
+// Drain implements sdkservices.Deployments.
+func (c *client) Drain(ctx context.Context, id sdktypes.DeploymentID) error {
+	resp, err := c.client.Drain(ctx, connect.NewRequest(&deploymentsv1.DrainRequest{DeploymentId: id.String()}))
 	if err != nil {
-		return nil, rpcerrors.TranslateError(err)
+		return rpcerrors.TranslateError(err)
```
efiShtain, line 77: can probably add a FromConnectError matching the ToConnectError to have it more zen
WDYT ?

backend/internal/db/dbgorm/gorm_test.go
ig-ra: testing. otherwise I need to create deployment, build, env for everything.
And there is specific test that ensures that foreign keys are working

backend/internal/db/dbgorm/deployments.go
ig-ra: I'm waiting for Itays new generic implementation. If this is something that would allow me to solve this, then good. Otherwise I may consider either remove TODO or add a ticket

backend/internal/db/dbgorm/builds.go
ig-ra: No, I did it due to my fix for foreign keys. Till now foreign keys weren't enforced for Deployment/Sesions. See Notes in the text. That's why later I've introduce the tests which don't require foreign keys

backend/internal/db/dbgorm/deployments.go
ig-ra: I  left the comment for myself. Once I'm finishing with deleting of everything (connecitons, projects etc) I will decide whether I need to keep generic delete or remove it. Right now we are not using it in deployments, only in session

backend/internal/db/dbgorm/deployments.go
ig-ra: see schema. Deployment cascading soft-delete to sessions

cmd/ak/cmd/deployments/delete.go
ig-ra: explain?

cmd/ak/cmd/deployments/delete.go
ig-ra: I'm following the logic of other cmd commands. If you think it's not needed, let's discuss and open a ticket to fix all commands?

proto/autokitteh/sessions/v1/svc.proto
ig-ra: explain?

sdk/sdkclients/sdkdeploymentsclient/client.go
ig-ra: sure, worth a ticker. This is the first step into this direction indeed. But we are still not there. Need to fix tests, then write tests for all commands and then ensure the right errors. 
I think I will finish with all commands first

backend/internal/deployments/deployments.go
ig-ra: this is the solution. My claim that we could delete inactive deployment not in transaction

backend/internal/db/dbgorm/gorm_test.go
ig-ra: yes, I will move all tests to use formkitteh. But not now. I will open a ticket
https://linear.app/autokitteh/issue/ENG-476/move-db-tests-to-use-gormkitteh

backend/internal/db/dbgorm/gorm_test.go
ig-ra: I will need this in next PRs. That's why it's still there

backend/internal/db/dbgorm/scheme/scheme.go
ig-ra: I think it's written in comment above. Without having Sessions reference in Deployment schema, gorm won't enforce it and will allow to add sessions w/o matching deployment. This was a bug I solved.

backend/internal/db/dbgorm/scheme/scheme.go
ig-ra: I prefer to be explicit in this case. But I think I could remove references.
What do u prefer?

backend/internal/db/dbgorm/scheme/scheme.go
ig-ra: This is DB implementation. The logic of checking whether we could delete deployment isn't here but in deployments.Delete. There I'm allowing to delete only inactive deployment. (drain->deactivate->delete). Which ensures that there are no sessions.

backend/internal/db/dbgorm/sessions.go
ig-ra: nil

sdk/sdkclients/sdkdeploymentsclient/client.go
ig-ra: I don't think I've changed any API here. My diff shows me that Create was returning ID. Am I wrong?

backend/internal/db/dbgorm/scheme/scheme.go
efiShtain: the relation we have is session belongs to deployment
we didn't need deployment has sessions
do we need the two way? what changes in the database after this change ?

backend/internal/db/dbgorm/scheme/scheme.go
efiShtain: minimum of options are better
I think it can even work without any of them

cmd/ak/cmd/deployments/delete.go
efiShtain: you say delete inactive deployments in the text
but delete allows deleting active deployment, it just moves it to inactive first by draining it

sdk/sdkclients/sdkdeploymentsclient/client.go
efiShtain: you are right, 
I read it wrong

proto/autokitteh/sessions/v1/svc.proto
efiShtain: why there are changes to the sessions svc ?
you added delete to the deployments svc

cmd/ak/cmd/deployments/delete.go
ig-ra: you are right. I already asked this question Haim and Itay. Got no answer meanwhile

ig-ra: rewritten. Now we are checking that deployment is inactive and then deleting, err otherwise

backend/internal/db/dbgorm/scheme/scheme.go
ig-ra: DB wont be changed. But without having sessions in Deployment foreign keys won't be forced by gorm.
Gorm is enforcing foreign keys only if you have relevant objects/pointers in the schema. 
It's semi-stated here https://gorm.io/docs/belongs_to.html, https://gorm.io/docs/has_many.html
see also https://medium.com/@chaewonkong/learn-go-understanding-and-implementing-foreign-keys-with-gorm-6d7608e1dbf6

proto/autokitteh/sessions/v1/svc.proto
ig-ra: just moved to the upper part of the file. refactoring

backend/internal/db/dbgorm/scheme/scheme.go
ig-ra: deleted references

itayd (approved):
preemptive to ungate. either efi or daniel must approve as well.
approved by Efi

efiShtain (approved):

cmd/ak/cmd/deployments/deployments.go
```diff
@@ -30,6 +30,7 @@ func init() {
 	deploymentsCmd.AddCommand(deactivateCmd)
 	deploymentsCmd.AddCommand(getCmd)
 	deploymentsCmd.AddCommand(listCmd)
+	deploymentsCmd.AddCommand(deleteCmd)
```
daabr: Cobra displays commands alphabetically, let's do the same when we add subcommands to a parent (I know we don't always do it yet, but we will).

cmd/ak/cmd/deployments/delete.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var deleteCmd = common.StandardCommand(&cobra.Command{
+	Use:     "delete <deployment ID> [--fail]",
+	Short:   "Delete inactive deployment",
+	Aliases: []string{"d"},
```
daabr: We already have 2 other commands that start with D: "drain" (dr), and "deactivate" (de). Please remove the alias "d", or change it to "del"?

cmd/ak/cmd/deployments/delete.go
```diff
@@ -0,0 +1,40 @@
...
+			return common.FailIfError(cmd, err, "deployment")
+		}
+
+		if err := common.FailIfNotFound(cmd, "deployment", d); err != nil { // test d != nil
+			return err
+		}
+
+		ctx, cancel := common.LimitedContext()
+		defer cancel()
+		err = deployments().Delete(ctx, id)
```
daabr: I'm not a big fan of empty lines like Itay, but I do see a need to emphasize that things with `defer` are "blocks" of code, no matter how small.

```suggestion
		defer cancel()

		err = deployments().Delete(ctx, id)
```

cmd/ak/cmd/deployments/delete.go
```diff
@@ -0,0 +1,40 @@
...
+		if err != nil { // don't ignore silently any error, expect not found above
+			return common.ToExitCodeError(err, "delete deployment")
```
daabr: `common.ToExitCodeError` already handles `err == nil`, why do you need this?

daabr (approved):

cmd/ak/cmd/deployments/delete.go
ig-ra: done

cmd/ak/cmd/deployments/deployments.go
ig-ra: done

cmd/ak/cmd/deployments/delete.go
ig-ra: done

PR #44, opened 2024-03-04
sdktypes rewrite

itayd:
This is the anticipated sdktypes rewrite.

- Refactor each type not to be an actual type and not just an alias. This allows to add methods on these types.
- IDs and Symbols can now be compared directly without `.String()`-ing them. Objects need to be compared using the `.Equal(other)` method though.
- Validations are part of the type now, not a member function, which was wrong.
- Get rid of sdktypes.Name. Everything "name-ish" is not a Symbol.
- Get rid of all icky oneofs in proto.

How to review:
- Start by looking at the proto first, not much changes there (mostly the oneofs good ridance).
- Then move to sdktypes: first read `doc.go`, then go over `object.go`, `id.go`, and `str.go`  by that order.
- Look how its being used everywhere in sdktypes.
- Go over the rest of the code which is just using the new sdktypes instead of the old.

"Enjoy".


backend/internal/connectionsgrpcsvc/svc.go
```diff
@@ -91,7 +91,7 @@ func (s *server) Get(ctx context.Context, req *connect.Request[connectionsv1.Get
 		return nil, sdkerrors.AsConnectError(err)
 	}
 
-	if c == nil {
+	if !c.IsValid() {
 		return connect.NewResponse(&connectionsv1.GetResponse{}), nil
```
tebeka, line 95: Why return an empty response with `nil` error when c is not valid?

backend/internal/db/dbgorm/dbgorm.go
```diff
@@ -107,20 +108,23 @@ func (db *gormdb) Debug() db.DB {
 	}
 }
 
-func getOneWTransform[T, R any](db *gorm.DB, ctx context.Context, f func(t T) (*R, error), where string, args ...any) (*R, error) {
-	var r T
+func getOneWTransform[T any, R sdktypes.Object](db *gorm.DB, ctx context.Context, f func(t T) (R, error), where string, args ...any) (R, error) {
```
tebeka, line 111: What does the `W` stand for?

go.mod
```diff
@@ -121,6 +121,7 @@ require (
 	github.com/go-sql-driver/mysql v1.7.1 // indirect
 	github.com/go-stack/stack v1.8.1 // indirect
 	github.com/go-viper/mapstructure/v2 v2.0.0-alpha.1 // indirect
+	github.com/gofrs/uuid/v5 v5.0.0 // indirect
```
tebeka, line 124: Why do use 3 different packages for UUID? (uuid, ulid, typeid)?

backend/internal/db/dbgorm/sessions_test.go
```diff
@@ -76,16 +76,16 @@ func TestListSessions(t *testing.T) {
...
 	sessions, cnt, err = f.gormdb.listSessions(f.ctx, flt)
-	assert.Nil(t, err)
+	assert.NoError(t, err)
 	assert.Equal(t, 1, cnt)
 	assert.Equal(t, 1, len(sessions))
 	assert.Equal(t, session.SessionID, sessions[0].SessionID)
 
 	// delete and ensure that list is empty
-	assert.Nil(t, f.gormdb.deleteSession(f.ctx, session.SessionID))
+	assert.NoError(t, f.gormdb.deleteSession(f.ctx, session.SessionID))
```
tebeka, line 85: Nitpick, I prefer to use `require` which stops the test on the first error.

integrations/http/http.go
```diff
@@ -233,27 +233,27 @@ func toStruct(r *http.Response) (sdktypes.Value, error) {
...
 
 	return sdktypes.NewStructValue(
 		sdktypes.NewStringValue("http_response"),
 		map[string]sdktypes.Value{
 			"url":         sdktypes.NewStringValue(r.Request.URL.String()),
 			"status_code": sdktypes.NewIntegerValue(int64(r.StatusCode)),
-			"headers": sdktypes.NewDictValue(
-				kittehs.TransformMapToList(r.Header, func(k string, vs []string) *sdktypes.DictValueItem {
-					return &sdktypes.DictValueItem{
+			"headers": kittehs.Must1(sdktypes.NewDictValue(
```
tebeka, line 245: I don't like using `Must` variants in a code that's not in `init` or `var`. Should return an error.

tebeka (commented):
Looks good in general.

I still don't like that the underlying type in business layer is protobuf.

sdk/sdktypes/object.go
```diff
@@ -13,207 +13,193 @@ import (
...
 type Object interface {
-	// all methods must be unexported.
-	// we want all nil guards effective for outsiders.
-	// this interface is just to form an algebric data type.
+	json.Marshaler
+	fmt.Stringer
+	isValider
+	stricter
 
 	isObject()
```
efiShtain, line 24: what's the point of an isObject method which not return a boolean indicating if this is an object?

efiShtain:
Still haven't reviewed everything,

I think @tebeka's comment sums the main point here
we have a tight coupling between the API (protos) and the business models we use inside (the sdk types)

maybe we should think of business models types for internal use and only provide an sdk wrappers for clients

sdk/sdktypes/id.go
```diff
@@ -2,176 +2,103 @@ package sdktypes
...
+	isValider
+	stricter
 
+	// Kind returns the type kind, which is the id prefix.
 	Kind() string
+
+	// Value returns the id value, meaning without the prefix.
 	Value() string
 
 	isID()
```
efiShtain, line 28: what's the purpose of isID()

sdk/sdktypes/id.go
```diff
@@ -2,176 +2,103 @@ package sdktypes
...
 
-func (id *id[T]) isID() {}
+func (i id[T]) isID()         {}
+func (i id[T]) IsValid() bool { var zero id[T]; return i != zero }
+func (i id[T]) Hash() string  { return hash(wrapperspb.String(i.String())) }
 
-func (id *id[T]) String() string {
-	if id == nil {
-		return ""
+func (i id[T]) Strict() error {
```
efiShtain, line 42: what does this do ?
IsValid tells us if this is valid

this only return a generic invalid error without any more information if it is not valid

sdk/sdktypes/id.go
```diff
@@ -2,176 +2,103 @@ package sdktypes
...
-	kind, value, ok := SplitRawID(raw)
-	if !ok {
-		return nil, fmt.Errorf("%w: no delimiter", sdkerrors.ErrInvalidArgument)
-	}
+func (i id[T]) MarshalJSON() ([]byte, error)           { return json.Marshal(i.tid) }
+func (i *id[T]) UnmarshalJSON(data []byte) (err error) { err = json.Unmarshal(data, &i.tid); return }
 
+func newID[ID id[T], T idTraits]() ID {
 	var t T
+	tid := kittehs.Must1(typeid.FromUUIDBytes[typeid.TypeID[T]](t.Prefix(), newUUID()))
```
efiShtain, line 79: we can't keep using panics in production on something which is user related
stuff will fail, the process should not panic because of it

cmd/ak/cmd/builds/delete.go
```diff
@@ -21,7 +21,7 @@ var deleteCmd = common.StandardCommand(&cobra.Command{
 		if err != nil {
 			return err
 		}
-		if b == nil {
+		if !b.IsValid() {
```
ig-ra, line 24: it feels that there should be 2 different errors. 
- b could be not found and b could be invalid
- how b could be invalid if it's found?

sdk/sdktypes/deployment_state.go
```diff
@@ -0,0 +1,40 @@
...
+type DeploymentState struct {
+	enum[deploymentStateTraits, deploymentsv1.DeploymentState]
+}
+
+func deploymentStateFromProto(e deploymentsv1.DeploymentState) DeploymentState {
+	return kittehs.Must1(DeploymentStateFromProto(e))
+}
+
+var (
+	PossibleDeploymentStatesStrings = AllEnumStrings[deploymentStateTraits]()
```
ig-ra: I sugget to rename PossibleDeploymentStatesStrings => DeploymentStateNames
- `Strings` is a (IMO) redundant hint for the type of variable, which could be easily understood/derived w/o stating it in the variable name. `Names` (IMO) is better for the variable name
- are there any `Impossible` StatesStrings? If no, then it just makes the variable name longer w/o providing any additional info

cmd/ak/cmd/events/records/add.go
```diff
@@ -21,22 +22,16 @@ var addCmd = common.StandardCommand(&cobra.Command{
...
 		}
 
-		record, err := sdktypes.EventRecordFromProto(&sdktypes.EventRecordPB{
-			EventId: args[0],
-			State:   sdktypes.ParseEventRecordState(state.String()).ToProto(),
-		})
-		if err != nil {
-			return fmt.Errorf("invalid event record: %w", err)
-		}
+		record := sdktypes.NewEventRecord(eid, kittehs.Must1(sdktypes.ParseEventState(state.String())))
```
ig-ra, line 34: I'm unsure what's going on here.
state is stateString, which is string in zero string, e.g. "". If so, I think there is no need to `state.String()` and for `ParseEventState` as well?

cmd/ak/common/flags.go
```diff
@@ -18,8 +18,8 @@ func AddFailIfError(cmd *cobra.Command) {
 	cmd.Flags().BoolP("fail", "f", false, "fail if error")
 }
 
-func FailIfNotFound[T any](cmd *cobra.Command, what string, v *T) error {
-	if v == nil {
+func FailIfNotFound(cmd *cobra.Command, what string, found bool) error {
```
ig-ra, line 21: - I think in many places we should first check that FailIfNotFound(isZeroObject)
- then FailIfError(obj.IsValid())
unexisting and invalid objects  IMO are two different errors

proto/autokitteh/sessions/v1/session.proto
```diff
@@ -38,16 +38,11 @@ message SessionState {
     values.v1.Value return_value = 3 [(buf.validate.field).required = true];
   }
 
-  google.protobuf.Timestamp t = 1;
-
-  oneof states {
```
ig-ra, line 43: From the PR description I understood that there is no more `oneof`s. Are they enforced anywhere?

backend/internal/db/dbgorm/scheme/scheme.go
```diff
@@ -51,7 +52,7 @@ func ParseBuild(b Build) (sdktypes.Build, error) {
 		CreatedAt: timestamppb.New(b.CreatedAt),
 	})
 	if err != nil {
-		return nil, fmt.Errorf("invalid record: %w", err)
+		return sdktypes.InvalidBuild, fmt.Errorf("invalid record: %w", err)
```
efiShtain, line 55: why not return a reference so you can return nil ?

runtimes/configrt/runtime/run.go
```diff
@@ -28,29 +28,29 @@ func (r *run) getFunc(ctx context.Context, args []sdktypes.Value, kwargs map[str
...
 	v, ok := r.exports[key]
 	if !ok {
-		return sdktypes.NewNothingValue(), nil
+		return sdktypes.Nothing, nil
 	}
 
 	return v, nil
 }
 
 func (r *run) Call(ctx context.Context, v sdktypes.Value, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
```
ig-ra, line 42: could we return pointer instead of invalid type? could u please elaborate on reasons why you prefer objects and not pointers here? How copying objects will/may influence performance?

Makefile
```diff
@@ -81,7 +81,7 @@ golangci_lint=$(shell which golangci-lint)
 $(OUTDIR)/tools/golangci-lint:
 	mkdir -p $(OUTDIR)/tools
 ifeq ($(golangci_lint),)
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(OUTDIR)/tools" v1.56.1
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(OUTDIR)/tools" v1.56.2
```
daabr, line 84: See comment in line 79: please update the CI too.

itayd:
@tebeka the idea of having it wrap the protobufs is that we don't need to describe what we store twice, and then write conversions for it. here it's all the same thing, with the possibility in the future to change it per use case if we want to.

backend/internal/connectionsgrpcsvc/svc.go
itayd: Get returns an Invalid object if it's not found, it's not really an error.

backend/internal/db/dbgorm/dbgorm.go
itayd: @ig-ra is this a mispelling on your end?

go.mod
itayd: we're using only typeid which uses UUIDv7, i think ULID is used for the integrations for connection tokens. @daabr am i correct?

backend/internal/db/dbgorm/sessions_test.go
itayd: personally i like to see all the errors :)

integrations/http/http.go
itayd: i use `Must`s only if there is no way this can fail. I think it's better than just ignoring the error. Does it make sense?

cmd/ak/cmd/builds/delete.go
itayd: it's not found, hence why it's Invalid. It replaced nil.

sdk/sdktypes/object.go
itayd: it's technique to deny implementing more concrete structs outside of the package, since nothing outside the package can implement `isObject`.

sdk/sdktypes/id.go
itayd: ditto

sdk/sdktypes/id.go
itayd: if this fails, something is seriously wrong and we should panic.

sdk/sdktypes/id.go
itayd: it makes sure that the id is not empty in addition for it being valid.

go.mod
daabr: Integrations use "shortuuid" to generate tokens. ULID was suggested previously as a replacement for the general usage of UUIDv4, but we decided to use TypeID (over UUIDv7) instead.

BTW, integrations will continue to use "shortuuid" because they need 100% random values which are **not** K-sortable by time.

backend/internal/db/dbgorm/scheme/scheme.go
itayd: kind of defeats the purpose as the actual object is just a wrap around a pointer. if we pass a pointer, the callee need to check it's not nil and that the object message is not nil. no need for that.

proto/autokitteh/sessions/v1/session.proto
itayd: each message that i removed the one of from is checking in its validation that at least one is specified.

cmd/ak/cmd/events/records/add.go
itayd: yeah we need to clean up this code - there's some mess here. will do it in a different PR.

cmd/ak/common/flags.go
itayd: you can read `IsValid` exactly as `IsNil`. This is essentially what it does. The only "invalid" object is a nil object.

runtimes/configrt/runtime/run.go
itayd: The objects are just wrappers around a pointer. See what I wrote in https://github.com/autokitteh/autokitteh/pull/44#discussion_r1511960443. I guess we'll have to discuss it in sync.

Makefile
itayd: done

cmd/ak/cmd/events/records/add.go
itayd: created https://linear.app/autokitteh/issue/ENG-484/clean-up-cmd-parsing-of-enums

sdk/sdktypes/deployment_state.go
itayd: agreed re the rename. didn't quite follow the Impossible point?

cmd/ak/cmd/builds/delete.go
daabr: @ig-ra see the subtlety in the resolver functions documentation: functions return (`obj, id, err`) - it's possible that they return (`nil, id, nil`) which means that the ID is syntactically correct but not found, and that's why `!obj.IsValid()` means "not found"

cmd/ak/cmd/builds/upload.go
```diff
@@ -25,7 +25,7 @@ var uploadCmd = common.StandardCommand(&cobra.Command{
 		ctx, cancel := common.LimitedContext()
 		defer cancel()
 
-		id, err := builds().Save(ctx, sdktypes.NewBuild(), data)
+		id, err := builds().Save(ctx, sdktypes.InvalidBuild, data)
```
daabr: Semantically, an `Invalid` object shouldn't be used as input.

tests/system/testdata/connections/get.txtar
```diff
@@ -1,26 +1,26 @@
...
-output equals 'connection_id: connection:00000000000000000000000000000002'
+output equals 'connection_id: connection_00000000000000000000000002'
 
 # Negative tests: get nonexistent connection name/ID, with/out --fail flag.
 ak connection get my_project/bad_connection
 return code == 0
 output equals ''
 
-ak connection get connection:deadbeefdeadbeefdeadbeefdeadbeef
+ak connection get connection_00000000000000000000000666
```
daabr: I prefer not to use "666". Firstly, it's not that unrealistic. Secondly, other people may not share our sense of humor. Isn't "deadbeef0deadbeef0deadbeef" valid for usage? Same for all instances in this and other files.

tests/system/testdata/runtimes/run.txtar
```diff
@@ -144,7 +144,7 @@ And quiet sleep and a sweet dream when the long trick’s over.
 {"values":{"num_cats":{"integer":{"v":"4"}},"num_dogs":{"integer":{"v":"2"}}}}
 
 -- multi_single_remote_error.txt --
-Error: create build: build error for "multi/main.star": runtime_build: open("multi/cats.star"): not a directory: cats.star: file does not exist
+Error: create build: rpc: build error for "multi/main.star": runtime_build: open("multi/cats.star"): not a directory: cats.star: file does not exist ([])
```
daabr, line 147: Let's not block this PR, but please open a bug on this specific error: build error should not be treated like RPC errors, not to mention that the `([])` suffix is annoying

sdk/sdktypes/doc.go
```diff
@@ -1,15 +1,33 @@
...
-// 1. Make the data immutable.
-// 2. Make invalid data unrepresantble.
-// There must be no way to instansiate an object or an id outside
-// of the SDK, other than other via SDK methods, such as <Object>FromProto or New<Object>ID.
-//
-// Strict* versions of object functions do not allow for missing required fields.
-// They do allow, though, for missing the entire object entirely (nil input).
-//
-// Strict* versions of ids and handles (non-objects) do not allow for missing values at all.
+// This package contains the types used by the SDK.
```
daabr: https://tip.golang.org/doc/comment:

```suggestion
// Package sdktypes contains all the data types used by the autokitteh SDK.
```

daabr (approved):

cmd/ak/cmd/builds/upload.go
itayd: done

tests/system/testdata/connections/get.txtar
itayd: done

sdk/sdktypes/doc.go
itayd: done

tests/system/testdata/runtimes/run.txtar
itayd: yeah i'm not sure what we should exactly do here. ENG-485

PR #45, opened 2024-03-04
stop autokitteh in an orderly manner at end of system tests

itayd:
this properly terminates temporal as well

### DIFF
--- a/cmd/ak/cmd/up.go
+++ b/cmd/ak/cmd/up.go
@@ -34,11 +34,9 @@ var upCmd = common.StandardCommand(&cobra.Command{
 		}
 
 		<-app.Wait()
+
 		fmt.Println() // End the output with "\n".
 
-		if err := app.Stop(ctx); err != nil {
-			return fmt.Errorf("fx app stop: %w", err)
-		}
 		return nil
 	},
 })
--- a/tests/system/ak_test.go
+++ b/tests/system/ak_test.go
@@ -19,16 +19,19 @@ package systest
 
 import (
 	"context"
+	"fmt"
 	"io"
 	"io/fs"
 	"os"
 	"path/filepath"
 	"strings"
 	"testing"
+	"time"
 )
 
 const (
-	rootDir = "testdata/"
+	rootDir     = "testdata/"
+	stopTimeout = 3 * time.Second
 )
 
 func TestSuite(t *testing.T) {
@@ -86,15 +89,22 @@ func setUpTest(t *testing.T) string {
 		w.Close()
 	}()
 
-	// Start the AK server, but in a goroutine rather than as a separate
+	// Start the AK server, but embedded rather than as a separate
 	// subprocess: to support breakpoint debugging, and measure test coverage.
-	ctx, cancel := context.WithCancel(context.Background())
-	go startAKServer(ctx)
-	t.Cleanup(cancel) // Stop the AK server's goroutine.
+	svc, err := startAKServer(context.Background())
+	if err != nil {
+		t.Fatalf("error: %v", err)
+	}
 
-	akAddr := waitForAKServer(t, combinedOutput)
+	t.Cleanup(func() {
+		ctx, cancel := context.WithTimeout(context.Background(), stopTimeout)
+		defer cancel()
+		if err := svc.Stop(ctx); err != nil {
+			t.Log(fmt.Errorf("fx app stop: %w", err))
+		}
+	})
 
-	return akAddr
+	return svc.Addr()
 }
 
 func runTestSteps(t *testing.T, steps []string, akPath, akAddr string) {
--- a/tests/system/ak_test.go
+++ b/tests/system/ak_test.go
@@ -89,7 +89,7 @@ func setUpTest(t *testing.T) string {
 		w.Close()
 	}()
 
-	// Start the AK server, but embedded rather than as a separate
+	// Start the AK server, but in-process rather than as a separate
 	// subprocess: to support breakpoint debugging, and measure test coverage.
 	svc, err := startAKServer(context.Background())
 	if err != nil {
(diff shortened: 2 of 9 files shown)


backend/basesvc/config.go
```diff
@@ -60,13 +61,15 @@ func LoadConfig(envVarPrefix string, confmapvs map[string]any, yamlPath string)
...
 	}
 
-	// Env variables shuold have the following convention to support hierarchical keys:
-	// PREFIX_PARENT1__CHILD1__REQUIRED_KEY_NAME
-	// meaning heirarchy is separated by `__` and the name it self use one `_``.
-	if err := k.Load(env.Provider(envVarPrefix, "__", func(s string) string {
-		return strings.ToLower(strings.TrimPrefix(s, envVarPrefix))
-	}), nil); err != nil {
-		return nil, fmt.Errorf("load env: %w", err)
+	if envVarPrefix != "" {
```
daabr: Why do we want to allow this condition in production code? Come to think about it, why do we even want this in tests?

tests/system/ak_test.go
```diff
@@ -86,15 +88,19 @@ func setUpTest(t *testing.T) string {
 		w.Close()
 	}()
 
-	// Start the AK server, but in a goroutine rather than as a separate
+	// Start the AK server, but embedded rather than as a separate
```
daabr: I'm not familiar with "embedded" as a well-defined/recognized Go term (in this context). How is this different from a goroutine?

tests/system/ak_test.go
```diff
@@ -86,15 +88,19 @@ func setUpTest(t *testing.T) string {
...
 	// subprocess: to support breakpoint debugging, and measure test coverage.
-	ctx, cancel := context.WithCancel(context.Background())
-	go startAKServer(ctx)
-	t.Cleanup(cancel) // Stop the AK server's goroutine.
+	stop := startAKServer(context.Background())
 
-	akAddr := waitForAKServer(t, combinedOutput)
+	t.Cleanup(func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
daabr: This should be a const, but it also feel very long. Is there a reason for `stop()` to not succeed within 1 second?

tests/system/server.go
```diff
@@ -1,48 +1,35 @@
...
 // subprocess: to support breakpoint debugging, and measure test coverage.
-func startAKServer(ctx context.Context) {
-	cmd.RootCmd.SetArgs([]string{
-		"up",
-		"--config", "db.dsn=sqlite:test.sqlite",
-		"--config", "http.addr=:0",
-		"--mode", "test",
-	})
+func startAKServer(ctx context.Context) (stop func(context.Context) error) {
+	cfg := kittehs.Must1(basesvc.LoadConfig("", map[string]any{
```
daabr: You shouldn't skip the error check here. If it fails we should return an error, not a panic.

More importantly, You're not really running the "up" command, you're simulating it. This is a red flag in integration testing: if someone breaks the "up" command or the root command's `Execute` or `PersistentPreRunE`, we wouldn't detect it.

If we didn't have a choice, I'd accept this, but since the current code works I'm curious why you think this is a better approach.

cmd/ak/cmd/up.go
```diff
@@ -32,11 +32,9 @@ var upCmd = common.StandardCommand(&cobra.Command{
 		}
 
 		<-app.Wait()
+
 		fmt.Println() // End the output with "\n".
 
-		if err := app.Stop(ctx); err != nil {
```
daabr, line 39: Isn't it correct to run `app.Stop` at this point anyway, regardless of system tests?

daabr (commented):
See my questions, let's discuss.

backend/basesvc/config.go
itayd: this is not dealing with this in a fundamental manner, i just don't want the tests to depend on env. that is what this does.

tests/system/ak_test.go
itayd: we just don't use goroutine to start anymore, start returns when the everything is initialized and up after all the hooks are fired. no need for complex wait-for-it-to-finish logic. i guess "embedded" might be confusing here, i meant "linked in with the test, but not running as a go routine".

tests/system/ak_test.go
itayd: yeah i guess, i will move it into a const. i think 1s might be tad too short, I'll do it 3s.

tests/system/server.go
itayd: i think it's better because it simplifies the code a lot. i think that testing if the up command is working or not can be done in a different test.

cmd/ak/cmd/up.go
itayd: not afaict. there is no reason to run stop after everything stops.

tests/system/ak_test.go
daabr: I accept, but I'd rather we keep the original wording. The important thing for me is that the reader of this code understands that this is a (separate) part of our process: not a subprocess, but also not in the same goroutine. "Embedded" hints at that, but isn't a clear term, and after all this is still a gorountine, even if we don't manage its lifecycles.

tests/system/server.go
daabr: Discussed offline. I accept your reasoning for the new code here, but please do use an explicit error check instead of `Must1`. "This Is the Way". A panic here would be very hard to debug based based on how it would look in the test log, and an error check will make the code here more clear, not less.

daabr (approved):
Approved after discussing offline, but please address the last 2 replies I added now.

tests/system/ak_test.go
itayd: changed to in-process

PR #48, opened 2024-03-04
Merge set-resources into build command

daabr:
Merge `set-resources` command into `build` command.

This has 2 reasons:
1. Simpler UX with the CLI (less typing)
2. Minimize chance for race condition (need to eliminate this risk completely in the future, but not yet)

En passant, support multiple paths.


itayd:
we have the set resource end point in projects, shouldn't it be in projects cli as well?

daabr:
> we have the set resource end point in projects, shouldn't it be in projects cli as well?

The CLI primarily is a tool for users, and only after that a tool to exercise our SDK/APIs, so in my opinion it should not necessarily mirror the SDK/APIs.

Specifically in this case, a real user would never call "set-resources" without immediately calling "build" - so they should be a single CLI command. The independent usage of "set-resources" (e.g. by VS Code) is not related to how CLI users interact with AK.

It's an interesting question whether "set-resources" needs to be separate from "build" in the API too, but that's beyond the scope of this PR, and not an urgent question anyway.

itayd:
i think that in the future we wont have set-resources at all, so that's not a big deal.

itayd (approved):

PR #49, opened 2024-03-05
reorg repo structure to prepare for ee features

itayd:
this puts all of backend into internal, exposing only a small surface of it to be used for embedding. this will allow to add an `ee` (temporary name) folder for enterprise features that have access to all the inards of the backend.


tests/system/testdata/manifest/apply.txtar
```diff
@@ -16,12 +16,12 @@ output equals ''
 # Successful dry run of a full project.
 ak manifest apply full.yaml --dry-run -J
 return code == 0
-output equals file expected_on_full_dry_run.json
+output equals file expected_on_full_dry_run.txt
```
daabr, line 19: I agree with changing the extension, but can you also remove the unnecessary `-J` from lines 17 and 22?

internal/xdg/xdg.go
```diff
@@ -1,5 +1,5 @@
 /*
-Package config manages autokitteh's configuration and data directories,
+This manages autokitteh's configuration and data directories,
```
daabr: Package documentation should begin with "Package <Name>..." - see https://tip.golang.org/doc/comment. Bug given the package name change, it should be "Package XDG manages..."

daabr (approved):
Approved with 2 tiny nits.

tests/system/testdata/manifest/apply.txtar
itayd: done

internal/xdg/xdg.go
itayd: done

PR #50, opened 2024-03-05
Shorten command name + add missing build tests

daabr:
This makes sense because resources are a property of projects, as opposed to builds, so this command won't be confused with the "build download" command.

En passant:
- Fix project check (checking the resolver's returned ID is not enough)
- Replace `fmt.Errorf` with `--fail` (commands shouldn't write to stdout/stderr directly anyway)
- Move some tests of `project build` and `project download` to system tests

### DIFF
--- a/cmd/ak/cmd/projects/download.go
+++ b/cmd/ak/cmd/projects/download.go
@@ -1,6 +1,7 @@
 package projects
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"path"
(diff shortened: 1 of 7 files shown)


itayd:
please rebase

cmd/ak/cmd/projects/download.go
```diff
@@ -14,36 +15,34 @@ import (
 
 var outputDirectory string
 
-var downloadResourcesCmd = common.StandardCommand(&cobra.Command{
-	Use:   "download-resources <project name or ID> [--output-dir=...]",
-	Short: "Download the project's resources",
+var downloadCmd = common.StandardCommand(&cobra.Command{
```
efiShtain, line 18: download a project is confusing, we might need it as download project definition or download project manifest or something

cmd/ak/cmd/projects/download.go
```diff
@@ -14,36 +15,34 @@ import (
...
+		if !p.IsValid() {
+			err = errors.New("project not found")
+			return common.NewExitCodeError(common.NotFoundExitCode, err)
 		}
 
 		ctx, cancel := common.LimitedContext()
 		defer cancel()
 
-		resources, err := projects().DownloadResources(ctx, pid)
+		rs, err := projects().DownloadResources(ctx, pid)
```
efiShtain: why do you need to change a readable name like resources to rs?

cmd/ak/cmd/projects/download.go
```diff
@@ -14,36 +15,34 @@ import (
...
 		if err != nil {
-			return fmt.Errorf("download resources: %w", err)
+			return err
 		}
-
-		if len(resources) == 0 {
-			fmt.Println("no resources found")
-			return nil
+		if err := common.FailIfNotFound(cmd, "resources", len(rs) > 0); err != nil {
+			return err
```
efiShtain: why fail?
project can have no resources, it's not an error
also, I'm not sure 
```go
common.FailIfNotFound(cmd, "resources", len(rs) > 0)
```
is more readable than
```go
if len(resources) == 0 {
```

efiShtain: @daabr Please open a ticket, we should resolve it somehow
not readable (I know its not on you :) )

efiShtain (approved):

itayd (approved):

cmd/ak/cmd/projects/download.go
daabr: https://linear.app/autokitteh/issue/ENG-506/clean-up-not-found-error-handling-across-all-cli-commands

PR #57, opened 2024-03-06
ENG-475: describe build endpoint

itayd:
- add describe endpoint to builds service, returns a json marshalling of the build sans the build compiled data content. there is currently no protobuf description of the actual build file, hence the json return, which should be ok for ui purposes. not sure if we need an actual proto defintion of the build file.
- makes sure builds being uploaded are actually a build.
- pull out sdkbuildfile from sdkruntimes.


cmd/ak/cmd/builds/describe.go
```diff
@@ -0,0 +1,41 @@
...
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var describeCmd = common.StandardCommand(&cobra.Command{
+	Use:     "describe <build ID> [--fail]",
+	Short:   "Get build details from server",
+	Aliases: []string{"g"},
```
daabr: "g" is already used by the "get" command. Maybe "desc", or no alias at all?

cmd/ak/cmd/builds/describe.go
```diff
@@ -0,0 +1,41 @@
...
+
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var describeCmd = common.StandardCommand(&cobra.Command{
+	Use:     "describe <build ID> [--fail]",
+	Short:   "Get build details from server",
```
daabr: This is the exact description of the "get" command. Need to distinguish between them. Just a suggestion to get you started:

Get - "Get build creation details"
Describe - "Get build structure details"

internal/backend/builds/builds.go
```diff
@@ -56,3 +63,21 @@ func (b *Builds) Save(ctx context.Context, build sdktypes.Build, data []byte) (s
...
 	return b.DB.DeleteBuild(ctx, id)
 }
+
+func (b *Builds) Describe(ctx context.Context, bid sdktypes.BuildID) (*sdkbuildfile.BuildFile, error) {
+	r, err := b.Download(ctx, bid)
+	if err != nil {
+		return nil, fmt.Errorf("download: %w", err)
+	}
+
+	defer r.Close()
```
daabr: `defer` should be as close as possible to the thing it's attached to.
```suggestion
	}
	defer r.Close()
```

daabr (commented):
The major issue in this PR is that the schema of the JSON output of Describe is undefined, and therefore hard to guess and hard to safeguard. But I'm ignoring this issue for this PR.

There are a few tiny nits - please address them.

But that thing I want to review before approving is the integration tests - it makes much more sense to implement them as `txtar` files instead of `txt`.

cmd/ak/cmd/builds/describe.go
itayd: removed

cmd/ak/cmd/builds/describe.go
itayd: changed, ptal

internal/backend/builds/builds.go
itayd: fixed

daabr (approved):

PR #61, opened 2024-03-07
allow to cancel sessions, add tests

itayd:
- fix stop session work, with and without force
- allow to create sessions without a deployment, based on build alone. these will have no env, no connections and no triggers ("one shot").
- add sessions test
- add to cli ability to watch sessions in various ways (needed for testing and usually helpful)
- system test to ignore tests that end with '.disabled'

This currently does not pass tests because of @ig-ra 's change re foreign keys. Will discuss in sync. (it worked before rebase)

NOTE: me and @daabr found flakiness in system testing that come out from the fact that it does not run each test in separate process. This will be dealt with in another PR (see ENG-447, ENG-483).


internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -323,7 +323,8 @@ func ParseSessionCallAttemptComplete(c SessionCallAttempt) (d sdktypes.SessionCa
 
 type Session struct {
 	SessionID        string `gorm:"primaryKey"`
-	DeploymentID     string `gorm:"index;foreignKey:DeploymentID"`
+	DeploymentID     string `gorm:"index"`
+	BuildID          string `gorm:"index"`
```
ig-ra, line 327: do we want to cascade delete builds on session delete?
I think that it would be better and logical to remove buildID from deployment then and keep it in session only

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -103,12 +104,23 @@ func (w *sessionWorkflow) addPrint(ctx workflow.Context, print string) {
 func (w *sessionWorkflow) updateState(ctx workflow.Context, state sdktypes.SessionState) {
 	w.z.Debug("update state", zap.Any("state", state))
 
+	if ctx.Err() == workflow.ErrCanceled {
+		goCtx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
```
ig-ra, line 108: should be defined elsewhere as a constant that we could change.

internal/backend/temporalclient/config.go
```diff
@@ -47,10 +47,9 @@ var Configs = configset.Set[Config]{
 		},
 	},
 	Test: &Config{
-		CheckHealthInterval: time.Minute,
-		CheckHealthTimeout:  10 * time.Second,
-		LogLevel:            zap.NewAtomicLevelAt(zapcore.WarnLevel),
-
+		CheckHealthInterval:  time.Minute,
+		CheckHealthTimeout:   10 * time.Second,
```
ig-ra, line 51: should be defined elsewhere, and all constants above as well

sdk/sdktypes/session.go
```diff
@@ -30,11 +34,13 @@ func (SessionTraits) Validate(m *SessionPB) error {
 }
 
 func (SessionTraits) StrictValidate(m *SessionPB) error {
+	if m.BuildId == "" && m.DeploymentId == "" {
```
ig-ra, line 37: just one more comment on the same topic. I think maybe we should remove buildID from deployment. Then session will always have buildID but not always deployment. Model and concept of what is deployment should be redefined as well

ig-ra (changes requested):

PR #62, opened 2024-03-07
ENG-504: autogenerate project name if not given

itayd:
also in test mode, predictively generate names


internal/backend/projects/projects.go
```diff
@@ -29,6 +29,10 @@ func New(p Projects) sdkservices.Projects { return &p }
 func (ps *Projects) Create(ctx context.Context, project sdktypes.Project) (sdktypes.ProjectID, error) {
 	project = project.WithNewID()
 
+	if !project.Name().IsValid() {
+		project = project.WithName(sdktypes.NewRandomSymbol())
```
itayd, line 33: add random suffix just to be sure

itayd: added random 4 digits suffix

cmd/ak/cmd/projects/create.go
```diff
@@ -9,14 +9,20 @@ import (
...
-	Args:    cobra.ExactArgs(1),
+	Args:    cobra.NoArgs,
 
 	RunE: func(cmd *cobra.Command, args []string) error {
-		p, err := sdktypes.ProjectFromProto(&sdktypes.ProjectPB{Name: args[0]})
+		if _, err := sdktypes.ParseSymbol(name); err != nil {
+			return fmt.Errorf("invalid project name: %w", err)
+		}
+
+		p, err := sdktypes.ProjectFromProto(&sdktypes.ProjectPB{Name: name})
```
efiShtain: please add sdktypes.NewProject(name string) (sdktypes.Project,error) to sdktypes
we added this layer so no one would have to use protos directly, 
but we do use them so it kinds of miss the purpose

cmd/ak/cmd/projects/create.go
```diff
@@ -9,14 +9,20 @@ import (
...
-	Use:     "create <project name>",
+	Use:     "create [--name project-name]",
 	Short:   "Create new project",
 	Aliases: []string{"c"},
-	Args:    cobra.ExactArgs(1),
+	Args:    cobra.NoArgs,
 
 	RunE: func(cmd *cobra.Command, args []string) error {
-		p, err := sdktypes.ProjectFromProto(&sdktypes.ProjectPB{Name: args[0]})
+		if _, err := sdktypes.ParseSymbol(name); err != nil {
```
efiShtain: we don't really need this check
hide it in sdktypes.NewProject (see next comment) and it should fail validation on name

cmd/ak/cmd/projects/create.go
```diff
@@ -29,7 +35,20 @@ var createCmd = common.StandardCommand(&cobra.Command{
...
+		if name != "" {
+			common.RenderKV("project_id", id)
+			return nil
+		}
+
+		if p, err = projects().GetByID(ctx, id); err != nil {
+			return fmt.Errorf("get project: %w", err)
+		}
+
+		common.RenderKV("project", p)
```
efiShtain: line 47 and line 39 gives different output structure on success (project_id vs project)

as I see it, the cli output is an API
we should have one output structure to the API not based on the input (name == "" or not)

internal/catnames/catnames.go
```diff
@@ -0,0 +1,27 @@
...
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+)
+
+var caser = cases.Title(language.English)
+
+func NewGenerator(pick func(int) int) func() string {
+	if pick == nil {
+		pick = rand.New(rand.NewSource(time.Now().UnixNano())).Intn
+	}
```
efiShtain, line 12: yagni? 
or for testing purposes ?
if for testing, maybe using fx here is better ?

cmd/ak/cmd/projects/create.go
itayd: done

internal/catnames/catnames.go
itayd: catnames is pretty standalone and lean. this is for testing pruprosed so we can plug the sequential generator in sdktypes into it.

cmd/ak/cmd/projects/create.go
itayd: i changed the code, now we're feeding it into NewProject.

cmd/ak/cmd/projects/create.go
itayd: you know what, i will just return the id.

cmd/ak/cmd/projects/create.go
```diff
@@ -9,18 +9,22 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+var name string
+
 var createCmd = common.StandardCommand(&cobra.Command{
-	Use:     "create <project name>",
+	Use:     "create [--name project-name]",
```
daabr, line 15: Why did make the project name a flag? The usual behavior for project commands is to treat it as an arg.

internal/catnames/catnames.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func NewGenerator(pick func(int) int) func() string {
+	if pick == nil {
+		pick = rand.New(rand.NewSource(time.Now().UnixNano())).Intn
+	}
+
+	return func() string {
+		return fmt.Sprintf(
+			"%s the %s",
```
daabr: Better english: reverse the order and drop the "the". That's how all the other animal/adjective generators work too.

internal/catnames/data.go
```diff
@@ -0,0 +1,772 @@
+package catnames
+
+import (
+	"golang.org/x/text/cases"
+	"golang.org/x/text/language"
+)
+
+func init() {
```
daabr: This is wasteful. Why not do it once and fix the file itself?

cmd/ak/cmd/projects/create.go
itayd: because it's optional now, so it makes sense.

internal/catnames/catnames.go
itayd: done

internal/catnames/data.go
itayd: done

internal/backend/projects/projects.go
daabr: "How much random is random enough" rarely has a good answer. I don't like random junk in autogenerated names because it's junk, but it doesn't really matter anyway. A *robust* solution would be to regenerate or append an auto-incrementing ID if there's a duplicate, but it's not necessary for this PR.

daabr (approved):

internal/backend/projects/projects.go
itayd: yeah we could be saints, but it's not needed here. i actually like adding the number suffix since it makes it even more clear that this stuff is random.

PR #64, opened 2024-03-07
program error contains a value, not a message

itayd:
a program error can now contain a value instead of just a message. will allow integrations and other functions to return more detailed and structured errors (for example, HTTP errors with codes etc).

also splits module from program proto to prevent cyclical dependencies.

based on #63.


runtimes/starlarkrt/runtime/builtins.go
```diff
@@ -20,17 +22,69 @@ var builtins = map[string]func(thread *starlark.Thread, fn *starlark.Builtin, ar
 
 		value, err := starlark.Call(thread, fn, args[1:], kwargs)
 		if err != nil {
+			if perr, ok := sdktypes.FromError(err); ok {
+				v := perr.Value()
+				vctx := values.Context{}
+				if slv, cerr := vctx.ToStarlarkValue(v); cerr == nil {
```
efiShtain, line 28: if cerr is not nil ?
you want to return just the original err ?
if it's intentional you can resolve this thread

runtimes/starlarkrt/runtime/builtins.go
```diff
@@ -20,17 +22,69 @@ var builtins = map[string]func(thread *starlark.Thread, fn *starlark.Builtin, ar
...
+
+		vctx := values.Context{}
+
+		if len(args) == 1 {
+			v, err := vctx.FromStarlarkValue(args[0])
+			if err != nil {
+				return nil, fmt.Errorf("cannot convert value from starlark: %w", err)
+			}
+			return nil, sdktypes.NewProgramError(v, nil, nil).ToError()
+		} else if len(args) > 1 {
```
efiShtain: since you are returning from the if condition, please remove the else
it makes code less indented and more readable

runtimes/starlarkrt/runtime/builtins.go
```diff
@@ -20,17 +22,69 @@ var builtins = map[string]func(thread *starlark.Thread, fn *starlark.Builtin, ar
...
+			})
+			if err != nil {
+				return nil, fmt.Errorf("cannot convert values from starlark: %w", err)
+			}
+
+			v := sdktypes.NewDictValueFromStringMap(vs)
+			return nil, sdktypes.NewProgramError(v, nil, nil).ToError()
+		}
+
+		return nil, errors.New("user triggered error")
```
efiShtain: move this to the top after the check that not both kwargs and args len > 0
and then you can remove this last condition as it is implicit
again less indentation

pseudo code
so if both > 0 fail
if both == 0 fail

if args == 1 return do something
if args > 1 return do something 2
do something 3 // implicit len(args)=0 and kwargs len > 0

sdk/sdktypes/program_error.go
```diff
@@ -56,11 +61,11 @@ func StrictProgramErrorFromProto(m *ProgramErrorPB) (ProgramError, error) {
 	return Strict(ProgramErrorFromProto(m))
 }
 
-func NewProgramError(msg string, callstack []CallFrame, extra map[string]string) ProgramError {
+func NewProgramError(v Value, callstack []CallFrame, extra map[string]string) ProgramError {
 	return kittehs.Must1(ProgramErrorFromProto(
```
efiShtain, line 65: this should really move to the sdktypes part as sdktypes.NewProgramError(msg, extra,value,callstack) and we should stop using proto directly in backend

efiShtain (approved):
nits but I don't see reason to stop the PR

runtimes/starlarkrt/runtime/builtins.go
itayd: if cerr is not nil it means that the value is unconvertable to starlark, so i just string it into the error, no info lost.

runtimes/starlarkrt/runtime/builtins.go
itayd: done

runtimes/starlarkrt/runtime/builtins.go
itayd: rearranged to be more clear

sdk/sdktypes/program_error.go
itayd: not sure i follow - there is no exposing of protobuf here? it is sdktypes.Value.

efiShtain: this part, proto stuff
```go
&ProgramErrorPB{
			Message:   msg,
			Extra:     extra,
			Value:     v.ToProto(),
			Callstack: kittehs.Transform(callstack, func(f CallFrame) *CallFramePB { return f.ToProto() }),
		},
```

PR #69, opened 2024-03-07
Configurable secrets manager

daabr:
Fixes ENG-436 and ENG-145

### DIFF
--- a/internal/backend/secrets/aws.go
+++ b/internal/backend/secrets/aws.go
@@ -11,15 +11,14 @@ import (
 	"go.uber.org/zap"
 )
 
-type amazonSecrets struct {
+type awsSecrets struct {
 	client  *secretsmanager.Client
 	logger  *zap.Logger
 	timeout time.Duration
 }
 
-// NewAmazonSecrets initializes a client connection to Amazon
-// Secrets Manager (https://aws.amazon.com/secrets-manager/).
-func NewAmazonSecrets(l *zap.Logger, c *Config) (Secrets, error) {
+// NewAWSSecrets initializes a client connection to AWS Secrets Manager.
+func NewAWSSecrets(l *zap.Logger, c *Config) (Secrets, error) {
 	cfg, err := config.LoadDefaultConfig(context.Background())
 	if err != nil {
 		l.Error("AWS config initialization", zap.Error(err))
@@ -27,22 +26,17 @@ func NewAmazonSecrets(l *zap.Logger, c *Config) (Secrets, error) {
 	}
 
 	client := secretsmanager.NewFromConfig(cfg)
-	timeout := parseTimeout(l, c.TimeoutDuration)
-	return &amazonSecrets{client: client, logger: l, timeout: timeout}, nil
+	return &awsSecrets{client: client, logger: l, timeout: c.Timeout}, nil
 }
 
 // The data size limit is 64 KiB, according to this link:
 // https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_CreateSecret.html
-func (s *amazonSecrets) Set(scope, name string, data map[string]string) error {
+func (s *awsSecrets) Set(scope, name string, data map[string]string) error {
 	ctx, cancel := limitedContext(s.timeout)
 	defer cancel()
 
 	d, err := json.MarshalIndent(data, "", "  ")
 	if err != nil {
-		s.logger.Error("JSON marshal",
-			zap.String("name", secretPath(scope, name)),
-			zap.Error(err),
-		)
 		return err
 	}
 
@@ -51,16 +45,12 @@ func (s *amazonSecrets) Set(scope, name string, data map[string]string) error {
 		SecretString: aws.String(string(d)),
 	})
 	if err != nil {
-		s.logger.Error("AWS Secrets Manager CreateSecret",
-			zap.String("name", secretPath(scope, name)),
-			zap.Error(err),
-		)
 		return err
 	}
 	return nil
 }
 
-func (s *amazonSecrets) Get(scope, name string) (map[string]string, error) {
+func (s *awsSecrets) Get(scope, name string) (map[string]string, error) {
 	ctx, cancel := limitedContext(s.timeout)
 	defer cancel()
 
@@ -68,26 +58,17 @@ func (s *amazonSecrets) Get(scope, name string) (map[string]string, error) {
 		SecretId: aws.String(secretPath(scope, name)),
 	})
 	if err != nil {
-		s.logger.Error("AWS Secrets Manager GetSecretValue",
-			zap.String("name", secretPath(scope, name)),
-			zap.Error(err),
-		)
 		return nil, err
 	}
 
 	data := make(map[string]string)
 	if err := json.Unmarshal([]byte(*sec.SecretString), &data); err != nil {
-		s.logger.Error("JSON unmarshal",
-			zap.String("name", secretPath(scope, name)),
-			zap.Error(err),
-		)
 		return nil, err
 	}
-
 	return data, nil
 }
 
-func (s *amazonSecrets) Append(scope, name, token string) error {
+func (s *awsSecrets) Append(scope, name, token string) error {
 	ctx, cancel := limitedContext(s.timeout)
 	defer cancel()
 
@@ -99,10 +80,6 @@ func (s *amazonSecrets) Append(scope, name, token string) error {
 	data[token] = time.Now().UTC().Format(time.RFC3339)
 	d, err := json.MarshalIndent(data, "", "  ")
 	if err != nil {
-		s.logger.Error("JSON marshal",
-			zap.String("name", secretPath(scope, name)),
-			zap.Error(err),
-		)
 		return err
 	}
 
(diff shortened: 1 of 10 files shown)


internal/backend/secrets/svc.go
```diff
@@ -3,29 +3,54 @@ package secrets
...
+var Configs = configset.Set[Config]{
+	Default: &Config{},
+}
+
+var defaultTimeout = 1 * time.Minute
+
+func New(l *zap.Logger, cfg *Config) (sdkservices.Secrets, error) {
+	var impl Secrets
+	var err error
+
```
itayd, line 32: ParseURL

internal/backend/secrets/svc.go
```diff
@@ -3,29 +3,54 @@ package secrets
...
 	"go.autokitteh.dev/autokitteh/internal/xdg"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 )
 
-type secrets struct {
-	impl   Secrets
-	logger *zap.Logger
+type Config struct {
+	ManagerURL      string `koanf:"manager_url"`
+	TimeoutDuration string `koanf:"timeout_duration"`
```
itayd: time.Duration

internal/backend/secrets/vault.go
```diff
@@ -12,55 +11,93 @@ import (
...
+	timeout time.Duration
 }
 
-// NewVaultSecrets initializes a client connection to HashiCorp Vault
-// (https://www.vaultproject.io/).
-func NewVaultSecrets(l *zap.Logger) (Secrets, error) {
-	config := vault.DefaultConfig()
-	client, err := vault.NewClient(config)
+// NewVaultSecrets initializes a client connection to
+// HashiCorp Vault (https://www.vaultproject.io/).
```
itayd:
```suggestion
// NewVaultSecrets initializes a client connection to HashiCorp Vault.
```

internal/backend/secrets/vault.go
```diff
@@ -12,55 +11,93 @@ import (
...
+func NewVaultSecrets(l *zap.Logger, c *Config) (Secrets, error) {
+	cfg := vault.DefaultConfig()
+	cfg.Address = c.ManagerURL
+
+	client, err := vault.NewClient(cfg)
 	if err != nil {
-		l.Error("Vault client initialization error",
-			zap.Error(err),
-		)
+		l.Error("Vault client initialization", zap.Error(err))
```
itayd, line 27:
```suggestion
```

internal/backend/secrets/svc.go
daabr: I had to add a type field, which made parsing the URL unnecessary anyway.

internal/backend/secrets/svc.go
daabr: done

internal/backend/secrets/vault.go
daabr: done

internal/backend/secrets/vault.go
daabr: done

internal/backend/secrets/aws.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+
+	client := secretsmanager.NewFromConfig(cfg)
+	return &awsSecrets{client: client, logger: l, timeout: c.Timeout}, nil
+}
+
+// The data size limit is 64 KiB, according to this link:
+// https://docs.aws.amazon.com/secretsmanager/latest/apireference/API_CreateSecret.html
+func (s *awsSecrets) Set(scope, name string, data map[string]string) error {
+	ctx, cancel := limitedContext(s.timeout)
```
itayd: not sure why not just passing the context from the caller?

internal/backend/secrets/svc.go
itayd: it could have just been the scheme, but oh well

internal/backend/secrets/aws.go
daabr: 1. Contexts are an implementation-specific detail, not all implementations use them
2. The context for these calls needs to be stricter (short timeout by default, even if the caller is a long-running runtime workflow)

daabr: done

internal/backend/secrets/aws.go
```diff
@@ -0,0 +1,108 @@
...
+	defer cancel()
+
+	sec, err := s.client.GetSecretValue(ctx, &secretsmanager.GetSecretValueInput{
+		SecretId: aws.String(secretPath(scope, name)),
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	data := make(map[string]string)
```
efiShtain, line 64: I think secret can be main 
Or just plain string
Do we support it? Want to support it?

Binary is out of scope

internal/backend/secrets/aws.go
```diff
@@ -0,0 +1,108 @@
...
+	return nil
+}
+
+func (s *awsSecrets) Delete(ctx context.Context, scope, name string) error {
+	ctx, cancel := limitContext(ctx, s.timeout)
+	defer cancel()
+
+	_, err := s.client.DeleteSecret(ctx, &secretsmanager.DeleteSecretInput{
+		SecretId:                   aws.String(secretPath(scope, name)),
+		ForceDeleteWithoutRecovery: aws.Bool(true),
```
efiShtain, line 102: Why?
This should probably be a flag in production for force delete 

I agree in dev it is better to force delete

internal/backend/secrets/secrets.go
```diff
@@ -23,17 +24,17 @@ type Secrets interface {
...
-	Set(scope, name string, data map[string]string) error
+	Set(ctx context.Context, scope, name string, data map[string]string) error
 	// Get retrieves the key-value data associated with a named secret.
 	// If the name does not exist then we return nothing, not an error.
-	Get(scope, name string) (map[string]string, error)
+	Get(ctx context.Context, scope, name string) (map[string]string, error)
 	// Append a token (as a key, with the current timestamp as the value)
 	// to an existing secret, or create it if it doesn't exist already.
-	Append(scope, name, token string) error
+	Append(ctx context.Context, scope, name, token string) error
```
efiShtain, line 33: Why do we need append?
Is it thread safe? Multiple instance safe?

internal/backend/secrets/svc.go
```diff
@@ -3,36 +3,62 @@ package secrets
...
 }
 
-func New(l *zap.Logger) (sdkservices.Secrets, error) {
-	// TODO(ENG-145): Allow the user to select the implementation via CLI.
-	impl, err := NewFileSecrets(l, xdg.DataHomeDir())
+func New(l *zap.Logger, cfg *Config) (sdkservices.Secrets, error) {
+	var impl Secrets
+	var err error
+
+	switch cfg.Type {
```
efiShtain, line 33: You can use fx to provide the relevant implementation instead of this switch condition

Wdyt?

internal/backend/secrets/svc.go
```diff
@@ -99,6 +125,10 @@ func (s secrets) List(ctx context.Context, scope, key string) ([]string, error)
 	return maps.Keys(data), nil
 }
 
+func limitContext(ctx context.Context, timeout time.Duration) (context.Context, context.CancelFunc) {
```
efiShtain, line 128: Don't you have a generic fund that already does that?

internal/backend/secrets/aws.go
daabr: Not sure what you mean by "main". Plain strings are technically possible, but secrets managers are usually used to save JSON (`map[string]any`) and not just arbitrary strings, and we use a subset of JSON: `map[string]string`. Specifically, we store multiple strings per connection, not just a single one. That's why the API is defined this way, and the AWS implementation simply conforms to that.

internal/backend/secrets/aws.go
daabr: For security reasons, user secrets need to be completely purged on demand, unlike AK entities (e.g. connections, which use secrets) which may be soft-deleted. In other words, it's OK to keep an unused/unusable AK entity in the DB for book-keeping/auditing purposes, but it's not OK to keep an unused secret.

We do the same in Vault - we completely destroy all versions and metadata of a secret when it's deleted.

internal/backend/secrets/secrets.go
daabr: `Append` is used specifically for reverse-mapping of connections. It's thread-safe and multiple-instance-safe in all non-AWS implementations, but not multiple-instance-safe in AWS. I opened a ticket to fix this later: ENG-513.

internal/backend/secrets/svc.go
daabr: I really prefer to rely on explicit checking and initialization which is traceable, and less on code injection that isn't easy to trace.

internal/backend/secrets/svc.go
daabr: There's something similar in `cmd/ak/common` but it's not exactly the same: this func takes an existing context and limits it, instead of creating a new one. We could think about a common implementation for both in the future, but it's beyond the scope of this PR and not a high priority at this time.

internal/backend/secrets/aws.go
efiShtain: I meant json, the auto correct changed it 

So this implementation is for connections only?
No other secrets?

daabr: Yes

daabr: See https://github.com/autokitteh/autokitteh/pull/69#discussion_r1520067876

itayd (approved):

PR #71, opened 2024-03-08
ENG-508: Add secrets table to DB

daabr:


internal/backend/db/dbgorm/secrets.go
```diff
@@ -0,0 +1,71 @@
...
+		return err
+	}
+
+	data[token] = time.Now().UTC().Format(time.RFC3339)
+	jsonData, err := json.Marshal(data)
+	if err != nil {
+		return err
+	}
+
+	result := db.db.WithContext(ctx).Model(&scheme.Secret{}).Where("name = ?", name).Update("data", jsonData)
```
efiShtain, line 57: You can just try to update and if rows affected is zero it means secret does not exists 

You can save a call to db

efiShtain (approved):
I didn't see any usage so it's ok to add it

Please see the point where you can save one db call on update

internal/backend/db/dbgorm/secrets.go
daabr: Yes, but unfortunately I need `Get()`. This function adds the `token` input into the existing data (see line 51), it doesn't overwrite existing data.

That said, this reminds me that I need to make sure if this function is expected to succeed in the data is not found, i.e. create a new row is it doesn't exist. I'll update the unit tests either way.

As for using this table - I'll reference this PR in the next PR which will do that. Not merging it yet, because it's stacked over #69.

efiShtain: So if this is a json data type and you want to append stuff you can still do it with one sql command in Postgres
No need to fetch the data before (same for delete if it's by key)

PR #74, opened 2024-03-09
ENG-477: feat(db): delete projects

ig-ra:
ENG-477

### DIFF
--- a/cmd/ak/cmd/deployments/delete.go
+++ b/cmd/ak/cmd/deployments/delete.go
@@ -20,7 +20,7 @@ var deleteCmd = common.StandardCommand(&cobra.Command{
 			return common.FailIfError(cmd, err, "deployment")
 		}
 
-		if err := common.FailIfNotFound(cmd, "deployment", d.IsValid()); err != nil { // test d != nil
+		if err := common.FailIfNotFound(cmd, "deployment", d.IsValid()); err != nil {
 			return err
 		}
 
--- a/cmd/ak/cmd/projects/delete.go
+++ b/cmd/ak/cmd/projects/delete.go
@@ -1,0 +1,40 @@
+package projects
+
+import (
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var deleteCmd = common.StandardCommand(&cobra.Command{
+	Use:     "delete <project ID> [--fail]",
+	Short:   "Delete inactive project",
+	Aliases: []string{"d"},
+	Args:    cobra.ExactArgs(1),
+
+	RunE: func(cmd *cobra.Command, args []string) error {
+		r := resolver.Resolver{Client: common.Client()}
+		p, id, err := r.ProjectNameOrID(args[0])
+		if err != nil {
+			return common.FailIfError(cmd, err, "project")
+		}
+
+		if err := common.FailIfNotFound(cmd, "project", p); err != nil { // test p != nil
+			return err
+		}
+
+		ctx, cancel := common.LimitedContext()
+		defer cancel()
+		err = projects().Delete(ctx, id)
+		if err != nil { // report any other than "not found" error (which was handled above)
+			return common.ToExitCodeError(err, "delete project")
+		}
+		return nil
+	},
+})
+
+func init() {
+	// Command-specific flags.
+	common.AddFailIfError(deleteCmd)
+}
--- a/cmd/ak/cmd/projects/delete.go
+++ b/cmd/ak/cmd/projects/delete.go
@@ -20,17 +20,15 @@ var deleteCmd = common.StandardCommand(&cobra.Command{
 			return common.FailIfError(cmd, err, "project")
 		}
 
-		if err := common.FailIfNotFound(cmd, "project", p); err != nil { // test p != nil
+		if err := common.FailIfNotFound(cmd, "project", p.IsValid()); err != nil {
 			return err
 		}
 
 		ctx, cancel := common.LimitedContext()
 		defer cancel()
+
 		err = projects().Delete(ctx, id)
-		if err != nil { // report any other than "not found" error (which was handled above)
-			return common.ToExitCodeError(err, "delete project")
-		}
-		return nil
+		return common.ToExitCodeError(err, "delete project")
 	},
 })
 
--- a/cmd/ak/cmd/projects/delete.go
+++ b/cmd/ak/cmd/projects/delete.go
@@ -9,7 +9,7 @@ import (
 
 var deleteCmd = common.StandardCommand(&cobra.Command{
 	Use:     "delete <project ID> [--fail]",
-	Short:   "Delete inactive project",
+	Short:   "Delete project",
 	Aliases: []string{"d"},
 	Args:    cobra.ExactArgs(1),
 
--- a/internal/backend/db/dbgorm/deployments.go
+++ b/internal/backend/db/dbgorm/deployments.go
@@ -47,22 +47,31 @@ func (db *gormdb) GetDeployment(ctx context.Context, id sdktypes.DeploymentID) (
 }
 
 func (db *gormdb) deleteDeployment(ctx context.Context, deploymentID string) error {
-	// delete deployment and relevant sessions
 	return db.deleteDeploymentsAndDependents(ctx, []string{deploymentID})
 }
 
-func (db *gormdb) deleteDeploymentsAndDependents(ctx context.Context, ids []string) error {
+// delete deployments and relevant sessions
+func (db *gormdb) deleteDeploymentsAndDependents(ctx context.Context, depIDs []string) error {
 	// NOTE: should be transactional
 
-	if len(ids) == 0 {
+	if len(depIDs) == 0 {
 		return nil
 	}
 	gormDB := db.db.WithContext(ctx)
+	s := scheme.Session{}
+	b := scheme.Build{}
+	d := scheme.Deployment{}
 
-	if err := gormDB.Where("deployment_id IN ?", ids).Delete(&scheme.Session{}).Error; err != nil {
+	if err := gormDB.Delete(&s, "deployment_id IN ?", depIDs).Error; err != nil {
 		return err
 	}
-	return gormDB.Where("deployment_id IN ?", ids).Delete(&scheme.Deployment{}).Error
+
+	buildIDs := gormDB.Model(&d).Select("build_id").Where("deployment_id IN (?)", depIDs)
+	if err := gormDB.Delete(&b, "build_id IN (?)", buildIDs).Error; err != nil {
+		return err
+	}
+
+	return gormDB.Delete(&d, "deployment_id IN ?", depIDs).Error
 }
 
 func (db *gormdb) DeleteDeployment(ctx context.Context, deploymentID sdktypes.DeploymentID) error {
--- a/internal/backend/db/dbgorm/deployments.go
+++ b/internal/backend/db/dbgorm/deployments.go
@@ -58,20 +58,17 @@ func (db *gormdb) deleteDeploymentsAndDependents(ctx context.Context, depIDs []s
 		return nil
 	}
 	gormDB := db.db.WithContext(ctx)
-	s := scheme.Session{}
-	b := scheme.Build{}
-	d := scheme.Deployment{}
 
-	if err := gormDB.Delete(&s, "deployment_id IN ?", depIDs).Error; err != nil {
+	if err := gormDB.Delete(&scheme.Session{}, "deployment_id IN ?", depIDs).Error; err != nil {
 		return err
 	}
 
-	buildIDs := gormDB.Model(&d).Select("build_id").Where("deployment_id IN (?)", depIDs)
-	if err := gormDB.Delete(&b, "build_id IN (?)", buildIDs).Error; err != nil {
+	buildIDs := gormDB.Model(&scheme.Deployment{}).Select("build_id").Where("deployment_id IN (?)", depIDs)
+	if err := gormDB.Delete(&scheme.Build{}, "build_id IN (?)", buildIDs).Error; err != nil {
 		return err
 	}
 
-	return gormDB.Delete(&d, "deployment_id IN ?", depIDs).Error
+	return gormDB.Delete(&scheme.Deployment{}, "deployment_id IN ?", depIDs).Error
 }
 
 func (db *gormdb) DeleteDeployment(ctx context.Context, deploymentID sdktypes.DeploymentID) error {
--- a/internal/backend/db/dbgorm/deployments.go
+++ b/internal/backend/db/dbgorm/deployments.go
@@ -57,13 +57,13 @@ func (db *gormdb) deleteDeploymentsAndDependents(ctx context.Context, ids []stri
 		return nil
 	}
 
-	dbb := db.db.WithContext(ctx)
 	return db.transaction(ctx, func(tx *tx) error {
-		if err := dbb.Where("deployment_id IN ?", ids).Delete(&scheme.Session{}).Error; err != nil {
+		db := tx.db.WithContext(ctx)
+		if err := db.Where("deployment_id IN ?", ids).Delete(&scheme.Session{}).Error; err != nil {
 			return err
 		}
 
-		return dbb.Where("deployment_id IN ?", ids).Delete(&scheme.Deployment{}).Error
+		return db.Where("deployment_id IN ?", ids).Delete(&scheme.Deployment{}).Error
 	})
 }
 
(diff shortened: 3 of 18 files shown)


cmd/ak/cmd/deployments/delete.go
```diff
@@ -20,7 +20,7 @@ var deleteCmd = common.StandardCommand(&cobra.Command{
 			return common.FailIfError(cmd, err, "deployment")
 		}
 
-		if err := common.FailIfNotFound(cmd, "deployment", d.IsValid()); err != nil { // test d != nil
+		if err := common.FailIfNotFound(cmd, "deployment", d.IsValid()); err != nil {
```
efiShtain, line 23: could you elaborate more on this FailIfNotFound method ? it is not very clear

cmd/ak/cmd/projects/delete.go
```diff
@@ -0,0 +1,38 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var deleteCmd = common.StandardCommand(&cobra.Command{
+	Use:     "delete <project ID> [--fail]",
+	Short:   "Delete inactive project",
```
efiShtain: we don't have a concept of inactive project
so just delete a project

internal/backend/db/dbgorm/deployments.go
```diff
@@ -48,15 +48,29 @@ func (db *gormdb) GetDeployment(ctx context.Context, id sdktypes.DeploymentID) (
...
+
+func (db *gormdb) deleteDeploymentsAndDependents(ctx context.Context, ids []string) error {
+	// delete deployments and related sessions
+	if len(ids) == 0 {
+		return nil
+	}
+
+	dbb := db.db.WithContext(ctx)
+	return db.transaction(ctx, func(tx *tx) error {
+		if err := dbb.Where("deployment_id IN ?", ids).Delete(&scheme.Session{}).Error; err != nil {
```
efiShtain: if you don't use the tx object, you're not actually running in the transaction context

internal/backend/db/dbgorm/deployments.go
```diff
@@ -48,15 +48,29 @@ func (db *gormdb) GetDeployment(ctx context.Context, id sdktypes.DeploymentID) (
...
+			return err
+		}
+
+		return dbb.Where("deployment_id IN ?", ids).Delete(&scheme.Deployment{}).Error
```
efiShtain: you can use create a generic scope method IN like this
```go
func IN(field string, values []string) func(db *gorm.DB) *gorm.DB {
  return func(db *gorm.DB) *gorm.DB {
    return db.Where("(?) IN (?)", field, values)
  }
}
```

and change 
```go
dbb.Where("deployment_id IN ?", ids).Delete(&scheme.Deployment{}).Error
``` 
to 

```go
dbb.Scopes(IN("deployment_id", ids).Delete(&scheme.Deployment{}).Error
```


not mandatory, worth considering since you are making a lot of nice cleaning and refactoring in the db area

internal/backend/db/dbgorm/envs.go
```diff
@@ -39,9 +39,12 @@ func (db *gormdb) CreateEnv(ctx context.Context, env sdktypes.Env) error {
 	return translateError(db.createEnv(ctx, e))
 }
 
+func (db *gormdb) deleteEnvs(ctx context.Context, ids []string) error {
+	return db.db.WithContext(ctx).Where("env_id IN ?", ids).Delete(&scheme.Env{}).Error
```
efiShtain: if you'll change to use scopes, you can reuse the function here I stated above

internal/backend/db/dbgorm/envs.go
```diff
@@ -39,9 +39,12 @@ func (db *gormdb) CreateEnv(ctx context.Context, env sdktypes.Env) error {
...
+}
+
 func (db *gormdb) deleteEnv(ctx context.Context, envID string) error {
-	e := scheme.Env{EnvID: envID}
-	return db.db.WithContext(ctx).Delete(&e).Error
+	return db.db.WithContext(ctx).Delete(&scheme.Env{EnvID: envID}).Error
```
efiShtain: do you think calling deleteEnvs with one id in the list is a valid option ?

internal/backend/db/dbgorm/gorm_test.go
```diff
@@ -45,7 +47,7 @@ func setupDB(dbName string) *gorm.DB {
 		},
 	)
 	if dbName == "" {
-		dbName = ":memory:"
+		dbName = "file::memory:?cache=shared"
```
efiShtain: why do you need to share stuff between tests ?

internal/backend/db/dbgorm/projects.go
```diff
@@ -28,6 +30,44 @@ func (db *gormdb) CreateProject(ctx context.Context, p sdktypes.Project) error {
...
 }
 
+func (db *gormdb) deleteProject(ctx context.Context, projectID string) error {
+	p := scheme.Project{ProjectID: projectID}
+	return db.db.WithContext(ctx).Delete(&p).Error
+}
+
+func (db *gormdb) deleteProjectAndDependents(ctx context.Context, projectID string) error {
+	return db.transaction(ctx, func(tx *tx) error {
+		depsEnvs, err := db.getProjectDeployments(ctx, projectID)
```
efiShtain: you need to use the tx object

internal/backend/db/dbgorm/projects.go
```diff
@@ -28,6 +30,44 @@ func (db *gormdb) CreateProject(ctx context.Context, p sdktypes.Project) error {
...
+		if err != nil {
+			return err
+		}
+		var deplIDs, envIDs []string
+		for _, de := range depsEnvs {
+			if de.State != int32(sdktypes.DeploymentStateInactive.ToProto()) {
```
efiShtain: does the new sdktypes provide a better way to do it?
like parse deployment state or something? I'm not sure
I tend to dislike this part:
```
 int32(sdktypes.DeploymentStateInactive.ToProto())
```

internal/backend/db/dbgorm/projects.go
```diff
@@ -28,6 +30,44 @@ func (db *gormdb) CreateProject(ctx context.Context, p sdktypes.Project) error {
...
+			}
+			deplIDs = append(deplIDs, de.DeploymentID)
+			envIDs = append(envIDs, de.EnvID)
+		}
+
+		if err = db.deleteDeploymentsAndDependents(ctx, deplIDs); err != nil {
+			return err
+		}
+
+		if err = db.deleteEnvs(ctx, envIDs); err != nil {
```
efiShtain: this assumes all envs has a deployment, 
but if you just create an env, and don't deploy anything, you won't get this envID in the list and won't delete it
so you have to check for all project envs directly, not only through deployments

internal/backend/db/dbgorm/projects.go
```diff
@@ -28,6 +30,44 @@ func (db *gormdb) CreateProject(ctx context.Context, p sdktypes.Project) error {
...
+			if de.State != int32(sdktypes.DeploymentStateInactive.ToProto()) {
+				return fmt.Errorf("%w: project <%s>: cannot delete non-inactive deployment <%s> in state <%s>",
+					sdkerrors.ErrFailedPrecondition, projectID, de.DeploymentID,
+					deploymentsv1.DeploymentState(de.State).String())
+			}
+			deplIDs = append(deplIDs, de.DeploymentID)
+			envIDs = append(envIDs, de.EnvID)
+		}
+
+		if err = db.deleteDeploymentsAndDependents(ctx, deplIDs); err != nil {
```
efiShtain: will this use the same transaction ?

internal/backend/db/dbgorm/projects.go
```diff
@@ -41,24 +81,61 @@ func (db *gormdb) UpdateProject(ctx context.Context, p sdktypes.Project) error {
...
 
+func (db *gormdb) getProject(ctx context.Context, projectID string) (*scheme.Project, error) {
+	return getOne(db.db, ctx, scheme.Project{}, "project_id = ?", projectID)
+}
+
+func (db *gormdb) getProjectByName(ctx context.Context, projectName string) (*scheme.Project, error) {
+	return getOne(db.db, ctx, scheme.Project{}, "name = ?", projectName)
+}
+
+func schemaToSDKProject(p *scheme.Project, err error) (sdktypes.Project, error) {
```
efiShtain, line 98: we have parseProject in the scheme file by convention 
any reason it is here instead of there ?

internal/backend/db/dbgorm/projects.go
```diff
@@ -41,24 +81,61 @@ func (db *gormdb) UpdateProject(ctx context.Context, p sdktypes.Project) error {
...
 func (db *gormdb) GetProjectByName(ctx context.Context, ph sdktypes.Symbol) (sdktypes.Project, error) {
-	return getOneWTransform(db.db, ctx, scheme.ParseProject, "name = ?", ph.String())
+	return schemaToSDKProject(db.getProjectByName(ctx, ph.String()))
 }
 
-func (db *gormdb) ListProjects(ctx context.Context) ([]sdktypes.Project, error) {
-	var rs []scheme.Project
-	q := db.db.WithContext(ctx)
+type DepEnv struct {
+	DeploymentID string `gorm:"primaryKey"`
```
efiShtain: why do you need this gorm tag here ?

internal/backend/db/dbgorm/projects.go
```diff
@@ -41,24 +81,61 @@ func (db *gormdb) UpdateProject(ctx context.Context, p sdktypes.Project) error {
...
+	res := db.db.WithContext(ctx).Model(&scheme.Deployment{}).
+		Joins("join Envs on Envs.env_id = Deployments.env_id").
+		Where("Envs.project_id = ?", pid).
+		Select("Deployments.deployment_id, Deployments.state, Envs.env_id").
+		Find(&pds)
+	return pds, res.Error
+}
+
+func (db *gormdb) listProjects(ctx context.Context) ([]scheme.Project, error) {
+	q := db.db.WithContext(ctx).Order("project_id")
```
efiShtain, line 134: do you really need to sort by project_id here ?
anyhow, I think it's the default sort anyway so it has no meaning

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -211,12 +223,26 @@ func ParseEventRecord(e EventRecord) (sdktypes.EventRecord, error) {
 
 type Env struct {
 	EnvID     string `gorm:"primaryKey"`
-	ProjectID string `gorm:"index"`
+	ProjectID string `gorm:"index;foreignKey"`
```
efiShtain, line 216: do you need the foreign key tag even if you added the project field ?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -211,12 +223,26 @@ func ParseEventRecord(e EventRecord) (sdktypes.EventRecord, error) {
...
 	MembershipID string `gorm:"uniqueIndex"`
+
+	// just for the foreign key. Wihtout it gorm won't enforce it
+	Project Project
+}
+
+// enforce foreign keys constrains while soft-deleting
+func (e *Env) BeforeDelete(db *gorm.DB) error {
+	var count int64
+	db.Model(&Deployment{}).Where("deleted_at is NULL and env_id = ?", e.EnvID).Count(&count)
```
efiShtain: make sure you have the relevant indexes on deployment

cmd/ak/cmd/deployments/delete.go
ig-ra: it's not my invention actually :) There is a --fail flag. So cmd will finish with exit status 0, if the error is not found. Unless --fail flag is supplied. In this case cmd will report error if such an object isn't found

internal/backend/db/dbgorm/projects.go
ig-ra: I kept the existing code logic. Not added by me, so prefer not to remove it

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: If I remove it gorm will guess it I think, but explicit IMO is better then implicit.

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: added 2 indexes

internal/backend/db/dbgorm/projects.go
ig-ra: no need, u r right. removed

internal/backend/db/dbgorm/projects.go
ig-ra: it's still there. Just exctracted common logic from getByID and getByName

internal/backend/db/dbgorm/envs.go
ig-ra: actually there is a problem. Since deleteEnv will enforce foreign key constrains, but deleteEnvs won't. Thanks for causing to rethink it. Rewritten

cmd/ak/cmd/projects/delete.go
ig-ra: done

internal/backend/db/dbgorm/deployments.go
ig-ra: fixed. please recheck

internal/backend/db/dbgorm/deployments.go
ig-ra: hmm.. IMO it changes nothing in terms of clarity or length of the gorm statements. So, I think I will skip

internal/backend/db/dbgorm/envs.go
ig-ra: see my note for the scopes

cmd/ak/cmd/deployments/delete.go
efiShtain: I understand that, but you're not really checking for not found, you check if d.IsValid which is really not readable 

can you just do 
```go
if !d.IsValid() {
  return errors.New("invalid deployment id")
}
```

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -393,7 +393,7 @@ func ParseSession(s Session) (sdktypes.Session, error) {
 
 type Deployment struct {
 	DeploymentID string `gorm:"primaryKey"`
-	EnvID        string `gorm:"foreignKey"`
+	EnvID        string `gorm:"index;foreignKey"`
 	BuildID      string `gorm:"foreignKey"`
```
efiShtain, line 377: if you are going to query deployments by build id, you will need an index as well

internal/backend/db/dbgorm/envs.go
efiShtain: WDYM ?
foreign key constraint should be validated even if you delete the table it self

internal/backend/db/dbgorm/deployments.go
ig-ra: rewritten once again.
at the end, I did what u suggested - extracted transactions enclosure to API. it was simpler than introducing internal methods

cmd/ak/cmd/deployments/delete.go
ig-ra: this is due to new changes to sdk that were introduced. Previously all methods were returning nil, so it was clear and readable. Now, when not found, zero object will be returned and it should fail on `isValid()`. So actually failing on `IsValid` means not found

internal/backend/db/dbgorm/gorm_test.go
ig-ra: fixed. It seems the problem was with transactions. They were opening new db connection (closing old one), which was causing to all data to disappear in db. Once we got rid of transactions, the problem was gone

internal/backend/db/dbgorm/projects.go
ig-ra: rewritten

internal/backend/db/dbgorm/projects.go
ig-ra: I'm not sure myself. What's the new way to get int32 from sdktype enum?

internal/backend/db/dbgorm/projects.go
ig-ra: rewritten. please recheck

cmd/ak/cmd/deployments/delete.go
efiShtain: I understand why it should fail, I just think this is not readable to use fail if not found on not valid id

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: no, we are not querying deployments by buildID. The opposite - gathering buildIDs by deployments

internal/backend/db/dbgorm/envs.go
ig-ra: rewritten as u suggested - via deleteEnvs

cmd/ak/cmd/deployments/delete.go
ig-ra: agree. But it's not something introduced by this PR. Actually it's something that was unreadable for me in the original sdk2 PR as well

efiShtain (approved):

PR #79, opened 2024-03-13
Add support for optional GHES URL

daabr:
This PR implements ENG-510.

It allows the user to set the OS env var `GITHUB_ENTERPRISE_URL` to `http[s]://my-on-prem-github-address` - and if so, the AK server will initialize its GitHub client accordingly when it executes GitHub API calls.

~This also affects the creation of GitHub OAuth connections - the OAuth flow will be redirected to the GHES URL instead of `github.com`.~ This is now in PR #80.

This PR also updates the version of the `go-github` package in order to use the new builder-pattern initialization for enterprise clients instead of the the deprecated `NewEnterpriseClient` constructor.


integrations/github/jwt.go
```diff
@@ -101,7 +107,11 @@ func (i integration) NewClientWithAppJWTFromGitHubID(appID int64) (*github.Clien
 	}
 
 	// Initialize a client with the generated JWT injected into outbound requests.
-	return github.NewClient(&http.Client{Transport: atr}), nil
+	client := github.NewClient(&http.Client{Transport: atr})
+	if enterpriseURL := os.Getenv(enterpriseURLEnvVar); enterpriseURL != "" {
```
itayd, line 111: tip: you could use LookupEnv, but just fyi

itayd (approved):

integrations/github/jwt.go
daabr: ack, I prefer `Getenv` because if the result is "" I don't care why that is (i.e. env var missing or empty)

PR #82, opened 2024-03-13
container values: verifies members are not invalid

itayd:
Container values need to verify that their members are all valid (meaning not the zero value, which is a nil message) before constructing new one.

### DIFF
--- a/runtimes/starlarkrt/internal/values/convert.go
+++ b/runtimes/starlarkrt/internal/values/convert.go
@@ -141,9 +141,9 @@ func (vctx *Context) FromStarlarkValue(v starlark.Value) (sdktypes.Value, error)
 	case starlarktime.Time:
 		return sdktypes.NewTimeValue(time.Time(v)), nil
 	case *starlark.List, starlark.Tuple:
-		return vctx.fromSequence(v.(starlark.Sequence), sdktypes.NewListValue)
+		return vctx.fromSequence(v.(starlark.Sequence), kittehs.Must11(sdktypes.NewListValue))
 	case *starlark.Set:
-		return vctx.fromSequence(v, sdktypes.NewSetValue)
+		return vctx.fromSequence(v, kittehs.Must11(sdktypes.NewSetValue))
 	case *starlark.Dict:
 		ks := v.Keys()
 		items := make([]sdktypes.DictItem, len(ks))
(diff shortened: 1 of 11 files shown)


tebeka:
It'll help me review if you'll explain what is the purpose of this PR.

tebeka:
> Yes, I know I'm using Musts here, but only in places where there must never be any failures. What should I do otherwise—ignore the errors?

IMO return errors. The question about panic is weather we're catching them. A panic in a goroutine won't be caught in a different goroutine.

runtimes/starlarkrt/internal/values/convert.go
```diff
@@ -141,9 +141,9 @@ func (vctx *Context) FromStarlarkValue(v starlark.Value) (sdktypes.Value, error)
 	case starlarktime.Time:
 		return sdktypes.NewTimeValue(time.Time(v)), nil
 	case *starlark.List, starlark.Tuple:
-		return vctx.fromSequence(v.(starlark.Sequence), sdktypes.NewListValue)
+		return vctx.fromSequence(v.(starlark.Sequence), kittehs.Must11(sdktypes.NewListValue))
```
tebeka, line 144: `v(.starlark.Sequence)` might panic as well. If you are sure it won't it's OK to keep it that way.
IMO return error instead of using `Must`, but not critical.

runtimes/starlarkrt/runtime/run.go
```diff
@@ -43,6 +43,21 @@ type run struct {
 
 func (r *run) ID() sdktypes.RunID { return r.runID }
 
+func makeStruct(th *starlark.Thread, b *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
tebeka: Why is this function required? In what cases? Maybe a design change will be better?
Disclaimer: Don't know enough about the code.

tebeka (approved):
Should work.

As a general note, we should redesign the system that starlark code & type system won't go beyond the runtime boundary. Then the rest of the AutoKitteh code will be native Go types.

runtimes/starlarkrt/internal/values/convert.go
itayd: it wont since List and Tuple implement sequence.

runtimes/starlarkrt/runtime/run.go
itayd: ah, forgot to remove this.

itayd:
We've spoken before about the runtime boundry and starlark. For now it's a heavy lift which I'm not convinced we need to do at all. I guess time will tell.

Re the errors - sometimes it makes the function a lot less wieldy for no benefit - since in some cases we *know* they wont error out, it makes sense.

PR #83, opened 2024-03-13
do not use miniredis for prod, warn instead

itayd:


internal/backend/storegrpcsvc/svc.go
```diff
@@ -15,6 +16,8 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+var errNotConfigured = connect.NewError(http.StatusNotImplemented, fmt.Errorf("store service not configured"))
```
daabr: I think `connect.CodeUnavailable` is a more appropriate error code.

itayd: changed to `connect.Unimplemented` as discussed. Unavailable is transient.

daabr (approved):

PR #84, opened 2024-03-13
run service as an external process for system tests in CI

itayd:
this should keep the tests deterministic until we solve the underlying in-process issue and still allow debugging tests easily locally

### DIFF
--- a/backend/svc/svcproc.go
+++ b/backend/svc/svcproc.go
@@ -2,30 +2,118 @@ package svc
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
+	"path/filepath"
+	"strings"
+	"time"
+
+	"go.uber.org/fx"
+
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 )
 
 type svcProc struct {
-	cfg *Config
+	binPath string
+	ropts   RunOptions
+	cfg     *Config
+	addr    string
+	cmd     *exec.Cmd
+	wait    chan ShutdownSignal
 }
 
 func (s *svcProc) Start(ctx context.Context) error {
-	cmd := exec.Command("bin/ak" /*args...*/)
-	cmd.Stdout, cmd.Stderr = os.Stdout, os.Stderr
+	tmpDir, err := os.MkdirTemp("", "ak-svcproc-*")
+	if err != nil {
+		return fmt.Errorf("mkdirtemp: %w", err)
+	}
+
+	defer func() {
+		if err := os.RemoveAll(tmpDir); err != nil {
+			fmt.Fprintf(os.Stderr, "rm(%q): %v\n", tmpDir, err)
+		}
+	}()
+
+	addrFilePath := filepath.Join(tmpDir, "addr")
+	readyFilePath := filepath.Join(tmpDir, "ready")
+
+	// TODO: pass user configuration to executable.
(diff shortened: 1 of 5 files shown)


cmd/ak/cmd/up.go
```diff
@@ -33,6 +37,12 @@ var upCmd = common.StandardCommand(&cobra.Command{
 			return fmt.Errorf("fx app start: %w", err)
 		}
 
+		if readyFile != "" {
```
daabr, line 41: I disagree with this approach of using a file as a signal, not to mention as a boolean signal. As a hack, it's a little too "dirty".

Furthermore, this flag shouldn't exist just for our testing purposes, and normal users shouldn't care (or at least shouldn't need to care) about it even if they also run AK as a subprocess.

For posterity (when we remove this block): there are more reliable and standard ways to advertise server state, and they should be used whenever AK is in "lame-duck mode" - both when it starts, and when it stops.

cmd/ak/cmd/up.go
```diff
@@ -33,6 +37,12 @@ var upCmd = common.StandardCommand(&cobra.Command{
...
+			if err := os.WriteFile(readyFile, []byte("ready"), 0o644); err != nil {
+				fmt.Fprintf(os.Stderr, "write ready file: %v\n", err)
```
daabr: Please don't use `os.Stdout` or `os.Stderr` directly in CLI commands. Right now it doesn't actually matter, but it may in the future, so we should minimize this. Instead, use `cmd.OutOrStdout` or `cmd.ErrOrStderr` - this allows Cobra to inject its own streams between/instead stdout and stderr.

cmd/ak/cmd/up.go
```diff
@@ -44,4 +54,5 @@ var upCmd = common.StandardCommand(&cobra.Command{
 func init() {
 	// Command-specific flags.
 	upCmd.Flags().StringVarP(&mode, "mode", "m", "", "run mode: {default|dev|test}")
+	upCmd.Flags().StringVarP(&readyFile, "ready-file", "r", "", "write a file when the server is ready")
```
daabr, line 59:
```suggestion

	upCmd.Flags().StringVarP(&readyFile, "ready-file", "r", "", "write a file when the server is ready")
	kittehs.Must0(upCmd.MarkFlagFilename("ready-file"))
```

backend/svc/svcproc.go
```diff
@@ -0,0 +1,119 @@
+package svc
```
daabr, line 1: Self reminder after offline discussion:

we can keep this file here for now, but I don't think it's appropriate to put it in the "backend" directory. It's not AK backend code.

The only caller is the system test - so that's a more logical place, but it's not unreasonable that some AK users may want to use this too as a wrapper for running AK.

`internal` is also not a great choice for the same reason.

daabr (approved):

cmd/ak/cmd/up.go
itayd: i think that it's sufficient for now. better way would have been to reach to the server via http and ask if it's ok, but that'll be part of ENG-530

cmd/ak/cmd/up.go
itayd: oh cool didn't know about this.

cmd/ak/cmd/up.go
itayd: done

PR #85, opened 2024-03-13
allow to specify callopts, among them an easier catch syntax

itayd:
not sure i like the syntax, let's discuss

### DIFF
--- a/examples/testprograms/http/http.star
+++ b/examples/testprograms/http/http.star
@@ -7,7 +7,18 @@ def foo(x):
     return "meow, " + x
 
 def on_http_get():
-    print(activity(lambda: foo("mooo")))
+    resp, err = http1.get("https://httpbin3212.org/status/404", ak.callopts(catch=True))
+
+    # also works:
+    # resp, err = catch(lambda: http1.get("https://httpbin3212.org/status/404"))
+    # resp, err = catch(http1.get, "https://httpbin3212.org/status/404")
+    # resp, err = http1.get("https://httpbin3212.org/status/404", ak = ak.callopts(catch=True))
+    # resp, err = http1.get("https://httpbin3212.org/status/404", ak_catch=True)
+    # resp, err = http1.get("https://httpbin3212.org/status/404", ak = {"catch": True})
+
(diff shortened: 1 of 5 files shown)


examples/testprograms/http/http.star
```diff
@@ -4,10 +4,19 @@ load("env", "MEOW")
...
+    # resp, err = catch(lambda: http1.get("https://httpbin3212.org/status/404"))
+    # resp, err = catch(http1.get, "https://httpbin3212.org/status/404")
+    # resp, err = http1.get("https://httpbin3212.org/status/404", ak = ak.callopts(catch=True))
+    # resp, err = http1.get("https://httpbin3212.org/status/404", ak_catch=True)
+    # resp, err = http1.get("https://httpbin3212.org/status/404", ak = {"catch": True})
+
     print(resp, err)
     print(err.op)
 
+
```
daabr, line 22: Nit: Not sure if this line was added by mistake. Unlike Google Python style, Starlark style is to have only 1 empty line between top-level blocks (at least according to Google's Bazel formatter).

examples/testprograms/http/http.star
```diff
@@ -4,10 +4,19 @@ load("env", "MEOW")
 print(MEOW)
 
 def on_http_get():
-    resp, err = catch(lambda: http1.get("https://httpbin3212.org/status/404"))
+    resp, err = http1.get("https://httpbin3212.org/status/404", ak.callopts(catch=True))
```
daabr, line 10: This "trick" of multiple assignment happens to be supported in Python and Starlark, and sort-of in JS/TS but not so nicely. But we can't rely on it. Also, none of these options really feels "intuitive" or "easy to read".

Instead, why not just return a single `result` object that contains optional error information alongside an optional `resp` struct property, e.g. a Boolean `ok` property and/or a string `error` property.

daabr (changes requested):

examples/testprograms/http/http.star
itayd:
```suggestion

```

daabr (approved):

PR #86, opened 2024-03-13
feat(db): delete connections

ig-ra:
delete (hard) connections when deleting a project

### DIFF
--- a/cmd/ak/cmd/connections/delete.go
+++ b/cmd/ak/cmd/connections/delete.go
@@ -1,13 +1,11 @@
 package connections
 
 import (
-	"errors"
-	"fmt"
-
 	"github.com/spf13/cobra"
 
 	"go.autokitteh.dev/autokitteh/cmd/ak/common"
 	"go.autokitteh.dev/autokitteh/internal/resolver"
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 )
 
 var deleteCmd = common.StandardCommand(&cobra.Command{
@@ -20,23 +18,16 @@ var deleteCmd = common.StandardCommand(&cobra.Command{
 		r := resolver.Resolver{Client: common.Client()}
 		c, id, err := r.ConnectionNameOrID(args[0])
 		if err != nil {
-			if errors.As(err, resolver.NotFoundErrorType) {
-				err = common.NewExitCodeError(common.NotFoundExitCode, err)
-			}
-			return err
+			return common.ToExitCodeError(err, "connection")
 		}
 		if !c.IsValid() {
-			err = errors.New("connection not found")
-			return common.NewExitCodeError(common.NotFoundExitCode, err)
+			return common.ToExitCodeError(sdkerrors.ErrNotFound, "connection")
 		}
 
 		ctx, cancel := common.LimitedContext()
 		defer cancel()
 
-		if err = connections().Delete(ctx, id); err != nil {
-			return fmt.Errorf("delete connection: %w", err)
-		}
-
-		return nil
+		err = connections().Delete(ctx, id)
+		return common.ToExitCodeError(err, "delete connection")
 	},
 })
--- a/cmd/ak/common/flags.go
+++ b/cmd/ak/common/flags.go
@@ -7,6 +7,7 @@ import (
 	"github.com/spf13/cobra"
 
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 )
 
@@ -45,10 +46,10 @@ func ToExitCodeError(err error, what string) error {
 	case errors.Is(err, sdkerrors.ErrFailedPrecondition):
 		msg = fmt.Sprintf("on %s", what)
 		code = FailedPrecondition
-	case errors.As(err, resolver.NotFoundErrorType) {
+	case errors.As(err, resolver.NotFoundErrorType):
 		msg = fmt.Sprintf("%s not found", what)
 		code = NotFoundExitCode
-    }
+	}
 	return NewExitCodeError(code, fmt.Errorf("%w: %s", err, msg))
 }
 
--- a/cmd/ak/common/flags.go
+++ b/cmd/ak/common/flags.go
@@ -41,14 +41,12 @@ func ToExitCodeError(err error, what string) error {
 	var code int = GenericFailure
 	switch {
 	case errors.Is(err, sdkerrors.ErrNotFound):
-		msg = fmt.Sprintf("%s not found", what)
-		code = NotFoundExitCode
+		return NewExitCodeError(NotFoundExitCode, fmt.Errorf("%s not found", what))
 	case errors.Is(err, sdkerrors.ErrFailedPrecondition):
 		msg = fmt.Sprintf("on %s", what)
 		code = FailedPrecondition
 	case errors.As(err, resolver.NotFoundErrorType):
-		msg = fmt.Sprintf("%s not found", what)
-		code = NotFoundExitCode
+		return NewExitCodeError(NotFoundExitCode, fmt.Errorf("%s not found", what))
 	}
 	return NewExitCodeError(code, fmt.Errorf("%w: %s", err, msg))
 }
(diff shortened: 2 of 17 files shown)


cmd/ak/cmd/connections/delete.go
```diff
@@ -20,23 +18,16 @@ var deleteCmd = common.StandardCommand(&cobra.Command{
...
-			if errors.As(err, resolver.NotFoundErrorType) {
-				err = common.NewExitCodeError(common.NotFoundExitCode, err)
-			}
-			return err
+			return common.ToExitCodeError(err, "connection")
 		}
 		if !c.IsValid() {
-			err = errors.New("connection not found")
-			return common.NewExitCodeError(common.NotFoundExitCode, err)
+			return common.ToExitCodeError(sdkerrors.ErrNotFound, "connection")
```
efiShtain, line 24: if you fix this line, maybe we should fix the error?

c.IsValid == not found? or invalid id ?

cmd/ak/cmd/connections/delete.go
```diff
@@ -20,23 +18,16 @@ var deleteCmd = common.StandardCommand(&cobra.Command{
...
 		ctx, cancel := common.LimitedContext()
 		defer cancel()
 
-		if err = connections().Delete(ctx, id); err != nil {
-			return fmt.Errorf("delete connection: %w", err)
-		}
-
-		return nil
+		err = connections().Delete(ctx, id)
+		return common.ToExitCodeError(err, "delete connection")
```
efiShtain, line 31: this make me read it as "there is an error, convert it to exit code"
which is kind of confusing

cmd/ak/common/flags.go
```diff
@@ -45,7 +45,10 @@ func ToExitCodeError(err error, what string) error {
 	case errors.Is(err, sdkerrors.ErrFailedPrecondition):
 		msg = fmt.Sprintf("on %s", what)
 		code = FailedPrecondition
-	}
+	case errors.As(err, resolver.NotFoundErrorType) {
```
efiShtain: why errors.As instead of Is?

internal/backend/db/dbgorm/projects.go
```diff
@@ -59,11 +59,14 @@ func (db *gormdb) deleteProjectAndDependents(ctx context.Context, projectID stri
...
 
-	// delete triggers
 	if err = gormDB.Delete(&scheme.Trigger{}, "project_id = ?", projectID).Error; err != nil {
 		return err
 	}
 
+	if err = gormDB.Delete(&scheme.Connection{}, "project_id = ?", projectID).Error; err != nil {
+		return err
+	}
+
```
efiShtain, line 82: how do we know no one is referencing this connection ?
delete order ?

efiShtain (dismissed):
Please review my comments before merge and update if you see it fit
Looks ok to me

cmd/ak/cmd/connections/delete.go
ig-ra: after sdk2 there are no nulls. So !c.IsValid() is not found actually

cmd/ak/common/flags.go
ig-ra: I moved the code without changing the logic. So it was `errors.As` and since there is no tests for this I prefer to leave it as is

internal/backend/db/dbgorm/projects.go
ig-ra: Connection could be referenced by Triggers or Signals. Triggers are deleted before Connections. And now since it's a hard delete, if there is anyone referencing this connection, delete will fail on foreign key

efiShtain: connection are also referenced by events
so this is something we should think about
not sure we can delete them now

cmd/ak/cmd/connections/delete.go
```diff
@@ -20,23 +18,16 @@ var deleteCmd = common.StandardCommand(&cobra.Command{
 		r := resolver.Resolver{Client: common.Client()}
 		c, id, err := r.ConnectionNameOrID(args[0])
 		if err != nil {
-			if errors.As(err, resolver.NotFoundErrorType) {
-				err = common.NewExitCodeError(common.NotFoundExitCode, err)
-			}
-			return err
+			return common.ToExitCodeError(err, "connection")
```
daabr, line 21: I understand the motivation for the change here and under `!c.IsValid()` below, but I think it leads to a wrong result, cosmetically.

The expected output is "Error: connection not found" (if the problem is that the ID wasn't found), but `common.ToExitCodeError()` prints something more complicated (error's string + error message = unnecessary duplication).

And as a result, there's a legitimate failure in a system test. And because it's less user-friendly than before I wouldn't want to fix the test, but rather leave the error handling as-is.

We could definitely improve this systematically later, but it isn't worth the effort right now.

internal/backend/db/dbgorm/projects.go
daabr: > Connection could be referenced by Triggers or Signals. Triggers are deleted before Connections. And now since it's a hard delete, if there is anyone referencing this connection, delete will fail on foreign key

Please add this information as a comment in the code. This will help future readers to understand the code and the flow much more quickly.

cmd/ak/cmd/connections/delete.go
ig-ra: plz suggest better name?
we already have for example translateError which doing the same (converts nil to nil of there is no error)
I don't think we shoudl split it to if/else, but better name could do the job

cmd/ak/cmd/connections/delete.go
ig-ra: I fixed a ToExitCodeError wrapper. Now it prints what was expected

cmd/ak/cmd/connections/delete.go
efiShtain: standard stuff, 
check if there is an error use common.ToExitCodeError(err, 'delete connection')
otherwise return null

cmd/ak/common/flags.go
```diff
@@ -40,15 +41,13 @@ func ToExitCodeError(err error, what string) error {
...
-		code = NotFoundExitCode
+		return NewExitCodeError(NotFoundExitCode, fmt.Errorf("%s not found", what))
 	case errors.Is(err, sdkerrors.ErrFailedPrecondition):
 		msg = fmt.Sprintf("on %s", what)
 		code = FailedPrecondition
-	case errors.As(err, resolver.NotFoundErrorType) {
-		msg = fmt.Sprintf("%s not found", what)
-		code = NotFoundExitCode
-    }
+	case errors.As(err, resolver.NotFoundErrorType):
```
efiShtain, line 48: we have both sdk and resolver not found errors ?

internal/backend/db/dbgorm/projects.go
```diff
@@ -59,10 +59,23 @@ func (db *gormdb) deleteProjectAndDependents(ctx context.Context, projectID stri
...
+	var signalIDs []string
+	if err := gormDB.Model(&scheme.Signal{}).
+		Joins("join Connections on Connections.connection_id = Signals.connection_id").
+		Where("Connections.project_id = ?", projectID).
+		Pluck("Signals.signal_id", &signalIDs).Error; err != nil {
+		return err
+	}
+	if err = gormDB.Delete(&scheme.Signal{}, "signal_id IN ?", signalIDs).Error; err != nil {
+		return err
+	}
```
efiShtain, line 77: you can just delete without getting the signals id back

(unless somehow gorm convert this two statements to one query with sub query)

internal/backend/db/dbgorm/projects.go
```diff
@@ -59,10 +59,23 @@ func (db *gormdb) deleteProjectAndDependents(ctx context.Context, projectID stri
...
+
 	if err = gormDB.Delete(&scheme.Connection{}, "project_id = ?", projectID).Error; err != nil {
```
efiShtain, line 79: deleting a connection has to perform some other logic to delete the secret
is it handled here ?

efiShtain (approved):

cmd/ak/common/flags.go
ig-ra: yup. I just moved the code. No new logic introduction

internal/backend/db/dbgorm/projects.go
ig-ra: I talked to Haim and it seems that we understood this PRD too wide - so secrets are not part of this PRD. We want only to hide info from the user, not to deal or actually clean the DB.

internal/backend/db/dbgorm/projects.go
ig-ra: rewritten

PR #87, opened 2024-03-13
fix paths manipulation in reader to work on cross-platforms

RonenMars:
To support server implementation on Windows, we have to implement a path separator adjusted to each OS.

Using this fix, we'll be able to work with the same OS for the server and the client, e.g. 
windows client -> windows server
linux client -> linux server, etc...

The next step is to allow full cross-platform support, to make the server OS agnostic.

### DIFF
--- a/sdk/sdkbuildfile/reader.go
+++ b/sdk/sdkbuildfile/reader.go
@@ -8,6 +8,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"os"
 	"path/filepath"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


tebeka (approved):
LGTM in general. I only wonder about the case where the server runs on Linux and the extension is on Windows - will this work?

Maybe normalize the string: strings.Replace(path,  `\`, "/", -1)

itayd (approved):

RonenMars:
@tebeka Hey, it won't work (mentioned in the description), we should think of a better solution, but at this moment it solves the support for both platforms (windows/linux/mac).

PR #89, opened 2024-03-14
support activities calling back to scripts

itayd:
this also allow running a specific script function as an activity, as long as it does not try to invoke other activities.

### DIFF
--- a/internal/backend/sessions/sessionworkflows/workflow.go
+++ b/internal/backend/sessions/sessionworkflows/workflow.go
@@ -52,6 +52,8 @@ type sessionWorkflow struct {
 	callSeq uint32
 
 	signals map[string]uint64 // map signals to next sequence number
+
+	state sdktypes.SessionState
 }
 
 func runWorkflow(
@@ -102,6 +104,8 @@ func (w *sessionWorkflow) addPrint(ctx workflow.Context, print string) {
 
 func (w *sessionWorkflow) updateState(ctx workflow.Context, state sdktypes.SessionState) {
 	w.z.Debug("update state", zap.Any("state", state))
+
+	w.state = state
 
 	if err := workflow.ExecuteLocalActivity(ctx, w.ws.svcs.DB.UpdateSessionState, w.data.SessionID, state).Get(ctx, nil); err != nil {
 		w.z.Panic("update session", zap.Error(err))
@@ -370,8 +374,7 @@ func (w *sessionWorkflow) removeEventSubscription(ctx context.Context, signalID 
 }
 
 func (w *sessionWorkflow) run(ctx workflow.Context) (prints []string, err error) {
-	// set before r.Call is called.
-	var callValue sdktypes.Value
+	workflowContextKey := struct{ _ string }{"autokitteh_workflow_context"}
 
 	newRunID := func() (runID sdktypes.RunID) {
 		if err := workflow.SideEffect(ctx, func(workflow.Context) any {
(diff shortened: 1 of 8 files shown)


internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -370,8 +374,7 @@ func (w *sessionWorkflow) removeEventSubscription(ctx context.Context, signalID
 }
 
 func (w *sessionWorkflow) run(ctx workflow.Context) (prints []string, err error) {
-	// set before r.Call is called.
-	var callValue sdktypes.Value
+	workflowContextKey := struct{ _ string }{"autokitteh_workflow_context"}
```
tebeka: Nitpick. This is not going to guaranteed to be unique in the context:

```go
	s1 := struct{ _ string }{"Hello"}
	s2 := struct{ _ string }{"Hello"}

	fmt.Println(s1 == s2) // true

	m := make(map[any]any)
	m[s1] = 1
	m[s2] = 2
	fmt.Println(m) // map[{}:2]
```

The way to make sure nobody runs over the key is to create a type.
```go
	type ctxKeyType1 int
	var ctxKey1 ctxKeyType1 = 42

	type ctxKeyType2 int
	var ctxKey2 ctxKeyType2 = 42

	m2 := make(map[any]any)
	m2[ctxKey1] = 1
	m2[ctxKey2] = 2
	fmt.Println(m2) // map[42:1 42:2]
```

runtimes/starlarkrt/runtime/builtins.go
```diff
@@ -5,107 +5,181 @@ import (
...
-		}
+	return tlsContext.vctx.ToStarlarkValue(rv)
+}
 
-		if len(args) > 1 {
-			vs, err := kittehs.TransformError(args, vctx.FromStarlarkValue)
-			if err != nil {
-				return nil, fmt.Errorf("cannot convert values from starlark: %w", err)
-			}
+func catchBuiltinFunc(th *starlark.Thread, _ *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
tebeka, line 80: How is this related to the PR?

runtimes/starlarkrt/runtime/builtins.go
```diff
@@ -5,107 +5,181 @@ import (
...
+	})
+	if err != nil {
+		return nil, fmt.Errorf("cannot convert values from starlark: %w", err)
+	}
+
+	v := sdktypes.NewDictValueFromStringMap(vs)
+	return nil, sdktypes.NewProgramError(v, nil, nil).ToError()
+}
+
+func globalsBuiltinFunc(th *starlark.Thread, bi *starlark.Builtin, args starlark.Tuple, kwargs []starlark.Tuple) (starlark.Value, error) {
```
tebeka, line 167: How is this related to the PR?

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -103,6 +105,8 @@ func (w *sessionWorkflow) addPrint(ctx workflow.Context, print string) {
 func (w *sessionWorkflow) updateState(ctx workflow.Context, state sdktypes.SessionState) {
```
tebeka, line 105: I find it confusion that you call the workflow context `ctx` and there's the built-in context which is traditionally called `ctx`. Maybe rename the parameter to `wctx` ?

tebeka (approved):
Go go gadget activity.

internal/backend/sessions/sessionworkflows/workflow.go
itayd: ugh, thanks

runtimes/starlarkrt/runtime/builtins.go
itayd: did a little refactoring that simplified the structure as part of it.

runtimes/starlarkrt/runtime/builtins.go
itayd: ditto

internal/backend/sessions/sessionworkflows/workflow.go
itayd: done. i think from now on for workflow context we'll use wctx

PR #92, opened 2024-03-15
give builtin integrations readable name

itayd:
this helped me debugging #91 , it gives builtin integrations/modules an id where we can easily deduce from site which is it.

### DIFF
--- a/sdk/sdktypes/integration_id.go
+++ b/sdk/sdktypes/integration_id.go
@@ -2,6 +2,7 @@ package sdktypes
 
 import (
 	"fmt"
+	"strings"
 
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
 )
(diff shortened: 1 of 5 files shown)


sdk/sdktypes/integration_id.go
```diff
@@ -23,7 +24,16 @@ func StrictParseIntegrationID(s string) (IntegrationID, error) { return Strict(P
 func IsIntegrationID(s string) bool { return IsIDOf[integrationIDTraits](s) }
 
 func IntegrationIDFromName(name string) IntegrationID {
-	txt := fmt.Sprintf("%s_%026x", integrationIDKind, kittehs.HashString64(name))
+	const chars = "0123456789abcdefghjkmnpqrstvwxyz"
```
daabr, line 27: I have a couple of technical comments on how you implemented this, but more importantly this function doesn't guarantee uniqueness, so we can't use it in production.

Example: "httpi", "httpl", and "httpx" would all be converted to "httpx".

itayd: yeah hence why it's only for builtin integrations, where it's not such an issue imo

PR #93, opened 2024-03-15
when running from make, always use ak as a proc 

itayd:
it's not running in a debugger anyway, so why flake.


Makefile
```diff
@@ -37,6 +37,8 @@ endif
 VERSION_PKG_PATH="go.autokitteh.dev/autokitteh/internal/version"
 LDFLAGS+=-X '${VERSION_PKG_PATH}.Version=${VERSION}' -X '${VERSION_PKG_PATH}.Time=${TIMESTAMP}' -X '${VERSION_PKG_PATH}.Commit=${COMMIT}$(shell git diff --quiet || echo '(dirty)')' -X '${VERSION_PKG_PATH}.User=$(shell whoami)' -X '${VERSION_PKG_PATH}.Host=$(shell hostname)'
 
+export AK_SYSTEST_USE_PROC_SVC=1
+
```
daabr, line 41: I agree with the principle, but once you run `make`, and then start VS Code, doesn't it inherit this environment variable?

itayd: nope? i think it's saved in the shell spawned by make. shouldn't leap out?

itayd:
```
$ export | grep AK
$ make lint
bin/tools/golangci-lint run
$ export | grep AK
```

daabr (approved):

PR #97, opened 2024-03-17
GRPC Integration minimal implementation

efiShtain:
this is a simplified implementation for grpc integration (with example)
support only actions to grpc and not listening for triggers


examples/testprograms/grpc/autokitteh-grpc.yaml
```diff
@@ -0,0 +1,15 @@
+version: v1
+
+project:
```
itayd: i don't think you need to suffix the autokitteh with -grpc, just drop it

examples/testprograms/grpc/grpc.star
```diff
@@ -0,0 +1,7 @@
+load("@grpc", "grpc1")
+
+def on_http_get():
+    # evs = grpc1.call({"host":"localhost:9980", "service":"autokitteh.sessions.v1.SessionsService", "method":"List"})
+    evs2 = grpc1.call(host="localhost:9980", service="autokitteh.events.v1.EventsService", method="List", payload={"event_type":"get"})
```
itayd, line 4: we need to think how to support auth here, also i think we should do something like
```
svc = grpc1.with(host='...', service='...')
svc('List', {...})
```

integrations/grpc/client.go
```diff
@@ -60,6 +75,7 @@ func newGRPCClientWithProtos(conn *grpc.ClientConn, protos protoFiles) (*grpccli
 	return c, nil
 }
 
+//lint:ignore U1000 Ignore unused to keep this code until we decide if we need it
 func (r *grpcclient) getServiceDescriptor(service string) (*desc.ServiceDescriptor, error) {
```
itayd: i say kill it now, we can get it from the git history.

itayd: or you can just add another function to the integration now to list methods on a remote service.

examples/testprograms/grpc/grpc.star
efiShtain: We agreed to start with the simplest implementation for now, therefore this is it

We can extend it later in other PRs

So unless you think this is unusable I think we should merge it and keep iterating later

itayd: sure, create a ticket and assign to yourself though. i didn't approve mostly because of https://github.com/autokitteh/autokitteh/pull/97#discussion_r1527618409

examples/testprograms/grpc/autokitteh-grpc.yaml
efiShtain: fixed

integrations/grpc/client.go
efiShtain: removed

examples/testprograms/grpc/grpc.star
efiShtain: Opened ENG-546

efiShtain: I also think this can be easily implemented in startlark, which is much more flexible and requires less code (we don't have to add anything)

@haimzlato WDYT?

```python
svc = lambda(method, payload: grpc1.call(host="localhost:9980", service="autokitteh.events.v1.EventsService", method=method, payload=payload))

svc('List', {})
```

examples/testprograms/grpc/grpc.star
```diff
@@ -0,0 +1,7 @@
+load("@grpc", "grpc1")
+
+def on_http_get():
+    # evs = grpc1.call({"host":"localhost:9980", "service":"autokitteh.sessions.v1.SessionsService", "method":"List"})
```
itayd:
```suggestion
```

examples/testprograms/grpc/grpc.star
itayd: i think having it in the integration is more ergonomic, maybe we can add a starlark function called `with` to do it for everything.

itayd (approved):

examples/testprograms/grpc/grpc.star
itayd: see https://github.com/autokitteh/autokitteh/pull/108

PR #103, opened 2024-03-18
sessions: properly handle self calls from runtimes

itayd:
depends on #98

### DIFF
--- a/internal/backend/sessions/sessionworkflows/workflow.go
+++ b/internal/backend/sessions/sessionworkflows/workflow.go
@@ -94,14 +94,6 @@ func runWorkflow(
 	return nil
 }
 
-func (w *sessionWorkflow) addPrint(ctx workflow.Context, print string) {
-	w.z.Debug("add print", zap.String("print", print))
-
-	if err := workflow.ExecuteLocalActivity(ctx, w.ws.svcs.DB.AddSessionPrint, w.data.SessionID, print).Get(ctx, nil); err != nil {
-		w.z.Panic("add print", zap.Error(err))
-	}
-}
-
 func (w *sessionWorkflow) updateState(ctx workflow.Context, state sdktypes.SessionState) {
 	w.z.Debug("update state", zap.Any("state", state))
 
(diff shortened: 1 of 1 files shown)


internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -389,19 +390,37 @@ func (w *sessionWorkflow) run(wctx workflow.Context) (prints []string, err error
 	cbs := sdkservices.RunCallbacks{
 		NewRunID: newRunID,
 		Load:     w.load,
-		Call: func(callCtx context.Context, rid sdktypes.RunID, v sdktypes.Value, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
-			if callCtx.Value(workflowContextKey) == nil {
+		Call: func(callCtx context.Context, runID sdktypes.RunID, v sdktypes.Value, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
+			if xid := v.GetFunction().ExecutorID(); xid.ToRunID() == runID && w.executors.GetCaller(xid) == nil {
+				// This happens only when we are during initial evaluation (the first run because invoking the entrypoint function),
+				// and the runtimes tries to call itself in order to start an activity with its own functions.
```
daabr:
```suggestion
				// This happens only during initial evaluation (the first run because invoking the entrypoint function),
				// and the runtime tries to call itself in order to start an activity with its own functions.
```

daabr (commented):
Other than the nit typo fixes, `updateState` is too long and unreadable. A hint for that is the fact that you're using multiple contexts with different names that aren't really clear.

To fix all this, please split the closures into real separate (private) functions, and call these functions intead.

internal/backend/sessions/sessionworkflows/workflow.go
itayd: fixed

daabr (approved):

PR #105, opened 2024-03-18
decouple deployments and sessions; sessions tests

itayd:
- deployment and sessions are decoupled: while a deployment can be specified for a session, it does not require it. session depends only on build and env.
- change cli to accommodate for that
- add sessions test functionality a-la runtimes test. in the future will be converted into unit tests (ENG-545).
- removed foreign key to deployments from sessions. need to add a constraint instead (ENG-547).

btw, i would have loved to break this PR down a bit - but it's not too bad. most of the lines are proto gen. also breaking it up is a bit problematic in that there are changes in cmd that would have needed to change again later, so isn't worth it.


proto/autokitteh/sessions/v1/session.proto
```diff
@@ -106,8 +106,9 @@ message SessionLog {
 
 message Session {
   string session_id = 1;
-  string deployment_id = 2 [(buf.validate.field).string.min_len = 1];
-  string event_id = 3; // TODO: eliminate? move to inputs? or memo?
+  string build_id = 2;
+  string env_id = 3;
```
ig-ra, line 110: maybe I'm missing somethifn - there is event_id and deployment_id in validate(). Will they be empty?

sdk/sdktypes/session.go
```diff
@@ -32,10 +34,10 @@ func (SessionTraits) Validate(m *SessionPB) error {
 func (SessionTraits) StrictValidate(m *SessionPB) error {
 	return errors.Join(
 		mandatory("created_at", m.CreatedAt),
-		mandatory("deployment_id", m.DeploymentId),
 		mandatory("entrypoint", m.Entrypoint),
```
ig-ra, line 37: why env_id isn't mandatory?

sdk/sdktypes/session.go
```diff
@@ -67,15 +72,37 @@ func (p Session) WithInputs(inputs map[string]Value) Session {
...
-			ParentSessionId: parentSessionID.String(),
+			BuildId:    buildID.String(),
+			Entrypoint: ToProto(ep),
+			Inputs:     kittehs.TransformMapValues(inputs, ToProto),
+			Memo:       memo,
 		},
 	))
 }
+
+func (s Session) WithParentSessionID(id SessionID) Session {
```
ig-ra, line 86: just curious - why do we need parent session id?

sdk/sdktypes/session_state_type.go
```diff
@@ -38,3 +38,7 @@ func SessionStateTypeFromProto(e sessionsv1.SessionStateType) (SessionStateType,
 func ParseSessionStateType(raw string) (SessionStateType, error) {
 	return ParseEnum[SessionStateType](raw)
 }
+func (e SessionStateType) IsFinal() bool {
```
ig-ra, line 41: maybe cancelled as well?

tests/sessions/run.sh
```diff
@@ -0,0 +1,78 @@
...
+fi
+
+logdir="tmp/testcli"
+
+rm -fR "${logdir}"
+mkdir -p "${logdir}"
+
+addr_filename="/tmp/addr-$$"
+
+up() {
```
ig-ra, line 29: are we using this method in other scripts? could we extract common parts and source?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -345,17 +345,16 @@ func ParseSessionCallAttemptComplete(c SessionCallAttempt) (d sdktypes.SessionCa
...
+	BuildID          string `gorm:"index"`
+	EnvID            string `gorm:"index"`
+	DeploymentID     string `gorm:"index"`
 	EventID          string `gorm:"index"`
 	CurrentStateType int    `gorm:"index"`
 	Entrypoint       string
 	Inputs           datatypes.JSON
 	CreatedAt        time.Time      //`gorm:"default:current_timestamp"`
 	UpdatedAt        time.Time      //`gorm:"default:current_timestamp"`
 	DeletedAt        gorm.DeletedAt `gorm:"index"`
```
ig-ra, line 357: let's add a Note/TODO/FIXME regarding what should be foreign key and what should be a constraint?
and assign a ticket

internal/backend/db/dbgorm/sessions.go
```diff
@@ -67,6 +67,8 @@ func (db *gormdb) CreateSession(ctx context.Context, session sdktypes.Session) e
 
 	s := scheme.Session{
 		SessionID:        session.ID().String(),
+		BuildID:          session.BuildID().String(),
+		EnvID:            session.EnvID().String(),
```
ig-ra, line 71: adding a comment here since I don't know how to add a comment on unchanged parts :)
what about listSessions and filters?
- maybe we should add to the filter buildID and envID?
- sessions without deployment will remain undeleted, and will be shown always in list. So maybe we should delete sessions via deleteBuild now?

examples/testprograms/http/http.star
```diff
@@ -7,7 +7,7 @@ def foo(x):
     return "meow, " + x
 
 def on_http_get():
-    resp, err = http1.get("https://httpbin3212.org/status/404", ak.callopts(catch=True))
+    resp, err = http1.get("https://httpbin.org/status/404", ak.callopts(catch=True))
```
ig-ra, line 10: Is there any explanation to the user about all this catch/callopts things?

cmd/ak/cmd/sessions/start.go
```diff
@@ -15,70 +14,53 @@ import (
...
 )
 
 var (
-	entryPoint   string
-	memos        []string
-	wait         bool
-	waitInterval time.Duration
+	entryPoint string
+	memos      []string
+	build      string
```
ig-ra: inconsistent naming. buildID?

cmd/ak/cmd/sessions/start.go
```diff
@@ -15,70 +14,53 @@ import (
...
+	deployID   string
 )
 
 var startCmd = common.StandardCommand(&cobra.Command{
-	Use:   "start <--deployment-id=...> <--event-id=...> <--entrypoint=...> [--memo=...] [--wait]",
+	Use:   "start [--build-id=...] [--env=...] [--deployment-id=...] <--entrypoint=...> [--memo=...] [--watch] [--watch-timeout=...] [--poll-interval=...] [--no-timestamps] [--quiet]",
```
ig-ra, line 28: it feels that according to the command description all 3 are optional: bid, env, did
maybe < did | bid env> ?
is env optional if bid is specified?

cmd/ak/cmd/sessions/start.go
```diff
@@ -15,70 +14,53 @@ import (
...
 		}
-		if !e.IsValid() {
-			err = fmt.Errorf("event ID %q not found", eventID)
-			return common.NewExitCodeError(common.NotFoundExitCode, err)
-		}
 
-		ep, err := sdktypes.StrictParseCodeLocation(entryPoint)
-		if err != nil {
-			return fmt.Errorf("invalid entry-point %q: %w", entryPoint, err)
+		if !ep.IsValid() {
```
ig-ra, line 46: maybe to move this check in sessionsArgs as well?

sdk/sdktypes/session.go
itayd: because we don't have to have envs. if we don't have them, we wont use env vars and connections.

sdk/sdktypes/session.go
itayd: there is a way to start a session from a session.

sdk/sdktypes/session_state_type.go
itayd: probably, but in a different pr.

tests/sessions/run.sh
itayd: there is a ticket to turn these into a unit test, so it'll be done there.

examples/testprograms/http/http.star
itayd: @haimzlato will work on it.

internal/backend/db/dbgorm/scheme/scheme.go
itayd: done

cmd/ak/cmd/sessions/start.go
itayd: fixed

cmd/ak/cmd/sessions/start.go
itayd: it doesn't have to be specified in other use of sessionArgs.

internal/backend/db/dbgorm/sessions.go
itayd: don't think we should delete sessions from delete builds. imo we shouldn't be deleting anything, we'll discuss.

envID is already in the filter. I'll add build as well.

cmd/ak/cmd/sessions/start.go
itayd: @daabr what do you usually do here?

ig-ra (approved):

cmd/ak/cmd/sessions/start.go
itayd: going to merge this, @daabr i'll fix this according to whatever you say

PR #108, opened 2024-03-19
starlark: partial

itayd:
not very different (if at all) from https://docs.python.org/3/library/functools.html#functools.partial.

allows:
```
def dump(*args, **kwargs):
    print(args)
    print(sorted(kwargs.items()))
    return len(args) + len(kwargs)

f = partial(dump, 1, 2, a=4, b=5)

print(f())
print(f(3, c=6))
print(f(1, b=9))
```

### DIFF
--- a/runtimes/starlarkrt/internal/bootstrap/bootstrap.star
+++ b/runtimes/starlarkrt/internal/bootstrap/bootstrap.star
@@ -69,3 +69,10 @@ def activity(fn):
     return struct(
         run=lambda *args, **kwargs: run_activity(fn, *args, **kwargs),
     )
+
+# EXPORT: withargs
+def withargs(f, *args, **kwargs):
+    def wrapper(*args2, **kwargs2):
+        kwargs.update(kwargs2)
+        return f(*(args + args2), **kwargs)
+    return wrapper
--- a/runtimes/starlarkrt/internal/bootstrap/bootstrap.star
+++ b/runtimes/starlarkrt/internal/bootstrap/bootstrap.star
@@ -70,8 +70,8 @@ def activity(fn):
         run=lambda *args, **kwargs: run_activity(fn, *args, **kwargs),
     )
 
-# EXPORT: withargs
-def withargs(f, *args, **kwargs):
+# EXPORT: partial
+def partial(f, *args, **kwargs):
     def wrapper(*args2, **kwargs2):
         kwargs.update(kwargs2)
         return f(*(args + args2), **kwargs)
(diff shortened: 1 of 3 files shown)


runtimes/starlarkrt/internal/bootstrap/bootstrap.star
```diff
@@ -69,3 +69,10 @@ def activity(fn):
     return struct(
         run=lambda *args, **kwargs: run_activity(fn, *args, **kwargs),
     )
+
+# EXPORT: partial
+def partial(f, *args, **kwargs):
+    def wrapper(*args2, **kwargs2):
```
tebeka, line 75: Not sure if starlark has it, but maybe do  what the [wraps](https://docs.python.org/3/library/functools.html#functools.wraps) decorator does? (Copy `__doc__` and other meta)

runtimes/starlarkrt/internal/bootstrap/bootstrap.star
```diff
@@ -69,3 +69,10 @@ def activity(fn):
     return struct(
         run=lambda *args, **kwargs: run_activity(fn, *args, **kwargs),
     )
+
+# EXPORT: partial
+def partial(f, *args, **kwargs):
```
tebeka, line 74: Why is that required?

runtimes/starlarkrt/internal/bootstrap/bootstrap.star
```diff
@@ -69,3 +69,10 @@ def activity(fn):
...
+    def wrapper(*args2, **kwargs2):
+        kwargs.update(kwargs2)
```
tebeka, line 76: In Python this will mutate the original `kwargs` passed in line 74. It's safer to do a copy.

tebeka (approved):
Can merge. See my comment on copying `kwargs`

efiShtain:
I don't see any value in this,
workflow developers can handle not using advanced programming languages capabilities like partials
especially where they can just create a closure in starlark and do what ever they want in python

runtimes/starlarkrt/internal/bootstrap/bootstrap.star
itayd: huh interesting, but i don't think starlark has it.

runtimes/starlarkrt/internal/bootstrap/bootstrap.star
itayd: motivation: https://github.com/autokitteh/autokitteh/pull/97#discussion_r1527618164

PR #111, opened 2024-03-21
fix: disable foreign keys for automigrate

ig-ra:
due to limitations in `alter` commands, GORM SQLite implementation moves data to temp tables and then recreates schema, table by table. Thus On DROP of original table (in persistent DB) we will fail on foreign key constrains.

Solution is to temporarily disable foreign keys for migration


efiShtain (approved):
LGTM

PR #112, opened 2024-03-21
Remove obsolete docs - transfered to docs.autokitteh.com

daabr:
https://docs.autokitteh.com/config/integrations


itayd (approved):
lgtm, but i think we need to leave a README here instructing where to find the docs.

PR #113, opened 2024-03-21
Dev mode's default log level: info -> debug

daabr:


internal/backend/logger/logger.go
```diff
@@ -15,7 +15,7 @@ type Config struct {
 	Zap zap.Config `koanf:"zap"`
 }
 
-var defaultZapLevel = zap.NewAtomicLevelAt(zap.InfoLevel)
+var defaultZapLevel = zap.NewAtomicLevelAt(zap.DebugLevel)
```
itayd, line 18: see WithDebug, looks like we're using it somewhere and it would do nothing now. Maybe we can just pass '--debug' or something instead?

efiShtain: why would it do nothing?
I agree with Daniel
this should be default to info in normal mode and debug in dev mode

daabr: After checking (https://github.com/search?q=repo%3Aautokitteh%2Fautokitteh%20WithDebug&type=code), it looks like `WithDebug` isn't actually used anywhere, so I'm removing it.

itayd (approved):

PR #106, opened 2024-03-19
stop sessions

itayd:
- implement session stop, graceful and non-graceful.
- add STOPPED state

PR #115, opened 2024-03-23
allow to specify cel expressions in triggers

itayd:
allows us to more precisely specify what to trigger on. this makes event_type optional. if event_type and filter are both specifies, event is filtered against both.

this also changes the event waiting logic (subscribe/next_event) work with cel only.

tests TBD after the sessions test in #106 are in.


sdk/sdktypes/event_test.go
```diff
@@ -0,0 +1,41 @@
...
+func TestEventFilter(t *testing.T) {
+	e := kittehs.Must1(sdktypes.EventFromProto(
+		&sdktypes.EventPB{
+			Data: map[string]*valuev1.Value{
+				"foo": sdktypes.NewStringValue("meow").ToProto(),
+			},
+		},
+	))
+
+	matches, err := e.Matches("")
```
daabr, line 22: Lines 22-40 look like a natural fit for a table-based test.

internal/manifest/manifest.go
```diff
@@ -64,14 +64,25 @@ type Trigger struct {
 	EnvKey string `yaml:"-" json:"-"` // associated with env.
 
 	ConnectionKey string `yaml:"connection" json:"connection" jsonschema:"required"` // coming from connection.
-	EventType     string `yaml:"event_type" json:"event_type" jsonschema:"required"`
+	EventType     string `yaml:"event_type" json:"event_type"`
 	Entrypoint    string `yaml:"entrypoint" json:"entrypoint" jsonschema:"required"`
+	Filter        string `yaml:"filter,omitempty" json:"filter,omitempty"`
 }
 
 func (t Trigger) GetKey() string {
```
daabr, line 72: No need for this in this PR, but FYI `Key()` is a better and more idiomatic function name than `GetKey()`.

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -251,10 +251,16 @@ func (w *sessionWorkflow) initGlobalModules() (map[string]sdktypes.Value, error)
...
 
-func (w *sessionWorkflow) createEventSubscription(ctx context.Context, connectionName, eventType string) (string, error) {
+func (w *sessionWorkflow) createEventSubscription(ctx context.Context, connectionName, filter string) (string, error) {
 	wctx := sessioncontext.GetWorkflowContext(ctx)
 	workflowID := workflow.GetInfo(wctx).WorkflowExecution.ID
-	signalID := fmt.Sprintf("wid_%s_cn_%s_et_%s", workflowID, connectionName, eventType)
+
+	var minSequence uint64
+	if err := workflow.ExecuteLocalActivity(wctx, w.ws.svcs.DB.GetLatestEventSequence).Get(wctx, &minSequence); err != nil {
+		w.z.Panic("get current sequence", zap.Error(err))
```
daabr: I see you just moved this block from below, so it's not a deal-breaker, but I'm curious: since the `createEventSubscription` function can return errors, why do you convert errors from `ExecuteLocalActivity` to panics? Same question for the panics below in lines 275 and 280.

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -73,11 +73,20 @@ func (d *dispatcher) getEventSessionData(ctx context.Context, event sdktypes.Eve
...
 
-		if triggerEventType != eventType {
+		if triggerEventType != "" && eventType != triggerEventType {
 			z.Debug("irrelevant event type")
 			continue
 		}
 
+		if relevant, err := event.Matches(t.Filter()); err != nil {
+			z.Debug("filter error", zap.Error(err))
+			// TODO: alert user their filter is bad. Integrate with alerting and monitoring.
```
daabr: Is this an actual task, or "someone needs to think about this sometime in the future"? If it's the first, please add an issue ID.

Either way, alerting and monitoring are for the server, not for the user, no? What if this is happening on AK cloud?

daabr (commented):
A few comments and questions - please respond to the questions and ping me, and then I'll approve.

General comments for the PR:

1. Where/how is usage of CEL in filter documented, i.e. how can users know **how** to use this field exactly?
2. It's a safe (i.e. good) approach to support both `event_type` and `filter` in tandem, but please add tickets to:
   - Fix all samples and doc'n to use only filters - for consistency, and role-modeling of the recommended usage
   - Deprecate `event_type` - the longer it lingers, the harder it will be to get rid of it

sdk/sdktypes/event_test.go
itayd: not worth it for such a short test

internal/manifest/manifest.go
itayd: ack

internal/backend/sessions/sessionworkflows/workflow.go
itayd: you're right, these shouldn't be panics. i guess the intent was to make the workflow retry, but since the failures are activity failures, i think we should be good. fixed.

internal/backend/dispatcher/eventsworkflow.go
itayd: these failures might be a result of a misconfiguration by the user, even in multi tenant when the filter is bad. we should alert the user (consumer in this case?) that they should check it. will add an issue.

sdk/sdktypes/event_test.go
daabr: On the contrary, it's a waste to write each case separately, but I'll approve.

daabr (approved):

PR #117, opened 2024-03-24
fix: prevent orphaned temporal processes

ig-ra:
kill both ak and temporal processes when running ak as svcproc

### DIFF
--- a/tests/internal/svcproc/svcproc.go
+++ b/tests/internal/svcproc/svcproc.go
@@ -109,7 +109,7 @@ func (s *svcProc) Stop(ctx context.Context) error {
 		err = syscall.Kill(-pgid, syscall.SIGTERM)
 	}
 	if err != nil { // if anything was wrong just kill ak
-		s.cmd.Process.Kill()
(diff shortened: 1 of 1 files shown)


tests/internal/svcproc/svcproc.go
```diff
@@ -100,7 +103,17 @@ func (s *svcProc) Start(ctx context.Context) error {
...
 func (s *svcProc) Stop(ctx context.Context) error {
 	// TODO: wait until verified stopped?
-	return s.cmd.Process.Kill()
+
+	pgid, err := syscall.Getpgid(s.cmd.Process.Pid)
+	if err == nil {
+		err = syscall.Kill(-pgid, syscall.SIGTERM)
+	}
+	if err != nil { // if anything was wrong just kill ak
+		s.cmd.Process.Kill()
```
efiShtain: you need to handle this error as well

ig-ra: by design.

efiShtain (approved):

PR #119, opened 2024-03-25
new manifest scheme for http

itayd:
- add `http` section to project in a manifest. this allows to directly declare routes. these translate into a `default_http` connection and appropriate triggers.
- add optional `name` field to all triggers. this helps to identify the trigger fired by the program and also allows the manifest planning to determine which trigger exactly is referenced.
- if `default_http` connection exists and `http` does not, the session will pass it to the runtime as `http`. `http` is always put into the global namespace, so load is not required for it.
- http server route prefix is now `project_name/env_name`. this also changes the dispatcher method to have that as an input if desired. killed the need for http token. 

on the way (= en passant, but sounds a bit less snobby):
- kill `kittehs.JoinMaps`, use `maps.Copy` instead.
- values unwrap into can now ignore fields that do not exist in target struct

TODO:
- Tests, will be done before merge.


examples/testprograms/http/http.star
```diff
@@ -1,39 +1,5 @@
...
-    n += 1000
-    print("zzzz3", n, time.now())
-    sleep(3)
-    resp = http1.get("http://example.com")
-    print(resp)
+def on_http_get_meow(data, trigger, event):
+    print(data)
+    print(trigger)
+    print(event)
+    print(http.get("http://example.com/{}".format(trigger.data.params['who'])))
```
ig-ra, line 5: there were many several examples with ak_catch, ak = , ak.callopts.
are we supporting them? should we leave them  in the code for reference?

integrations/http/server.go
```diff
@@ -23,16 +29,21 @@ type HTTPHandler struct {
...
 }
 
 func (h HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	l := h.logger.With(zap.String("url", r.URL.String()))
 
-	l.Info("Incoming request")
+	l.Info("incoming request")
+
+	ns := r.PathValue("ns")
+	env := strings.ReplaceAll(ns, ".", "/")
```
ig-ra, line 41: any possible problem with consequent dots and thus consequent slashes?

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -24,6 +32,62 @@ func (d *dispatcher) createEventRecord(ctx context.Context, eventID sdktypes.Eve
...
 }
 
+func (d *dispatcher) resolveEnv(ctx context.Context, env string) (envID sdktypes.EnvID, err error) {
+	if env == "" {
+		return sdktypes.InvalidEnvID, nil
+	}
+
+	parts := strings.Split(env, "/")
+	switch len(parts) {
+	case 1:
```
ig-ra, line 42: since there are only 2 cases, I would suggest to use simple `if` and remove nesting for the second part

examples/testprograms/http/http.star
itayd: there are appropiate tests for these in `tests/runs` and `tests/sessions`.

itayd:
breaking this down and changing scheme.

PR #120, opened 2024-03-26
DRAFT: user and token auth

itayd:
Implements user and token auth. Users need to be created before they are allowed to log in. Also kill svc, and basesvc -> svc to simplify.

This does:
- allows to create tokens and use them.
- protects grpc and http.

This doesn't:
- associate any object to a user. it only allowed to log in.

Not tested yet.


proto/autokitteh/users/v1/svc.proto
```diff
@@ -0,0 +1,83 @@
...
+
+message DeleteRequest {
+  string user_id = 1;
+}
+
+message DeleteResponse {}
+
+message GetRequest {
+  option (buf.validate.message).cel = {
+    id: "user_id_xor_name",
```
efiShtain: do we really need this logic here instead of just get by ID like any other types ?
we can have a getByName method instead of xor logic

proto/autokitteh/users/v1/svc.proto
```diff
@@ -0,0 +1,83 @@
...
+  };
+  option (buf.validate.message).cel = {
+    id: "user.name_required",
+    message: "name must be specified",
+    expression: "has(this.user) && this.user.name != ''",
+  };
+
+  // org.org_id is ignored.
+  User user = 1 [(buf.validate.field).required = true];
+}
```
efiShtain: I think we should stop with this pattern
if all we require to create a user is username, we should only get username and not a user object with buf validation on other fields like id

proto/autokitteh/users/v1/svc.proto
```diff
@@ -0,0 +1,83 @@
...
+message UpdateRequest {
+  option (buf.validate.message).cel = {
+    id: "user.user_id_required",
+    message: "user_id must be specified",
+    expression: "has(this.user) && this.user.user_id == ''",
+  };
+
+  User user = 1 [(buf.validate.field).required = true];
+  google.protobuf.FieldMask field_mask = 2;
+}
```
efiShtain: what is this ?

proto/autokitteh/users/v1/svc.proto
```diff
@@ -0,0 +1,83 @@
...
+
+message GetResponse {
+  User user = 1;
+}
+
+message BatchGetRequest {
+  repeated string user_ids = 1;
+  google.protobuf.FieldMask field_mask = 2;
+}
```
efiShtain: in which case would you batch get users ?

sdk/sdkclients/sdkusersclient/client.go
```diff
@@ -0,0 +1,77 @@
...
+	}))
+	if err != nil {
+		return sdktypes.InvalidUserID, sdkerrors.AsConnectError(err)
+	}
+
+	if err := internal.Validate(resp.Msg); err != nil {
+		return sdktypes.InvalidUserID, err
+	}
+
+	oid, err := sdktypes.Strict(sdktypes.ParseUserID(resp.Msg.UserId))
```
efiShtain: not sure the client should parse the user id, it shouldn't exists in this request

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -0,0 +1,166 @@
...
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type Config struct {
+	Required bool `koanf:"required"`
+
+	UseAnonymous bool `koanf:"use_anonymous"` // set user as "anonymous" if !required and no auth given.
+
+	AllowFakeTestAuth bool `koanf:"allow_fake_test_auth"` // allow passing raw user names in auth header for testing.
```
efiShtain: I don't like this parameter, why do we need it ?

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -0,0 +1,166 @@
...
+	Default: &Config{
+		Required: true,
+	},
+	Dev: &Config{},
+	Test: &Config{
+		AllowFakeTestAuth: true,
+	},
+}
+
+type WrapFunc func(http.Handler) http.Handler
```
efiShtain, line 24: is this exposed intentionally ? cause if so the name is not very clear

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -0,0 +1,166 @@
...
+func newTokensMiddleware(next http.Handler, tokens authtokens.Tokens) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		if userID := authcontext.GetAuthnUserID(ctx); !userID.IsValid() {
+			if auth := r.Header.Get("Authorization"); auth != "" {
+				kind, payload, _ := strings.Cut(auth, " ")
+
+				switch kind {
+				case "Bearer":
```
efiShtain, line 43: do you plan to support other types of tokens ?
I think we should stick with Bearer for now, no need for switch on kind

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -0,0 +1,166 @@
...
+}
+
+func newSessionsMiddleware(next http.Handler, sessions authsessions.Store) http.HandlerFunc {
+	return func(w http.ResponseWriter, r *http.Request) {
+		ctx := r.Context()
+
+		if userID := authcontext.GetAuthnUserID(ctx); !userID.IsValid() {
+			session, err := sessions.Get(r)
+			if err != nil {
+				http.Redirect(w, r, fmt.Sprintf("/error.html?err=%s", url.QueryEscape(err.Error())), http.StatusTemporaryRedirect)
```
efiShtain: not sure this is still valid, need to check

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -0,0 +1,166 @@
...
+		if deps.Sessions != nil {
+			f = newSessionsMiddleware(f, deps.Sessions)
+		}
+
+		if deps.Tokens != nil {
+			f = newTokensMiddleware(f, deps.Tokens)
+		}
+
+		if deps.Cfg.AllowFakeTestAuth && deps.Users != nil {
+			f = newFakeTestAuth(f, deps.Users)
```
efiShtain: this should really be set by fx and not a configuration option for everyone to use

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -472,3 +474,26 @@ type Signal struct {
...
+}
+
+func ParseUser(r User) (sdktypes.User, error) {
+	return sdktypes.StrictUserFromProto(&sdktypes.UserPB{
+		UserId: r.UserID,
+		Name:   r.Name,
+	})
+}
+
+type UserExternalIdentitiy struct {
```
efiShtain: I wanted to rename this to UserIdentity so we can use it for native user as well

PR #126, opened 2024-03-27
Ignore Temporal broken pipe errors

daabr:
As noted in the changed code: there's no useful information and no recourse for an error at that point. It only confuses people when they see this unrelated error at the end of system tests.

Also, convert an unnecessary closure to a regular function.

### DIFF
--- a/internal/backend/temporalclient/client.go
+++ b/internal/backend/temporalclient/client.go
@@ -88,7 +88,7 @@ func New(cfg *Config, z *zap.Logger) (Client, error) {
 					return nil, fmt.Errorf("temporal client: %w", err)
 				}
 
-				z.Info("cannot connect to temporal, starting dev server")
+				z.Info("Cannot connect to Temporal, starting Temporal dev server")
 
 				if err := impl.startDevServer(ctx, cfg, opts); err != nil {
 					return nil, err
@@ -107,7 +107,7 @@ func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Opti
 	if c.srv, err = testsuite.StartDevServer(ctx, cfg.DevServer); err != nil {
 		return fmt.Errorf("start Temporal dev server: %w", err)
 	}
-	c.z.Info("started Temporal dev server", zap.String("address", c.srv.FrontendHostPort()))
+	c.z.Info("Started Temporal dev server", zap.String("address", c.srv.FrontendHostPort()))
 
 	c.client = c.srv.Client()
 
@@ -162,8 +162,13 @@ func (c *impl) Start(context.Context) error {
 	}
 
 	go func() {
+		ok := false
 		for {
-			if err := c.healthCheck(context.Background()); err != nil {
+			err := c.healthCheck(context.Background())
+			if err == nil && !ok {
+				c.z.Info("Connection to Temporal is healthy")
+				ok = true
+			} else if err != nil {
 				// TODO: stats.
 				c.z.Error("Temporal health check error", zap.Error(err))
 			}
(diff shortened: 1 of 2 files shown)


internal/backend/temporalclient/client.go
```diff
@@ -98,15 +82,15 @@ func New(cfg *Config, z *zap.Logger) (Client, error) {
...
 		if cfg.StartDevServerIfNotUp {
-			if err := impl.healthcheck(ctx); err != nil {
+			if err := impl.healthCheck(ctx); err != nil {
 				var unavailable *serviceerror.Unavailable
 				if !errors.As(err, &unavailable) {
 					return nil, fmt.Errorf("temporal client: %w", err)
 				}
 
-				z.Info("cannot connect to server, starting dev server")
+				z.Info("cannot connect to temporal, starting dev server")
```
ig-ra: do you want to use `Temporal`  or `temporal`?

internal/backend/temporalclient/client.go
```diff
@@ -116,21 +100,45 @@ func New(cfg *Config, z *zap.Logger) (Client, error) {
...
 }
 
+func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Options) error {
+	cfg.DevServer.ClientOptions = &opts
+
+	var err error
+	if c.srv, err = testsuite.StartDevServer(ctx, cfg.DevServer); err != nil {
+		return fmt.Errorf("start Temporal dev server: %w", err)
+	}
+	c.z.Info("started Temporal dev server", zap.String("address", c.srv.FrontendHostPort()))
```
ig-ra: here Temporal starts with capital `T` and in several places above with `t`

internal/backend/temporalclient/client.go
```diff
@@ -142,28 +150,22 @@ func (c *impl) healthcheck(ctx context.Context) error {
...
 		for {
-			if err := c.healthcheck(context.Background()); err != nil {
+			if err := c.healthCheck(context.Background()); err != nil {
 				// TODO: stats.
-				ok = false
-				c.z.Error("temporal check health error", zap.Error(err))
-			} else if !ok {
-				ok = true
-				c.z.Info("temporal reports healthy")
+				c.z.Error("Temporal health check error", zap.Error(err))
```
ig-ra, line 173: IMO, in terms of logs it's nice to have what u've removed - the single report that connection is back.

internal/backend/temporalclient/client.go
```diff
@@ -116,21 +100,45 @@ func New(cfg *Config, z *zap.Logger) (Client, error) {
...
+
+	return nil
+}
+
 func (c *impl) Temporal() client.Client { return c.client }
 
 func (c *impl) Stop(context.Context) error {
 	close(c.done)
+
+	if c.client != nil {
```
ig-ra, line 122: could c.client be nil?

ig-ra (approved):

internal/backend/temporalclient/client.go
daabr: Standardized "Temporal" in all log messages.

internal/backend/temporalclient/client.go
daabr: See previous comment

internal/backend/temporalclient/client.go
daabr: Technically yes - right at the beginning. Anyway, it's smarter to check rather than assume, in order to avoid a panic.

internal/backend/temporalclient/client.go
daabr: I get your point, added it back.

PR #128, opened 2024-03-28
feat(db): hard- and soft-foreign keys constraints

ig-ra:
This PR adds foreign and soft-foreign keys constrains to the model (ENG-547)
Soft foreign key is actual foreign key pointing to existing asset or NULL


tebeka:
Maybe we should stop using foreign keys? See [here](https://code.openark.org/blog/mysql/things-that-dont-work-well-with-mysqls-foreign-key-implementation#more-8226) for some reasoning (the author was responsible for GitHubs MySQL infra among other things).

internal/backend/db/dbgorm/deployments_test.go
```diff
@@ -167,3 +179,9 @@ func TestDeleteDeployment(t *testing.T) {
 	// TODO: meanwhile builds are not deleted when deployment is deleted
 	// assertBuildDeleted(t, f, b.BuildID)
 }
+
```
tebeka, line 181: Delete commented out code.

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -17,6 +17,21 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+// aux methods to convert nil <-> empty strting
+func PtrOrNil(s string) *string {
```
tebeka, line 21: Why not use [sql.NullString](https://pkg.go.dev/database/sql#NullString) ?

ig-ra: *string is simpler in usage. sql.NullString will do the same in background, but require more changes in the code. Anyway we need to convert empty string from sdktypes to ptr-like type in DB, but access to *string is direct, while for sql.NullString you need to use sql.NullString.String

ig-ra:
@tebeka 
> Maybe we should stop using foreign keys? 
we are using soft foreign keys where possible. Moreover there are no cascade deletes and most of the asserts are soft deleted. Hard foreign keys defined only where appropriate, e.g. for `SessionLogRecord` to the `SessionID`, since the record shouldn't exist without session and session is soft-deleted. So it mostly used to ensure valid insertion and proper data and not for cascading deletion. I prefer foreign keys to prevent us from inserting inconsistent data into DB

efiShtain (approved):

PR #129, opened 2024-03-28
ENG-575: Fix OAuth redirect for GitHub Enterprise Server

daabr:
This change ensures that the value in the env var "GITHUB_ENTERPRISE_URL" is a valid URL, i.e. with a scheme prefix (`http[s]://`), and without a suffix after the host address.

### DIFF
--- a/internal/backend/oauth/oauth.go
+++ b/internal/backend/oauth/oauth.go
@@ -19,6 +19,7 @@ import (
 	googleoauth2 "google.golang.org/api/oauth2/v2"
 	"google.golang.org/api/sheets/v4"
 
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 )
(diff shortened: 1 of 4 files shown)


internal/kittehs/strings.go
```diff
@@ -52,3 +53,29 @@ func PadLeft(s string, r rune, n int) string {
 
 	return strings.Repeat(string(r), n-len(s)) + s
 }
+
+// NormalizeURL ensures that the given URL has the right scheme
+// prefix, and no suffix (e.g. path) after the host address.
+func NormalizeURL(rawURL string, secure bool) (string, error) {
+	// Normalize the URL's scheme prefix.
+	scheme := "http://"
```
itayd, line 61: instead of parsing it yourself prior to Parse below, just let the lib parse it, then set the Scheme to whatever you want. Then convert to string.

internal/backend/oauth/oauth.go
```diff
@@ -45,6 +46,14 @@ func New(l *zap.Logger) sdkservices.OAuth {
 	if githubBaseURL == "" {
 		githubBaseURL = "https://github.com"
 	}
+	var err error
+	githubBaseURL, err = kittehs.NormalizeURL(githubBaseURL, true)
+	if err != nil {
+		l.Fatal("Invalid environment variable value",
+			zap.String("name", "GITHUB_ENTERPRISE_URL"),
```
itayd, line 53: don't you want to include the value?

internal/kittehs/strings.go
daabr: This won't work, if the scheme is missing in the rawURL, `url.Parse` will think that the host address is actually just a relative path. Handling this corner case is more complex and less readable.

internal/backend/oauth/oauth.go
daabr: I can, but the value is already included in the error message (see `strings.go` line 76), and I want to keep it in the error for other callers that return the error and don't print a log line, e.g. `jwt.go`. WDYT?

itayd (approved):

internal/kittehs/strings.go
itayd: sg

internal/backend/oauth/oauth.go
itayd: sg

PR #132, opened 2024-03-30
dispatcher options include env, not just envid

itayd:
- dispatcher option receives env, not just env id. this allows for integrations to direct events specifically to certain envs without knowing their id, without using tokens.
- http integration use of ^.
- token is not mandated in manifest and cmd.
- fix some old StrictParseXID to Strict(ParseXID).


integrations/http/server.go
```diff
@@ -23,16 +29,21 @@ type HTTPHandler struct {
...
 	h := HTTPHandler{dispatcher: d, logger: l}
-	mux.Handle(routePrefix, http.StripPrefix(routePrefix, h))
+	mux.Handle(routePrefix("{ns}")+"*", h)
 }
 
 func (h HTTPHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	l := h.logger.With(zap.String("url", r.URL.String()))
 
-	l.Info("Incoming request")
+	l.Info("incoming request")
```
daabr, line 38: We need to standardize on logging conventions throughout the code - levels, text, and args - but not a concern for this PR.

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -24,6 +25,62 @@ func (d *dispatcher) createEventRecord(ctx context.Context, eventID sdktypes.Eve
...
+	if env == "" {
+		return sdktypes.InvalidEnvID, nil
+	}
+
+	parts := strings.Split(env, "/")
+	switch len(parts) {
+	case 1:
+		if sdktypes.IsEnvID(parts[0]) {
+			return sdktypes.ParseEnvID(parts[0])
+		}
```
daabr: It's hard to follow the logic in this branch, i.e. need to skip to the end of the function. Why not call `return` here?

Also, why do we even support IDs? They're a purely internal implementation detail which users should not be exposed to or rely upon. It should be just names.

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -24,6 +25,62 @@ func (d *dispatcher) createEventRecord(ctx context.Context, eventID sdktypes.Eve
 	return nil
 }
 
+func (d *dispatcher) resolveEnv(ctx context.Context, env string) (envID sdktypes.EnvID, err error) {
```
daabr: Not a single comment in ~50 LOC. Please add comments inside this function that explain its flow.

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -24,6 +25,62 @@ func (d *dispatcher) createEventRecord(ctx context.Context, eventID sdktypes.Eve
...
+
+	parts := strings.Split(env, "/")
+	switch len(parts) {
+	case 1:
+		if sdktypes.IsEnvID(parts[0]) {
+			return sdktypes.ParseEnvID(parts[0])
+		}
+	case 2:
+		var pid sdktypes.ProjectID
+		if sdktypes.IsProjectID(parts[0]) {
```
daabr: Same as above - I don't think we should support IDs in URLs.

tests/system/testdata/connections/create.txtar
```diff
@@ -4,32 +4,28 @@ return code == 0
 output equals 'project_id: prj_00000000000000000000000001'
 
 # Negative test: create connection without name.
-ak connection create --project my_project --integration http --connection-token my_url_path
+ak connection create --project my_project --integration http --connection-token tok
```
daabr: Please use a human-friendly string like `"my_token"` instead of the meaningless abbreviation `"tok"` (in all instances below too).

tests/system/testdata/connections/create.txtar
```diff
@@ -4,32 +4,28 @@ return code == 0
...
 output equals 'Error: required flag(s) "integration" not set'
 return code == 1
 
-ak connection create my_connection --project my_project --integration http
-output equals 'Error: required flag(s) "connection-token" not set'
-return code == 1
-
 # Create connection.
-ak connection create my_connection --project my_project --integration http --connection-token my_url_path
+ak connection create my_connection --project my_project --integration http --connection-token tok
```
daabr: Please add another test for creating a connection without a token.

daabr (commented):
I want to see your comments in the `resolveEnv` function to understand it better, and then I'll approve.

internal/backend/dispatcher/eventsworkflow.go
itayd: why not?

internal/backend/dispatcher/eventsworkflow.go
itayd: simplified now. no ids.

internal/backend/dispatcher/eventsworkflow.go
itayd: fixed

internal/backend/dispatcher/eventsworkflow.go
itayd: scratch that, you're right.

tests/system/testdata/connections/create.txtar
itayd: done

daabr (approved):

PR #136, opened 2024-03-31
ENG-592: allow to specify http routes in triggers

itayd:
- triggers have a new field: `data`. arbitrary data can be specified there which the dispatcher can read.
- dispatcher knows to expect `data` for triggers of http integrations. it would parse that and match the incoming http request against a pattern.
- cripser definition of what passed to event hander: instead of just `data`, now also `trigger` and `event. `data` is now the combined `trigger.data` and `event.data`.


integrations/http/pattern.go
```diff
@@ -0,0 +1,64 @@
...
+	"net/http"
+	"net/http/httptest"
+	"net/url"
+	"strings"
+
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+)
+
+func extractKeys(s string) (keys []string, err error) {
```
daabr: 1. Missing documentation, so it's unclear to me what is expected input and what isn't valid / supposed to be possible
2. Because of that, I'm not sure if this loop is safe, or if you have to parse `{}` characters with a queue to understand your nesting level

integrations/http/pattern.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	if inside {
+		err = sdkerrors.NewInvalidArgumentError("unmatched '{'")
+	}
+
+	return
+}
+
+func MatchPattern(pattern string, req string) (vs map[string]string, err error) {
```
daabr, line 52: Missing documentation: when/how/why is this function used?

examples/testprograms/http/autokitteh.yaml
```diff
@@ -2,19 +2,19 @@ version: v1
...
-      token: "meow"
+    - name: myhttp
       integration: "http"
   triggers:
     - event_type: get
-      connection: http1
+      connection: myhttp
       entrypoint: http.star:on_http_get
+      data:
+        pattern: "get"
```
daabr: 1. What does "data" actually mean/contain?
2. What is the valid/expected contents of "pattern"?

internal/backend/dispatcher/special_triggers.go
```diff
@@ -0,0 +1,64 @@
+package dispatcher
+
+import (
+	"errors"
+
+	httpint "go.autokitteh.dev/autokitteh/integrations/http"
```
daabr, line 6: Why are you aliasing this package? I don't see any other instance of "http"

tests/system/testdata/endtoend/http.txtar
```diff
@@ -0,0 +1,128 @@
+ak manifest apply project.yaml
+return code == 0
+
+ak project build myproject --from main.star
+return code == 0
+output equals 'build_id: bld_00000000000000000000000006'
+
+ak deployment create --build-id=bld_00000000000000000000000006 --env=myproject/default --activate
+return code == 0
```
daabr, line 9: Nit: you can use "ak project deploy", because it's not critical for this PR because I'll replace "ak manifest apply" and this anyway with "ak deploy".

internal/backend/dispatcher/special_triggers.go
itayd: `http` is confusing, this makes sure it's not.

daabr (approved):

integrations/http/pattern.go
itayd: documented.

integrations/http/pattern.go
itayd: documented.

examples/testprograms/http/autokitteh.yaml
itayd: documented

PR #137, opened 2024-03-31
http integration: return body functions

itayd:
instead of returning `body_json`, `body_bytes` etc, which is... not pretty, we return now functions that return the appropriate data. These are functions and not just values since in the future we'll do it lazily and not just duplicating the payload.

For that purpose, "const function" constructors are implemented that just return their data as values.

### DIFF
--- a/examples/testprograms/http/http.star
+++ b/examples/testprograms/http/http.star
@@ -19,7 +19,8 @@ def on_http_get():
     print(resp, err)
     if err:
         print(err.op)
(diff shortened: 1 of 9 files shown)


daabr (approved):
Ok... I guess?

I'm not quite comfortable with the concept of passing constant functions to scripts, it feels a little... Shaky.

However, I acknowledge the need for this improvement in HTTP.

PR #138, opened 2024-03-31
ENG-598: auto create default env for project

efiShtain:
- Create a default env for project on project's creation
- Stop creating default env in manifest
- Remove double transalte error in transaction wrapper
- fix tests to match


internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -68,7 +68,7 @@ func (db *gormdb) locked(f func(db *gormdb) error) error {
 		defer db.mu.Unlock()
 	}
 
-	return translateError(f(db))
+	return f(db)
```
ig-ra: I think in transaction the error won't be translated in all cases

ig-ra (approved):

PR #140, opened 2024-03-31
Eng-574: trigger workflow GitHub integration

efiShtain:


integrations/github/client.go
```diff
@@ -246,5 +246,18 @@ func New(sec sdkservices.Secrets) sdkservices.Integration {
 			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/users#get-a-user"),
 			sdkmodule.WithArgs("username"),
 		),
+		// Actions
+		sdkmodule.ExportFunction(
+			"list_workflows",
+			i.listWorkflows,
+			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/actions/workflows#create-a-workflow-dispatch-event"),
```
daabr:
```suggestion
			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/actions/workflows#list-repository-workflows"),
```

integrations/github/client.go
```diff
@@ -246,5 +246,18 @@ func New(sec sdkservices.Secrets) sdkservices.Integration {
...
+			sdkmodule.WithArgs("owner", "repo"),
```
daabr, line 254: Can you also add "per_page" and "page"? (Both optional integers). I expect realistic usage of this will easily go over the default (1 page of 30 results).
Anyway, not as urgent as the core capability, so feel free to do it in a separate PR.

integrations/github/workflows.go
```diff
@@ -0,0 +1,67 @@
...
+
+	_, err = gh.Actions.CreateWorkflowDispatchEventByFileName(ctx, owner, repo, workflowName, req)
+	if err != nil {
+		return sdktypes.InvalidValue, err
+	}
+
+	return sdktypes.Nothing, nil
+}
+
+func (i integration) listWorkflows(ctx context.Context, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
```
daabr, line 44:
```suggestion
// https://docs.github.com/en/rest/actions/workflows#list-repository-workflows
func (i integration) listWorkflows(ctx context.Context, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
```

daabr (approved):
Approved, but see 2 tiny suggestions, and 1 request (not urgent for this specific PR).

integrations/github/client.go
efiShtain: done

PR #142, opened 2024-04-01
ENG-582: start temporal ui; add easy redirection to it

itayd:
- when starting temporal dev server, enable ui
- print temporal and autokitteh UI addresses in clickable format in terminal
- print ipv4 address, if possible, for autokitteh server

image


internal/backend/httpsvc/http.go
```diff
@@ -103,6 +103,18 @@ func New(lc fx.Lifecycle, z *zap.Logger, cfg *Config, reflectors []string, extra
 			}
 
 			svc.addr = ln.Addr().String()
+			if host, port, err := net.SplitHostPort(svc.addr); err == nil {
+				ip := net.ParseIP(host)
+
+				var ipStr string
+				if ip.IsUnspecified() {
+					ipStr = "localhost"
```
daabr: Nit: this is not an IP address. Use `addr` instead of `ipStr`?

internal/backend/basesvc/hello.go
```diff
@@ -17,26 +17,39 @@ var hello string
...
 	eyeColor := color.New(color.FgGreen).Add(color.Bold).SprintFunc()
 
 	var mode string
 	if opts.Mode != "" {
-		mode = "Mode:      " + fieldColor(opts.Mode) + " "
+		mode = "Mode:        " + fieldColor(opts.Mode) + " "
+	}
+
+	if temporalFrontendAddr != "" {
+		temporalFrontendAddr = "Temporal:    " + fieldColor(temporalFrontendAddr) + " "
```
daabr: Maybe "Temporal RPC" to clarify the difference from the UI? (If you do change this, make sure to realign spaces).

internal/backend/temporalclient/client.go
```diff
@@ -119,6 +122,36 @@ func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Opti
...
+		return
+	}
+
+	nport, err := strconv.Atoi(port)
+	if err != nil {
+		return
+	}
+
+	// temporal's default is frontend+1000 for dev server.
+	nport += 1000
```
daabr: Problem: this is the default, but it's not necessary. Temporal can be started with this flag:

```
--ui-port value
Port for the Web UI (default: --port + 1000, e.g. 8233)
```

Is there a way to get the actual value? If not, I'm not sure guessing is a good idea.

internal/backend/temporalclient/config.go
```diff
@@ -38,6 +38,11 @@ type Config struct {
 }
 
 var (
+	defaultTemporalDevServerConfig = testsuite.DevServerOptions{
+		LogLevel: zapcore.WarnLevel.String(),
+		EnableUI: true,
```
ig-ra: will it enable temporal UI in tests as well?
if so, maybe we want to disable UI in tests?

efiShtain: I think it should be an **env variable** with default true and set to false in tests mode

internal/backend/basesvc/hello.go
efiShtain: I think they just call it temporal and temporal-ui
temporalFrontend is really confusing since frontend == UI most of the times

internal/backend/basesvc/hello.go
```diff
@@ -17,26 +17,39 @@ var hello string
 
 var helloTemplate = template.Must(template.New("hello").Parse(hello))
 
-func sayHello(opts RunOptions, addr string) {
+func sayHello(opts RunOptions, addr, temporalFrontendAddr, temporalUIAddr string) {
```
efiShtain: can we rename this to a more reasonable name?
I thought sayHello is a test util or an hello world

internal/backend/basesvc/hello.txt
```diff
@@ -1,12 +1,12 @@
...
-  '                        `      
+        _..--~~~~--.._             autokitteh   {{ .Version }}
+     _--              --_          PID:         {{ .PID }}
+    '                    `         gRPC/HTTP:   {{ .Addr }}
+   '                      `        UI:          {{ .UIAddr }}
+  '                        `       {{ .Mode }}
  |                    |\   |       
- |                    | \  | 
- |          .,    _,-~   \ | 
+ |                    | \  |       {{ .Temporal1 }}
```
efiShtain: Temporal
TemporalUI 

1 and 2 are not informative

internal/backend/basesvc/svc.go
```diff
@@ -162,9 +162,20 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
...
 		}),
-		fx.Invoke(func(lc fx.Lifecycle, z *zap.Logger, httpsvc httpsvc.Svc) {
+		fx.Invoke(func(mux *http.ServeMux, tclient temporalclient.Client) {
+			_, uiAddr := tclient.TemporalAddr()
+
+			if uiAddr != "" {
+				mux.HandleFunc("/temporal", func(w http.ResponseWriter, r *http.Request) {
+					http.Redirect(w, r, uiAddr, http.StatusFound)
+				})
+			}
```
efiShtain: NIT, remove indentation, reverse the if condition
```suggestion
			if uiAddr == "" {
			   return 
			}
		        mux.HandleFunc("/temporal", func(w http.ResponseWriter, r *http.Request) {
				http.Redirect(w, r, uiAddr, http.StatusFound)
			})
			
```

internal/backend/temporalclient/client.go
```diff
@@ -119,6 +122,36 @@ func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Opti
 
 func (c *impl) Temporal() client.Client { return c.client }
 
+func (c *impl) TemporalAddr() (frontend, ui string) {
+	if c.srv == nil {
+		if frontend = c.cfg.HostPort; frontend == "" || frontend == "localhost" {
+			// known temporal defaults.
+			frontend = "localhost:7233"
+			ui = fmt.Sprintf("http://%s:%d", "localhost", 8233)
```
efiShtain:
```go
ui := "http://localhost:8233"
```
?

internal/backend/temporalclient/client.go
```diff
@@ -119,6 +122,36 @@ func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Opti
...
+	frontend = c.srv.FrontendHostPort()
+
+	host, port, err := net.SplitHostPort(frontend)
+	if err != nil {
+		return
+	}
+
+	nport, err := strconv.Atoi(port)
+	if err != nil {
+		return
```
efiShtain: you just hide the error here?

internal/backend/httpsvc/http.go
itayd: fixed

internal/backend/basesvc/hello.go
itayd: agree with @efiShtain here

internal/backend/temporalclient/client.go
itayd: Unfortunately, there is no way to get it, so there is no way around it.

internal/backend/temporalclient/config.go
itayd: modes only give you a set of defaults. you can override it using dynamic config. i will disable it for tests though.

internal/backend/temporalclient/client.go
itayd: ugh you're right

internal/backend/basesvc/hello.go
itayd: what name do you suggest?

internal/backend/basesvc/svc.go
itayd: done

internal/backend/temporalclient/client.go
itayd: there isn't much i can do here. this will cause the UI addr to be empty and not displayed.

internal/backend/basesvc/hello.txt
itayd: done

internal/backend/basesvc/hello.go
daabr: Me too regarding "frontend" - it's an unfortunate decision on their part, that's why I suggested "RPC". Anyway, if you don't want to say "RPC", "Temporal" is better than "Temporal Frontend".

internal/backend/basesvc/hello.go
efiShtain: ShowBanner, ShowServerConfig, DisplayConfig, PrintConfig, StartupMessage ...

internal/backend/temporalclient/client.go
efiShtain: Is it user config based error? If so I prefer to fail and let him know something went wrong rather than ignoring it's config

internal/backend/temporalclient/client.go
daabr: That's my point: we're only guessing, and we could be wrong, so why give this info to the user if it can mislead them?

Same thing about the scheme: it **may** be HTTP, but it can also be HTTPS if it's remote.

Other better options, in decreasing order of how much I like them:

1. Don't print a UI URL at all
2. Print it, but also say it's a guess
3. Check the guess with an HTTPS query (with HTTP as fallback) to `/healthz` (expect to receive `{"status":"OK"}`), and print it if we get a response, and "Unknown" otherwise

internal/backend/basesvc/hello.go
daabr: MOTD? (My age is showing...) :)

itayd: Ha, went for `banner`.

itayd: (it does not change daily :P )

internal/backend/temporalclient/client.go
itayd: tbh there is no way this will fail - it comes from the frontend address temporal returns. not worth the effort doing anything there. i could just put it down as "unknown".

internal/backend/temporalclient/client.go
itayd: it can't be https since it's coming from the dev server - where it will always run locally over http.
since it's always coming from the devserver, it will always be right.

i could do the probe against its address, but is it worth it? i mean it's an amusing idea, but still...

daabr: The banner is not limited to dev mode, and even in dev mode creating a Temporal dev server is optional not a requirement.

internal/backend/temporalclient/client.go
```diff
@@ -119,6 +122,36 @@ func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Opti
 
 func (c *impl) Temporal() client.Client { return c.client }
 
+func (c *impl) TemporalAddr() (frontend, ui string) {
+	if c.srv == nil {
+		if frontend = c.cfg.HostPort; frontend == "" || frontend == "localhost" {
```
itayd: check that frontend is localhost:7233

daabr (approved):

PR #143, opened 2024-04-01
bundle libs internaly; implement twilio over http lib

itayd:


libs/libs/integrations/twilio.star
```diff
@@ -0,0 +1,65 @@
+_BASE_URL = "https://api.twilio.com/2010-04-01"
+
+def connect(conn, account_sid, auth_token=None):
```
efiShtain, line 3: This is a python reference for twilio sdk, I think we should stick with this standard instead of the connect option (rename connect to Client or NewClient)
```python
from twilio.rest import Client

# Your Account SID and Auth Token from console.twilio.com
account_sid = "ACXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX"
auth_token  = "your_auth_token"

client = Client(account_sid, auth_token)

message = client.messages.create(
    to="+15558675309",
    from_="+15017250604",
    body="Hello from Python!")

print(message.sid)
```

libs/libs/integrations/twilio.star
```diff
@@ -0,0 +1,65 @@
...
+            form_body=body, 
+            basic_auth=auth_token and (account_sid, auth_token) or None,
+        )
+        
+        json = resp.body.json()
+
+        if resp.status_code < 200 or resp.status_code >= 400:
+            fail(json)
+            
+        return json
```
efiShtain, line 21: This part is probably shared utils for all http clients, can be lib/integrations/http
usable by itself 
small effort to split it now and we can get a lot by reusing it

examples/testprograms/twilio/main.star
```diff
@@ -0,0 +1,16 @@
+load("env", "TWILIO_ACCOUNT_SID", "TWILIO_AUTH_TOKEN", "TWILIO_MSG_SERVICE_SID")
+load("libs/integrations/twilio.star", "twilio")
```
efiShtain, line 2: since this is a predefined lib, 
I think we can try (in a different ticket in the future) to add "libs" or "plugins" section to the manifest, and inject this code automatically 

so
```yaml
project:
  libs:
    - name: twilio
      source: github/.... (optional, only if name is not auto resolved by ak core)
       data:
         account_sid: (from secret)
         twilio_auth_token: (from secret)
```

then the script can just declare what it needs and start using it
without loading env and lib and setting up a client (although it can)

PR #145, opened 2024-04-01
BUGFIX: deal with http routes that do not begin with '/'

itayd:
- add a leading '/' for paths that do not begin with it in the dispatcher special triggers handler
- do not error on env not found in dispatcher
- manifest schema fixes

### DIFF
--- a/internal/backend/dispatcher/special_triggers.go
+++ b/internal/backend/dispatcher/special_triggers.go
@@ -25,7 +25,7 @@ func processHTTPTrigger(trigger sdktypes.Trigger, event sdktypes.Event) (bool, m
 	// Get expected path pattern from the trigger.
 	pathValue, ok := trigger.Data()["path"]
 	if !ok {
-		// No pattern means we don't need to match anything.
+		// No path means we don't need to match anything.
 		return true, nil, nil
 	}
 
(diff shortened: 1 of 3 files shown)


internal/backend/dispatcher/special_triggers.go
```diff
@@ -35,6 +35,15 @@ func processHTTPTrigger(trigger sdktypes.Trigger, event sdktypes.Event) (bool, m
 
 	triggerPath := pathValue.GetString().Value()
 
+	if triggerPath == "" {
+		// Empty path means we don't need to match anything.
+		return true, nil, nil
+	}
+
+	if triggerPath[0] != '/' {
```
efiShtain: it is a bit low level but this only works since '/' is represented as 1 byte in utf-8, otherwise the comparison would fail

probably better to use something like
```go
if !strings.HasPrefix(triggerPath, "/") {
...
}
```

efiShtain (approved):
Nit we might consider but LGTM

daabr (approved):

internal/backend/dispatcher/special_triggers.go
itayd: fixed

PR #146, opened 2024-04-01
fix: de-duplicate error message

ig-ra:
de duplicate error message in connect -> SDK error conversion

### DIFF
--- a/sdk/internal/rpcerrors/errors.go
+++ b/sdk/internal/rpcerrors/errors.go
@@ -8,7 +8,7 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 )
 
-func TranslateError(err error) error {
+func ToSDKError(err error) error {
 	connectErrorCode := connect.CodeOf(err)
 	if err == nil {
 		return err
--- a/sdk/internal/rpcerrors/errors.go
+++ b/sdk/internal/rpcerrors/errors.go
@@ -44,9 +44,8 @@ func ToSDKError(err error) error {
 	}
 
 	// err is a connect error (checked in connect.CodeOf), so we can safely cast it
-	connErr := err.(*connect.Error)
-	if len(connErr.Details()) != 0 {
-		return fmt.Errorf("%w: (%v)", sdkErr, connErr.Details())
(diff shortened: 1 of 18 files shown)


sdk/internal/rpcerrors/errors.go
```diff
@@ -9,33 +9,82 @@ import (
...
+	   	err1 = sdkerrors.ErrUnauthenticated
+	   // ErrUnauthenticated    = errors.New("unauthenticated")
+	   // 		return "unauthenticated"
+
+	   case connect.CodePermissionDenied:
+	   	err1 = sdkerrors.ErrUnauthorized
+	   // ErrUnauthorized       = errors.New("unauthorized")
+	   // 		return "permission_denied"
+
+	   case connect.CodeResourceExhausted:
```
efiShtain: remove the comments

sdk/internal/rpcerrors/errors.go
```diff
@@ -9,33 +9,82 @@ import (
...
+		return connectErr.Unwrap()
 	default:
-		err1 = sdkerrors.ErrRPC
+		sdkErr = sdkerrors.ErrRPC
 	}
 
-	if cerr := new(connect.Error); errors.As(err, &cerr) {
-		return fmt.Errorf("%w: %s (%v)", err1, cerr.Message(), cerr.Details())
+	// err is a connect error (checked in connect.CodeOf), so we can safely cast it
+	connErr := err.(*connect.Error)
```
efiShtain: I think you removed the connect.CodeOf check so is this still valid conversion ?
also, why do we need this part if you already converted it before to sdk error ?

efiShtain (approved):
quick questions but looks ok to me if you're sure this doesn't break CLI or tests

sdk/internal/rpcerrors/errors.go
ig-ra: yes, conversion is done above in errors.As. There err is converted to connect.Error. If failed we return an original error

ig-ra: ah. I see. you are right. unneeded conversion

PR #150, opened 2024-04-03
quickstart flow

itayd:
Motivation: allow a user to quickly play with their automation when doing a first tutorial. This will be available only in dev mode.

- introduce webtools route, which supports both a terminal-in-browser tool and also messages.
- to be used for quickstart tutorial and ability to "chat" with your automations.

image

image


haimzlato (commented):
I'm not sure about the look and feel but the flow is OK.
I would put the send at the top and the messages lower because the user will first send a request to ak and then see results not the other way around.

internal/backend/webtools/db.go
```diff
@@ -0,0 +1,66 @@
...
+
+	if err := db.db(ctx).Where("addr = ?", addr).Find(&msgs).Error; err != nil {
+		return nil, err
+	}
+
+	return msgs, nil
+}
+
+func (db *terminalDB) AddMessage(ctx context.Context, addr, text string) (uint, error) {
+	r := Message{Addr: addr, Text: text}
```
tebeka, line 49: Validate `addr` and `text` (not empty, maybe other checks)

internal/backend/webtools/db.go
```diff
@@ -0,0 +1,66 @@
...
+}
+
+func newDB(db db.DB) *terminalDB { return &terminalDB{db} }
+
+func (db *terminalDB) Setup(ctx context.Context) error {
+	return db.db(ctx).AutoMigrate(&Message{})
+}
+
+func (db *terminalDB) db(ctx context.Context) *gorm.DB {
+	return (*gorm.DB)(db.DB.GormDB()).WithContext(ctx)
```
tebeka: I don't like this code. Is there a better way to do it without type conversion?

internal/backend/webtools/web/webmessages/app.js
```diff
@@ -0,0 +1,116 @@
...
+  try {
+    const resp = await fetch(url, {
+      method: 'POST',
+      body: body,
+    })
+
+    const rtext = await resp.text()
+
+    newMessage({
+      addr: `You &#8594; <span class="like-pre">${url}</span>`,
```
tebeka: Use `→` directly instead of the HTML code.

internal/backend/webtools/web/webmessages/messages.go
```diff
@@ -0,0 +1,6 @@
+package webmessages
+
+import "embed"
+
+//go:embed *
```
tebeka: This will include this go file as well.

internal/backend/webtools/web/webterminal/index.html
```diff
@@ -0,0 +1,31 @@
...
+		<div class="command" id=input-cont>
+			<div class=prompt>
+				<span id=main-stdin-msg class=prompt-msg></span>
+				<span id=main-stdin-symbol class=prompt-symbol></span>
+			</div>
+			<input type=text id=main-stdin class=stdin autofocus autocomplete spellcheck=false />
+		</div>
+		<div id=bottom-offset></div>
+	</div>
+	<script data-main=js/initialize.js src=js/require.js></script>
```
tebeka: Shouldn't file names be in quotes?

internal/backend/webtools/web/webterminal/js/commands/internal.js
```diff
@@ -0,0 +1,67 @@
...
+              .forEach(command => print(`${command}`))
+            newLine()
+            print(`Type 'help [command]' for more info`)
+            newLine()
+          } else {
+            try {
+              const help = await terminal.get_command_help(args[0])
+              const lines = help
+                .split('~')
+                .map(line => line.replace(/\r?\n|\r/, ''))
```
tebeka, line 51: Comment what this regexp does

internal/backend/webtools/webtools.go
```diff
@@ -0,0 +1,127 @@
...
+func New(cfg *Config, db db.DB) (Svc, error) {
+	if !cfg.Enabled {
+		return &svc{cfg: cfg}, nil
+	}
+
+	return &svc{db: newDB(db), cfg: cfg}, nil
+}
+
+func (svc *svc) Setup(ctx context.Context) error {
+	if !svc.cfg.Enabled {
```
tebeka, line 34: Instead of checking this every method, return a `NopSvc` that does nothing?

tebeka (commented):
Can you explain *why* do we need this tool? Look like the user can use Postman or similar tool.

internal/backend/webtools/web/webmessages/app.js
itayd: done

internal/backend/webtools/web/webmessages/messages.go
itayd: fixed

internal/backend/webtools/web/webterminal/index.html
itayd: fixed

internal/backend/webtools/webtools.go
itayd: it's not on "every method", just on the Init.

internal/backend/webtools/web/webterminal/js/commands/internal.js
itayd: not really my code :P adapted it from https://github.com/autokitteh/browser-terminal, but sure.

internal/backend/webtools/db.go
itayd: turned out i didn't need the casting :P

internal/backend/webtools/db.go
itayd: added validation for empty address, nothing much else to validate...

itayd:
> Can you explain _why_ do we need this tool? Look like the user can use Postman or similar tool.

added in description. talk to @haimzlato as well.

itayd:
> I'm not sure about the look and feel but the flow is OK. I would put the send at the top and the messages lower because the user will first send a request to ak and then see results not the other way around.

changed order

ig-ra:
> Can you explain _why_ do we need this tool? Look like the user can use Postman or similar tool.
kinda late, but I would like to join Miki's question. Why do we want to implement this as a webtool instead of implementing this as a subcommand in `ak` for example? for me it feels much more naturally to have 2 shells, in one I could run `ak up -t dev` and in the second `ak dev flow` or similar.. doing this via the web feels for me less intuitive. 
@itayd @haimzlato

itayd:
add Experimental banner

tebeka (approved):

PR #151, opened 2024-04-03
ENG-580: dev mode is persistent by default

itayd:
maintains persistency by two files: `autokitteh.sqlite` and `temporal_dev.sqlite`.

Note: this does not take care of redis, since it's volatile by definition - unless using an external redis.

Qs:
- Should we use the regular `dev` mode for this, or create another `persistent-dev`? or `volatile-dev`?
- Should we add a command to clean up stuff?
- Should we save the files in XDG_DATA or current dir?

### DIFF
--- a/internal/backend/db/dbfactory/factory.go
+++ b/internal/backend/db/dbfactory/factory.go
@@ -2,6 +2,7 @@ package dbfactory
 
 import (
 	"context"
+	"path/filepath"
 	"testing"
 	"time"
 
@@ -11,6 +12,7 @@ import (
 	"go.autokitteh.dev/autokitteh/internal/backend/db"
 	"go.autokitteh.dev/autokitteh/internal/backend/db/dbgorm"
 	"go.autokitteh.dev/autokitteh/internal/backend/gormkitteh"
+	"go.autokitteh.dev/autokitteh/internal/xdg"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
@@ -19,7 +21,13 @@ type Config = gormkitteh.Config
 var Configs = configset.Set[Config]{
 	Default: &Config{
 		SlowQueryThreshold: 300 * time.Millisecond,
+		Type:               gormkitteh.RequireExplicitDSNType,
 	},
+	Dev: &Config{
+		Type: "sqlite",
+		DSN:  "file:" + filepath.Join(xdg.DataHomeDir(), "autokitteh.sqlite"),
+	},
+	Test: &Config{},
 }
 
 func New(z *zap.Logger, cfg *Config) (db.DB, error) {
--- a/internal/backend/temporalclient/config.go
+++ b/internal/backend/temporalclient/config.go
@@ -1,6 +1,7 @@
 package temporalclient
 
 import (
+	"path/filepath"
 	"time"
 
 	"go.temporal.io/sdk/testsuite"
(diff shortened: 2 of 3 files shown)


internal/backend/db/dbfactory/factory.go
```diff
@@ -20,6 +20,11 @@ var Configs = configset.Set[Config]{
 	Default: &Config{
 		SlowQueryThreshold: 300 * time.Millisecond,
 	},
+	Dev: &Config{
+		Type: "sqlite",
+		DSN:  "file:autokitteh.sqlite",
```
daabr: 1. Shouldn't this be in the default mode as well? After it should be equally or *more* reliable than dev
2. By default we should use the XDG data directory, not the current directory

internal/backend/temporalclient/config.go
```diff
@@ -52,8 +52,9 @@ var (
 			Monitor:               defaultMonitorConfig,
 			StartDevServerIfNotUp: true,
 			DevServer: testsuite.DevServerOptions{
-				LogLevel: zapcore.WarnLevel.String(),
-				EnableUI: true,
+				LogLevel:   zapcore.WarnLevel.String(),
+				EnableUI:   true,
+				DBFilename: "temporal_dev.sqlite",
```
daabr: Same as 2 in the previous comment.

daabr (commented):
1. Should we use the regular dev mode for this, or create another persistent-dev? or volatile-dev?

Let's start with a single dev mode, and make if persistent, if there's a need in the future we could create a volatile/ephemeral mode

2. Should we add a command to clean up stuff?

Yeah, under "server"? Please file a ticket.

3. Should we save the files in XDG_DATA or current dir?

XDG data. The current dir isn't a persistent detail in itself, so it would be too easy to "lose" data by starting from a different dir, plus databases are the very definition of what belongs in XDG_DATA.

ig-ra:
> 1. Should we use the regular dev mode for this, or create another persistent-dev? or volatile-dev?
> 
> Let's start with a single dev mode, and make if persistent, if there's a need in the future we could create a volatile/ephemeral mode
> 

For me it feels that the default mode should be non-persistent and pressing Ctrl-C should clean everything. This is what I expect when running command in foreground and not as a daemon.  Therefore I would definitely vote for creating another `persistent-dev` mode. 

Moreover, I would suggest to specify paths of persistent DB in the banner to facilitate migration from the dev to prod (another solution is to write it in config and add `ak config show` command)

internal/backend/temporalclient/config.go
itayd: here by default it tries to reach local temporal, so no need to deal with 1.

fixed 2.

internal/backend/db/dbfactory/factory.go
itayd: fixed 2

itayd: 1 - now in default mode explicit DSN is required.

itayd:
> > 1. Should we use the regular dev mode for this, or create another persistent-dev? or volatile-dev?
> > 
> > Let's start with a single dev mode, and make if persistent, if there's a need in the future we could create a volatile/ephemeral mode
> 
> For me it feels that the default mode should be non-persistent and pressing Ctrl-C should clean everything. This is what I expect when running command in foreground and not as a daemon. Therefore I would definitely vote for creating another `persistent-dev` mode.

A product decision was made to make it persistent. @haimzlato to elaborate.

> 
> Moreover, I would suggest to specify paths of persistent DB in the banner to facilitate migration from the dev to prod (another solution is to write it in config and add `ak config show` command)

It already shows up in `ak config show`.

daabr (approved):

PR #153, opened 2024-04-03
fix: simpify invalid argument error

ig-ra:


cmd/ak/cmd/projects/get.go
```diff
@@ -20,8 +23,8 @@ var getCmd = common.StandardCommand(&cobra.Command{
 			return err
 		}
 
-		if err := common.FailIfNotFound(cmd, "project", p.IsValid()); err != nil {
-			return err
+		if !p.IsValid() {
+			return common.FailIfError(cmd, sdkerrors.ErrNotFound, fmt.Sprintf("project <%s>", args[0]))
```
itayd:
```suggestion
			return common.FailIfError(cmd, sdkerrors.ErrNotFound, fmt.Sprintf("project %q", args[0]))
```

cmd/ak/common/flags.go
```diff
@@ -33,27 +34,31 @@ func FailNotFound(cmd *cobra.Command, what string) error {
...
 		return nil
 	}
-	msg := what
+	msg := strings.Join(whats, " ")
 	var code int = GenericFailure
+
 	switch {
 	case errors.Is(err, sdkerrors.ErrNotFound):
-		return NewExitCodeError(NotFoundExitCode, fmt.Errorf("%s not found", what))
+		return NewExitCodeError(NotFoundExitCode, fmt.Errorf("%s not found", msg))
```
itayd, line 46:
```suggestion
		return NewExitCodeError(NotFoundExitCode, fmt.Errorf("%q not found", msg))
```

and elsewhere

sdk/sdkclients/sdkprojectsclient/client.go
```diff
@@ -86,8 +88,9 @@ func (c *client) GetByID(ctx context.Context, pid sdktypes.ProjectID) (sdktypes.
 
 	project, err := sdktypes.StrictProjectFromProto(resp.Msg.Project)
 	if err != nil {
-		// TODO: "errors.Is(err, sdkerrors.ErrInvalidArgument)" doesn't work.
-		if err.Error() == "invalid argument: zero object" {
+		// FIXME: ENG-626: why we check and override errInvalid for project only?
+		var errInvalid sdkerrors.ErrInvalidArgument
+		if err.Error() == "zero object" && errors.As(err, &errInvalid) {
```
itayd, line 93: i think you can use errors.Is?

itayd: also don't check strings, just make a new error type if you need to.

sdk/sdktypes/symbol.go
```diff
@@ -18,7 +18,7 @@ var symbolRE = kittehs.Must1(regexp.Compile(`^[a-zA-Z_][\w]*$`))
 
 func (symbolTraits) Validate(s string) error {
 	if s != "" && !symbolRE.MatchString(s) {
-		return errors.New("invalid")
+		return errors.New("illegal symbol (expected [a-zA-Z_][\\w]*)")
```
itayd: not sure putting the regex here is a good idea :)

daabr: I wouldn't expect users to guess what a valid symbol is, so this seems to me like a succinct explanation.

If we want the code to be cleaner, we can define the regex as a const string near `symbolRE` above, and add it to the error here.

cmd/ak/common/flags.go
daabr: IIRC, it should be `%s` here because it's a type (e.g. `"project"`, `"connection"`, etc.) and not a specific instance.

cmd/ak/cmd/projects/get.go
```diff
@@ -20,8 +23,8 @@ var getCmd = common.StandardCommand(&cobra.Command{
 			return err
 		}
 
-		if err := common.FailIfNotFound(cmd, "project", p.IsValid()); err != nil {
-			return err
+		if !p.IsValid() {
```
daabr, line 26: I don't think this change is correct: the default behavior should be that all get commands return nothing if the requested entity isn't found (not even an error).

If the user specifies the `--fail` flag, then we return the "not found" error.

BTW - same thing for all list commands.

sdk/sdkclients/sdkprojectsclient/client.go
ig-ra: 1. InvalidArgument isn't an instance, but a struct. so, I think errors.Is won't work here 
2. it's not my code, so I just left it how it was, including the check of specific string. and added this FIXME: ENG-626 to understand this better and fix in all relevant places

cmd/ak/common/flags.go
ig-ra: yes, @daabr is right. That's the whole message which IMO shouldn't be formatted as %q

sdk/sdktypes/symbol.go
ig-ra: @itayd @dabbr please suggest the way you want it. I could either remove it as @itayd  suggested or use a constant string as @dabbr suggested. I think it's important to provide user with enough and clear info.

cmd/ak/cmd/projects/get.go
tebeka: > the default behavior should be that all get commands return nothing if the requested entity isn't found

Why? If the use explicitly asked for something that's not there - it should be an error.

sdk/sdktypes/symbol.go
tebeka: Use raw string:

```go
`illegal symbol (expected [a-zA-Z_][\w]*)`
```

ig-ra: I could, but why do we need raw string in this specific case? Is there any difference for the resulting error or I'm missing anything?

cmd/ak/cmd/projects/get.go
ig-ra: @tebeka right now this is how our cli is defined. even if used is asking for specific entity, cmd will fail silently if the entity isn't found, unless `--fail` flag is provided
@daabr - maybe I'm wrong, but IMO I introduced no changes. I retested with the new code and it produces the same results - silent fail if there is no such project, and error if `--fail` is specified

sdk/sdktypes/symbol.go
tebeka: Raw string lets you use a single `\` in the regular expressions.

daabr (approved):

sdk/sdktypes/symbol.go
ig-ra: done

PR #155, opened 2024-04-04
ENG-601: properly return stopped sessions stats

itayd:


daabr (approved):

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -478,6 +478,7 @@ type DeploymentWithStats struct {
 	Running   uint32
 	Error     uint32
 	Completed uint32
+	Stopped   uint32
```
tebeka, line 481: Where is this value changed?

itayd: it's written to into in  `listDeploymentsWithStats`. When i did it there i forgot to do it here, this fixes that.

PR #156, opened 2024-04-04
ENG-617: Detailed manifest documentation

itayd:
- added "setenv" to system tests
- removed --dry-run from manifest apply (not really needed, we got plan)
- added --quiet to various manifest commands

### DIFF
--- a/docs/autokitteh.yaml
+++ b/docs/autokitteh.yaml
@@ -16,10 +16,6 @@ project:
       name: myvar
       # Variable value.
       value: "myvalue"
-    - name: myvar2
-      # This value is used if env_var is not set in the OS.
-      value: "default"
-      env_var: "MY_VAR2"
     - name: airshield_combination_code
       # This is a secret, so the value is taken from the OS environment
       # variable named in `env_var`. If `env_var` is not specified,
--- a/docs/autokitteh.yaml
+++ b/docs/autokitteh.yaml
@@ -76,8 +76,10 @@ project:
   triggers:
     - # Optional: event_type. For HTTP this is the method.
       event_type: get
-      # Optional: event filter.
-      filter: "data.method == 'GET'"
+      # Optional: event filter, specified in CEL (https://github.com/google/cel-spec).
+      # All event data is available to the filter via the `data` field.
+      # The event type is available via the `type` field.
+      filter: "data.path == 'meow' && type == 'GET'"
       # Name of the connection the event originated from.
       connection: myhttp
       # Function to call when the event is received.
--- a/docs/autokitteh.yaml
+++ b/docs/autokitteh.yaml
@@ -1,0 +1,92 @@
+# Currently only v1 is supported.
+version: v1
+
+# Project configuration.
+project:
+  # Project name is used to identify a specific project.
+  # Unique system wide.
+  name: http
+  # Variables are used to store project specific configuration, which
+  # can be accessed from the project's code using:
+  #  load("env", "<ENV_VAR_NAME>")
+  # example:
+  #  load("env", "myvar")
+  vars:
+    - # Variable name.
+      name: myvar
+      # Variable value.
+      value: "myvalue"
+    - name: myvar2
+      # This value is used if env_var is not set in the OS.
+      value: "default"
+      env_var: "MY_VAR2"
+    - name: airshield_combination_code
+      # This is a secret, so the value is taken from the OS environment
+      # variable named in `env_var`. If `env_var` is not specified,
+      # it is considered to be the upper case of the variable name.
+      is_secret: true
+      env_var: "AIRSHIELD_COMBINATION_CODE"
+  # Connections map integrations to a specific project.
+  # A connection is used in project code using:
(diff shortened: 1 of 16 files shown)


internal/manifest/manifest.go
```diff
@@ -45,7 +45,7 @@ type Connection struct {
 	ProjectKey string `yaml:"-" json:"-"` // belongs to project.
 
 	Name           string `yaml:"name" json:"name" jsonschema:"required"`
-	Token          string `yaml:"token,omitempty" json:"token,omitempty"`
+	Token          string `yaml:"token,omitempty" json:"token"`
```
daabr: Why `omitempty` in YAML but not JSON?

tebeka (approved):
I agree we need more tests and validation to the manifest.
Better to validate it when building (even better - write a linter and show errors in VSCode).
Maybe we can use [cue](https://cuelang.org/) for validation?

Nitpick: I don't like "myvar" and other "my" in examples, I prefer more realistic names.

docs/autokitteh.yaml
```diff
@@ -0,0 +1,92 @@
...
+  # Triggers specify how to route incoming events generated from the connections to this project.
+  triggers:
+    - # Optional: event_type. For HTTP this is the method.
+      event_type: get
+      # Optional: event filter.
+      filter: "data.method == 'GET'"
+      # Name of the connection the event originated from.
+      connection: myhttp
+      # Function to call when the event is received.
+      call: http.star:on_http_get
```
daabr, line 82: Is `call` going to replace `entrypoint`, or is it just an alias?

docs/autokitteh.yaml
```diff
@@ -0,0 +1,92 @@
...
+      data:
+        # Optional: path pattern. Can include named parameters.
+        # Each parameter is enclosed in {}. Last parameter
+        # can be suffixed with "..."", which will cause it to
+        # capture the rest of the path.
+        # Last parameter can also be "{$}", which specifies
+        # it expects an exact match.
+        path: "/test/{a}/{b...}"
```
daabr, line 90: Is there a formal definition of this syntax anywhere?

docs/autokitteh.yaml
```diff
@@ -0,0 +1,92 @@
...
+    - name: google
+      integration: "google"
+      # Generated from <autokitteh_base_url>/i/google.
+      token: "xxxx_xxxx_xxxx"
+  # Triggers specify how to route incoming events generated from the connections to this project.
+  triggers:
+    - # Optional: event_type. For HTTP this is the method.
+      event_type: get
+      # Optional: event filter.
+      filter: "data.method == 'GET'"
```
daabr: Is there a formal definition of the syntax anywhere?

docs/autokitteh.yaml
```diff
@@ -0,0 +1,92 @@
...
+      # Generated from <autokitteh_base_url>/i/twilio.
+      token: "xxxx_xxxx_xxxx"
+    # Google
+    - name: google
+      integration: "google"
+      # Generated from <autokitteh_base_url>/i/google.
+      token: "xxxx_xxxx_xxxx"
+  # Triggers specify how to route incoming events generated from the connections to this project.
+  triggers:
+    - # Optional: event_type. For HTTP this is the method.
```
daabr, line 73: About `event_type` being optional: what if the event type isn't exposed in `data`? Wouldn't that make it required because `filter` can't replace it?

internal/manifest/plan.go
```diff
@@ -186,9 +187,23 @@ func planDefaultEnv(ctx context.Context, mvars []*EnvVar, client sdkservices.Ser
 		})
 
 		val := mvar.Value
-		if mvar.EnvVar != "" {
+		envVar := mvar.EnvVar
```
daabr, line 189: The var names in this section aren't clear: what's the purpose/difference between `val`, `envVar`, and `mvar`?

internal/manifest/manifest.go
itayd: fixed

docs/autokitteh.yaml
itayd: currently an alias, but i think we'll get rid of it in the future.

itayd: sorry, i mean we'll get rid of entrypoint in the future. call will replace it.

docs/autokitteh.yaml
itayd: that's what i kinda tried to do in the comment.

docs/autokitteh.yaml
itayd: added reference to CEL.

docs/autokitteh.yaml
itayd: added type to the filter. event type is in the `type` var now.

itayd: we really need somewhere to specify the structures of the events for each integration.

internal/manifest/plan.go
itayd: see the autokitteh.yaml doc

PR #158, opened 2024-04-05
Misc tweaks split from other PRs for cleanliness

daabr:
- Remove -v from go build to simplify GitHub action output (we don't need that level of detail, and it may speed up the runtime a bit)
- Simplify logo hint in all integration HTML pages (done during work on HTTP secrets, but split to avoid PR pollution)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -65,7 +65,7 @@ bin/ak:
 .PHONY: build
 build:
 	mkdir -p $(OUTDIR)
-	$(GO) build -v $(GO_BUILD_OPTS) ./...
+	$(GO) build $(GO_BUILD_OPTS) ./...
 
 .PHONY: debug
 debug:
--- a/web/github/connect/index.html
+++ b/web/github/connect/index.html
@@ -31,7 +31,7 @@
     </a>
 
     <table class="title">
-      <td><img class="logo" src="/static/images/github.svg" alt="GitHub" /></td>
+      <td><img class="logo" src="/static/images/github.svg" alt="Logo" /></td>
       <td><h1>GitHub - Create a New Connection</h1></td>
     </table>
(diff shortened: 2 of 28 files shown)


itayd (approved):

Makefile
```diff
@@ -65,7 +65,7 @@ bin/ak:
 .PHONY: build
 build:
 	mkdir -p $(OUTDIR)
-	$(GO) build -v $(GO_BUILD_OPTS) ./...
+	$(GO) build $(GO_BUILD_OPTS) ./...
```
tebeka, line 68: I disagree with that and prefer to see the tests as they go by. @efiShtain @itayd @ig-ra ?

web/github/connect/index.html
```diff
@@ -31,7 +31,7 @@
     </a>
 
     <table class="title">
-      <td><img class="logo" src="/static/images/github.svg" alt="GitHub" /></td>
```
tebeka, line 34: In general we should have one error and success HTML template and have the connection populate them.

tebeka (approved):
I'm OK with the change but would love some input from others on the `-v` thingie.

Makefile
itayd: i want the -v in

ig-ra: I would prefer `-v` as well for meanwhile at least.

daabr: To be clear - this is for the **build**, not the execution of tests. This change means we don't print 1000+ lines of which Go packages we built **before** starting to run tests.

web/github/connect/index.html
daabr: I agree, I hope to do that once we have more hands on deck :)

PR #160, opened 2024-04-07
ENG-261: store HTTP secrets in connection

daabr:
This PR includes:
- "New connection" web page for the HTTP integration
- Add "Authorization" HTTP header to outgoing requests, if the connection has a token, and the token refers to basic or bearer secrets

A few misc fixes along the way:

- Don't treat content-type as special, it's just an HTTP header
- Correct gRPC logo in the integrations page
- Added a unit test for a function that needed it
- Removed inconsistent args:
  - `sdkmodule.ExportFunction`: `"form_encoding?"`, `"auth?"`
  - `sdkmodule.UnpackArgs`: `"content_type?"`, `"basic_auth=?"`, `"oauth2=?"`


integrations/http/client.go
```diff
@@ -10,62 +10,56 @@ import (
...
 var IntegrationID = sdktypes.IntegrationIDFromName("http")
 
 var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.IntegrationPB{
 	IntegrationId: IntegrationID.String(),
 	UniqueName:    "http",
-	DisplayName:   "HTTP",
+	DisplayName:   "HTTP / REST",
 	LogoUrl:       "/static/images/http.svg",
-	// TODO: Integration documentation link.
+	ConnectionUrl: "/httprest/connect",
```
efiShtain: technically it doesn't have to be rest at all, it can be anything over http

integrations/http/server.go
```diff
@@ -12,28 +12,44 @@ import (
...
-	logger     *zap.Logger
-}
-
-func Start(l *zap.Logger, mux *http.ServeMux, d sdkservices.Dispatcher) {
-	h := HTTPHandler{dispatcher: d, logger: l}
+func Start(l *zap.Logger, mux *http.ServeMux, s sdkservices.Secrets, d sdkservices.Dispatcher) {
+	h := handler{logger: l, secrets: s, dispatcher: d, scope: "http"}
 	mux.Handle(routePrefix("{ns}")+"*", h)
+
+	// Save new autokitteh connections with user-submitted Twilio secrets.
```
efiShtain: Twilio -> Http

integrations/http/server.go
```diff
@@ -12,28 +12,44 @@ import (
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+	"go.autokitteh.dev/autokitteh/web/static"
 )
 
+const (
+	// Save new autokitteh connections with user-submitted secrets.
+	uiPath = "/httprest/connect/"
```
efiShtain: a though, let me know what do you think,
maybe if this is a rest supporting integration
it should be just /http
get will get the ui
post will save a new connection ?

integrations/http/save.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+// handleAuth saves a new autokitteh connection with user-submitted data.
+func (h handler) handleAuth(w http.ResponseWriter, r *http.Request) {
+	l := h.logger.With(zap.String("urlPath", r.URL.Path))
+
+	// Check "Content-Type" header.
+	ct := r.Header.Get(HeaderContentType)
+	if ct != ContentTypeForm {
+		l.Warn("Unexpected header value",
```
efiShtain, line 25: probably better to log unexpected content type than just header value

integrations/http/save.go
```diff
@@ -0,0 +1,87 @@
...
+			zap.String("header", HeaderContentType),
+			zap.String("got", ct),
+			zap.String("want", ContentTypeForm),
```
efiShtain, line 28: since this is not user facing, there is no real need to print got and want

integrations/http/save.go
```diff
@@ -0,0 +1,87 @@
...
+		e := "Form parsing error: " + err.Error()
+		u := fmt.Sprintf("%serror.html?error=%s", uiPath, url.QueryEscape(e))
+		http.Redirect(w, r, u, http.StatusFound)
+		return
+	}
+	basic := r.Form.Get("basic_username") + ":" + r.Form.Get("basic_password")
+	bearer := r.Form.Get("bearer_access_token")
+
+	authz := ""
+	switch {
```
efiShtain, line 51: do you want to error if both provided ?

integrations/http/client.go
daabr: technically I agree, but:
1. Even though all REST APIs are HTTP APIs, a majority of people refer to them simply as REST APIs without mentioning HTTP (even if they're wrong)
2. The real reason I added "rest" to the URL path was to avoid conflict with all `/http/` webhooks, e.g. in case someone creates a project called "connect"

integrations/http/server.go
daabr: fixed

integrations/http/server.go
daabr: See my comment above about why I wanted a different path from `http`.

I don't think your idea will work because I also use POST - when the user loads the web page in `/httprest/connect`, it contains a web form, which the user submits as a POST to `/httprest/save` (which points to my webhook in `save.go`).

integrations/http/client.go
daabr: I'm fine with changing the integration to use `/webhook/...` instead of `/http/...` for webhooks, and then change the connection links to `/http/connect` and `/http/save` - but that's a user-facing behavior change that should be discussed in a larger forum. WDYT?

integrations/http/save.go
daabr: I agree, but because this is happening in all integrations that use a web form to save data, I created a ticket to handle all of them separately from this PR.

https://linear.app/autokitteh/issue/ENG-645/logging-tweaks-based-on-code-review

integrations/http/save.go
daabr: This is related to the previous comment: the point was to have generic checks for various headers, but effectively I'm checking only Content-Type, and I always expect a web form.

So yes - added to ENG-645.

integrations/http/save.go
daabr: When a user uses the HTML form they can't submit more than one.

If someone want to send a handcrafted POST with both options that's their problem, I only care that they can't create damage/risks.

integrations/http/client.go
tebeka: Don't have a strong opinion on `http` vs `webhook` but IMO at this time point we should avoid it.

integrations/http/http.go
```diff
@@ -65,30 +66,31 @@ func request(method string) sdkexecutor.Function {
 			return sdktypes.InvalidValue, err
 		}
 
+		// Add the Authorization HTTP header?
+		if authz := i.getConnection(ctx)[authzHeader]; authz != "" {
```
tebeka: Why the `z` prefix?

integrations/http/client.go
efiShtain: regarding user projects, we have to have some prefix that split out endpoints from user endpoints so we can have what ever routes we need without worrying about project names users can have

regarding the renaming, I think we should do it, just need to sync with haim

regarding rest, I agree people use the term incorrectly, doesn't mean we have to :)

good way to test this sync between docs and code

integrations/http/http.go
```diff
@@ -148,25 +167,20 @@ func setBody(req *http.Request, rawBody string, formBody map[string]string, cont
...
+	// JSON body.
 	if jsonBody.IsValid() && !jsonBody.IsNothing() {
 		if formBody != nil {
 			return errMutuallyExclusive
 		}
 
-		if contentType == "" {
-			contentType = contentTypeJSON
+		// Set the Content-Type header only if it's not already set.
+		if req.Header.Get(contentTypeHeader) == "" {
```
tebeka, line 184: Maybe `setIfMissing` utility function? It's used in the form as well.

integrations/http/server.go
efiShtain: maybe all integrations should start with /integrations/<integration-name>/ ? 
and all web hooks should be received at /webhooks ?

integrations/http/save.go
tebeka: It's not just `their problem`, it becomes our problem. Also there might be a security issues. But CSRF is for another PR :)

web/static/httprest/connect/error.html
```diff
@@ -0,0 +1,46 @@
+<!DOCTYPE html>
```
tebeka: Can't we have a general template for all the error pages?

web/static/httprest/connect/index.html
```diff
@@ -0,0 +1,119 @@
...
+        Basic
+      </button>
+      <span class="toggle-sep"></span>
+      <button class="toggle" id="toggletab3" onclick="toggleTab('tab3')">
+        Bearer
+      </button>
+    </div>
+
+    <form method="post" action="/httprest/save">
+      <div class="tab" id="tab1"></div>
```
tebeka: Maybe better names than `tab1`, `tab2` ...?

web/static/httprest/connect/index.html
```diff
@@ -0,0 +1,119 @@
...
+      <td><h1>HTTP / REST - Create a New Connection</h1></td>
+    </table>
+
+    <div class="toggle-container">
+      <button class="toggle" id="toggletab1" onclick="toggleTab('tab1')">
+        No Auth
+      </button>
+      <span class="toggle-sep"></span>
+      <button class="toggle" id="toggletab2" onclick="toggleTab('tab2')">
+        Basic
```
tebeka: Change to `Basic (user/password)`, not all users know what is basic authentication.

web/static/httprest/connect/index.html
```diff
@@ -0,0 +1,119 @@
...
+      </button>
+      <span class="toggle-sep"></span>
+      <button class="toggle" id="toggletab3" onclick="toggleTab('tab3')">
+        Bearer
```
tebeka: Change to `Bearer (token)`

web/static/httprest/connect/index.html
```diff
@@ -0,0 +1,119 @@
...
+      </button>
+    </div>
+
+    <form method="post" action="/httprest/save">
```
tebeka: CSRF?

web/static/httprest/connect/styles.css
```diff
@@ -0,0 +1,142 @@
+body {
```
tebeka: There's a top level `web/static/styles.css` with a lot of common content.

web/static/httprest/connect/success.html
```diff
@@ -0,0 +1,45 @@
+<!DOCTYPE html>
```
tebeka: Let's write a success template and use it everywhere. Currently there are 9 different files.

web/static/httprest/connect/success.html
```diff
@@ -0,0 +1,45 @@
...
+      <td>
+        <img class="logo" src="/static/images/http.svg" alt="Logo" />
+      </td>
+      <td><h1>HTTP / REST - Connection Created</h1></td>
+    </table>
+
+    <p class="info">Your <strong>autokitteh</strong> connection token is:</p>
+    <p><span id="token" class="copyable"></span></p>
+
+    <p class="notif" id="notif">Copied into the clipboard</p>
```
tebeka: What is `notif`?

integrations/http/client.go
itayd: this shouldn't be called rest. this is a generic http endpoint.

integrations/http/server.go
itayd: we might opt to give a root mux to all integrations where they will build on top of it. we should give them an `/integrations/{integration_name}/` root mux imo.

integrations/http/http.go
itayd: yeah, drop the z.

integrations/http/client.go
daabr: Another alternative without renaming `/http`: rename connection pages from `/<integration>/connect` to `/i/<integration>/connect`

daabr: Done:
- Removed all mentions of "REST" (in web pages and URL paths)
- HTTP new-connection web pages now use `/i/http/...` paths
- No impact on existing webhooks, which continue to use `/http/...`
- In the future, all integrations will be under `/i` - but this is beyond the scope of this specific PR

integrations/http/server.go
daabr: done - see https://github.com/autokitteh/autokitteh/pull/160#discussion_r1564591312

integrations/http/http.go
daabr: done

web/static/httprest/connect/index.html
daabr: actually, this is an intentional choice suggested by Ronen - for ease of use across different integrations that use the same tab mechanism. Think of it as common boilerplate code. The **contents** of tabs (i.e. input fields) have unique IDs of course.

web/static/httprest/connect/index.html
daabr: Tab titles need to be very short.

I added an info link to each tab to give context. Also, it's feels a little silly to specify the names of input fields in the tab title when they appear right below that title.

But anyway, my rule of thumb for HTTP auth schemes is that if you don't know what it is, you probably don't need to use it.

web/static/httprest/connect/index.html
daabr: See previous comment about Basic.

web/static/httprest/connect/index.html
daabr: This isn't really a CSRF attack vector because we don't trust the input coming from this form in any meaningful way.

We could add additional security, e.g. an "Origin" HTTP header, but:
1. This can be forged too
2. We don't particularly mind where the data is coming from
3. Any attack that can be carried out against the "/save" webhook can be carried against "index.html" as well

web/static/httprest/connect/styles.css
daabr: Notice that different integrations have subtle differences in their "style.css" files. This is intentional - future integrations may not be part of the AK core, and their UIs will be served independently, so they shouldn't rely on AK's `/styles.css` file either.

web/static/httprest/connect/success.html
daabr: I agree, but will do it in a separate horizontal PR.

daabr: https://linear.app/autokitteh/issue/ENG-677/improvements-in-integration-web-uis

web/static/httprest/connect/error.html
daabr: https://linear.app/autokitteh/issue/ENG-677/improvements-in-integration-web-uis

web/static/httprest/connect/success.html
daabr: It's for the JS code that handles copying tokens into the clipboard (see `form.js`)

integrations/http/client.go
daabr: https://linear.app/autokitteh/issue/ENG-677/improvements-in-integration-web-uis

integrations/http/http.go
daabr: Added to https://linear.app/autokitteh/issue/ENG-677/improvements-in-integration-web-uis

efiShtain (approved):

tebeka (approved):

PR #161, opened 2024-04-08
Quote paths if they contain spaces, in "config where" CLI command

daabr:
This is a tiny optimization for users who want to copy & paste it in the terminal.

Also replace `fmt.Println(...)` with `fmt.Fprintln(cmd.OutOrStdout(), ...)` to allow Cobra to override the OS's stdout with its own (which is what we do in most CLI outputs).

### DIFF
--- a/cmd/ak/cmd/configuration/where.go
+++ b/cmd/ak/cmd/configuration/where.go
@@ -24,7 +24,6 @@ var whereCmd = common.StandardCommand(&cobra.Command{
 
 		data := xdg.DataHomeDir()
 		if strings.Contains(data, " ") {
-			data = fmt.Sprintf("%q", data)
 			data = `"` + data + `"`
 		}
 
--- a/cmd/ak/cmd/configuration/where.go
+++ b/cmd/ak/cmd/configuration/where.go
@@ -2,6 +2,7 @@ package configuration
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/spf13/cobra"
 
@@ -16,8 +17,19 @@ var whereCmd = common.StandardCommand(&cobra.Command{
 	Args:    cobra.NoArgs,
 
 	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Println("Config home directory:", xdg.ConfigHomeDir())
-		fmt.Println("Data home directory:  ", xdg.DataHomeDir())
+		cfg := xdg.ConfigHomeDir()
+		if strings.Contains(cfg, " ") {
+			cfg = `"` + cfg + `"`
+		}
+
+		data := xdg.DataHomeDir()
+		if strings.Contains(data, " ") {
+			data = fmt.Sprintf("%q", data)
+			data = `"` + data + `"`
+		}
+
+		fmt.Println("Config home directory:", cfg)
+		fmt.Println("Data home directory:  ", data)
 		fmt.Println()
 		fmt.Println("Override environment variable names:")
 		fmt.Println(xdg.ConfigEnvVar)
--- a/cmd/ak/cmd/configuration/where.go
+++ b/cmd/ak/cmd/configuration/where.go
@@ -2,6 +2,7 @@ package configuration
 
 import (
 	"fmt"
+	"strconv"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


cmd/ak/cmd/configuration/where.go
```diff
@@ -16,12 +17,22 @@ var whereCmd = common.StandardCommand(&cobra.Command{
...
 
 	RunE: func(cmd *cobra.Command, args []string) error {
-		fmt.Println("Config home directory:", xdg.ConfigHomeDir())
-		fmt.Println("Data home directory:  ", xdg.DataHomeDir())
-		fmt.Println()
-		fmt.Println("Override environment variable names:")
-		fmt.Println(xdg.ConfigEnvVar)
-		fmt.Println(xdg.DataEnvVar)
+		cfg := xdg.ConfigHomeDir()
+		if strings.Contains(cfg, " ") {
```
efiShtain, line 22: I think you can always wrap it with " "

cmd/ak/cmd/configuration/where.go
```diff
@@ -16,12 +17,22 @@ var whereCmd = common.StandardCommand(&cobra.Command{
...
+		if strings.Contains(data, " ") {
+			data = `"` + data + `"`
+		}
+
+		fmt.Fprintln(cmd.OutOrStdout(), "Config home directory:", cfg)
+		fmt.Fprintln(cmd.OutOrStdout(), "Data home directory:  ", data)
+		fmt.Fprintln(cmd.OutOrStdout(), "")
+		fmt.Fprintln(cmd.OutOrStdout(), "Override environment variable names:")
+		fmt.Fprintln(cmd.OutOrStdout(), xdg.ConfigEnvVar)
+		fmt.Fprintln(cmd.OutOrStdout(), xdg.DataEnvVar)
```
efiShtain, line 36: don't we support json output flag and this breaks it ?

daabr: The json flags only affect `common.Render()`.

It's unfortunate that they have no effect in this command, but I don't have a good idea how to show this output with JSON in a **useful** way, so I'm fine with this.

cmd/ak/cmd/configuration/where.go
tebeka: If so, you can use the `%q` verb,

tebeka (approved):

cmd/ak/cmd/configuration/where.go
daabr: We could... But I'd rather not **always** add `""` unless it's really needed?

cmd/ak/cmd/configuration/where.go
efiShtain: ?
```json
{
  homeDir: '',
  dataDir: '',
  envVars: {
    }
  dataEnvVars: {

  }
}
```

efiShtain (approved):

cmd/ak/cmd/configuration/where.go
```diff
@@ -16,12 +17,22 @@ var whereCmd = common.StandardCommand(&cobra.Command{
...
-		fmt.Println(xdg.ConfigEnvVar)
-		fmt.Println(xdg.DataEnvVar)
+		cfg := xdg.ConfigHomeDir()
+		if strings.Contains(cfg, " ") {
+			cfg = `"` + cfg + `"`
+		}
+
+		data := xdg.DataHomeDir()
+		if strings.Contains(data, " ") {
+			data = `"` + data + `"`
```
itayd: use [strconv.Quote](https://pkg.go.dev/strconv#Quote)

cmd/ak/cmd/configuration/where.go
daabr: Weird... I'll check

daabr: Oh, I thought you meant this is the current situation, not a proposal for JSON output. I'll do that in a separate PR.

cmd/ak/cmd/configuration/where.go
daabr: done

PR #164, opened 2024-04-09
ENG-651: Fix GHES transport

daabr:
Also generalize PR #154 after testing with GHES - `/github-apps/` (instead of `/apps/`) is a GHES feature rather than a configuration.

### DIFF
--- a/integrations/github/jwt.go
+++ b/integrations/github/jwt.go
@@ -100,7 +100,7 @@ func (i integration) getConnection(ctx context.Context) (map[string]string, erro
 func (i integration) NewClientWithAppJWTFromGitHubID(appID int64) (*github.Client, error) {
 	// Shared transport to reuse TCP connections.
 	tr := http.DefaultTransport
-	entURL, err := enterpriseURL()
+	enterpriseURL, err := enterpriseURL()
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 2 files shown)


integrations/github/jwt.go
```diff
@@ -100,21 +100,24 @@ func (i integration) getConnection(ctx context.Context) (map[string]string, erro
 func (i integration) NewClientWithAppJWTFromGitHubID(appID int64) (*github.Client, error) {
 	// Shared transport to reuse TCP connections.
 	tr := http.DefaultTransport
+	entURL, err := enterpriseURL()
```
ig-ra: enterpriseURL? in many cases I do prefer short names, but in this case I would advice to use the full one

integrations/github/jwt.go
```diff
@@ -100,21 +100,24 @@ func (i integration) getConnection(ctx context.Context) (map[string]string, erro
...
 
 	// Initialize a client with the generated JWT injected into outbound requests.
 	client := github.NewClient(&http.Client{Transport: atr})
-	u, err := enterpriseURL()
-	if err != nil {
-		return nil, err
-	}
-	if u != "" {
-		client, err = client.WithEnterpriseURLs(u, u)
+	if entURL != "" {
```
ig-ra: maybe in this case it would be more readable if we split this method for 2? enterprice and not enterprise?

integrations/github/jwt.go
daabr: done

integrations/github/jwt.go
daabr: Either way I don't think we can improve much, so let's think about refactoring this later.

internal/backend/oauth/oauth.go
```diff
@@ -54,9 +54,9 @@ func New(l *zap.Logger) sdkservices.OAuth {
 		)
 	}
 
-	githubAppsURLPart := os.Getenv("GITHUB_APPS_URL_PART")
-	if githubAppsURLPart == "" {
-		githubAppsURLPart = "apps"
+	appsDir := "apps"
+	if os.Getenv("GITHUB_ENTERPRISE_URL") != "" {
+		appsDir = "github-apps"
```
efiShtain, line 59: Maybe this should be env var as well? Sounds like something dynamic

efiShtain (approved):

internal/backend/oauth/oauth.go
daabr: It was an env var - but it turns out that it isn't dynamic, it's either "apps" for regular GH, or "github-apps" not GitHub Enterprise.

ig-ra (approved):

PR #167, opened 2024-04-10
reunite svc and basesvc, lay groundwork for auth work

itayd:
- unite svc and basesvc into a single module
- lay groundwork for multiple muxes that one of them is authed and the other is not. each component chooses which one it needs.

this extracts the relevant code from #120 .


internal/backend/svc/svc.go
```diff
@@ -53,38 +177,102 @@ func NewOpts(cfg *basesvc.Config, ropts basesvc.RunOptions) []fx.Option {
...
 				if err != nil {
-					return nil, nil, err
+					return "", nil, nil, err
+				}
+
+				// Replace the original mux with a main mux and also expose the original mux as the no-auth one.
+				wrapped := http.NewServeMux()
+
+				wrap := func(h http.Handler) http.Handler {
+					// TODO: Wrap the handler with auth middleware.
```
itayd: add ticket

daabr (approved):

internal/backend/svc/svc.go
itayd: doneish

PR #170, opened 2024-04-10
[ENG-516] Simplify Python Callbacks

tebeka:
This remove the old thread with message queue in Python that was there due to how the gRPC protocol worked.
It also creates a more "format" communication protocol between Python & Go. In the future this protocol between Go runtime and a language server can be used to create new runtime only by implementing the target language server.

Messages are now format `{"type": "run", "payload": ...}` so Go can use `json.RawMessage` and decode the underlying message type according to top level type (see `comm.go`).

Fixes ENG-516

### DIFF
--- a/runtimes/pythonrt/ak_runner.py
+++ b/runtimes/pythonrt/ak_runner.py
@@ -44,6 +44,9 @@ BUILTIN = {v for v in dir(builtins) if callable(getattr(builtins, v))}
 
 class Transformer(ast.NodeTransformer):
     """Replace 'fn(a, b)' with '_ak_call(fn, a, b)'."""
+    def __init__(self, file_name):
+        self.file_name = file_name
+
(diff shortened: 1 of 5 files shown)


runtimes/pythonrt/ak_runner.py
```diff
@@ -46,6 +44,9 @@ def name_of(node):
 
 class Transformer(ast.NodeTransformer):
     """Replace 'fn(a, b)' with '_ak_call(fn, a, b)'."""
+    def __init__(self, file_name):
```
daabr, line 47:
```suggestion

    def __init__(self, file_name):
```

daabr (approved):

runtimes/pythonrt/ak_runner.py
tebeka: Why?

PR #171, opened 2024-04-10
ENG-508: save secrets in DB by default

daabr:


internal/backend/secrets/svc.go
```diff
@@ -26,18 +27,19 @@ var Configs = configset.Set[Config]{
...
 	case "aws":
 		impl, err = NewAWSSecrets(l, cfg)
-	// TODO(ENG-508): case "db"
+	case "file":
+		impl, err = NewFileSecrets(l, xdg.DataHomeDir())
 	case "vault":
 		impl, err = NewVaultSecrets(l, cfg)
 	default:
-		impl, err = NewFileSecrets(l, xdg.DataHomeDir())
+		impl, err = NewDatabaseSecrets(l, db)
```
itayd: this should be when cfg.Type is empty or some specific string, not by default.

itayd (commented):
are you sure you pushed everything?

internal/backend/secrets/svc.go
daabr: done

daabr:
> are you sure you pushed everything?

Yes, the DB part was already done in #71.

itayd (approved):
lgtm, but would love to see a system test

PR #172, opened 2024-04-11
Misc tweaks in system tests infra

daabr:
Done during ongoing investigation of flakiness.

### DIFF
--- a/tests/system/README.md
+++ b/tests/system/README.md
@@ -49,18 +49,18 @@ be split by slash characters (`/`) into a sequence of regular expressions,
 and each part of a test's identifier (i.e. the txtar relative path under
 `testdata`) must match the corresponding element in the sequence.
 
-For example, to run only with the txtar files in/under `testdata/*foo*/*bar*`,
-including `testdata/*foo*/*bar*.txtar`:
+For example, to run only with the txtar files in/under `testdata/*subdir*/*filename*`,
+including `testdata/*subdir*/*filename*.txtar`:
 
 ```
-gotestsum -f testname ./tests/system -run /foo/bar
+gotestsum -f testname ./tests/system -run /subdir/filename
 ```
 
 To repeat a test `N` times when investigating flakiness, append the `-count`
 flag:
 
 ```
-gotestsum -f testname ./tests/system -run /foo/bar -count N
+gotestsum -f testname ./tests/system -run /subdir/filename -count N
 ```
 
 ## Txtar File Structure
--- a/tests/system/README.md
+++ b/tests/system/README.md
@@ -53,7 +53,14 @@ For example, to run only with the txtar files in/under `testdata/*foo*/*bar*`,
 including `testdata/*foo*/*bar*.txtar`:
 
 ```
-gotestsum -f testname -run /foo/bar
+gotestsum -f testname ./tests/system -run /foo/bar
+```
+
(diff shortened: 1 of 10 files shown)


tests/system/README.md
```diff
@@ -53,7 +53,14 @@ For example, to run only with the txtar files in/under `testdata/*foo*/*bar*`,
 including `testdata/*foo*/*bar*.txtar`:
 
 ```
-gotestsum -f testname -run /foo/bar
+gotestsum -f testname ./tests/system -run /foo/bar
```
itayd: I think you should give a concrete example here, not just `/foo/bar`

tests/system/ak_test.go
```diff
@@ -75,6 +75,7 @@ func setUpTest(t *testing.T, akPath string) string {
 
 	// Redirect the OS's stdout and stderr through a pipe, to
 	// detect when the AK server is ready for the test to begin.
+	// TODO: Replace with /healthz or /readyz once they're dependable.
```
itayd: ticket? also i think they're dependable right now for this use

itayd (approved):

tests/system/README.md
tebeka: I agree, no `foo`, `bar` and friends in any user facing documentation.

tests/system/README.md
```diff
@@ -53,7 +53,14 @@ For example, to run only with the txtar files in/under `testdata/*foo*/*bar*`,
...
+```
+
+To repeat a test `N` times when investigating flakiness, append the `-count`
+flag:
+
+```
+gotestsum -f testname ./tests/system -run /foo/bar -count N
```
tebeka: IMO document with `go test`, its more familiar.

tests/system/server.go
```diff
@@ -17,7 +17,8 @@ var useProcSvc, _ = strconv.ParseBool(os.Getenv("AK_SYSTEST_USE_PROC_SVC"))
 // subprocess: to support breakpoint debugging, and measure test coverage.
 func startAKServer(ctx context.Context, akPath string) (svc.Service, error) {
 	cfg := kittehs.Must1(svc.LoadConfig("", map[string]any{
```
tebeka, line 19: Why `Must` here if the function returns an error?

tests/system/ak_test.go
daabr: I prefer to use opentelemetry's endpoint, but you're right, we can use the one we currently have too. Will do in a separate PR.

tests/system/README.md
daabr: done

tests/system/ak_test.go
daabr: done (added ticket ID)

tests/system/README.md
daabr: I understand, but I strongly prefer users to use `gotestsum -f testname` instead of `go test -v` for this purpose. The output is MUCH more useful this way. So I prefer to steer them in this way without having to think about replacing `go test` with `gotestsum`.

tests/system/server.go
daabr: `svc.LoadConfig` can return an error only if the input is invalid, but since the input here is hard-coded, that should be impossible.

And if someone modifies the hard-coded input in the future and it becomes invalid, I think it's reasonable to panic in this case.

PR #177, opened 2024-04-12
ENG-657: cmd: --debug to enable debug logs easily

itayd:


daabr (changes requested):
I confirmed that this works:
```
ak up --config logger.zap.level=debug
```
So no need for this flag. However, I filed this bug:

https://linear.app/autokitteh/issue/ENG-671/logger-config-not-exposed-by-ak-config-cli

Because setting this persistently is my real goal.

PR #180, opened 2024-04-12
ENG-669: in multi instance, do not always fail when executors are nil

itayd:
When running in multiple instances, an activity might start in an instance different than the workflow. In such case, do not fail when executors are nil (see ENG-335), but fail only when the activity is trying to invoke non-integration call.

```
	// For every instance there are two worker pools created:
	// - generalWorker: used only for activities that do not need to run on
	//                  a specific instance. These activities are only used
	//                  to invoke integrations calls that are not built-in.
	// - uniqueWorker: used only for activities that need to run on a specific
	//                 instance. These activities have to run on these instances
	//                 since they must be able to access instance specific state.
	//                 This state is the workflow state as constructed by the
```


internal/backend/sessions/config.go
```diff
@@ -28,6 +28,7 @@ var defaultConfig = Config{
 			ActivityScheduleToCloseTimeout: time.Hour,
 			ActivityStartToCloseTimeout:    time.Minute * 30,
 			LocalScheduleToCloseTimeout:    time.Second * 5,
+			ActivityScheduleToStartTimeout: time.Second * 5,
```
daabr, line 31: 5s is reasonable for local activities (because the worker is already running by definition), but for potentially-remote activities it means a likelihood of activity loss because we won't tolerate worker restarts.

Also note the implication for `ActivityScheduleToCloseTimeout`:

`ActivityScheduleToStartTimeout` + `ActivityStartToCloseTimeout`  = 30m5s << 1h = `ActivityScheduleToCloseTimeout`

daabr (commented):
1. One comment about timeouts
2. There's almost no documentation in this PR, and it's **needed** at least in `calls.go`
3. I don't have any specific objections, but I don't understand the approach, please explain in the PR description the idea and how it's guaranteed to work

tebeka:
> When running in multiple instances, an activity might start in an instance different than the workflow. 

An instance of what?

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -32,10 +38,13 @@ type Calls interface {
...
-	config Config
-	worker worker.Worker
-	svcs   *sessionsvcs.Svcs
+	z             *zap.Logger
+	config        Config
+	generalWorker worker.Worker // general worker that can serve any workflows.
+	uniqueWorker  worker.Worker // unique worker that can server workflows that run only this this process.
+	svcs          *sessionsvcs.Svcs
+
+	uniqueWorkerCallTaskQueueName string
```
tebeka: What is a `uniqueWorker`? The fact we have two different types of worker is a `design smell`.

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -47,12 +56,28 @@ func New(z *zap.Logger, config Config, svcs *sessionsvcs.Svcs) Calls {
 	opts := config.Temporal.Worker
 	opts.DisableRegistrationAliasing = true
 	opts.OnFatalError = func(err error) { z.Error("temporal worker error", zap.Error(err)) }
+	opts.DisableWorkflowWorker = true
```
tebeka: Comment why

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -47,12 +56,28 @@ func New(z *zap.Logger, config Config, svcs *sessionsvcs.Svcs) Calls {
...
+
+	uniqueWorkerCallTaskQueueName := fmt.Sprintf("%s_%s", taskQueueName, strings.Replace(uuid.NewString(), "-", "", -1))
```
tebeka: If this is unique per `New`, what happens when the server is restarted and  there are tasks in this queue?

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -63,12 +88,19 @@ func New(z *zap.Logger, config Config, svcs *sessionsvcs.Svcs) Calls {
...
+		return err
+	}
+
+	return cs.uniqueWorker.Start()
+}
 
 func (cs *calls) Call(ctx workflow.Context, params *CallParams) (sdktypes.SessionCallAttemptResult, error) {
 	spec := params.CallSpec
 
 	fnv, _, _ := spec.Data()
```
tebeka, line 116: The diff is missing `	fnv, _, _ := spec.Data()` for some reason.

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -85,34 +117,54 @@ func (cs *calls) Call(ctx workflow.Context, params *CallParams) (sdktypes.Sessio
...
+		xid := fnvf.ExecutorID()
 
-		actx := workflow.WithActivityOptions(
-			ctx,
-			workflow.ActivityOptions{
+		// neither a non-integration (like a run) or an integration that is builtin.
+		local := !xid.IsIntegrationID() || fixtures.IsBuiltinIntegrationID(xid.ToIntegrationID())
+
+		if local {
+			executorsForSessions[params.SessionID.String()] = params.Executors // HACK
```
tebeka: Add an issue for the `HACK`?

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -85,34 +117,54 @@ func (cs *calls) Call(ctx workflow.Context, params *CallParams) (sdktypes.Sessio
...
 				TaskQueue:              taskQueueName,
 				ActivityID:             fmt.Sprintf("session_call_%s_%d", params.SessionID.Value(), seq),
 				ScheduleToCloseTimeout: cs.config.Temporal.ActivityScheduleToCloseTimeout,
 				HeartbeatTimeout:       cs.config.Temporal.ActivityHeartbeatTimeout,
-			},
-		)
+			}
+
+			if local {
+				if aopts.TaskQueue = cs.uniqueWorkerCallTaskQueueName; aopts.TaskQueue == "" {
```
tebeka: Assign to `aopts` only after you check for empty string.

internal/backend/sessions/config.go
itayd: > 5s is reasonable for local activities (because the worker is already running by definition), but for potentially-remote activities it means a likelihood of activity loss because we won't tolerate worker restarts.

This is used only for activities on unique workers.

> 
> Also note the implication for `ActivityScheduleToCloseTimeout`:
> 
> `ActivityScheduleToStartTimeout` + `ActivityStartToCloseTimeout` = 30m5s << 1h = `ActivityScheduleToCloseTimeout`

increased `StartToClose` to 45m.

itayd:
> 1. One comment about timeouts

Addressed

> 2. There's almost no documentation in this PR, and it's **needed** at least in `calls.go`

Now there is

> 3. I don't have any specific objections, but I don't understand the approach, please explain in the PR description the idea and how it's guaranteed to work

Ok

itayd:
> > When running in multiple instances, an activity might start in an instance different than the workflow.
> 
> An instance of what?

Autokitteh server.

internal/backend/sessions/sessioncalls/calls.go
itayd: commented

internal/backend/sessions/sessioncalls/calls.go
itayd: commented

internal/backend/sessions/sessioncalls/calls.go
itayd: now commented.

internal/backend/sessions/sessioncalls/calls.go
itayd: ?

internal/backend/sessions/sessioncalls/calls.go
itayd: no longer a hack

internal/backend/sessions/sessioncalls/calls.go
itayd: not sure it'll make a more readable code. not everything is local.

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -47,12 +71,28 @@ func New(z *zap.Logger, config Config, svcs *sessionsvcs.Svcs) Calls {
 	opts := config.Temporal.Worker
 	opts.DisableRegistrationAliasing = true
 	opts.OnFatalError = func(err error) { z.Error("temporal worker error", zap.Error(err)) }
+	opts.DisableWorkflowWorker = true // these workers serve only activities.
 
-	worker := worker.New(svcs.Temporal, taskQueueName, opts)
+	uniqueWorkerCallTaskQueueName := fmt.Sprintf("%s_%s", taskQueueName, strings.Replace(uuid.NewString(), "-", "", -1))
```
daabr:
```suggestion
	uniqueWorkerCallTaskQueueName := fmt.Sprintf("%s_%s", taskQueueName, strings.ReplaceAll(uuid.NewString(), "-", ""))
```

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -85,34 +132,57 @@ func (cs *calls) Call(ctx workflow.Context, params *CallParams) (sdktypes.Sessio
...
+				// When a local execution is required, it is scheduled on a unique worker. These local executions
+				// need access to the state that is built by the workflow.
+
+				if aopts.TaskQueue = cs.uniqueWorkerCallTaskQueueName; aopts.TaskQueue == "" {
+					return sdktypes.InvalidSessionCallAttemptResult, errors.New("local session worker for activities is not registerd")
+				}
+
+				aopts.ScheduleToStartTimeout = cs.config.Temporal.ActivityScheduleToStartTimeout
+				if aopts.ScheduleToStartTimeout == 0 {
+					z.Warn("call activity schedule to start timeout is 0 and local exec is needed")
```
daabr: Tiny readability nit:
```suggestion
					z.Warn("call activity schedule-to-start timeout is 0 and local exec is needed")
```

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -123,15 +193,27 @@ func (cs *calls) Call(ctx workflow.Context, params *CallParams) (sdktypes.Sessio
 					Poller:    params.Poller,
 				},
 			).Get(ctx, &ret); err != nil {
+				var terr *temporal.TimeoutError
```
daabr, line 196: For readability, please convert:
```
if err := workflow.ExecuteActivity(
    ...
).Get(ctx, &ret); err != nil {
    ...
}
```
To:
```
future := workflow.ExecuteActivity(
    ...
)
if err := future.Get(ctx, &ret); err != nil {
    ...
}
```

daabr (approved):
Approved with a few nits

internal/backend/sessions/sessioncalls/calls.go
itayd: done

PR #181, opened 2024-04-13
start three ak servers simultaniously on a local machine

itayd:
use nginx as the load balancer. 

9980 -> nginx -> 9981,9982,9983

TODO:
- do not automigrate on start, else is flaky as several trying to setup in parallel

### DIFF
--- a/internal/backend/svc/svc.go
+++ b/internal/backend/svc/svc.go
@@ -298,10 +298,13 @@ func NewOpts(cfg *Config, ropts RunOptions) []fx.Option {
 	return append(opts, fx.Populate(svcs))
 }
 
-func StartDB(ctx context.Context, cfg *Config) (db.DB, error) {
+func StartDB(ctx context.Context, cfg *Config, ropts RunOptions) (db.DB, error) {
+	setFXRunOpts(ropts)
+
 	var db db.DB
 
 	if err := fx.New(
+		fx.NopLogger,
 		fx.Supply(cfg),
 		LoggerFxOpt(),
 		DBFxOpt(),
--- a/scripts/start-multi.sh
+++ b/scripts/start-multi.sh
@@ -1,0 +1,31 @@
+#!/bin/bash
+
+# Requirements:
+# - nginx installed
+# - temporal server start-dev
+
+set -euo pipefail
+
+WD="${WD-tmp/multi}"
+
+rm -fR "${WD}" && mkdir -p "${WD}"
+
+./bin/ak server setup -m dev
+
+akup() {
+    port="$1"
+    ./bin/ak up -m dev \
+        -c "temporalclient.start_dev_server_if_not_up=false" \
+        -c "http.addr=0.0.0.0:${port}" \
+        -c "logger.zap.outputPaths=${WD}/${port}.log" \
+        >& "${WD}/${port}.out" &
+    echo $!
+}
+
+p1=$(akup 9981)
+p2=$(akup 9982)
+p3=$(akup 9983)
+
+trap 'kill $p1 $p2 $p3' 0
+
+nginx -c "$(pwd)/scripts/nginx.conf"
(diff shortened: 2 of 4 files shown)


internal/backend/svc/svc.go
```diff
@@ -298,10 +298,13 @@ func NewOpts(cfg *Config, ropts RunOptions) []fx.Option {
...
 }
 
-func StartDB(ctx context.Context, cfg *Config) (db.DB, error) {
+func StartDB(ctx context.Context, cfg *Config, ropts RunOptions) (db.DB, error) {
+	setFXRunOpts(ropts)
+
 	var db db.DB
 
 	if err := fx.New(
+		fx.NopLogger,
```
tebeka, line 307: Why `NopLogger`? We might miss some errors.

scripts/start-multi.sh
```diff
@@ -0,0 +1,31 @@
...
+
+WD="${WD-tmp/multi}"
+
+rm -fR "${WD}" && mkdir -p "${WD}"
+
+./bin/ak server setup -m dev
+
+akup() {
+    port="$1"
+    ./bin/ak up -m dev \
```
tebeka, line 17: Why `./bin/ak` and not `ak`? A user might want to use that without building.

scripts/start-multi.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/bash
```
tebeka, line 1: Why not a docker compose file?

scripts/start-multi.sh
```diff
@@ -0,0 +1,31 @@
...
+    echo $!
+}
+
+p1=$(akup 9981)
+p2=$(akup 9982)
+p3=$(akup 9983)
+
+trap 'kill $p1 $p2 $p3' 0
+
+nginx -c "$(pwd)/scripts/nginx.conf"
```
tebeka, line 31: This fails for me with:
```
./scripts/start-multi.sh 
2024/04/14 16:51:52 [emerg] 199325#199325: mkdir() "/var/lib/nginx/client-body" failed (13: Permission denied)
```

scripts/start-multi.sh
```diff
@@ -0,0 +1,31 @@
...
+        -c "logger.zap.outputPaths=${WD}/${port}.log" \
+        >& "${WD}/${port}.out" &
+    echo $!
+}
+
+p1=$(akup 9981)
+p2=$(akup 9982)
+p3=$(akup 9983)
+
+trap 'kill $p1 $p2 $p3' 0
```
tebeka, line 29: When running and the hitting `CTRL-C` I get:

```
./scripts/start-multi.sh: line 1: kill: (199474) - No such process
./scripts/start-multi.sh: line 1: kill: (199476) - No such process
./scripts/start-multi.sh: line 1: kill: (199478) - No such process
```

tebeka (commented):
The logs don't say which `ak` is handling a request. Should maybe add the server port to the intercept logs? Maybe give the server an ID somehow?

efiShtain:
do you handle the multiple sqlite instances ? I didn't get where

anyhow, docker compose is the way to go here
otherwise people can decide how they want to run ak scaled

scripts/start-multi.sh
ig-ra: I think we cannot assume that there is nginx installed.

scripts/start-multi.sh
ig-ra: I think we shouldn't reinvent the wheel and manage subprocesses, load balancer and such.
docker compose?

ig-ra:
> do you handle the multiple sqlite instances ? I didn't get where
> 
> anyhow, docker compose is the way to go here otherwise people can decide how they want to run ak scaled

agree with @efiShtain and @tebeka regarding docker compose

efiShtain:
@itayd 
https://github.com/autokitteh/autokitteh/pull/186

have a look at this change
let me know if you think we should merge yours / mine / both / other options

anyhow, I don't think we should provide / promise scaling capabilities until we have the base lines of load testing

itayd:
replaced by #186

PR #185, opened 2024-04-14
[ENG-680] Inject HTTP body

tebeka:
This is a temporary fix to inject HTTP request body to Python.
We'll remove this once ENG-624 is implemented.

Fixes ENG-680


runtimes/pythonrt/Makefile
```diff
@@ -17,7 +17,7 @@ test-py:
 
 test-go:
 	golangci-lint run ./...
-	go test -v
+	go test $(SHORT) -v
```
itayd: where is this coming from?

runtimes/pythonrt/ak_runner.py
```diff
@@ -379,5 +382,14 @@ def module_entries(mod):
     event = message.get('event')
     event = {} if event is None else event
 
+    # Inject HTTP body
+    # TODO (ENG-624) change this once we support callbacks to autokitteh
+    body = event.get('data', {}).get('body')
```
itayd, line 387: oh wow this is very specific in a very generic place

itayd (approved):
super hacky, but will do in the meantime

runtimes/pythonrt/Makefile
tebeka: You can set environment variable to run only short tests, this disables `Test_createVEnv`  which takes a long time.
Changed to `TESTOPTS` and added a comment.

runtimes/pythonrt/ak_runner.py
tebeka: Yes, I'll re-do it once we have stabler event API to Python.

PR #186, opened 2024-04-14
add compose-postgres.yaml to allow persistent / scale ak

efiShtain:
## Start AK with scaled version using compose

```bash
docker compose up --scale server=3 # can be any number
```

- This uses an nginx as a reverse proxy
- It uses AK as temporal dev server
- Uses postgres as a persistent db

### DIFF
--- a/compose-postgres.yaml
+++ b/compose-postgres.yaml
@@ -47,6 +47,7 @@ services:
       AK_TEMPORALCLIENT__HOSTPORT: "temporal:7233"
       AK_DB__TYPE: "postgres"
       AK_DB__DSN: "host=db dbname=postgres user=postgres password=postgres"
+      AK_STORE__SERVER_URL: "redis:6379"
     restart: unless-stopped
   temporal:
     build:
@@ -62,11 +63,22 @@ services:
       AK_TEMPORALCLIENT__HOSTPORT: "0.0.0.0:7233"
       AK_DB__TYPE: "postgres"
       AK_DB__DSN: "host=db dbname=postgres user=postgres password=postgres"
+      AK_STORE__SERVER_URL: "redis:6379"
     healthcheck:
       test: ["CMD-SHELL", "wget -q --spider http://localhost:9980"]
       interval: 1s
       timeout: 5s
       retries: 5
+  redis:
+    image: redis:7.2.4-alpine
+    restart: always
+    ports:
+      - '6379:6379'
+    command: redis-server --save 20 1 --loglevel warning
+    volumes: 
+      - redis:/data
 volumes:
   db:
+    driver: local
+  redis:
     driver: local
--- a/compose-postgres.yaml
+++ b/compose-postgres.yaml
@@ -1,0 +1,66 @@
+# Comments are provided throughout this file to help you get started.
+# If you need more help, visit the Docker compose reference guide at
+# https://docs.docker.com/compose/compose-file/
+
+# Here the instructions define your application as a service called "server".
+# This service is built from the Dockerfile in the current directory.
+# You can add other services your application may depend on here, such as a
+# database or a cache. For examples, see the Awesome Compose repository:
+# https://github.com/docker/awesome-compose
+services:
+  db:
+    image: postgres:16.1-alpine
+    restart: always
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+    ports:
+      - "5432:5432"
+    volumes:
+      - db:/var/lib/postgresql/data
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 5s
+      timeout: 5s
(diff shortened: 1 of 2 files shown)


compose-postgres.yaml
```diff
@@ -0,0 +1,66 @@
...
+services:
+  db:
+    image: postgres:16.1-alpine
+    restart: always
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+    ports:
+      - "5432:5432"
+    volumes:
```
ig-ra, line 19: do we need persistency? I think we could skip the volume part

compose-postgres.yaml
```diff
@@ -0,0 +1,66 @@
...
+# database or a cache. For examples, see the Awesome Compose repository:
+# https://github.com/docker/awesome-compose
+services:
+  db:
+    image: postgres:16.1-alpine
+    restart: always
+    environment:
+      - POSTGRES_USER=postgres
+      - POSTGRES_PASSWORD=postgres
+    ports:
```
ig-ra, line 17: random port? or anything that will prevent with other dockers/ real PG running on the same machine?

itayd:
not sure what "It uses AK as temporal dev server" means?

compose-postgres.yaml
itayd: you must have persistency with multi instance, that's how state is shared

compose-postgres.yaml
```diff
@@ -0,0 +1,66 @@
...
+      - "5432:5432"
+    volumes:
+      - db:/var/lib/postgresql/data
+    healthcheck:
+      test: ["CMD-SHELL", "pg_isready -U postgres"]
+      interval: 5s
+      timeout: 5s
+      retries: 5
+  nginx:
+    image: nginx:latest
```
itayd: do you think it might be worthwhile to pin to a specific version?

itayd (approved):

compose-postgres.yaml
efiShtain: technically we share the same postgres, so you don't have to have persistency

but it is normally expected from postgres to be persistence 
so I prefer to have it

compose-postgres.yaml
efiShtain: I think it is standard (at least as far as I saw) to expose the standard port
users can change it if they want

compose-postgres.yaml
efiShtain: since this is only development / local playground purposes, I don't really care about the version
we don't use any advanced logic here, just basic of the basic and nginx is super stable, so latest == less to maintain

but I'll change it to stable instead of latest to be on the safe side

efiShtain:
> not sure what "It uses AK as temporal dev server" means?

I didn't want to have a custom docker image to run temporal, I also didn't find any public temporal docker image that runs temporal dev-server

since I wanted to reuse the temporal server between all AK instances, I start another ak just to use it's temporal dev-server, the nginx doesn't pass any requests to this ak, only other ak servers talk to this container's temporal instance (and they don't start their own instances)

itayd:
> > not sure what "It uses AK as temporal dev server" means?
> 
> I didn't want to have a custom docker image to run temporal, I also didn't find any public temporal docker image that runs temporal dev-server
> 
> since I wanted to reuse the temporal server between all AK instances, I start another ak just to use it's temporal dev-server, the nginx doesn't pass any requests to this ak, only other ak servers talk to this container's temporal instance (and they don't start their own instances)

ah i see. interesting. you could just install temporal using curl and then run the start-dev though.
https://github.com/temporalio/cli?tab=readme-ov-file#curl

less... convoluted :p

efiShtain:
> > > not sure what "It uses AK as temporal dev server" means?
> > 
> > 
> > I didn't want to have a custom docker image to run temporal, I also didn't find any public temporal docker image that runs temporal dev-server
> > since I wanted to reuse the temporal server between all AK instances, I start another ak just to use it's temporal dev-server, the nginx doesn't pass any requests to this ak, only other ak servers talk to this container's temporal instance (and they don't start their own instances)
> 
> ah i see. interesting. you could just install temporal using curl and then run the start-dev though. https://github.com/temporalio/cli?tab=readme-ov-file#curl
> 
> less... convoluted :p

you can do the same with postgres and use the other compose file running just ak

depends what we want to achieve
I'm ok with all options

itayd:
oh - can you add redis to the party here?

itayd (changes requested):
now i see that all these files are at root :/ can we put it anywhere else? maybe scripts?

efiShtain:
@itayd 
I've added redis

could you verify it works as expected on your end?

if so, please ack the pr

regarding location of docker-compose, as far as I saw, it is mostly on the main directory, especially in kind of monorepos, since it might involve multiple components

I don't see any value in moving it, I'll move it if you insist

PR #188, opened 2024-04-15
webtools/msgs tuning

itayd:
- added default paths, so no need to specify addr in API. default addr is "default".
- body is not required. added semi-sane default.
- target URL in UI can be anything under localhost/http.


haimzlato (approved):
No comments

PR #190, opened 2024-04-16
named integration ids to contain integration name hint

itayd:
it is useful for integration ids that are built based on names to include a hint for the name inside them. this change tries to embed the original name in the id itself.

for example, for http:

`int_3kth00httpf1201a7ed83f7cd5`

this allows for easier debugging when need to know what integration this id belongs to.


sdk/sdktypes/integration_id.go
```diff
@@ -22,8 +23,33 @@ func StrictParseIntegrationID(s string) (IntegrationID, error) { return Strict(P
 
 func IsIntegrationID(s string) bool { return IsIDOf[integrationIDTraits](s) }
 
-func IntegrationIDFromName(name string) IntegrationID {
-	txt := fmt.Sprintf("%s_%026x", integrationIDKind, kittehs.HashString64(name))
+const integrationIDFromNamePrefix = "3kth"
```
daabr, line 26: I still think this prefix isn't useful. It's obvious that this kind of ID is special and looks different from regular UUIDs, and while you see "❤️🐱", I see "3000th".

Perosnally, I think it's better to use 0's to separate the name and the hash, i.e. `"<prefix> _ 00 %06s 00 %016s"`

But it's a stylistic choice, so I won't block on it.

sdk/sdktypes/integration_id.go
```diff
@@ -22,8 +23,33 @@ func StrictParseIntegrationID(s string) (IntegrationID, error) { return Strict(P
...
+			return 'v'
+		default:
+			return r
+		}
+	}, name)
+
+	idName = fmt.Sprintf("%06s", idName)
+	if len(idName) > 6 {
+		idName = idName[:6]
+	}
```
daabr, line 45: Limit the name's length before the character mapping to avoid wasted work.

daabr (approved):
Please limit the integration name's length to 6 bytes before the character mapping into a valid base32 string, to avoid wasted work.

sdk/sdktypes/integration_id.go
itayd: you can't put 0 in the beginning of an id, uuid complains. so i chose 3. i needed some more chars to make it a bit more distinguishable than a random 0000.

sdk/sdktypes/integration_id.go
itayd: done

PR #191, opened 2024-04-16
allow to override project name on manifest apply/deploy/etc

itayd:
this allows to take a manifest off-the-shelf and apply it to an arbitrary project name.


efiShtain (approved):
LGTM

PR #193, opened 2024-04-16
Fix for sending form over HTTP without specifying content type

daabr:
See lines 210-213 (not modified by this PR): if there's no custom content-type, we set the default one in the HTTP request we're preparing (`req.Header`), but not the helper variable (`contentType`). So in the switch block (which evaluates at least a part of the content-type value) we still need to consider that `contentType` may be `""`.

### DIFF
--- a/integrations/http/http.go
+++ b/integrations/http/http.go
@@ -214,7 +214,7 @@ func setBody(req *http.Request, rawBody string, formBody map[string]string, json
 
 		// Ignore (but allow the user to set) the charset in the Content-Type header.
 		switch strings.Split(contentType, ";")[0] {
-		case contentTypeForm:
+		case "", contentTypeForm:
 			s := form.Encode()
 			req.Body = io.NopCloser(strings.NewReader(s))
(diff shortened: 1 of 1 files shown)


integrations/http/http.go
```diff
@@ -214,7 +214,7 @@ func setBody(req *http.Request, rawBody string, formBody map[string]string, json
 
 		// Ignore (but allow the user to set) the charset in the Content-Type header.
 		switch strings.Split(contentType, ";")[0] {
-		case contentTypeForm:
+		case "", contentTypeForm:
```
itayd, line 217: explain?

daabr: See lines 210-213 (not modified by this PR): if there's no custom content-type, we set the default one in the HTTP request we're preparing (`req.Header`), but not the helper variable (`contentType`). So in the switch block (which evaluates at least a part of the content-type value) we still need to consider that `contentType` may be `""`.

itayd (approved):

PR #194, opened 2024-04-16
session: when triggered by a connection, pass the connection to the entrypoint

itayd:


internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -478,6 +482,21 @@ func (w *sessionWorkflow) run(wctx workflow.Context) (prints []string, err error
...
+				return c.ID().String() == trigger.ConnectionId
+			}); conn.IsValid() {
+				if kwargs["conn"], err = sdktypes.NewModuleValue(
+					conn.Name(),
+					w.executors.GetValues(conn.ID().String()),
+				); err != nil {
+					return prints, err
+				}
+			}
+		}
```
efiShtain, line 498: maybe it would be more readable if you convert the if checks and exit faster

less indentations are better, 
here you check that trigger unwraps and has connectionId and find it in data.connections and is valid and conn is kwargs and you getValues by connID 
which is really hard to follow and get what you try to do

examples/testprograms/http/http.star
```diff
@@ -3,11 +3,12 @@ load("@http", "myhttp")
 def foo(x):
     return "meow, " + x
 
-def on_http_get():
+def on_http_get(conn):
```
efiShtain, line 6: in AWS they pass (event, context) to any lambda
context can have multiple fields in it and is extendable 
event usually mean a string of the event that triggered this lambda

maybe we should have something similar since multiple events can point to the same entrypoint and additional information like the connection (if exists) and be added to the context

another option is just context and setting everything inside it

examples/testprograms/http/http.star
```diff
@@ -3,11 +3,12 @@ load("@http", "myhttp")
...
+    print(conn)
     print(time.now())
     print(store.set("x", "meow"))
 
-    resp, err = myhttp.get("https://httpbin.org/status/404", ak.callopts(catch=True))
+    resp, err = conn.get("https://httpbin.org/status/404", ak.callopts(catch=True))
```
efiShtain, line 11: so you pass an actual reference to the connection and not just an id ? do we really want to do it ?
not sure starlark lsp could understand the types here vs getting connectionID and then selecting the correct loaded one

examples/testprograms/http/http.star
tebeka: We need to formally define how events are passed to event handlers. We can borrow from serverless:
- In AWS it's `handler(event, context)`
- In Google it's `handler(request)`
- In OpenFass it's `handler(request)`
- ...

Currently in Python the handler gets a single `event` where the trigger data is in a field called `data`

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -478,6 +482,21 @@ func (w *sessionWorkflow) run(wctx workflow.Context) (prints []string, err error
 			kittehs.ContainedIn(argNames...),
 		)
 
+		// If session was triggered by a connection, pass the connection to the entrypoint.
+		var trigger struct{ ConnectionId string }
+		if err := w.data.Session.Inputs()["trigger"].UnwrapInto(&trigger); err == nil && trigger.ConnectionId != "" {
+			if _, conn := kittehs.FindFirst(w.data.Connections, func(c sdktypes.Connection) bool {
```
tebeka, line 488: I find this style of coding, where the `; conn.IsValid` a couple of line below, confusing.
I prefer

```go
_, conn = kittehs.FindFirst(w.data.Connections, func(c sdktypes.Connection) bool {
    return c.ID().String() == trigger.ConnectionId
})
if !conn.IsValid() {
...
```

tests/system/testdata/endtoend/http.txtar
```diff
@@ -46,10 +46,11 @@ project:
         path: "/test/{a}/{b...}"
 
 -- main.star --
-def on_http(data, trigger, event):
+def on_http(data, trigger, event, conn):
```
tebeka, line 49: Where are all the valid argument names for handlers documented?
This is something users must know.

PR #198, opened 2024-04-16
ENG-702: convert db id type from string to uuid

efiShtain:
All ids in db are now uuids


Makefile
```diff
@@ -160,5 +160,5 @@ pythonrt:
 .PHONY: generate-migrations
 generate-migrations:
```
ig-ra, line 161: why do we need to ask user for migration_name? instead of creating migrations for everything = sqlite pg?

sdk/sdktypes/integration_id.go
```diff
@@ -50,6 +50,6 @@ func NewIntegrationIDFromName(name string) IntegrationID {
 	return kittehs.Must1(ParseIntegrationID(txt))
 }
 
-func IsIntegrationIDFromName(id IntegrationID) bool {
-	return strings.HasPrefix(id.Value(), integrationIDFromNamePrefix)
-}
+// func IsIntegrationIDFromName(id IntegrationID) bool {
```
ig-ra: remove/explain why this commented out?

internal/backend/db/dbgorm/gorm_test.go
```diff
@@ -239,23 +253,23 @@ func assertDeleted[T any](t *testing.T, f *dbFixture, m T) {
 
 var (
 	// testSessionID    = "ses_00000000000000000000000001"
-	testBuildID = "bld_00000000000000000000000001"
+	testBuildID = kittehs.Must1(uuid.FromBytes([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1}))
```
ig-ra, line 256: let's define this as const?

internal/backend/db/dbgorm/connections.go
```diff
@@ -15,10 +15,10 @@ func (db *gormdb) createConnection(ctx context.Context, conn *scheme.Connection)
 
 func (db *gormdb) CreateConnection(ctx context.Context, conn sdktypes.Connection) error {
 	c := scheme.Connection{
-		ConnectionID:     conn.ID().String(),
-		IntegrationID:    scheme.PtrOrNil(conn.IntegrationID().String()), // TODO(ENG-158): need to verify integration id
+		ConnectionID:     *conn.ID().UUIDValue(),
+		IntegrationID:    conn.IntegrationID().UUIDValue(),
```
ig-ra: you removed TODO comment which is still relevant

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -238,15 +224,15 @@ type EventRecord struct {
...
 		State:     eventsv1.EventState(e.State),
 		CreatedAt: timestamppb.New(e.CreatedAt),
 	})
 }
 
 type Env struct {
-	EnvID     string  `gorm:"primaryKey"`
-	ProjectID *string `gorm:"index;foreignKey"`
+	EnvID     sdktypes.UUID  `gorm:"primaryKey;type:uuid"`
+	ProjectID *sdktypes.UUID `gorm:"index;type:uuid"`
```
ig-ra, line 244: foreign key?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -344,7 +330,7 @@ func ParseTrigger(e Trigger) (sdktypes.Trigger, error) {
 }
 
 type SessionLogRecord struct {
-	SessionID string `gorm:"index;foreignKey:SessionID"`
+	SessionID sdktypes.UUID `gorm:"index;type:uuid"`
```
ig-ra, line 342: foreignKey?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -443,9 +429,9 @@ func ParseSession(s Session) (sdktypes.Session, error) {
 }
 
 type Deployment struct {
-	DeploymentID string  `gorm:"primaryKey"`
-	EnvID        *string `gorm:"index;foreignKey"`
-	BuildID      *string `gorm:"foreignKey"`
+	DeploymentID sdktypes.UUID  `gorm:"primaryKey;type:uuid"`
+	EnvID        *sdktypes.UUID `gorm:"index;foreignKey;type:uuid"`
```
ig-ra: fk

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -443,9 +429,9 @@ func ParseSession(s Session) (sdktypes.Session, error) {
...
+	BuildID      *sdktypes.UUID `gorm:"foreignKey;type:uuid"`
```
ig-ra: fk

internal/backend/db/dbgorm/connections.go
```diff
@@ -27,44 +27,44 @@ func (db *gormdb) CreateConnection(ctx context.Context, conn sdktypes.Connection
 
 func (db *gormdb) UpdateConnection(ctx context.Context, conn sdktypes.Connection) error {
 	c := scheme.Connection{
-		ConnectionID:     conn.ID().String(),
-		IntegrationID:    scheme.PtrOrNil(conn.IntegrationID().String()), // TODO(ENG-158): need to verify integration id
+		ConnectionID:     *conn.ID().UUIDValue(),
+		IntegrationID:    conn.IntegrationID().UUIDValue(),
```
ig-ra: todo comment which is still relevant

internal/backend/db/dbgorm/projects_test.go
```diff
@@ -70,7 +70,7 @@ func TestGetProjects(t *testing.T) {
 	assert.ErrorIs(t, err, gorm.ErrRecordNotFound)
 
 	// test getProjectByName after delete
-	_, err = f.gormdb.getProject(f.ctx, p.Name)
+	_, err = f.gormdb.getProjectByName(f.ctx, p.Name)
```
ig-ra, line 73: getProject knows how to recognize whether it's name or ID. Not anymore?

internal/backend/db/dbgorm/sessions_test.go
```diff
@@ -68,7 +69,7 @@ func TestCreateSessionForeignKeys(t *testing.T) {
 
 	// negative test with non-existing assets
 	s := f.newSession(sdktypes.SessionStateTypeCompleted)
-	unexisting := "unexisting"
+	unexisting := uuid.New()
```
ig-ra, line 72: in other tests u did something different. please choose either, but consistent way

internal/backend/db/dbgorm/projects_test.go
efiShtain: this is getProject, doesn't look like it works with either name or id
plus, id is now uuid and not a string 
```go
func (db *gormdb) getProject(ctx context.Context, projectID sdktypes.UUID) (*scheme.Project, error) {
	return getOne(db.db, ctx, scheme.Project{}, "project_id = ?", projectID)
}
```

Makefile
efiShtain: is is common to give a migration a name, like added-created-at-to-project-id
so you have the meaning of the migration from the file name

sdk/sdktypes/integration_id.go
efiShtain: I'll remove it, it is not used

internal/backend/db/dbgorm/gorm_test.go
efiShtain: didn't change the behavior from before

internal/backend/db/dbgorm/connections.go
efiShtain: correct, restored

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: not required,
attaching the generated schema, foreign key is created correctly

```sql
CREATE TABLE "envs" (
  "env_id" uuid NOT NULL,
  "project_id" uuid NULL,
  "name" text NULL,
  "deleted_at" timestamptz NULL,
  "membership_id" text NULL,
  PRIMARY KEY ("env_id"),
  CONSTRAINT "fk_envs_project" FOREIGN KEY ("project_id") REFERENCES "projects" ("project_id") ON UPDATE NO ACTION ON DELETE NO ACTION
);
```

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: see as other comment

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: same as other comment

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: same

internal/backend/db/dbgorm/connections.go
efiShtain: correct, fixed

internal/backend/db/dbgorm/sessions_test.go
efiShtain: fixed, hope it makes sense now

where unexisting is reusable, I create it once, otherwise I create it directly

Makefile
ig-ra: ah.. my mistake. sorry.

internal/backend/db/dbgorm/gorm_test.go
ig-ra: before there was prefix in each ID. Now it's const bytes. I suggest to define those const bytes as a const var

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: consistency. Let's either write it everywhere, or remove it everywhere?

internal/backend/db/dbgorm/projects_test.go
ig-ra: ah, actually u fixed the bug. there is getProjectByName in the beginning of the test, and I mistakenly used getProject later after delete. Thanks!

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: removed all foreign key references

internal/backend/db/dbgorm/gorm_test.go
efiShtain: this is not constable (can't be defined const)

```go
kittehs.Must1(uuid.FromBytes([]byte{0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1}))
```

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: perfect!

ig-ra (approved):

sdk/sdktypes/id.go
```diff
@@ -73,6 +76,16 @@ func (i id[T]) Value() string {
 	return i.tid.Suffix()
 }
 
+func (i id[T]) UUIDValue() *UUID {
+
```
itayd:
```suggestion
```

sdk/sdktypes/id.go
```diff
@@ -27,6 +28,8 @@ type ID interface {
 	// Value returns the id value, meaning without the prefix.
 	Value() string
 
+	UUIDValue() *UUID
```
itayd: why not just return `UUID`? why do we need the ptr?

efiShtain: it makes life easier when the UUID is optional, like in db, I can just use the UUIDValue() instead of checking if it is zero value and then assigning null in the db

itayd: in this case return an empty array

efiShtain: fixed

PTAL

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -17,26 +17,20 @@ import (
...
-}
-
-func stringFromPtrOrEmpty(s *string) string {
-	if s == nil {
-		return ""
-	}
-	return *s
-}
-
 // TODO(ENG-192): use proper foreign keys and normalize model.
```
itayd, line 20: @ig-ra is this still in progress?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -57,15 +51,15 @@ var Tables = []any{
 }
 
 type Build struct {
-	BuildID   string `gorm:"primaryKey"`
+	BuildID   sdktypes.UUID `gorm:"primaryKey;type:uuid"`
```
itayd, line 54: why `sdktypes.UUID` and not google's UUID?

sdk/sdktypes/id.go
```diff
@@ -73,6 +76,14 @@ func (i id[T]) Value() string {
...
 }
 
+func (i id[T]) UUIDValue() UUID {
+	if !i.IsValid() {
+		return UUID{}
+	}
+
+	return uuid.UUID(i.tid.UUIDBytes())
+}
+
```
itayd, line 86: i think this is nicer

```suggestion
func (i id[T]) UUIDValue() (uuid UUID) {
	if i.IsValid() {
	    uuid = i.tid.UUIDBytes()
	}
	return
}

```

itayd (approved):

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: IDK, you exposed it, I assumed this is what we want to use

PR #200, opened 2024-04-17
add lock around executorForSessions access

efiShtain:
# Overview

There is a concurrency issue when running high volume of sessions (can be replicated using apache bench, ```ab -n 10000 -c 100  http://localhost:9980/http/http/get  ``` on a Mac, after applying the standard http project)

this solves the issue

not sure this is optimal, but seems to handle it ok

### DIFF
--- a/internal/backend/sessions/sessioncalls/calls.go
+++ b/internal/backend/sessions/sessioncalls/calls.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/sdk/activity"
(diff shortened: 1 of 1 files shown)


internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -58,6 +59,7 @@ type calls struct {
 	//                 user script.
 	generalWorker worker.Worker
 	uniqueWorker  worker.Worker
+	mu            *sync.Mutex
```
itayd, line 62: i think we need to use a RWMutex, and protect also the reads?

itayd: also i would call it sessionsMutex, and since executorsForSessions is global it needs to be in the global scope.

itayd:
replaced with #202

PR #202, opened 2024-04-17
lock sessions executors

itayd:
replaces #200

### DIFF
--- a/internal/backend/sessions/sessioncalls/calls.go
+++ b/internal/backend/sessions/sessioncalls/calls.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"sync"
 
 	enumspb "go.temporal.io/api/enums/v1"
 	"go.temporal.io/sdk/activity"
@@ -20,10 +21,6 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdkexecutor"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
-
-// Pass executors for session specific calls (functions that are
-// defined in runtime script, as opposed to integrations).
-var executorsForSessions = make(map[string]*sdkexecutor.Executors, 16)
 
 type CallParams struct {
 	SessionID     sdktypes.SessionID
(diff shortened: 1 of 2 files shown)


internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -58,6 +55,11 @@ type calls struct {
 	//                 user script.
 	generalWorker worker.Worker
 	uniqueWorker  worker.Worker
+
+	// Pass executors for session specific calls (functions that are
+	// defined in runtime script, as opposed to integrations).
+	executorsForSessions   map[sdktypes.SessionID]*sdkexecutor.Executors
+	executorsForSessionsMu sync.RWMutex
```
tebeka, line 61: The usual practice is to place the lock variable on top of the variables it guards.

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -76,11 +78,12 @@ func New(z *zap.Logger, config Config, svcs *sessionsvcs.Svcs) Calls {
...
-		config:        config,
-		svcs:          svcs,
-		generalWorker: worker.New(svcs.Temporal, taskQueueName, opts),
-		uniqueWorker:  worker.New(svcs.Temporal, uniqueWorkerCallTaskQueueName(), opts),
+		z:                    z,
+		config:               config,
+		svcs:                 svcs,
+		generalWorker:        worker.New(svcs.Temporal, taskQueueName, opts),
+		uniqueWorker:         worker.New(svcs.Temporal, uniqueWorkerCallTaskQueueName(), opts),
+		executorsForSessions: make(map[sdktypes.SessionID]*sdkexecutor.Executors, 16),
```
tebeka, line 86: Why 16?

tebeka (approved):

internal/backend/sessions/sessioncalls/calls.go
itayd: no real reason. just to prep the map with some capacity.

internal/backend/sessions/sessioncalls/calls.go
itayd: news to me, but ok

PR #203, opened 2024-04-17
add `/temporal` and `/version` routes

itayd:
this returns something that was lost due to a bad rebase before merge.


itayd:
force pushing this, as this was already reviewed in the past. just added `/version` as well. lmk if i should revert.

PR #204, opened 2024-04-18
ENG-723: add manifest.schema.yaml to code + ci to validate it when changed

efiShtain:


manifest.schema.yaml
```diff
@@ -0,0 +1,136 @@
+{
```
ig-ra, line 1: it would be nice to have a reference to git hash from whcih this schema was generated

ig-ra (approved):

manifest.schema.yaml
efiShtain: it doesn't really matter, it always match current commit

PR #205, opened 2024-04-18
ENG-688: sessions pagination

efiShtain:
Not done yet
please review the basic idea

This is stacked PR on ENG-702


sdk/sdkservices/sessions.go
```diff
@@ -13,12 +13,22 @@ type ListSessionsFilter struct {
...
 	CountOnly    bool
+
+	PageSize  int32
+	Skip      int32
+	PageToken string
+}
+
+type ListSessionResult struct {
+	Sessions      []sdktypes.Session
+	Count         int
```
itayd, line 24: do we really need this? is there a way where we can ask only for the count?

itayd: oh there is. right.

sdk/sdkservices/sessions.go
```diff
@@ -13,12 +13,22 @@ type ListSessionsFilter struct {
 	BuildID      sdktypes.BuildID
 	StateType    sdktypes.SessionStateType
 	CountOnly    bool
+
+	PageSize  int32
+	Skip      int32
+	PageToken string
```
itayd: i think these should be common to all paginated results. put these in a common struct and just reuse it all over.

sdk/sdkservices/sessions.go
```diff
@@ -13,12 +13,22 @@ type ListSessionsFilter struct {
...
+	Sessions      []sdktypes.Session
+	Count         int
+	NextPageToken string
 }
 
 type Sessions interface {
 	Start(ctx context.Context, session sdktypes.Session) (sdktypes.SessionID, error)
 	Stop(ctx context.Context, sessionID sdktypes.SessionID, reason string, force bool) error
-	List(ctx context.Context, filter ListSessionsFilter) ([]sdktypes.Session, int, error)
+	List(ctx context.Context, filter ListSessionsFilter) (ListSessionResult, error)
```
itayd, line 28:
```suggestion
	List(ctx context.Context, filter ListSessionsFilter) ([]sdktypes.Session, PaginationInfo, error)
```

then you could reuse this everywhere.

itayd:
didn't we talk also about secondary sorting?

sdk/sdkservices/sessions.go
efiShtain: WDYM ? why can't I reuse it everywhere anyway?

internal/backend/db/dbgorm/sessions.go
```diff
@@ -150,21 +150,43 @@ func (db *gormdb) listSessions(ctx context.Context, f sdkservices.ListSessionsFi
...
+	if f.Skip != 0 {
+		q = q.Offset(int(f.Skip))
+	}
+
+	if f.PageToken != "" {
+		q = q.Where("session_id < ?", f.PageToken)
+	}
+
+	// Double order in case we have two rows with the same created_at, then solve order by session_id
+	if err := q.
```
efiShtain, line 166: @itayd 
This is the double sorting

sdk/sdkservices/sessions.go
itayd: add `TotalCount`

sdk/sdkservices/sessions.go
itayd: `TotalCount`

efiShtain: Changed

sdk/sdkservices/sessions.go
efiShtain: I did some changes, 
created an embeddable struct 
PTAL and let me know if that makes sense for you

in any case, let's ack this PR and open another ticket to change it if we want a different design

itayd: i still think these should be separate return vals, no need for the results struct.

I'll approve, but please fix that.

itayd (approved):

PR #206, opened 2024-04-19
change HTTP integration API

ig-ra:
- removed `json_body`, `raw_body` and `form_body` params
- replaced by single `body` param which will be autoparsed to the relevant type (type match and/or headers)
- added tests
- disallow passing args, except url. All other params should be passed as kwargs
- remove form() from response body

### DIFF
--- a/integrations/http/http.go
+++ b/integrations/http/http.go
@@ -68,19 +68,16 @@ func (i integration) request(method string) sdkexecutor.Function {
 
 		// GET request doesn't have user-defined body
 		if method != http.MethodGet && body.IsValid() {
-			if bodyType == "" {
-				bodyType = "missing"
-			}
 
 			var err error
 			bodyType = strings.ToLower(bodyType)
 			switch bodyType {
-			case bodyTypeForm:
-				err = body.UnwrapInto(&formBody)
+			case "", bodyTypeRaw:
+				err = body.UnwrapInto(&rawBody)
 			case bodyTypeJSON:
 				jsonBody = body
-			case bodyTypeRaw:
-				err = body.UnwrapInto(&rawBody)
+			case bodyTypeForm:
+				err = body.UnwrapInto(&formBody)
 			default:
 				err = errors.New("body_type must be one of <form|json|raw>")
 			}
@@ -184,7 +181,7 @@ func (i integration) getConnection(ctx context.Context) map[string]string {
 
 func setBody(req *http.Request, bodyType string, rawBody string, formBody map[string]string, jsonBody sdktypes.Value) error {
 	switch bodyType {
-	case bodyTypeRaw:
+	case "", bodyTypeRaw:
 		if rawBody == "" {
 			return nil
 		}
--- a/integrations/http/http.go
+++ b/integrations/http/http.go
@@ -58,6 +58,8 @@ func (i integration) request(method string) sdkexecutor.Function {
 			"form_body?", &formBody,
 			"json_body?", &jsonBody,
 		)
+		// NOTE: for HTTP.GET {raw,form,json}_body will be ignored
+
 		if err != nil {
 			return sdktypes.InvalidValue, err
 		}
--- a/integrations/http/http.go
+++ b/integrations/http/http.go
@@ -65,32 +65,37 @@ func (i integration) request(method string) sdkexecutor.Function {
 			"params?", &params,
 			"headers?", &headers,
 			"body?", &body,
-			"body_type?", &bodyType,
 		); err != nil {
 			return sdktypes.InvalidValue, err
 		}
 
-		// GET request doesn't have user-defined body
-		if method != http.MethodGet && body.IsValid() {
-			var err error
-			bodyType = strings.ToLower(bodyType)
-			switch bodyType {
-			case "", bodyTypeRaw:
-				err = body.UnwrapInto(&rawBody)
-			case bodyTypeJSON:
-				jsonBody = body
-			case bodyTypeForm:
-				err = body.UnwrapInto(&formBody)
-			default:
-				err = errors.New("body_type must be one of <form|json|raw>")
-			}
-			if err != nil {
-				return sdktypes.InvalidValue, err
-			}
-		}
-
 		if headers == nil {
 			headers = make(map[string]string)
+		}
+
+		// GET request doesn't have user-defined body
+		if method != http.MethodGet && body.IsValid() {
+
+			// parse as RAW
+			bodyType = bodyTypeRaw
+			if err := body.UnwrapInto(&rawBody); err != nil {
+
+				// parse as FORM. Check Content-Type header, if present to distinguish between form and simple JSON
+				bodyType = bodyTypeForm
+				asJSON := false
+				if contentType, ok := headers[contentTypeHeader]; ok {
+					asJSON = contentType == contentTypeJSON
+				}
+				if err = body.UnwrapInto(&formBody); err != nil || asJSON {
+
+					// finally, parse as JSON
+					bodyType = bodyTypeJSON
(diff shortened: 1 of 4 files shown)


integrations/http/http.go
```diff
@@ -35,9 +35,14 @@ var args = sdkmodule.WithArgs(
 	"url",
 	"params?",
 	"headers?",
-	"raw_body?",
```
itayd, line 38: i think you might want to leave these in for backward compatibility unless @haimzlato doesn't have an issue with that

integrations/http/http.go
```diff
@@ -46,22 +51,49 @@ func (i integration) request(method string) sdkexecutor.Function {
...
 		var (
-			rawURL, rawBody           string
+			rawURL, rawBody, bodyType string
 			headers, params, formBody map[string]string
-			jsonBody                  sdktypes.Value
+			jsonBody, body            sdktypes.Value
 		)
-		err := sdkmodule.UnpackArgs(args, kwargs,
+
+		// if body, found := kwargs["body"]; found {
```
itayd: ???

integrations/http/http.go
```diff
@@ -155,13 +187,11 @@ func (i integration) getConnection(ctx context.Context) map[string]string {
...
-func setBody(req *http.Request, rawBody string, formBody map[string]string, jsonBody sdktypes.Value) error {
-	errMutuallyExclusive := errors.New("raw_body, form_body, and json_body are mutually exclusive")
-
-	// Raw body with unknown content type.
-	if rawBody != "" {
-		if formBody != nil || (jsonBody.IsValid() && !jsonBody.IsNothing()) {
-			return errMutuallyExclusive
+func setBody(req *http.Request, bodyType string, rawBody string, formBody map[string]string, jsonBody sdktypes.Value) error {
+	switch bodyType {
+	case bodyTypeRaw:
```
itayd, line 148: this should also be the default

```suggestion
	case "", bodyTypeRaw:
```

integrations/http/http.go
```diff
@@ -171,15 +201,17 @@ func setBody(req *http.Request, rawBody string, formBody map[string]string, json
...
-		return nil
-	}
 
-	// JSON body.
-	if jsonBody.IsValid() && !jsonBody.IsNothing() {
-		if formBody != nil {
-			return errMutuallyExclusive
+		// Set the Content-Type header only if it's not already set.
+		if req.Header.Get(contentTypeHeader) == "" {
+			req.Header.Set(contentTypeHeader, "text/plain") // or "application/octet-stream"
```
itayd: do we need the comment?

itayd: i wonder if we need this header at all. since it's a raw body, it's the user's responsibility. you can add body_type `text` if you want to.

integrations/http/http.go
```diff
@@ -171,15 +201,17 @@ func setBody(req *http.Request, rawBody string, formBody map[string]string, json
...
 		}
+		return nil
 
+	case bodyTypeJSON:
+		if !jsonBody.IsValid() || jsonBody.IsNothing() {
+			return nil
```
itayd: i think in this case the body content should be `null`

integrations/http/http.go
ig-ra: discussed at sync - no backward compatibility. 
resolving.

integrations/http/http.go
ig-ra: removed

integrations/http/http.go
ig-ra: fixed, in one additional place as well

integrations/http/http.go
ig-ra: httpbin won't respond if the content type isn't set. We define raw_body to be string and not bytes, so I think it's nicer to set content type and allow user to override it if he wants.
But I removed the comment.

integrations/http/http.go
ig-ra: could u please explain what you meant? 
I didn't change any logic here - it's copied from the original method - if json isn't valid (or empty) we are not setting the body.

integrations/http/http.go
haimzlato: No need for backward compatibility, but must change all samples and example.

integrations/http/http.go
```diff
@@ -46,26 +51,53 @@ func (i integration) request(method string) sdkexecutor.Function {
...
-		err := sdkmodule.UnpackArgs(args, kwargs,
+
+		if len(args) > 1 {
+			return sdktypes.InvalidValue, errors.New("pass non-URL arguments as kwargs only")
+		}
+
+		if err := sdkmodule.UnpackArgs(args, kwargs,
 			"url", &rawURL,
 			"params?", &params,
 			"headers?", &headers,
```
itayd:
```suggestion
			"params=?", &params,
			"headers=?", &headers,
```

integrations/http/http.go
```diff
@@ -46,26 +51,53 @@ func (i integration) request(method string) sdkexecutor.Function {
...
-			"raw_body?", &rawBody,
-			"form_body?", &formBody,
-			"json_body?", &jsonBody,
-		)
-		if err != nil {
+			"body?", &body,
```
itayd:
```suggestion
			"body=?", &body,
```

ig-ra: done

integrations/http/http.go
ig-ra: done

integrations/http/http.go
ig-ra: by default if not set, req.Body will be nil. you could see similar case in tests, which I added

integrations/http/http.go
```diff
@@ -35,37 +35,83 @@ var args = sdkmodule.WithArgs(
...
+			"body=?", &body,
+		); err != nil {
 			return sdktypes.InvalidValue, err
 		}
 
 		if headers == nil {
 			headers = make(map[string]string)
 		}
 
+		if method != http.MethodGet && body.IsValid() { // NOTE: GET request shouldn't have user-defined body
```
itayd: throw error if GET with body? what does `requests` does?

integrations/http/http_test.go
```diff
@@ -88,3 +92,80 @@ func TestSetQueryParams(t *testing.T) {
...
+		},
+		{
+			name:     "string",
+			body:     sdktypes.NewStringValue("meow"),
+			bodyType: bodyTypeRaw,
+			reqBody:  "meow",
+		},
+		{
+			name:     "json as string => raw",
+			body:     sdktypes.NewStringValue(`{"k":"v"}`),
```
itayd: just pass string

integrations/http/http_test.go
```diff
@@ -88,3 +92,80 @@ func TestSetQueryParams(t *testing.T) {
...
+			bodyType: bodyTypeRaw,
+			reqBody:  `{"k":"v"}`,
+		},
+		{
+			name:     "map[string]string => form",
+			body:     map[string]string{"k": "v"},
```
itayd, line 122: what does requets do?

itayd:
image

use data instead of body
add json

itayd:
https://requests.readthedocs.io/en/latest/api/#requests.post

itayd (approved):

integrations/http/http_test.go
ig-ra: done

integrations/http/http_test.go
ig-ra: form

integrations/http/http.go
ig-ra: ignores body

PR #207, opened 2024-04-19
[ENG-706] Find also "python3" in PATH

tebeka:
Look for `python3` as well in `PATH`. Most common in OSX.

Fixes ENG-706


runtimes/pythonrt/py.go
```diff
@@ -41,8 +41,20 @@ type execInfo struct {
 	Version string
 }
 
+func findPython() (string, error) {
+	names := []string{"python3", "python"}
```
ig-ra, line 67: I think you should start with python and not python3. Python (at least in linux) is usually a link to some python, which could be python3, python 3.12 etc.. actually python3 also will be a link. But I would suggest to change the order

runtimes/pythonrt/py_test.go
```diff
@@ -78,3 +78,34 @@ func processEnv(t *testing.T, pid int) map[string]string {
 	}
 	return env
 }
+
+func genExe(t *testing.T, name string) {
+	file, err := os.Create(name)
+	require.NoError(t, err)
+	file.Close()
+	err = os.Chmod(name, 0o766)
```
ig-ra, line 91: why do you need to chmod?

runtimes/pythonrt/py.go
itayd: it makes more sense to me to start with python3, since we want specifically 3.

itayd: i also think you should make sure that the python you found is python3 using `python --version` or something?

tebeka: @ig-ra: IMO `python3` first, on old Linux systems `python` is still Python 2.
@itayd: I'll check Python version in different PR, see ENG-731

runtimes/pythonrt/py_test.go
tebeka: `LookPath` checks that the file is executable.

ig-ra: this will run in tests, but may fail on actual system
```shell
lrwxrwxrwx - root 12 Feb 17:50  /usr/bin/python3 -> python3.11
```
python may be a link , thus with 777 mod

runtimes/pythonrt/py_test.go
```diff
@@ -78,3 +82,34 @@ func processEnv(t *testing.T, pid int) map[string]string {
...
+	require.NoError(t, err)
+}
+
+func Test_findPython(t *testing.T) {
```
ig-ra, line 95: just curious. If test_findPython failed, could we skip other tests, e.g. venv and run?

runtimes/pythonrt/py_test.go
```diff
@@ -17,10 +17,14 @@ import (
 )
 
 func skipIfNoPython(t *testing.T) {
-	_, err := exec.LookPath("python")
-	if err != nil {
+	_, err := pyExeInfo(context.Background())
```
ig-ra, line 20: could you use the same func in test_createVEnv as well?

tebeka: Done

runtimes/pythonrt/py_test.go
tebeka: That's what's `skipIfNoPython`  for

runtimes/pythonrt/py.go
ig-ra: check that this is a python? version?

tebeka: > check that this is a python? version?

What do you mean? There is a check for minimal python version in #220

runtimes/pythonrt/py_test.go
tebeka: It will work with a link since the link is an executable.

ig-ra (approved):

PR #211, opened 2024-04-20
os module; call timeouts

itayd:
introduces:
- initial os module, to be active only in dev mode or when explicitly specified in config. allows to shell out. useful in many scenarios, such as running terraform and such. also good for demos. does not execute in a jailed environment, might be added in the future.
- call timeouts, can be specified with `ak_timeout=<duration>` on any connection call.


integrations/os/os.go
```diff
@@ -0,0 +1,98 @@
...
+	return fixtures.NewBuiltinExecutor(
+		ExecutorID,
+		sdkmodule.ExportFunction("command", command, sdkmodule.WithArgs("cmd", "*args")),
+		sdkmodule.ExportFunction("shell", shell, sdkmodule.WithArgs("cmd", "sh?")),
+	)
+}
+
+func execute(ctx context.Context, name string, args ...string) (sdktypes.Value, error) {
+	cmd := exec.CommandContext(ctx, name, args...)
+	cmd.WaitDelay = 250 * time.Millisecond // without this, cancellations do not work properly.
```
tebeka: TIL :)

integrations/os/os.go
```diff
@@ -0,0 +1,98 @@
...
+	var exit *exec.ExitError
+	if errors.As(err, &exit) {
+		rc = exit.ExitCode()
+
+		if rc < 0 && (errors.Is(ctx.Err(), context.DeadlineExceeded) || errors.Is(ctx.Err(), context.Canceled)) {
+			// rc could not be determined and the context was canceled or timed out.
+			err = ctx.Err()
+		} else {
+			// rc could be determined or no error context occured.
+			err = nil
```
tebeka: You mask the error returned from `cmd.CombinedOutput`, why?
If the process exited normally, then the returned error from `cmd.CombinedOutput` will be `nil`.

Also, do we want to account for very large output? Maybe use `io.LimitReader` on `cmd.Stdout`?

integrations/os/os.go
```diff
@@ -0,0 +1,98 @@
...
+		}
+	}
+
+	if err == nil {
```
tebeka: Please flip the logic to make it more idiomatic.

```
if err != nil {
    return sdktypes.InvalideValue, err
}
// rest of code
```

integrations/os/os.go
```diff
@@ -0,0 +1,98 @@
...
+	var cmd, sh string
+
+	err := sdkmodule.UnpackArgs(args, kwargs, "cmd", &cmd, "sh?", &sh)
+	if err != nil {
+		return sdktypes.InvalidValue, err
+	}
+
+	var cmdArgs []string
+
+	if runtime.GOOS == "windows" {
```
tebeka: Instead of runtime check, use a separate file with a built tag.
See https://github.com/tebeka/desktop for example.

internal/backend/akmodules/akmodules.go
```diff
@@ -2,14 +2,15 @@ package akmodules
 
 import (
 	"go.autokitteh.dev/autokitteh/internal/backend/akmodules/ak"
+	"go.autokitteh.dev/autokitteh/internal/backend/akmodules/os"
```
tebeka, line 5: Why do we need both this and `bootstrap.star`?

internal/backend/akmodules/os/os.go
```diff
@@ -0,0 +1,98 @@
+package os
```
tebeka, line 1: A general comment. The fact we need both this one and `os` in `integrations` shows a design issue.

internal/backend/akmodules/os/os.go
```diff
@@ -0,0 +1,98 @@
...
+
+func New() sdkexecutor.Executor {
+	return fixtures.NewBuiltinExecutor(
+		ExecutorID,
+		sdkmodule.ExportFunction("command", command, sdkmodule.WithArgs("cmd", "*args")),
+		sdkmodule.ExportFunction("shell", shell, sdkmodule.WithArgs("cmd", "sh?")),
+	)
+}
+
+func execute(ctx context.Context, name string, args ...string) (sdktypes.Value, error) {
```
tebeka: This is pretty much a duplication of the code in `integrations`, any way to combine them?

internal/backend/sessions/sessioncalls/call.go
```diff
@@ -77,7 +81,41 @@ func (cs *calls) invoke(ctx context.Context, callv sdktypes.Value, args []sdktyp
...
+					<-ctx.Done()
+					cancel()
+				}()
+			}
+		}
+	}
+
+	v, err := caller.Call(ctx, callv, args, kwargs)
+
+	if err != nil && errors.Is(err, workflow.ErrCanceled) && errors.Is(ctx.Err(), context.DeadlineExceeded) {
```
tebeka: Use this condition in line 107 and get rid of this section?

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -238,6 +239,12 @@ func (w *sessionWorkflow) initGlobalModules() (map[string]sdktypes.Value, error)
 
 	vs := make(map[string]sdktypes.Value, len(execs))
 
+	if w.ws.cfg.OSModule {
```
tebeka, line 242: Yet another option to support. What is the reason we don't want the `os` module?

tebeka (commented):
Apart from the comments, I prefer the Python API in https://docs.python.org/3/library/subprocess.html#subprocess.run

integrations/os/os.go
itayd: i thought about it, but tbh - not sure it's worth it? what's wrong in this approach?

itayd:
well shit, i need to drop the one in integrations. i left it there by mistake.

internal/backend/sessions/sessionworkflows/workflow.go
itayd: when deploying in cloud

internal/backend/akmodules/os/os.go
itayd: mistake.

internal/backend/akmodules/akmodules.go
itayd: not sure i follow this q?

integrations/os/os.go
itayd: you and me both when i saw the cancellation doesn't work. quite surprising.

integrations/os/os.go
itayd: done

internal/backend/akmodules/os/os.go
itayd: mistake.

internal/backend/sessions/sessioncalls/call.go
itayd: it's comparing two different things, err and ctx.Err, so not so simple

integrations/os/os.go
itayd: i don't really mask it. rc is determined and that's what is returned. the rc variable will appear in the result.

itayd:
added some features, changed return value to a struct. we should all discuss this.

internal/backend/akmodules/akmodules.go
tebeka: There's an import here and we also have an entry for the module in `bootstrap.star`, I was wondering why we have both.

internal/backend/sessions/sessionworkflows/workflow.go
tebeka: IMO the right solution is the general "isolation" discussion we have. But for now OK.

tebeka (approved):

internal/backend/akmodules/akmodules.go
itayd: ah - starlark has a concept of 'predeclared' variables, it needs to know the symbols it's going to use.

PR #213, opened 2024-04-21
introduce vdev mode

itayd:
vdev = volatile dev, same as dev - but without persistency.

vdev inherits default config.
dev inherits vdev config.
test inherits vdev config.


internal/backend/db/dbfactory/factory.go
```diff
@@ -23,6 +23,7 @@ var Configs = configset.Set[Config]{
 		SlowQueryThreshold: 300 * time.Millisecond,
 		Type:               gormkitteh.RequireExplicitDSNType,
 	},
+	VolatileDev: &Config{},
```
itayd: add SlowQueryThreashold

itayd: done

PR #215, opened 2024-04-21
[ENG-709] Better detection of functions from `builtins` module

tebeka:


tebeka:
This change is also in #216. We'll probably merge the other before and then I'll close this one.

tebeka:
Done in #216

PR #216, opened 2024-04-21
[ENG-703] Support Replay

tebeka:
Fixed the Go <-> Python flow where Python sends result back to Go and then receives it. This enables support for temporal replays.

NOTE: This also incorporates #215 since it's required for the fix.


tebeka:
Currently untested. There's a `replay` flow but need to find out how to run `ak` in persistent mode.

tebeka:
Tested with the `replay` workflow.

```
NOW: 2024-04-24 12:55:36.004120
NOW: 2024-04-24 12:55:36.004120
```

efiShtain (approved):
Not sure I fully understand it since I'm not involved in the pythonrt but it seems ok to me

PR #217, opened 2024-04-22
remove env var API

ig-ra:


internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -262,9 +262,9 @@ func ParseEnv(e Env) (sdktypes.Env, error) {
 }
 
 type EnvVar struct {
-	EnvID sdktypes.UUID `gorm:"index;type:uuid"`
-	Name  string
-	Value string // not set if is_secret.
+	EnvID sdktypes.UUID `gorm:"primaryKey;index;type:uuid"`
```
efiShtain, line 263: Maybe you should add *Env to set the foreign key correctly

efiShtain (approved):

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: there is foreign key a couple of lines below. I added combined primary to facilitate model access through gorm

PR #221, opened 2024-04-24
ENG-745: fix server migrate in cli not working

efiShtain:


cmd/ak/cmd/server/setup.go
```diff
@@ -21,14 +18,14 @@ var setupCmd = common.StandardCommand(&cobra.Command{
...
 func Setup() error {
-	db, err := InitDB(mode)
-	if err != nil {
-		return err
-	}
-
-	if err := db.Setup(context.Background()); err != nil {
-		return fmt.Errorf("setup: %w", err)
-	}
+	// db, err := InitDB(mode)
```
ig-ra: why this is commented out? no setup is needed?

internal/backend/svc/svc.go
```diff
@@ -320,7 +320,9 @@ func NewOpts(cfg *Config, ropts RunOptions) []fx.Option {
 	return append(opts, fx.Populate(svcs))
 }
 
-func StartDB(ctx context.Context, cfg *Config) (db.DB, error) {
+func StartDB(ctx context.Context, cfg *Config, ropt RunOptions) (db.DB, error) {
+	setFXRunOpts(ropt)
```
ig-ra, line 324: how mode is influence db? it's always persistent, no? I thought that the only things are relevant are DSN and type

efiShtain: Mode dev hides the default db location 
So I used the same logic, instead of duplicating it
That's the same usage when starting the server so I wanted to have the same here experience

cmd/ak/cmd/server/setup.go
efiShtain: Setup does nothing
I'll remove the command entirely

ig-ra (approved):

PR #223, opened 2024-04-28
[ENG-544] Stop temporal log spam

tebeka:
I don't love this fix (see comments in ENG-544), but it works :)


tebeka:
Got some unrelated changes here, will repoen a new one.

PR #226, opened 2024-04-28
fix: wrap and unwrap nil

ig-ra:


sdk/sdktypes/value_wrap_test.go
```diff
@@ -66,10 +66,15 @@ func TestValueWrapper(t *testing.T) {
 			w:   sdktypes.NewStringValue("meow"),
 			unw: "meow",
 		},
+		// {
```
itayd: kill

itayd (approved):

itayd (approved):
oh, please add a test to what happens with an empty struct as well.

sdk/sdktypes/value_wrap_test.go
ig-ra: added `struct{}{}` test

PR #227, opened 2024-04-28
[ENG-552] Logging from Python

tebeka:
Logging from Python to `ak`. 
The user will call `logging.info` and friends to get logs out. 
Standard output not captured (see ENG-708)


runtimes/pythonrt/ak_runner.py
```diff
@@ -18,13 +18,31 @@
...
+    def __init__(self, comm):
+        level = logging.getLogger().getEffectiveLevel()
+        super().__init__(level)
+        self.comm = comm
+        self.formatter = logging.Formatter()
+
+    def emit(self, record: logging.LogRecord):
+        level = record.levelname
+        if level == 'CRITICAL':
+            level = 'ERROR'
```
daabr: Just a suggestion, up to you:
```
level = 'ERROR' if record.levelname == 'CRITICAL' else record.levelname

daabr (approved):
Approved with 1 tiny comment

PR #230, opened 2024-04-30
[ENG-708] Log Pythons stdout/stderr to autokitteh

tebeka:
Done from Go's level by replacing the Python command stdout & stderr with custom io.Writer that uses the callback `Print`.
Adding an `[stdout]` prefix to stdout messages and `[stderr] ` to stderr.


ig-ra (approved):

runtimes/pythonrt/ak_runner.py
```diff
@@ -18,30 +18,13 @@
 from pathlib import Path
 from socket import AF_UNIX, SOCK_STREAM, socket
 
-# Must come first before any logging.
+# Must come first before any logging or loading of user code.
 logging.basicConfig(
     format='[PYTHON] %(asctime)s - %(levelname)s - %(filename)s:%(lineno)d - %(message)s',
```
ig-ra: I suggest not to change the default format

PR #232, opened 2024-05-01
envvars becomes var, can also store vars for connections

itayd:


internal/backend/db/dbgorm/vars.go
```diff
@@ -0,0 +1,116 @@
...
+	"github.com/google/uuid"
+	"gorm.io/gorm/clause"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/db/dbgorm/scheme"
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func (db *gormdb) SetVars(ctx context.Context, vars []sdktypes.Var) error {
```
ig-ra, line 21: please split all relevant functions to SDK/schema for testability and add `vars_test.go` (see env_tests.go removed lines)

ig-ra:
I think we should try to open much smaller PRs, i.e. ~300 lines (excluding auto-generated files, of coz).
For example in this PR all db-related changes (EnvVar -> Var) could be extracted and handled in separate PR.

internal/backend/db/dbgorm/vars.go
```diff
@@ -0,0 +1,116 @@
...
+			}
+		}
+
+		return translateError(tx.db.Clauses(clause.OnConflict{
+			UpdateAll: true,
+		}).Create(&dbvs).Error)
+	})
+}
+
+func (db *gormdb) GetVars(ctx context.Context, reveal bool, sid sdktypes.VarScopeID, ns []sdktypes.Symbol) ([]sdktypes.Var, error) {
```
ig-ra: ns -> names?

internal/backend/db/dbgorm/vars.go
```diff
@@ -0,0 +1,116 @@
...
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func (db *gormdb) SetVars(ctx context.Context, vars []sdktypes.Var) error {
+	if i, err := kittehs.ValidateList(vars, func(_ int, v sdktypes.Var) error {
+		return v.Strict()
+	}); err != nil {
+		return fmt.Errorf("#%d: %w", i, err)
```
ig-ra, line 25: let's add a string to make this error more human-friendly?
fmt.Errorf("setVars: #%d: %w", i, err)

internal/backend/db/dbgorm/vars.go
```diff
@@ -0,0 +1,116 @@
...
+	}
+
+	return db.transaction(ctx, func(tx *tx) error {
```
ig-ra, line 36: could we enclose in db.transaction only db-related ops?
e.g. first create (not under transaction) list of scheme.Var with everything precomputed and separated list of indexes + cids to fetch iid(s), 
then under transcation only fetch and update?

internal/backend/db/dbgorm/vars.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		return translateError(tx.db.Clauses(clause.OnConflict{
+			UpdateAll: true,
+		}).Create(&dbvs).Error)
+	})
+}
+
+func (db *gormdb) GetVars(ctx context.Context, reveal bool, sid sdktypes.VarScopeID, ns []sdktypes.Symbol) ([]sdktypes.Var, error) {
+	q := db.db.Where("scope_id = ?", sid.UUIDValue())
```
ig-ra: should we check sid.IsValid() and return error? this is what we did in EnvVars

cmd/ak/cmd/vars/delete.go
```diff
@@ -35,7 +29,7 @@ var removeCmd = common.StandardCommand(&cobra.Command{
 		ctx, cancel := common.LimitedContext()
 		defer cancel()
 
-		if err := envs().RemoveVar(ctx, id, k); err != nil {
+		if err := vars().Delete(ctx, id, k); err != nil {
 			return fmt.Errorf("remove environment variable: %w", err)
```
ig-ra: delete?

internal/backend/db/dbgorm/vars.go
itayd: https://linear.app/autokitteh/issue/ENG-768/db-vars-testing

internal/backend/db/dbgorm/vars.go
itayd: done

internal/backend/db/dbgorm/vars.go
itayd: i think that adding the `setVars` perfix is wrong, the calling function should add it.

cmd/ak/cmd/vars/delete.go
itayd: ?

itayd: oh right! done

internal/backend/db/dbgorm/vars.go
itayd: i want to do a one over of all of these in a different pr

internal/backend/db/dbgorm/vars.go
itayd: now getting all connections in a batch

manifest.schema.yaml
```diff
@@ -13,6 +13,12 @@
         },
         "integration": {
           "type": "string"
+        },
```
daabr, line 16: Before merging this, please also prepare a PR that fixes the manifests in https://github.com/autokitteh/samples

internal/backend/db/dbgorm/sessions.go
```diff
@@ -183,7 +183,7 @@ func (db *gormdb) ListSessions(ctx context.Context, f sdkservices.ListSessionsFi
 
 	// Only if we have a full page, there might be more sessions
 	nextPageToken := ""
-	if len(sessions) == int(f.PageSize) {
+	if len(sessions) == int(f.PageSize) && len(sessions) > 0 {
```
daabr: More efficient:
```suggestion
	if len(sessions) == int(f.PageSize) && f.PageSize > 0 {
```

internal/backend/db/dbfactory/factory.go
```diff
@@ -24,8 +24,7 @@ var Configs = configset.Set[Config]{
 		Type:               gormkitteh.RequireExplicitDSNType,
 	},
 	Dev: &Config{
-		Type: "sqlite",
-		DSN:  "file:" + filepath.Join(xdg.DataHomeDir(), "autokitteh.sqlite"),
+		DSN: "sqlite:file:" + filepath.Join(xdg.DataHomeDir(), "autokitteh.sqlite"),
```
daabr, line 27: I remember that somewhere there was an explicit check of the type value. Do you know where? Is that changed?

cmd/ak/cmd/vars/vars.go
```diff
@@ -0,0 +1,96 @@
...
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+// Flags shared by all the subcommands.
+var env, project, conn string
+
+var varsCmd = common.StandardCommand(&cobra.Command{
+	Use:     "vars",
+	Short:   "Variable subcommands",
+	Aliases: []string{"var", "vs", "v"},
```
daabr: In accordance with #229:

```suggestion
	Use:   "var",
	Short: "Connection variable subcommands: set, get, delete",
```

internal/backend/db/dbgorm/sessions.go
itayd: doesn't belong in this pr

cmd/ak/cmd/manifest/apply.go
```diff
@@ -10,7 +10,7 @@ import (
 var projectName string
 
 var applyCmd = common.StandardCommand(&cobra.Command{
-	Use:     "apply [file] [--project-name <name>] [--no-validate] [--from-scratch] [--quiet]",
+	Use:     "apply [file] [--project-name <name>] [--no-validate] [--from-scratch] [--quiet] [--rm-unused-cvars]",
```
daabr, line 13: 1. This feels like a very specific and hard-to-explain/use flag, but I don't have a good idea to fix that for now
2. Why the focus on just connection vars?
3. "rm" is a very OS-specific term, maybe "del"?

cmd/ak/cmd/manifest/plan.go
```diff
@@ -7,10 +7,10 @@ import (
 	"go.autokitteh.dev/autokitteh/internal/manifest"
 )
 
-var noValidate, fromScratch bool
+var noValidate, fromScratch, rmUnusedConnVars bool
```
daabr: Vars for flags that are shared across commands belong in the parent command, i.e. `manifest.go`

cmd/ak/cmd/vars/set.go
```diff
@@ -0,0 +1,49 @@
...
+		if err := vars().Set(ctx, ev); err != nil {
+			return fmt.Errorf("set variable: %w", err)
+		}
+		return nil
+	},
+})
+
+func init() {
+	// Command-specific flags.
+	setCmd.Flags().BoolVarP(&secret, "secret", "s", false, "this is a secret")
```
daabr:
```suggestion
	setCmd.Flags().BoolVarP(&secret, "secret", "s", false, "this value is secret")
```

daabr (approved):
A few more small comments, but nothing substantial beyond what we discussed. Approving in advance.

internal/backend/db/dbgorm/vars.go
ig-ra: then we need to ensure that calling function is doing this, which is not the case :)
that's why I suggest to add it here at least

internal/backend/db/dbgorm/vars.go
ig-ra: @itayd merged the tests into your PR

ig-ra (approved):

internal/backend/db/dbfactory/factory.go
ig-ra: I think it's in gormkitteh, but there is a call for `explicit()` which will parse `sqlite` as type from DSN

itayd: yeah and he doesn't like it :) wdyt?

itayd: let's discuss https://linear.app/autokitteh/issue/ENG-769/gormkitteh-db-type-in-dsn

cmd/ak/cmd/manifest/apply.go
itayd: connection vars can be created dynamically by the integration. when applying a manifest - i don't want to delete them, so by default this is not removing any.

i actually think `rm` is nicer than `del`.

cmd/ak/cmd/manifest/plan.go
itayd: moved

manifest.schema.yaml
itayd: https://github.com/autokitteh/samples/pull/37

internal/backend/db/dbfactory/factory.go
daabr: See my comment there

cmd/ak/cmd/manifest/apply.go
daabr: This PR is already merged, and I don't have a *strong* opinion, but consider that:
- a Windows user has no idea what "rm" means; "room"? "rem"?
- a Unix user understands what "del" means, even if it's not a Unix command

PR #234, opened 2024-05-01
ENG-684: add authenticated mux + descope integration

efiShtain:
# Overview

Added integration on client side for descope
Added support for authenticated access using auth mix
Added an example for authenticated call (with token) and unauthorized call

New env variables:
```
        "AK_AUTHENTICATOR__ENABLED": "true", // default false
        "AK_AUTHENTICATOR__PROVIDER": "descope", // currently only descope is supported
        "AK_AUTHENTICATOR__PROJECT_ID": "" // depends on provider, currently only project_id is required
```

## Details
This is client side authentication mostly, server only verifies the token but is not part of the authentication process it self
we can guard any endpoint we want

```Internal/backend/auth ``` has an authenticator interface that exposes a middleware that suppose to handle authentication. 

Default authentication is noAuthAuthentication that verifies nothing and sets a default user in context

Descpose authentication verifies the JWT and set basic user details in context

### CODE REVIEW

File: internal/backend/auth/auth.go [hunk 1]
```diff
@@ -0,0 +1,50 @@
...
+	Name   string
+	Config map[string]string
+}
+
+type Authenticator interface {
+	Middleware(next http.Handler) http.Handler
+	Provider() ProviderDetails
+}
+
+const AuthenticatedUserCtxKey = "auth:user_details"
```
daabr: See lint error: define this key as a `struct{}` instead of a string to prevent collisions.

internal/backend/auth/auth.go
```diff
@@ -0,0 +1,50 @@
...
+		z.Info("authentication is disabled")
+		return &noAuthAuthenticator{}, nil
+	}
+
+	switch cfg.Provider {
+	case "descope":
+		return newDescopeAuthenticator(cfg, z)
+	}
+
+	return nil, ErrNotSupportedAuthProvider
```
daabr, line 53: Why not a `default` case? Does Go complain that you're not returning anything?

internal/backend/auth/descope.go
```diff
@@ -0,0 +1,83 @@
...
+func newDescopeAuthenticator(cfg *Config, z *zap.Logger) (Authenticator, error) {
+	if cfg.ProjectID == "" {
+		return nil, ErrInvalidAuthProviderConfiguration
+	}
+
+	ll := stringToLogLevel(cfg.LogLevel)
+
+	descopeClient, err := client.NewWithConfig(&client.Config{ProjectID: cfg.ProjectID, LogLevel: ll})
+	if err != nil {
+		z.Error("failed initiating descope: %w", zap.Error(err))
```
daabr: Remove `": %w"`, the `zap.Error()` is a separate field in the log, not a part of the error message.

internal/backend/svc/svc.go
```diff
@@ -207,20 +208,25 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
...
 
-				wrap := func(h http.Handler) http.Handler {
-					// TODO(ENG-3): Wrap the handler with auth middleware.
-					return h
-				}
-
 				mux = svc.Mux()
 
-				mux.Handle("/", wrap(authed))
+				mux.Handle("/", authenticator.Middleware(authed))
```
daabr: Question: if Descope is enabled, but I haven't logged in yet, or my login failed, won't this prevent me from seeing anything? Won't it affect images, CSS and scripts? Won't it affect integration webhooks?

daabr (commented):
General comments:
1. Please fix various lint errors about unchecked errors
2. Embedded HTML files should be under `/web` (I was surprised about the main AK `index.html`, you don't have to fix that, but at least move your new files)
3. See my question in `svc.go` - I'll approve after you answer this
4. FYI but no action item needed - I don't like to combine CSS and JS inside HTML files, and using `/web` helps with this (see integrations), but it's not important to fix this in this PR or even soon

internal/backend/svc/svc.go
efiShtain: I might have put the authenticated route on the wrong place,
technically we would only want to block APIs and not web pages / standard http which is not grpc
I'll have a look

internal/backend/auth/auth.go
efiShtain: it doesn't really matter if you return from the cases

efiShtain:
> General comments:
> 
> 1. Please fix various lint errors about unchecked errors
> 2. Embedded HTML files should be under `/web` (I was surprised about the main AK `index.html`, you don't have to fix that, but at least move your new files)
> 3. See my question in `svc.go` - I'll approve after you answer this
> 4. FYI but no action item needed - I don't like to combine CSS and JS inside HTML files, and using `/web` helps with this (see integrations), but it's not important to fix this in this PR or even soon

regarding 2, I'll move the files

regarding 4, everything here is under the assumption that this won't be the "real" UI, everything is only for dev purposes, so it doesn't really matter as long as it works
with that said, the direction the world goes with is everything on the same file (styling, backend logic, frontend logic, and more), look at any new frontend framework like nextjs / remix / svetle / vuejs and more. putting stuff together make it easier in lots of cases, especially sharing components. But again, it doesn't really matter for this case since it's not production UI, just dev stuff

internal/backend/auth/auth.go
efiShtain: fixed

internal/backend/auth/descope.go
efiShtain: good point, fixed

internal/backend/svc/svc.go
efiShtain: fixed it so only routes that begins with /api/ would be authenticated
next PR once this is approved is updating all our grpc endpoints to move to the authenticated routes

daabr (approved):

itayd:
going to merge this and incorporate into my changes

PR #235, opened 2024-05-02
ENG-693: Ready file --> /readyz

daabr:
System test:

- Removed the `ready-file` flag from the `up` CLI command
- Removed the usage of a mutex buffer as the AK server’s combined output (no need to check the output at all anymore)
- Notice: PR #45 deleted the logic to wait for the AK server’s readiness in the non-subprocess mode
  - Asked Itay: this was unintentional, but it’s unclear if it’s a problem
  - `fx.App.Start()` is blocking, but individual `Start()` functions aren’t?
- Wait for the server’s `/readyz` URL, no matter how it was started
- Indirect bug fix: SQLite filename prefix `sqlite:` → `file:`
- Indirect bug fix: `client.go` - run AK `event` and `session` instead of `events` and `sessions`

SvcProc:

- Remove the `ready-file` flag from the `up` CLI command
- No need to create a temp dir, the system test already did it in `parse.go`
- Removed unnecessary `Addr()` function from AK “service” interface
- Indirect bug fix: AK server in subprocess mode should also use a SQLite file in the test’s temp directory


tests/system/server.go
```diff
@@ -5,43 +5,102 @@ import (
...
+		}
+
+		// For now, the availability of "/readyz" is sufficient,
+		// no need to check the response body yet.
+		if resp.resp.StatusCode == 200 {
+			result <- true
+			return
+		}
+
+		time.Sleep(10 * time.Millisecond)
```
ig-ra: I think it will take more than 10 milli to start server anyway..
maybe we could move time.Sleep in the beginnning of the loop and remove it from 3 places?
or use 
```
for {
    if condition {
      break
    }
    ...
}
sleep
```

tests/system/server.go
```diff
@@ -5,43 +5,102 @@ import (
...
-		panic(fmt.Errorf("fx app start: %w", err))
+	return server, addr, nil
+}
+
+// Wait for the AK server to be ready, with a timeout,
+// and return its HTTP address for client connections.
+func waitForReadiness() (string, error) {
+	ready := make(chan bool, 1)
+	timer := time.NewTimer(serverReadyTimeout)
+	go queryReadyz(ready)
```
ig-ra, line 67: maybe to return file you already read. In case of success you are reading it twice

tests/system/server.go
```diff
@@ -5,43 +5,102 @@ import (
...
+		resp, err := sendRequest(addr, httpRequest{method: "GET", url: "/readyz"})
+		if err != nil {
+			time.Sleep(10 * time.Millisecond)
+			continue
+		}
+
+		// For now, the availability of "/readyz" is sufficient,
+		// no need to check the response body yet.
+		if resp.resp.StatusCode == 200 {
+			result <- true
```
ig-ra: maybe to push into the channel the addr you read? in the case of success you will read the file twice

tests/internal/svcproc/svcproc.go
```diff
@@ -4,128 +4,89 @@ import (
...
+func (s *svcProc) Start(ctx context.Context) error {
+	// TODO: Pass user configuration to executable.
+	s.cmd = exec.Command(
+		s.binPath, "up",
+		"--mode", string(s.ropts.Mode),
 
-	addrFilePath := filepath.Join(tmpDir, "addr")
-	readyFilePath := filepath.Join(tmpDir, "ready")
+		"--config", "db.type=sqlite",
+		"--config", "db.dsn=file:autokitteh.sqlite", // In the test's temporary directory.
```
ig-ra, line 47: plz see https://github.com/autokitteh/autokitteh/pull/232#discussion_r1587134677
I donno what's the right thing to do - either use `sqlite://file:` or just `file:`

ig-ra (changes requested):

tests/internal/svcproc/svcproc.go
```diff
@@ -4,128 +4,89 @@ import (
...
-		return fmt.Errorf("start: %w", err)
+	if err := s.cmd.Start(); err != nil {
+		return fmt.Errorf("start subprocess: %w", err)
 	}
 
-	wait := make(chan fx.ShutdownSignal, 1)
+	s.wait = make(chan fx.ShutdownSignal, 1)
 	go func() {
-		var rc int
+		var ec int
```
itayd: dude nobody knows what `ec` is by just looking at it

tests/internal/svcproc/svcproc.go
daabr: See my comment in https://linear.app/autokitteh/issue/ENG-769/gormkitteh-db-type-in-dsn

Also note that `sqlite://` is not a valid prefix anyway

tests/internal/svcproc/svcproc.go
daabr: See line 66: `ec = exitError.ExitCode()`, or line 73: `{ExitCode: ec}`.

"rc" is the acronym without explanation. :)

tests/system/server.go
daabr: I'll move it as you suggested, but note that 10ms is the wait time between checks, not the timeout (`serverReadyTimeout = 5 * time.Second` in line 18)

tests/system/server.go
daabr: good idea

tests/internal/svcproc/svcproc.go
ig-ra: ec -> exitCode? or ret?
Short var names help to minimize code clutter and thus increasing readability.
In this specific case IMO variable name `ec` reduces readability. Let's use meaningful names? e.g. `exitCode` or at least `ret` ?

ig-ra (approved):

tests/internal/svcproc/svcproc.go
daabr: makes sense, I'll use `exitCode` for consistency

PR #239, opened 2024-05-05
Auth

itayd:
Adding on @efiShtain 's #234 , this adds google and github auth. Also allows to create tokens that can be used in APIs as well as VSCode.


cmd/ak/cmd/auth/auth.go
```diff
@@ -0,0 +1,27 @@
...
+})
+
+// AddSubcommands adds this command, and its own subcommands, to the calling parent.
+func AddSubcommands(parentCmd *cobra.Command) {
+	parentCmd.AddCommand(authCmd)
+}
+
+func init() {
+	// Subcommands.
+	authCmd.AddCommand(whoamiCmd)
```
efiShtain, line 23: you need some way to set the token in the cli
something like 
```bash
ak auth set-token <token> # or it should ask the user to enter it if we don't want to keep it in history
```
or 
```bash 
ak auth login # this will open the UI in the create token page or something we should think of
```

cmd/ak/cmd/auth/create_token.go
```diff
@@ -0,0 +1,32 @@
...
+var createTokenCmd = common.StandardCommand(&cobra.Command{
+	Use:   "create-token",
+	Short: "Create auth token",
+	Args:  cobra.NoArgs,
+
+	RunE: func(cmd *cobra.Command, _ []string) error {
+		ctx, cancel := common.LimitedContext()
+		defer cancel()
+
+		tok, err := auth().CreateToken(ctx)
```
efiShtain, line 18: this can't really work, you can't create a token unless you are authenticated

web/static/embed.go
```diff
@@ -45,9 +45,3 @@ var SlackWebContent embed.FS
 
 //go:embed twilio/connect
 var TwilioWebContent embed.FS
-
-//go:embed descopelogin/index.html
-var DescopeIndex string
```
efiShtain, line 50: @daabr requested specifically to put the static pages here
do we have a convention ?

sdk/sdkservices/auth.go
```diff
@@ -0,0 +1,12 @@
...
+
+import (
+	"context"
+
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type Auth interface {
+	WhoAmI(ctx context.Context) (sdktypes.User, error)
+	CreateToken(ctx context.Context) (string, error)
```
efiShtain, line 11: not for now, but most of the times when creating tokens, there is a possibility to make it expire sometime in the future, we should probably support that

sdk/sdktypes/user.go
```diff
@@ -0,0 +1,37 @@
...
+
+func NewUser(provider string, data map[string]string) User {
+	return kittehs.Must1(UserFromProto(&UserPB{
+		Provider: provider,
+		Data:     data,
+	}))
+}
+
+func (u User) UniqueID() (id string) {
+	if id = u.Data()["email"]; id == "" {
```
efiShtain, line 34: are you sure we can count on email here?
if email doesn't exists, all users without email would get the same <unknown> id, won't that be a problem?

sdk/sdktypes/user.go
```diff
@@ -0,0 +1,37 @@
...
+
+type UserTraits struct{}
+
+func (UserTraits) Validate(m *UserPB) error       { return nil }
+func (UserTraits) StrictValidate(m *UserPB) error { return nil }
+
+func UserFromProto(m *UserPB) (User, error) { return FromProto[User](m) }
+
+func (u User) Data() map[string]string { return u.read().Data }
+func (u User) Provider() string        { return u.read().Provider }
```
efiShtain, line 23: why do we need the provider on the user type ?

proto/autokitteh/users/v1/user.proto
```diff
@@ -0,0 +1,8 @@
+syntax = "proto3";
+
+package autokitteh.users.v1;
+
+message User {
+  string provider = 1;
```
efiShtain, line 6: don't we need an id at least ?

proto/autokitteh/events/v1/svc.proto
```diff
@@ -48,12 +48,6 @@ message GetResponse {
 }
 
 message ListRequest {
-  option (buf.validate.message).cel = {
-    id: "list.mutuals",
-    message: "integration_token must be specified with integration_id",
-    expression: "this.integration_id != '' && this.integration_token != ''",
-  };
-
```
efiShtain, line 56: isn't this part of the connections change pr ?
otherwise why not remove the integration_token field as well?

internal/backend/webtools/webtools.go
```diff
@@ -40,36 +41,36 @@ func (svc *svc) Setup(ctx context.Context) error {
...
 	}
 
 	// messages api paths.
-	mux.HandleFunc("GET /webtools/api/msgs/{addr}", svc.getMessages)
-	mux.HandleFunc("GET /webtools/api/msgs", svc.getMessages)
-	mux.HandleFunc("POST /webtools/api/msgs/{addr}", svc.postMessage)
-	mux.HandleFunc("POST /webtools/api/msgs", svc.postMessage)
-	mux.HandleFunc("DELETE /webtools/api/msgs/{addr}/{id}", svc.deleteMessage)
-	mux.HandleFunc("DELETE /webtools/api/msgs/{addr}", svc.deleteMessage)
+	muxes.HandleFunc("GET /webtools/api/msgs/{addr}", svc.getMessages)
```
efiShtain: should all of this be authenticated API ?

internal/backend/svc/svc.go
```diff
@@ -212,79 +239,65 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 				// Replace the original mux with a main mux and also expose the original mux as the no-auth one.
 				apimux := http.NewServeMux()
 
-				mux = svc.Mux()
-				mux.Handle("/api/", http.StripPrefix("/api", authenticator.Middleware(apimux)))
+				mux := svc.Mux()
+				mux.Handle("/", wrapAuth(apimux))
+				mux.Handle("/api/", http.StripPrefix("/api", wrapAuth(mux)))
```
efiShtain: not sure why do you wrapAuth both / and /api ?
/ suppose to be not protected (for static resources and index page and such)
/api should be the prefix for all grpc endpoints

internal/backend/muxes/muxes.go
```diff
@@ -2,4 +2,23 @@ package muxes
...
+type Muxes struct {
+	API, Root *http.ServeMux
+	WrapAuth  func(http.Handler) http.Handler
+}
+
+func (m *Muxes) Handle(pattern string, handler http.Handler) {
+	m.Root.Handle(pattern, handler)
+}
+
+func (m *Muxes) HandleFunc(pattern string, handler func(http.ResponseWriter, *http.Request)) {
```
efiShtain: why do you need this wrapper if API and Root are publicly exposed ?

internal/backend/muxes/muxes.go
```diff
@@ -2,4 +2,23 @@ package muxes
...
+	m.Root.HandleFunc(pattern, handler)
+}
+
+func (m *Muxes) AuthHandle(pattern string, handler http.Handler) {
+	m.Root.Handle(pattern, m.WrapAuth(handler))
+}
+
+func (m *Muxes) AuthHandleFunc(pattern string, handler func(http.ResponseWriter, *http.Request)) {
+	m.Root.Handle(pattern, m.WrapAuth(http.HandlerFunc(handler)))
+}
```
efiShtain: API mux should be behind a middleware that verifies whatever we need, no need to wrap specific functions each time
is is already wrapped in the svc.go file, no need for this wrappers here

internal/backend/muxes/muxes.go
```diff
@@ -2,4 +2,23 @@ package muxes
 
 import "net/http"
 
-type Muxes struct{ Auth, NoAuth *http.ServeMux }
+type Muxes struct {
+	API, Root *http.ServeMux
```
efiShtain: maybe we should expose an type alias like AuthMux instead of this muxes wrapper
than everywhere we can DI AuthMux instead of muxes and use muxes.API

internal/backend/applygrpcsvc/svc.go
```diff
@@ -24,11 +24,11 @@ type server struct {
 
 var _ applyv1connect.ApplyServiceHandler = (*server)(nil)
 
-func Init(mux *http.ServeMux, client sdkservices.Services) {
+func Init(muxes *muxes.Muxes, client sdkservices.Services) {
 	srv := server{client: client}
 
 	path, namer := applyv1connect.NewApplyServiceHandler(&srv)
-	mux.Handle(path, namer)
+	muxes.Handle(path, namer)
```
efiShtain: should be under muxes.API

internal/backend/auth/authgrpcsvc/svc.go
```diff
@@ -0,0 +1,57 @@
...
+	authv1connect.UnimplementedAuthServiceHandler
+}
+
+var _ authv1connect.AuthServiceHandler = (*server)(nil)
+
+func Init(muxes *muxes.Muxes, auth sdkservices.Auth) {
+	srv := server{auth: auth}
+
+	path, namer := authv1connect.NewAuthServiceHandler(&srv)
+	muxes.AuthHandle(path, namer)
```
efiShtain: muxes.API, this has to be authenticated

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -0,0 +1,105 @@
...
+	"go.uber.org/fx"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/auth/authcontext"
+	"go.autokitteh.dev/autokitteh/internal/backend/auth/authsessions"
+	"go.autokitteh.dev/autokitteh/internal/backend/auth/authtokens"
+	"go.autokitteh.dev/autokitteh/internal/backend/configset"
+)
+
+type Config struct {
+	Required bool `koanf:"required"`
```
efiShtain, line 16: not sure we need this 
if we will always have a default user, 
the noop authenticator I implemented in eng-684 would just authenticate this by default and a real authenticator would work otherwise

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+var Configs = configset.Set[Config]{
+	Default: &Config{Required: true},
+	Dev:     &Config{},
+}
+
+type WrapFunc func(http.Handler) http.Handler
```
efiShtain: The standard term is Middleware for this type of functions, 
WrapFunction does not really mean anything 
Other option id Decorator or RequestDecorator to give it a better context

internal/backend/auth/authcontext/authcontext.go
```diff
@@ -0,0 +1,27 @@
+package authcontext
+
+import (
+	"context"
+
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type ctxKey string
```
efiShtain, line 9: I know it's my code, but I think I saw it's better to have a struct type for context keys instead of string aliases

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -0,0 +1,105 @@
...
+		}
+
+		next.ServeHTTP(w, r)
+	}
+}
+
+func New(deps Deps) WrapFunc {
+	return func(next http.Handler) http.Handler {
+		f := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if deps.Cfg.Required {
```
efiShtain, line 85: no real need to check this on every call, deps.Cfg.Required is known at start time of the server
better to extract this check out of the function and return a noauth wrapper and auth wrapper based on deps.Cfg.Required

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -0,0 +1,105 @@
...
+				if user := authcontext.GetAuthnUser(r.Context()); !user.IsValid() {
+					http.Redirect(w, r, "/login", http.StatusTemporaryRedirect)
+					return
+				}
+			}
+
+			next.ServeHTTP(w, r)
+		})
+
+		if deps.Sessions != nil {
```
efiShtain, line 95: do we still need both sessions and token ?

internal/backend/auth/authjwttokens/jwt.go
```diff
@@ -0,0 +1,91 @@
...
+
+const issuer = "ak"
+
+var (
+	method   = j.SigningMethodHS256
+	hashSize = method.Hash.Size()
+)
+
+type Config struct {
+	SignKey string `koanf:"sign_key"`
```
efiShtain, line 26: this should be mandatory non optional in default mode

internal/backend/auth/authjwttokens/jwt.go
```diff
@@ -0,0 +1,91 @@
...
+	return &tokens{signKey: key}, nil
+}
+
+func (js *tokens) Create(user sdktypes.User) (string, error) {
+	uj, err := user.MarshalJSON()
+	if err != nil {
+		return "", fmt.Errorf("marshal: %w", err)
+	}
+
+	uuid, err := uuid.NewV7()
```
efiShtain, line 58: what's the purpose of a new uuid each time user creates a token ?
id should be related to the user somehow, and deterministic

internal/backend/auth/authloginhttpsvc/svc.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+
+		bs, err := u.MarshalJSON()
+		if err != nil {
+			http.Error(w, err.Error(), http.StatusInternalServerError)
+			return
+		}
+
+		w.Header().Add("Content-Type", "application/json")
+		_, _ = w.Write(bs)
```
efiShtain: why do you ignore the errors ?
at least log them

internal/backend/auth/authloginhttpsvc/svc.go
```diff
@@ -0,0 +1,118 @@
...
+		}
+	}
+
+	if a.Cfg.GithubOAuth.Enabled {
+		if err := registerGithubOAuthRoutes(muxes.Root, a.Cfg.GithubOAuth, a.newSuccessLoginHandler); err != nil {
+			return err
+		}
+	}
+
+	if a.Cfg.Descope.Enabled {
```
efiShtain, line 56: I think we should not allow Descope and github/google simulatnously

internal/backend/auth/authloginhttpsvc/google.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+
+	"golang.org/x/oauth2"
+	googleOAuth2 "golang.org/x/oauth2/google"
+)
+
+func registerGoogleOAuthRoutes(mux *http.ServeMux, cfg oauth2Config, onSuccess func(sdktypes.User) http.Handler) error {
+	if cfg.ClientID == "" {
+		return errors.New("google login is enabled, but missing GOOGLE_CLIENT_ID")
```
efiShtain, line 17: We need to make sure this env variables are not mixed the with google integration env variables (which might have the same values, but should have different names)

internal/backend/auth/authjwttokens/jwt.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	if len(key) != hashSize {
+		return nil, fmt.Errorf("invalid key len: %d != desired %d", len(key), hashSize)
+	}
+
+	return &tokens{signKey: key}, nil
+}
+
+func (js *tokens) Create(user sdktypes.User) (string, error) {
```
efiShtain, line 52: we should have a ticket for expiring tokens, refreshing tokens, blocking tokens and so on

internal/backend/auth/authsessions/config.go
```diff
@@ -0,0 +1,30 @@
+package authsessions
+
+import (
+	"github.com/dghubble/sessions"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/configset"
+)
+
+type Config struct {
+	UnsafeCookie *sessions.CookieConfig `koanf:"cookie"`
```
efiShtain: why UnsafeCookie ?

internal/backend/oauth/svc.go
```diff
@@ -24,10 +24,10 @@ type server struct {
 
 var _ oauthv1connect.OAuthServiceHandler = (*server)(nil)
 
-func Init(mux *http.ServeMux, l *zap.Logger, oauth sdkservices.OAuth) {
+func Init(muxes *muxes.Muxes, l *zap.Logger, oauth sdkservices.OAuth) {
 	srv := server{logger: l, impl: oauth}
 	path, handler := oauthv1connect.NewOAuthServiceHandler(&srv)
-	mux.Handle(path, handler)
```
efiShtain, line 30: are you sure oauth flow can be authenticated fully ?

internal/backend/svc/svc.go
```diff
@@ -212,79 +239,65 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
...
 		}),
-		fx.Invoke(func(mux *http.ServeMux, tclient temporalclient.Client) {
+		fx.Invoke(func(muxes *muxes.Muxes, tclient temporalclient.Client) {
 			_, uiAddr := tclient.TemporalAddr()
 			if uiAddr == "" {
 				return
 			}
 
-			mux.HandleFunc("/temporal", func(w http.ResponseWriter, r *http.Request) {
+			muxes.HandleFunc("/temporal", func(w http.ResponseWriter, r *http.Request) {
```
efiShtain: this should be allowed in dev mode only

internal/backend/svc/svc.go
efiShtain: did you update the cli / sdk client to use the /api prefix for all grpc endpoints ?

what about the tests ?

web/static/embed.go
daabr: `embed.go` is reasonable, because it needs to be at the relative root. What doesn't belong here is go files that control/render dynamic templates, e.g. `github/index.html`.

cmd/ak/cmd/auth/auth.go
itayd: yup, should be done in a different PR though.

itayd: https://linear.app/autokitteh/issue/ENG-778/store-auth-token-for-cli-in-ak-login-or-some-such

cmd/ak/cmd/auth/create_token.go
itayd: yup, that's why i tested it with another token :) user might have a token generated by the UI, and then they can generate more using cli.

web/static/embed.go
itayd: what i did is added some templating logic, i'd rather it stay as now. unless @daabr insists.

sdk/sdkservices/auth.go
itayd: agree, this is just a starter.
https://linear.app/autokitteh/issue/ENG-779/support-expiry-for-auth-tokens

sdk/sdktypes/user.go
itayd: "id", it's more of a unique qualifier to display. maybe i should change it to "title". I've seen everywhere email set.

sdk/sdktypes/user.go
itayd: to know what the auth source is. you did the same on your pr?

proto/autokitteh/users/v1/user.proto
itayd: when we'll store them, yes. currently it's just passed around.

internal/backend/webtools/webtools.go
itayd: this is available only in dev. but you're right. I'll change.

proto/autokitteh/events/v1/svc.proto
itayd: the buf linter decided to start screaming at this in order to make the pr go. i can get this out, but let's just do it here?

internal/backend/svc/svc.go
itayd: i kinda brainfarted on this one. now fixed.

internal/backend/muxes/muxes.go
itayd: it's just ergonomics, i'll remove.

internal/backend/muxes/muxes.go
itayd: yeah yeah fixed

internal/backend/muxes/muxes.go
itayd: makes sense, but i'd rather refactor later if we really want it.

internal/backend/applygrpcsvc/svc.go
itayd: fixed

internal/backend/auth/authgrpcsvc/svc.go
itayd: it's under `.Auth.` now.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: you still need to know if you require login or not

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: fixed

internal/backend/auth/authcontext/authcontext.go
itayd: i actually think this is ok, see https://medium.com/@matryer/context-keys-in-go-5312346a868d.
@tebeka

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: not required doesn't mean not available. you can have it non-required but still see if anything is given.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: yes, for web ui.

internal/backend/auth/authjwttokens/jwt.go
itayd: yup - it is.

internal/backend/auth/authjwttokens/jwt.go
itayd: 1. consecutive tokens generated in the same second will differ.
2. ability to cancel it in the future, per specific token.

internal/backend/auth/authloginhttpsvc/svc.go
itayd: ok

internal/backend/auth/authloginhttpsvc/svc.go
itayd: hmmm... i see no technical reason why to forbid it. it's just another mechanism you can choose if it's configured.

internal/backend/auth/authloginhttpsvc/google.go
itayd: when we move to structured config for integrations it's going to be always different because they're all going to be prefixed.

internal/backend/auth/authjwttokens/jwt.go
itayd: all under https://linear.app/autokitteh/issue/ENG-780/tokens now

internal/backend/svc/svc.go
itayd: why?

internal/backend/oauth/svc.go
itayd: always authd now

internal/backend/svc/svc.go
itayd: i'll put this behind a flag

sdk/sdktypes/user.go
efiShtain: so maybe we should panic or error if email is not set ?
I worry about case where email is not set for multiple users and they get the same <unknown> id

internal/backend/oauth/svc.go
itayd: this is fixed in the next pr (#240 ), which I'll merge together, it's the other pr's concern.

sdk/sdktypes/user.go
efiShtain: oh, that's provider like google / GitHub not descope / oauth, fine then

internal/backend/svc/svc.go
itayd: https://linear.app/autokitteh/issue/ENG-783/temporal-route-to-be-exposed-only-when-a-flag-is-set

internal/backend/auth/authhttpmiddleware/authmiddleware.go
efiShtain: this requires jwt, and it should always be available even if no login
it should be auto generated if no login
that way we always have a user

internal/backend/auth/authsessions/config.go
itayd: i think i meant here that if you specify something here, it doesn't use `sessions` default safe cookie. this is not necessary. renamed to `Cookie`.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
efiShtain: not sure I follow what you mean about not requires != not available, 
but in any case, this is a redundant check on every function call since the condition is fixed
since this is hot path (will run on every call to our API), it is worth a simple optimization (I think)

```go
func my_func() {
   if condition {

  }
}

->

func my_func_without_condition() {


}

func my_func_with_condition() {

}

return condition ? my_func_with_condition() : my_func_without_condition()
```

internal/backend/auth/authjwttokens/jwt.go
efiShtain: not sure I get it

this is the id for the token ? 
you can just use the token it self as it is unique

if it's something that should reference the user - it should be deterministic by user characteristics

internal/backend/auth/authloginhttpsvc/google.go
efiShtain: ticket ?

itayd: https://linear.app/autokitteh/issue/ENG-481/use-config-instead-of-direct-env-vars-in-integrations

sdk/sdktypes/user.go
itayd: it's for display only - we're not using it for anything else. the function name is wrong - will change it to `Title`.

efiShtain (approved):

internal/backend/auth/authloginhttpsvc/svc.go
itayd: done

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: i agree with the idea, but i tried it and it turns out pretty complicated and unwieldy. let's see if we have a real performance issue and then do it.

PR #241, opened 2024-05-05
cmd: specify mode in root

itayd:
will allow any command to work on any mode

### DIFF
--- a/cmd/ak/cmd/server/migrate.go
+++ b/cmd/ak/cmd/server/migrate.go
@@ -6,7 +6,6 @@ import (
 
 	"github.com/spf13/cobra"
 
-	"go.autokitteh.dev/autokitteh/backend/svc"
 	"go.autokitteh.dev/autokitteh/cmd/ak/common"
 )
 
@@ -17,12 +16,12 @@ var migrateCmd = common.StandardCommand(&cobra.Command{
 	Args:    cobra.NoArgs,
 
 	RunE: func(cmd *cobra.Command, args []string) error {
-		m, err := svc.ParseMode(mode)
+		mode, err := common.GetMode()
 		if err != nil {
-			return fmt.Errorf("mode: %w", err)
+			return err
 		}
 
-		db, err := InitDB(common.Config(), m)
+		db, err := InitDB(common.Config(), mode)
 		if err != nil {
 			return fmt.Errorf("init db: %w", err)
 		}
@@ -34,3 +33,7 @@ var migrateCmd = common.StandardCommand(&cobra.Command{
 		return nil
 	},
 })
+
+func init() {
+	common.AddModeFlag(migrateCmd)
+}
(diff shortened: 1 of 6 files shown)


cmd/ak/cmd/server/migrate.go
```diff
@@ -17,12 +16,7 @@ var migrateCmd = common.StandardCommand(&cobra.Command{
 	Args:    cobra.NoArgs,
 
 	RunE: func(cmd *cobra.Command, args []string) error {
-		m, err := svc.ParseMode(mode)
-		if err != nil {
-			return fmt.Errorf("mode: %w", err)
-		}
-
-		db, err := InitDB(common.Config(), m)
+		db, err := InitDB(common.Config(), common.Mode())
```
efiShtain: I added the mode to initDB since it didn't work before
now that you change so every command can work in dev mode, maybe you can remove the mode from the signature

efiShtain (approved):
better to remove the mode from initDB if it will work without it now (it didn't before, since it didn't know about dev mode)
your call
otherwise LGTM

cmd/ak/common/svc.go
```diff
@@ -0,0 +1,23 @@
+package common
+
+import (
+	"go.autokitteh.dev/autokitteh/backend/svc"
+	"go.autokitteh.dev/autokitteh/internal/backend/configset"
+)
+
+var mode configset.Mode
+
+func SetMode(m string) (err error) {
```
daabr: An exported function that merely controls a private var? Isn't this a Go smell/antipattern?

cmd/ak/common/svc.go
```diff
@@ -0,0 +1,23 @@
...
+	return
+}
+
+func Mode() configset.Mode { return mode }
+
+func NewSvc() (svc.Service, error) {
+	return svc.New(Config(), svc.RunOptions{Mode: mode})
+}
+
+func NewSilentSvc() (svc.Service, error) {
```
daabr: What does this mean?

daabr (changes requested):
1. There are a lot of test failures that show that this PR wasn't tested
2. `cmd/ak/common/svc.go` is undocumented, I have no idea what is the purpose of all the functions there
3. Didn't we agree offline not to have SDK mode? If so, what's the meaning/purpose of mode in this PR?

cmd/ak/cmd/server/migrate.go
itayd: you can't :( InitDB needs to creates a service first, so it needs the mode.

itayd:
> 1. There are a lot of test failures that show that this PR wasn't tested

should be good now.

> 2. `cmd/ak/common/svc.go` is undocumented, I have no idea what is the purpose of all the functions there

just a common place to start the svc.

> 3. Didn't we agree offline not to have SDK mode? If so, what's the meaning/purpose of mode in this PR?

it's not an SDK mode, discussed.

itayd:
i split the mode up to different cmds. it appears that the cfg commands don't need mode at all. it's ok to just use dev mode - they don't need the values.

daabr (approved):

PR #242, opened 2024-05-05
feat(cron): distributed cron via temporal

ig-ra:
- in memory cron replaced by (legacy) temporal one
- TBD

### DIFF
--- a/integrations/scheduler/save.go
+++ b/integrations/scheduler/save.go
@@ -61,15 +61,8 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	timezone := r.Form.Get("timezone")
 	memo := r.Form.Get("memo")
 
-	// Test the validity of the schedule string.
-	id, err := cronTable.AddFunc(schedule, func() {})
-	if err != nil {
-		e := "Invalid schedule error: " + err.Error()
-		u := fmt.Sprintf("%serror.html?error=%s", uiPath, url.QueryEscape(e))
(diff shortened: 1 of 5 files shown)


internal/backend/dispatcher/dispatcher.go
```diff
@@ -59,7 +59,13 @@ func (d *dispatcher) Dispatch(ctx context.Context, event sdktypes.Event, opts *s
 
 	z.Debug("event saved")
 
-	if err := d.startWorkflow(ctx, eventID, opts); err != nil {
+	if event.Type() == "cron_trigger" {
```
efiShtain, line 62: probably should be const somewhere

integrations/scheduler/save.go
```diff
@@ -61,15 +61,8 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
-		u := fmt.Sprintf("%serror.html?error=%s", uiPath, url.QueryEscape(e))
-		http.Redirect(w, r, u, http.StatusFound)
-		return
-	}
-	cronTable.Remove(id)
+	// FIXME: ENG-771 Test the validity of the schedule string.
+	// use some go library to parse and normalize cron schedule (cronexpr?) and ensure that TZ is supported as well
 
 	// Save a new connection, and return to the user an autokitteh connection token.
 	connToken, err := h.createConnection(schedule, timezone, memo)
```
efiShtain, line 68: so no way to remove the schedule ?
update it ?

PR #246, opened 2024-05-06
WIP: vars and secrets integration

efiShtain:
This is a draft PR
please let me know what do you think

- added generic secret store independent of connection
- added aws secret store
- added db secret store (default in dev mode)
- create var isSecret true - creates the secret in store
- get var isSecret true - return var from db
- reveal - return var from db + secret value from store
- delete - fetch all secrets, if isSecret also remove from secret store

added some tests, need to add more

need to check edge cases (creating multiple secret, partial success, what should we do? )

on delete we delete from from db and only then from secret store so secrets can't be referenced 
but we might have to cleanup in case we failed to delete a secret and we have dangling ones

Next steps:
manifest support for valueFrom reference a secret


efiShtain:
Closed and moved to #251

PR #249, opened 2024-05-07
make authentication redirection work

itayd:
this simplifies the previous method, also working as intended now.


daabr (approved):
I think we need a more unique (i.e. AK-specific) cookie name than "redir", but ¯\_(ツ)_/¯

PR #251, opened 2024-05-07
ENG-712: vars and secrets integration

efiShtain:
This is a draft PR
please let me know what do you think

- added generic secret store independent of connection
- added aws secret store
- added db secret store (default in dev mode)
- create var isSecret true - creates the secret in store
- get var isSecret true - return var from db
- reveal - return var from db + secret value from store
- delete - fetch all secrets, if isSecret also remove from secret store

added some tests, need to add more

need to check edge cases (creating multiple secret, partial success, what should we do? )

on delete we delete from from db and only then from secret store so secrets can't be referenced 
but we might have to cleanup in case we failed to delete a secret and we have dangling ones

### Create secret

```bash
ak var create <key> <val> -p <project_name> -e <env_name> --secret
```

### Get secret value
 ```bash
ak var get <key> -p <project_name> -e <env_name> --reveal
```

### Create from manifest
NOT SUPPORTED - We don't want to allow people setting secrets in the manifest

### CODE REVIEW

File: internal/backend/secrets/svc.go [hunk 1]
```diff
@@ -0,0 +1,84 @@
...
+	"errors"
+	"fmt"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/db"
+	"go.uber.org/zap"
+)
+
+type Secrets interface {
+	Get(ctx context.Context, key string) (map[string]string, error)
+	Set(ctx context.Context, key string, value map[string]string) error
```
itayd: k->V string:string

internal/backend/secrets/svc.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+	case SecretProviderDatabase:
+		if provider, err = newDatabaseSecrets(z, db); err != nil {
+			z.Panic("invalid provider configuration", zap.Error(err))
+			return nil, ErrSecretsInvalidProviderConfiguration
+		}
+	}
+
+	if provider == nil {
+		return nil, ErrSecretsInvalidProvider
```
itayd: sdkerrors.NewInvalidArguments

internal/backend/secrets/svc.go
```diff
@@ -0,0 +1,84 @@
...
+func New(cfg *Config, z *zap.Logger, db db.DB) (Secrets, error) {
+	var (
+		provider Secrets
+		err      error
+	)
+
+	switch cfg.Provider {
+	case SecretProviderAWSSecretManager:
+		if provider, err = newAWSSecrets(z, cfg); err != nil {
+			z.Panic("invalid provider configuration", zap.Error(err))
```
itayd: kill

internal/backend/secrets/svc.go
```diff
@@ -0,0 +1,84 @@
...
+			return nil, ErrSecretsInvalidProviderConfiguration
+		}
+	case SecretProviderDatabase:
+		if provider, err = newDatabaseSecrets(z, db); err != nil {
+			z.Panic("invalid provider configuration", zap.Error(err))
```
itayd: kill

internal/backend/secrets/svc.go
efiShtain: done

internal/backend/secrets/svc.go
efiShtain: done

internal/backend/secrets/svc.go
efiShtain: done

internal/backend/secrets/svc.go
efiShtain: done

internal/backend/vars/vars.go
```diff
@@ -2,34 +2,90 @@ package vars
...
+func varSecretKey(secret sdktypes.Var) string {
+	return fmt.Sprintf("%s/%s", secret.ScopeID().AsID().UUIDValue().String(), secret.Name())
 }
 
 func (v *vars) Set(ctx context.Context, vs ...sdktypes.Var) error {
+	for i, va := range vs {
+		if va.IsSecret() {
+			key := varSecretKey(va)
+			if err := v.secrets.Set(ctx, key, va.Value()); err != nil {
+				//TODO: handle delete secrets that were already created
```
itayd: ticket

internal/backend/vars/vars.go
```diff
@@ -2,34 +2,90 @@ package vars
...
+			}
+		}
+	}
+
+	return err
 }
 
 func (v *vars) Get(ctx context.Context, sid sdktypes.VarScopeID, names ...sdktypes.Symbol) (sdktypes.Vars, error) {
-	return v.db.GetVars(ctx, false, sid, names)
+	return v.db.GetVars(ctx, sid, names)
```
itayd, line 68: this will not return the secrets keys right? i think we should include the keys here.

internal/backend/vars/vars.go
```diff
@@ -2,34 +2,90 @@ package vars
...
 func (v *vars) Set(ctx context.Context, vs ...sdktypes.Var) error {
+	for i, va := range vs {
+		if va.IsSecret() {
+			key := varSecretKey(va)
+			if err := v.secrets.Set(ctx, key, va.Value()); err != nil {
+				//TODO: handle delete secrets that were already created
+				return err
+			}
+
+			vs[i] = sdktypes.NewVar(va.Name(), key, true).WithScopeID(va.ScopeID())
```
itayd, line 38: ahhh i see you store the secret key. is this really necessary though?

itayd (approved):
lgtm, though i got a q inside

tests/cli/vars.txt
daabr, line 1: Could you convert this to a system test instead of CLI?

internal/backend/db/db.go
```diff
@@ -150,4 +150,9 @@ type DB interface {
 	GetSignal(ctx context.Context, signalID string) (scheme.Signal, error)
 	RemoveSignal(ctx context.Context, signalID string) error
 	ListSignalsWaitingOnConnection(ctx context.Context, connectionID sdktypes.ConnectionID) ([]scheme.Signal, error)
+
+	// -----------------------------------------------------------------------
+	SetSecret(ctx context.Context, key string, value string) error
+	GetSecret(ctx context.Context, key string) (string, error)
```
daabr, line 156: Question: do we also want to add an atomic update and/or upsert in a separate PR?

internal/backend/secrets/svc.go
```diff
@@ -0,0 +1,79 @@
...
+
+type Secrets interface {
+	Get(ctx context.Context, key string) (string, error)
+	Set(ctx context.Context, key string, value string) error
+	Delete(ctx context.Context, key string) error
+}
+
+const (
+	SecretProviderDatabase         = "db"
+	SecretProviderAWSSecretManager = "awsSecretManager"
```
daabr, line 20: Just my personal opinion, but I think "aws" is enough for configs, instead of "awsSecretManager".

internal/backend/secrets/svc.go
```diff
@@ -0,0 +1,79 @@
...
+	SecretProviderVault            = "vault"
```
daabr, line 21: I don't see the Vault implementation in this PR. Please add it, I'm relying on it.

daabr (approved):
Approved - but please see my comments. I don't mind if you implement them in a separate PR. Specifically not losing Vault support is very important for me.

internal/backend/vars/vars.go
efiShtain: it makes it simpler to list secrets and vars without calling secrets stores

internal/backend/secrets/svc.go
efiShtain: they also support system parameter store for secrets, otherwise I agree with you
this makes sure we know what we use

internal/backend/secrets/svc.go
efiShtain: https://linear.app/autokitteh/issue/ENG-815/re-add-vault-as-a-secret-store

will do in another pr

internal/backend/db/db.go
efiShtain: I change set in aws implementation to overwrite existing secrets
other than that it should be atomic as far as I see

tests/cli/vars.txt
efiShtain: https://linear.app/autokitteh/issue/ENG-816/convert-secrets-test-from-cli-test-to-system-tests

opened a ticket for that, since I'm not sure how to do it and I prefer to do it as another step

internal/backend/vars/vars.go
efiShtain: https://linear.app/autokitteh/issue/ENG-817/handle-dangling-secrets

this is a new kind of task, we should have back office maintenance jobs to clean up dangling secrets

PR #252, opened 2024-05-07
fix(db): shema, UUID and constrains

ig-ra:
- fixes uuid vs. text for events
- defines NON_NULL constrains on relevant fields
- fixes tests


internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -49,7 +49,7 @@ var Tables = []any{
 }
 
 type Build struct {
-	BuildID   sdktypes.UUID `gorm:"primaryKey;type:uuid"`
+	BuildID   sdktypes.UUID `gorm:"primaryKey;type:uuid;not null"`
```
efiShtain, line 52: primaryKey is by definition not null, 
you don't need to add it

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -189,9 +189,9 @@ type Secret struct {
 }
 
 type Event struct {
```
efiShtain, line 191: something is not right here
where is the primary key of the table ?
there was a composed primary key of 3 fields if I recall correctly

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: see this
https://github.com/autokitteh/autokitteh/blob/1af8cad26782595c9ba41d18b3f9e7a019980ddf/migrations/sqlite/20240506105938_baseline.sql#L33

efiShtain (approved):

PR #253, opened 2024-05-07
[ENG-761] Log exceptions when running user code

tebeka:
Once logging is merged, top level errors will show in autokitteh.


runtimes/pythonrt/ak_runner.py
```diff
@@ -236,6 +238,16 @@ def send_response(self, value):
     def extract_response(self, message):
         data = message['payload']['value']
         return pickle.loads(b64decode(data))
+    
+    def send_error(self, error, traceback):
+        message = {
```
ig-ra, line 258: unrelated to this PR. could we replace message (and payload) by serializable class or similar?

ig-ra (approved):

runtimes/pythonrt/ak_runner.py
tebeka: The ideal will be to have a schema for the messages and then generate code both for Go & Python (like protobuf).
But, I don't want to add any external dependency in Python (e.g. protobuf).
Opened ENG-821

ig-ra: we could go for less ideal solution, e.g. dataclass or similar

tebeka:
Closing since #230 will send the error to the session log.

PR #257, opened 2024-05-08
ENG-790: login via cli

itayd:
cli opens a browser at the server with a local port to respond to. server logs in, redirects back with token. token is persisted. everyone is happy.

also added the ability to store a token manually.

### DIFF
--- a/cmd/ak/cmd/auth/login.go
+++ b/cmd/ak/cmd/auth/login.go
@@ -1,0 +1,85 @@
+package auth
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"net/http"
+
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+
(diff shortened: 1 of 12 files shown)


cmd/ak/cmd/auth/login.go
```diff
@@ -0,0 +1,85 @@
...
+		}
+
+		url := fmt.Sprintf("%s/auth/cli-login?p=%d", common.ServerURL().String(), port)
+		if err := browser.OpenURL(url); err != nil {
+			fmt.Fprintf(cmd.ErrOrStderr(), "To start, open this address in your browser: %s\n", url)
+		} else {
+			fmt.Fprintf(cmd.ErrOrStderr(), "Please continue in browser (%s).\n", url)
+		}
+
+		token, err := wait(cmd.Context())
```
efiShtain, line 36: we should probably limit this for a minute or two for security reasons

efiShtain (approved):

PR #262, opened 2024-05-08
ENG-530: add health check

efiShtain:
added db and temporal health check to /healthz endpoint

Intended to support ecs / k8s health probes

### DIFF
--- a/internal/backend/db/dbgorm/health.go
+++ b/internal/backend/db/dbgorm/health.go
@@ -1,7 +1,6 @@
 package dbgorm
 
-func (db *gormdb) Report() []error {
+func (db *gormdb) Report() error {
 	s, _ := db.db.DB()
-	err := s.Ping()
-	return []error{err}
+	return s.Ping()
 }
--- a/internal/backend/db/dbgorm/health.go
+++ b/internal/backend/db/dbgorm/health.go
@@ -1,0 +1,7 @@
+package dbgorm
+
+func (db *gormdb) Report() []error {
+	s, _ := db.db.DB()
+	err := s.Ping()
+	return []error{err}
+}
--- a/internal/backend/health/healthchecker/svc.go
+++ b/internal/backend/health/healthchecker/svc.go
@@ -1,6 +1,8 @@
 package healthchecker
 
 import (
+	"errors"
+
 	"go.autokitteh.dev/autokitteh/internal/backend/db"
 	"go.autokitteh.dev/autokitteh/internal/backend/health/healthreporter"
 	"go.autokitteh.dev/autokitteh/internal/backend/temporalclient"
@@ -22,14 +24,14 @@ func New(db db.DB, z *zap.Logger, tc temporalclient.Client) healthreporter.Healt
 	return checker
 }
 
-func (h *healthChecker) Report() []error {
+func (h *healthChecker) Report() error {
 	allHealthErrors := []error{}
 	for name, reporter := range h.checks {
-		if reporterErrors := reporter.Report(); len(reporterErrors) > 0 {
-			h.z.Error("health check error", zap.Errors(name, reporterErrors))
-			allHealthErrors = append(allHealthErrors, reporterErrors...)
+		if err := reporter.Report(); err != nil {
+			h.z.Error("health check error", zap.String("service", name), zap.Error(err))
+			allHealthErrors = append(allHealthErrors, err)
 		}
 	}
 
-	return allHealthErrors
+	return errors.Join(allHealthErrors...)
 }
--- a/internal/backend/health/healthchecker/svc.go
+++ b/internal/backend/health/healthchecker/svc.go
@@ -1,0 +1,35 @@
+package healthchecker
+
+import (
+	"go.autokitteh.dev/autokitteh/internal/backend/db"
+	"go.autokitteh.dev/autokitteh/internal/backend/health/healthreporter"
+	"go.autokitteh.dev/autokitteh/internal/backend/temporalclient"
+	"go.uber.org/zap"
+)
+
+type healthChecker struct {
+	db     db.DB
+	z      *zap.Logger
+	tc     temporalclient.Client
+	checks map[string]healthreporter.HealthReporter
+}
+
+func New(db db.DB, z *zap.Logger, tc temporalclient.Client) healthreporter.HealthReporter {
+	checker := &healthChecker{db: db, z: z, tc: tc, checks: map[string]healthreporter.HealthReporter{}}
+
+	checker.checks["db"] = db
+	checker.checks["temporal"] = tc
+	return checker
+}
+
+func (h *healthChecker) Report() []error {
+	allHealthErrors := []error{}
+	for name, reporter := range h.checks {
+		if reporterErrors := reporter.Report(); len(reporterErrors) > 0 {
+			h.z.Error("health check error", zap.Errors(name, reporterErrors))
+			allHealthErrors = append(allHealthErrors, reporterErrors...)
+		}
+	}
+
+	return allHealthErrors
+}
--- a/internal/backend/health/healthreporter/reporter.go
+++ b/internal/backend/health/healthreporter/reporter.go
@@ -1,0 +1,5 @@
+package healthreporter
+
+type HealthReporter interface {
+	Report() []error
+}
--- a/internal/backend/health/healthreporter/reporter.go
+++ b/internal/backend/health/healthreporter/reporter.go
@@ -1,5 +1,5 @@
 package healthreporter
 
 type HealthReporter interface {
-	Report() []error
+	Report() error
 }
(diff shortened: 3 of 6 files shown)


internal/backend/health/healthreporter/reporter.go
```diff
@@ -0,0 +1,5 @@
+package healthreporter
+
+type HealthReporter interface {
```
itayd, line 3: I think a better way might be that instead an interface it will return a function with a specific type. Then the uber.Fx thingy can gather all of them and deliver them to the healthz endpoint, so it can iterate all of them.

see https://uber-go.github.io/fx/value-groups/

internal/backend/svc/svc.go
```diff
@@ -160,6 +162,8 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 		Component("triggers", configset.Empty, fx.Provide(triggers.New)),
 		Component("oauth", configset.Empty, fx.Provide(oauth.New)),
 		Component("runtimes", configset.Empty, fx.Provide(runtimes.New)),
+
```
daabr, line 181: Nit: any reason for the empty line?

internal/backend/svc/svc.go
```diff
@@ -285,11 +289,14 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 				http.Redirect(w, r, uiAddr, http.StatusFound)
 			})
 		}),
-		fx.Invoke(func(z *zap.Logger, lc fx.Lifecycle, muxes *muxes.Muxes) {
+		fx.Invoke(func(z *zap.Logger, lc fx.Lifecycle, muxes *muxes.Muxes, h healthreporter.HealthReporter) {
 			var ready atomic.Bool
 
 			muxes.NoAuth.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
-				// TODO(ENG-530): check db, temporal, etc.
+				if healthErrors := h.Report(); len(healthErrors) != 0 {
```
daabr: 1. What about `readyz`?
2. Can you document the intention behind this behavior?
   1. I'm not used to health pages ever returning HTTP 500, but rather they report many states and health-related metrics
   2. **Maybe** this behavior is appropriate for `/readyz`, I'm not sure, but even if so... Given that theses pages are (currently) used by k8s to determine whether they should restart/replace an AK pod, is there any nuance in the meaning of failures? Or does any failure mean we need to kill AK immediately?

daabr (commented):
Oops, I thought I commented on this during my flight, but it looks like I didn't... Sorry for the delay.

My comment/question is more general. Please answer in Linear, or by adding this info to the PR description, or in the code itself - wherever you think is more appropriate.

1. How does this PR/issue relate to, or affect, existing Temporal health checks? (https://github.com/search?q=repo%3Aautokitteh%2Fautokitteh+health+path%3Atemporal&type=code) - I see changes in `client.go` but I don't understand the impact on the existing check

2. How does this PR/issue relate to, or affect, the (currently empty) `/healthz` and `/readyz` handlers? (https://github.com/search?q=repo%3Aautokitteh%2Fautokitteh+healthz+readyz&type=code) - I see the addition in `svc.go` but I don't understand the overall idea

3. The new files under `/health` need documentation for developers to be able to understand, extend, and maintain this code

In other words, what is the **strategy** for health checks in or beyond this PR. Beyond the implementation details: which checks do we expect to have exactly? How sensitive/critical are they? Do they impact each other? How do we expect them to be used?

I don't expect all this to be implemented in this specific PR, but I do expect to document the direction.

internal/backend/svc/svc.go
efiShtain: this is not a health page, it's health endpoint that suppose to return 200 or 500 so deployments manager can decide if it should kill the instance or not

you are talking about status page that shows to users that status of every component in the system

k8s has 2 notions, health probe and readiness probe

**health** is to check if service is healthy, it is up to the service to decide what is healthy
it also has some variables like 
- initialDealy (start checking health X seconds after service startup)
- Number of consecutive success / failures to change the service state
- Period time

so it will check for example every 5 seconds, and after 3 errors it will mark the service as failed and restart it


**ready** is to check if service can serve traffic, assume it is healthy but loaded, or want to throttle requests for some reason, then it is healthy, but not ready to serve traffic

most of the times they use same implementation behind the scenes but they don't have to

internal/backend/svc/svc.go
efiShtain: nope, it just doesn't matter

efiShtain:
@daabr 
I've updated the linear ticket with more details

https://linear.app/autokitteh/issue/ENG-530/improve-health-route

internal/backend/db/dbgorm/health.go
```diff
@@ -0,0 +1,7 @@
+package dbgorm
+
+func (db *gormdb) Report() []error {
+	s, _ := db.db.DB()
+	err := s.Ping()
```
tebeka: Ping is not enough since it doesn't check permissions. Add a `SELECT 1` query.

internal/backend/health/healthchecker/svc.go
```diff
@@ -0,0 +1,35 @@
...
+
+func New(db db.DB, z *zap.Logger, tc temporalclient.Client) healthreporter.HealthReporter {
+	checker := &healthChecker{db: db, z: z, tc: tc, checks: map[string]healthreporter.HealthReporter{}}
+
+	checker.checks["db"] = db
+	checker.checks["temporal"] = tc
+	return checker
+}
+
+func (h *healthChecker) Report() []error {
```
tebeka: Return a single error, use https://pkg.go.dev/errors#Join to join several.

internal/backend/health/healthreporter/reporter.go
```diff
@@ -0,0 +1,5 @@
...
+	Report() []error
```
tebeka: Just a single error, use `errors.Join` if there are several.

internal/backend/svc/svc.go
daabr: Good overview!

I did mean health and not status pages: Google's internal `/healthz` **always** serve (relevant metrics) as long as the process is alive, and the control plane (Borg) decides what to do with overloaded or misbehaving instances. Otherwise, how can you know **what** is wrong?

The only thing I'd clarify in your overview is that, for AK, readiness failures mean replacing the pod as soon as the failure is noticed by the K8s control plane.

So taking this comment back to this PR, does that mean that this PR doesn't make any claims about readiness? That's the part I'm confused about. We both see "health" as an advanced and complex aspect of "readiness", and I interpret that as "if not healthy then obviously also not ready, i.e. replace it", but it seems like this PR says instead "bad = not healthy or not ready, which are unrelated to each other".

daabr: NM, I re-read your additions in Linear, and we had inverted terminology:

> This is not /ready implementation which mainly used to decide if service is ready to serve traffic (service could be healthy but under load in which case health would report ok but ready would report error)

I don't know for certain which one is correct, but either way it's not a blocker for this PR.

daabr (approved):
Approved - but please ack/resolve Itay's and Miki's comments.

internal/backend/db/dbgorm/health.go
efiShtain: select 1 doesn't check permissions as well, it only check I can select from a specific schema and specific table

internal/backend/health/healthreporter/reporter.go
efiShtain: this will require that we add additional code to each service we want to expose (fx.annotate to the health group)
I don't see how it makes it better / simpler

```   fx.Annotate(
      ServiceWithHealth,
      fx.As(new(HealthReporter)),
      fx.ResultTags(`group:"health_reporters"`),
    ),
```


if we plan to have loads of health reports, then I agree we should consider it
but for this use case, where we would probably have 2-3 (db/temporal/redis ? ) I think the explicit version is simpler and it also makes sure we know which health check are actually tested

WDYT?

internal/backend/svc/svc.go
efiShtain: this is an example from k8s docs (https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/)
although they return the error in the response as a string, the main check if the http status code and the text is ignored by the deployment controller

```go
http.HandleFunc("/healthz", func(w http.ResponseWriter, r *http.Request) {
    duration := time.Now().Sub(started)
    if duration.Seconds() > 10 {
        w.WriteHeader(500)
        w.Write([]byte(fmt.Sprintf("error: %v", duration.Seconds())))
    } else {
        w.WriteHeader(200)
        w.Write([]byte("ok"))
    }
})
```

you can see the definition for liveness (health) / readiness they have match what I know as standard

internal/backend/health/healthchecker/svc.go
efiShtain: fixed

internal/backend/health/healthreporter/reporter.go
efiShtain: fixed

internal/backend/health/healthreporter/reporter.go
itayd: meh ok for now, but if we add more we'll def need to add it

itayd (approved):

PR #263, opened 2024-05-08
fix descope login

itayd:
looks like the mgmt api is not working all the time. instead we're using now the `auth.me` method. for some reason it mandates refresh token. looks like the web component doesn't propagates the DSR to the server, so we do it here manually.

also kill the redir cookie once it's used.

### DIFF
--- a/internal/backend/auth/authloginhttpsvc/redir.go
+++ b/internal/backend/auth/authloginhttpsvc/redir.go
@@ -26,3 +26,12 @@ func getRedirect(r *http.Request) string {
 
 	return "/"
 }
+
+func killRedirect(w http.ResponseWriter) {
+	http.SetCookie(w, &http.Cookie{
+		Name:   redirCookieName,
+		Value:  "",
+		Path:   "/",
+		MaxAge: -1,
+	})
+}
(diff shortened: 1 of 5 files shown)


internal/backend/auth/authloginhttpsvc/web/descope.html
```diff
@@ -31,6 +31,7 @@ <h1>Descope Login</h1>
         const wcElement = document.getElementsByTagName("descope-wc")[0];
 
         const onSuccess = (e) => {
+            document.cookie = `DSR=${e.detail.refreshJwt}; path=/`;
```
efiShtain, line 34: not sure why you set the cookie manually, but you should at least set a secure cookie, same site only and such, better to let the server set the cookie

internal/backend/auth/authloginhttpsvc/redir.go
```diff
@@ -26,3 +26,12 @@ func getRedirect(r *http.Request) string {
 
 	return "/"
 }
+
+func killRedirect(w http.ResponseWriter) {
```
efiShtain, line 30: what does killRedirect mean ?

internal/backend/auth/authloginhttpsvc/web/descope.html
efiShtain, line 34: maybe just perform a post request to the backend to exchange the token (instead of storing it in a non secure cookie)

post would set the cookie to our jwt, using secure same site attributes
and then you can redirect again

efiShtain:
This change didn't work on my end, I never got the refresh token, it always failed

but it is also not required, 
you can add the email and user id to the session jwt and take it from there

please consider this PR instead 
https://github.com/autokitteh/autokitteh/pull/272

PR #264, opened 2024-05-09
fix redis integration

itayd:
refactored integration creation to be simpler, now detecting correctly if module is internal or external.


examples/testprograms/redis/main.star
```diff
@@ -0,0 +1,7 @@
+load("@redis", "myredis")
```
daabr, line 1: System tests are better than examples, because they're executed automatically. I'll add/move this in a separate PR.

daabr (approved):

PR #267, opened 2024-05-09
auth: restrict domains to allowlist if supplied

itayd:


efiShtain:
This is redundant, we can do it at the load balancer level and never reach the service

itayd:
there are also github and google - we'd like to limit there as well

itayd:
@efiShtain i'm going top push this in now since i need this deployed and I'm not sure i know the steps for descope dance needed. worst case I'll revert if we think it shouldn't be in.

PR #271, opened 2024-05-09
service config tweaks

itayd:
- don't show banner in default mode
- when silent, don't show non-prod warnings
- in default, don't supply temporal redir link

### DIFF
--- a/internal/backend/svc/svc.go
+++ b/internal/backend/svc/svc.go
@@ -140,6 +140,20 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 				HookOnStart(lc, c.Start)
 				HookOnStop(lc, c.Stop)
 			}),
+			fx.Invoke(func(cfg *temporalclient.Config, tclient temporalclient.Client, muxes *muxes.Muxes) {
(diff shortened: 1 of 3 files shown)


internal/backend/svc/svc.go
```diff
@@ -140,6 +140,20 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 				HookOnStart(lc, c.Start)
 				HookOnStop(lc, c.Stop)
 			}),
+			fx.Invoke(func(cfg *temporalclient.Config, tclient temporalclient.Client, muxes *muxes.Muxes) {
+				if !cfg.EnableHelperRedirect {
+					return
+				}
+
+				_, uiAddr := tclient.TemporalAddr()
```
ig-ra, line 148: just to ensure - what are the cases when uiAddr is empty? when temporal started by us, but without ui?

ig-ra (approved):

internal/backend/svc/svc.go
itayd: yup, banner knows how to deal with it

PR #272, opened 2024-05-09
simplified version instead of using management api

efiShtain:
this assume custom claims are setup in descope flow (a simple one time addition)

this is an alternative to #263, only one of them should be merged


internal/backend/auth/authloginhttpsvc/descope.go
```diff
@@ -42,29 +44,30 @@ func registerDescopeRoutes(mux *http.ServeMux, cfg descopeConfig, onSuccess func
 		}
 
 		var claims j.RegisteredClaims
+
 		if _, _, err = j.NewParser().ParseUnverified(tok.JWT, &claims); err != nil {
 			http.Error(w, err.Error(), http.StatusInternalServerError)
 			return
 		}
 
+		name := tok.CustomClaim("name").(string)
```
itayd: this is odd - i didn't see it in the token. will check again.

itayd:
simplified a bit further - no need for much of descope.html and the fetch

itayd (approved):

PR #282, opened 2024-05-12
[ENG-544] Make temporal dev server shut up

tebeka:
Currently using a replace directive to get the temporal sdk with out patch commited. Once they cut a new version, we'll switch (ENG-820)


internal/backend/temporalclient/client_test.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	opts := client.Options{}
+
+	c := &impl{
+		done: make(chan struct{}),
+		z:    logger,
+	}
+	err := c.startDevServer(ctx, cfg, opts)
+	require.NoError(t, err)
+	defer c.Stop(ctx)
```
ig-ra: fix linter

internal/backend/temporalclient/client.go
```diff
@@ -75,10 +79,19 @@ func New(cfg *Config, z *zap.Logger) (Client, error) {
 }
 
 func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Options) error {
-	cfg.DevServer.ClientOptions = &opts
+	temporalLog := path.Join(xdg.DataHomeDir(), "temporal_dev.log")
```
ig-ra, line 82: not sure, but maybe we should make it configurable?

internal/backend/temporalclient/client_test.go
```diff
@@ -0,0 +1,47 @@
...
+	require.NotNil(t, c.srvLog)
+
+	time.Sleep(100 * time.Millisecond) // Let the server emit some logs
+
+	file, err := os.Open(c.srvLog.Name())
+	require.NoError(t, err)
+	defer file.Close()
+	data, err := io.ReadAll(file)
+	require.NoError(t, err)
+	require.NotEmpty(t, data)
```
ig-ra: temporal log file could exists before the test and then this test will pass anyway.
I suggest either to make temporal log file name configurable, then you could ensure its uniqueness before the test
or search for specific message in data with time stamp or some random identifier that you could recognize

internal/backend/temporalclient/client_test.go
tebeka: doh! Done

internal/backend/temporalclient/client_test.go
tebeka: Added the following at the start of the test to get a unique directory:

```
	dataDir := t.TempDir()
	t.Setenv("XDG_DATA_HOME", dataDir)
```

internal/backend/temporalclient/client.go
tebeka: I general I prefer to keep configuration minimal. Every option you add is something you need to carry on and to test. If you have only 10 true/false option you need to test 1024 combinations. Unless there's a user requirement I prefer to keep it this way.

ig-ra (approved):

PR #285, opened 2024-05-12
ENG-815: re add vault secret store

efiShtain:


internal/backend/secrets/vault.go
```diff
@@ -0,0 +1,51 @@
...
+	"go.uber.org/zap"
+)
+
+type vaultSecrets struct {
+	client *vault.KVv2
+	logger *zap.Logger
+}
+
+// NewVaultSecrets initializes a client connection to HashiCorp Vault.
+func newVaultSecrets(l *zap.Logger, c *vaultConfig) (Secrets, error) {
```
tebeka: The rule of thumb is: `Accept interfaces, return types.`. This should be:

`func newVaultSecrets(l *zap.Logger, c *vaultConfig) (valutSecrets, error) {`

tebeka (approved):
Small comment, can commit without fixing it.

daabr (approved):

internal/backend/secrets/vault.go
efiShtain: that's a good rule of thumb, I fixed the internal objects I use (db/aws/vault stores) but not the external object other services requires

I think we have many places that returns interfaces instead of concrete types
but currently our fx configuration requires additional code if we'll return concrete types and not interfaces for the DI to work

PR #286, opened 2024-05-12
[ENG-828] Send Python logs to ak server logs

tebeka:


tebeka:
Bad merge, will do in another.

daabr (commented):
Doesn't it make more sense to send Python logs to the session log instead of the server log?

tebeka:
> Doesn't it make more sense to send Python logs to the session log instead of the server log?

These are the logs from `ak_runner.py`, not the user logs.

runtimes/pythonrt/pythonrt.go
```diff
@@ -280,22 +311,34 @@ func (py *pySvc) Run(
...
+		}
 
-	msg, err := py.comm.Recv()
-	if err != nil {
-		py.log.Error("initial message from python", zap.Error(err))
-		return nil, err
-	}
+		if msg.Type == "" {
+			py.log.Error("python can't load module", zap.String("module", mainPath))
+			return nil, fmt.Errorf("python can't load %q", mainPath)
```
efiShtain, line 324: why do you log and turn the exact same error ?

efiShtain (approved):

runtimes/pythonrt/pythonrt.go
tebeka: Not sure how `ak` logs it, helps me with debugging.

PR #288, opened 2024-05-12
distributed cron via temporal

ig-ra:


ig-ra:
all framework changes (proto/scheme/sdktypes) are already merged and this PR was rebased on current main.

cmd/ak/cmd/triggers/create.go
```diff
@@ -13,10 +13,11 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-var event, filter, loc, name string
+var event, filter, loc, name, schedule string
 
 var createCmd = common.StandardCommand(&cobra.Command{
-	Use:     "create <--name=...> <--env=...> <--connection=...> <--event=...> <--loc=...>",
+	Use: `create <--name=...> <--env=...> <--connection=...> <--event=...> <--loc=...>
+                 <--name=...> <--env=...> <--schedule=...> <--loc=...>`,
```
daabr: `Use` should be a single line. Also, I think we should remove `--env` but that's beyond the scope of this PR.

```suggestion
	Use: "create --name <...> --env <...> {--connection <...> --event <...> | --schedule <...>} --loc <...>",
```

manifest.schema.yaml
```diff
@@ -98,6 +98,9 @@
         "filter": {
           "type": "string"
         },
+        "schedule": {
```
daabr, line 101: Notice that connection is still required, according to the `required` below. Maybe use `oneOf` like above?

proto/autokitteh/triggers/v1/trigger.proto
```diff
@@ -9,7 +9,7 @@ import "buf/validate/validate.proto";
 message Trigger {
   string trigger_id = 1;
   string name = 2;
-  string connection_id = 3;
+  string connection_id = 3; // could be empty
```
daabr: Just making sure: you are going to add Buf validation, right? This comment isn't really visible/effective for users.

internal/backend/db/dbgorm/triggers.go
```diff
@@ -14,24 +14,33 @@ import (
...
 func triggerToRecord(ctx context.Context, tx *tx, trigger sdktypes.Trigger) (*scheme.Trigger, error) {
 	connID := trigger.ConnectionID()
+	var conn sdktypes.Connection
+	var err error
+	var projID sdktypes.ProjectID
 
-	conn, err := tx.GetConnection(ctx, connID)
-	if err != nil {
-		return nil, fmt.Errorf("get trigger connection: %w", err)
+	if connID.IsValid() {
```
daabr: Shouldn't there be any mention of schedule as being required, if `!connID.IsValid()`?

internal/backend/triggers/triggers.go
```diff
@@ -53,3 +71,81 @@ func (m *triggers) Get(ctx context.Context, triggerID sdktypes.TriggerID) (sdkty
...
+	return trigger.ID(), nil
+}
+
+func (m *triggers) deleteScheduledTrigger(ctx context.Context, triggerID sdktypes.TriggerID, data map[string]sdktypes.Value) error {
+	scheduleIDVal, found := data["schedule_id"]
+	scheduleID, err := scheduleIDVal.ToString()
+
+	if !found || err != nil {
+		m.z.Error("Failed delete scheduler workflow. No schedulerID found")
+		err = fmt.Errorf("delete scheduler workflow: scheduleID not found")
```
daabr: Lint error

daabr (commented):
I didn't finish the review, but here are a few comments in the meantime

proto/autokitteh/triggers/v1/trigger.proto
ig-ra: adding buffer validation will require either to define schedule in trigger or define data as a formal struct. Therefore there was no buffer validation added. Instead it was added in `sdktypes.Trigger`

internal/backend/db/dbgorm/triggers.go
ig-ra: the only validity check on db layer is foreign keys used. Other constrains are checked elsewhere, either in proto or in sdk layer. For trigger plz see `sdktypes.trigger`

cmd/ak/cmd/triggers/create.go
ig-ra: I could convert Use to single line, but it's less readable IMO. Why not multiline?

proto/autokitteh/triggers/v1/trigger.proto
ig-ra: plz note that all proto/scheme/sdktypes changes were already merged to main

sdk/sdktypes/trigger.go
```diff
@@ -19,10 +19,11 @@ type TriggerPB = triggerv1.Trigger
 type TriggerTraits struct{}
 
 func (TriggerTraits) Validate(m *TriggerPB) error {
+	// FIXME: where this check belongs? Validate or StrictValidate?
```
itayd: `StrictValidate`. In `Validate` it just check each field individually. In `Strict` it checks the combination.

sdk/sdktypes/consts.go
```diff
@@ -0,0 +1,6 @@
+package sdktypes
+
+const (
+	SchedulerEventTriggerType = "scheduler"
```
itayd, line 4: why not put this in `trigger.go`?

internal/backend/triggers/triggers.go
```diff
@@ -53,3 +97,98 @@ func (m *triggers) Get(ctx context.Context, triggerID sdktypes.TriggerID) (sdkty
 func (m *triggers) List(ctx context.Context, filter sdkservices.ListTriggersFilter) ([]sdktypes.Trigger, error) {
 	return m.db.ListTriggers(ctx, filter)
 }
+
+func (m *triggers) createScheduledWorkflow(ctx context.Context, trigger sdktypes.Trigger, schedule sdktypes.Value) (sdktypes.Trigger, error) {
```
itayd: I think this code should belong to another module that has access to triggers, not be in triggers themselves.

sdk/sdktypes/trigger.go
ig-ra: hmm.. just trying to clarify for myself what you wrote - according to the code both `Validate` and `strictValidate` check each field individually and return combination of errors, no? The only difference that StrictValidate won't allow empty fields

I actually think that this check should be in `Validate`. The natural place of this check is in `proto/buf` code. But we cannot check it unless we defined either schedule as a field or `data` as a structure. So, by replacing `buf/proto` check by `sdktype` check we are trying to achieve the same coverage, therefore IMO the logical place is in `Validate` function to mimic the unexising check from `buf/proto`, no?

cmd/ak/cmd/triggers/create.go
daabr: See Cobra's documentation: "Use is the one-line usage message..."

https://pkg.go.dev/github.com/spf13/cobra#Command.Use

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -159,6 +159,52 @@ func (d *dispatcher) getEventSessionData(ctx context.Context, event sdktypes.Eve
...
+		return nil, nil
+	}
+
+	deployments, err := d.services.Deployments.List(ctx,
+		sdkservices.ListDeploymentsFilter{State: sdktypes.DeploymentStateActive, EnvID: envID})
+	if err != nil {
+		z.Panic("could not fetch active deployments", zap.Error(err))
+	}
+
+	// FIXME: do we need test deployments?
```
daabr: `FIXME/TODO` without ticket number

internal/backend/triggers/triggers.go
```diff
@@ -3,45 +3,89 @@ package triggers
...
+		trigger, err = m.createScheduledWorkflow(ctx, trigger, schedule)
+	} else if isSchedulerPrevTrigger && !isSchedulerTrigger { // scheduler trigger -> trigger
+		err = m.deleteSchedulerWorkflow(ctx, prevData)
+	} else if isSchedulerPrevTrigger && isSchedulerTrigger { // scheduler trigger -> scheduler trigger
+		if schedule.String() != prevSchedule.String() { // schedule changed
+			trigger, err = m.updateSchedulerWorkflow(ctx, trigger, prevData, schedule.String())
+		}
+	}
+
+	// REVIEW: should we update trigger if scheduler workflow op failed?
```
daabr: Is this a task before merging, or does this need a ticket?

internal/backend/triggers/triggers.go
```diff
@@ -53,3 +97,98 @@ func (m *triggers) Get(ctx context.Context, triggerID sdktypes.TriggerID) (sdkty
...
+	if scheduleID == nil {
+		m.z.Error("Failed delete scheduler workflow. No scheduleID found")
+		return fmt.Errorf("delete scheduler workflow: no scheduleID")
+	}
+
+	scheduleHandle := m.tmprl.ScheduleClient().GetHandle(ctx, *scheduleID) // validity of scheduleID is not checked by temporal
+	if err := scheduleHandle.Delete(ctx); err != nil {
+		return fmt.Errorf("delete scheduler workflow: %w", err)
+	}
+	// REVIEW: create and save cancel schedule event?
```
daabr: Is this a task before merging, or does this need a ticket?

manifest.schema.yaml
ig-ra: discussed. see ENG-862.

sdk/sdktypes/consts.go
ig-ra: because it's related to both event, trigger and dispatcher. So I created a new file with consts

internal/backend/triggers/triggers.go
ig-ra: temporal client module?

internal/backend/dispatcher/eventsworkflow.go
ig-ra: @itayd  - I left this FIXME to ask you. Is it ok not to support test deployments?

internal/backend/triggers/triggers.go
ig-ra: it's an open question. What's the behavior we want?

internal/backend/triggers/triggers.go
ig-ra: one more open question. @itayd @daabr WDYT?

internal/backend/triggers/triggers.go
ig-ra: rewritten

internal/backend/triggers/triggers.go
ig-ra: moved

sdk/sdktypes/trigger.go
```diff
@@ -18,6 +19,12 @@ type TriggerPB = triggerv1.Trigger
 type TriggerTraits struct{}
 
 func (TriggerTraits) Validate(m *TriggerPB) error {
+	_, scheduleInData := m.Data["schedule"]
+	var scheduleOrConnectionErr error
+	if (scheduleInData && m.ConnectionId != "") || (!scheduleInData && m.ConnectionId == "") {
+		scheduleOrConnectionErr = fmt.Errorf("trigger should contain either valid schedule or connection")
```
itayd:
```suggestion
		scheduleOrConnectionErr = errors.New("trigger should contain either valid schedule or connection")
```

sdk/sdktypes/trigger.go
```diff
@@ -88,11 +96,20 @@ func (p Trigger) ToValues() map[string]Value {
...
 				forceFromProto[Value],
 			),
 		)),
 	}
 }
+
+func (p Trigger) WithUpdatedData(key string, val Value) Trigger {
+	if !val.IsValid() {
+		return p
+	}
```
itayd: If val is InvalidValue, just remove it from the map in data.

sdk/sdktypes/trigger.go
```diff
@@ -18,6 +19,12 @@ type TriggerPB = triggerv1.Trigger
 type TriggerTraits struct{}
 
 func (TriggerTraits) Validate(m *TriggerPB) error {
+	_, scheduleInData := m.Data["schedule"]
+	var scheduleOrConnectionErr error
+	if (scheduleInData && m.ConnectionId != "") || (!scheduleInData && m.ConnectionId == "") {
```
itayd:
```suggestion
	if !scheduleInData && m.ConnectionId == "" {
```

ig-ra: done

sdk/sdktypes/trigger.go
ig-ra: done

sdk/sdktypes/connection.go
```diff
@@ -28,10 +28,18 @@ func (ConnectionTraits) Validate(m *ConnectionPB) error {
 }
 
 func (ConnectionTraits) StrictValidate(m *ConnectionPB) error {
+	// FIXME: I think we should remove mandatory for project_id and integration_id since in DB we allow empty values
+
+	var errProjectID, errIntegrationID error
+	if m.Name != "cron" {
```
itayd: const

internal/manifest/plan.go
```diff
@@ -427,57 +429,57 @@ func planTriggers(ctx context.Context, mtriggers []*Trigger, client sdkservices.
 			return nil, fmt.Errorf("trigger %q: invalid entrypoint: %w", mtrigger.GetKey(), err)
 		}
 
+		connectionID := curr.ConnectionID()
+		// schedule could be specified in manifest either in dedicated or in `data' sections. Ensure schedule is present in data section
+		//
```
itayd: kill

internal/manifest/plan.go
```diff
@@ -427,57 +429,57 @@ func planTriggers(ctx context.Context, mtriggers []*Trigger, client sdkservices.
...
+				if schedule != mtrigger.Schedule {
+					return nil, fmt.Errorf("trigger %q: conflicting schedules specified: %q != %q", mtrigger.GetKey(), mtrigger.Schedule, schedule)
+				}
+			}
+			if mtrigger.Data == nil {
+				mtrigger.Data = make(map[string]any)
+			}
+			mtrigger.Data[sdktypes.ScheduleExpression] = mtrigger.Schedule // ensure that schedule is present in data section
+
+			if mtrigger.ConnectionKey == projPrefix { // no connection was specified, assume it's a scheduler trigger
```
itayd: we don't need to assume anything here

internal/backend/triggers/triggers.go
```diff
@@ -27,21 +30,71 @@ func (m *triggers) Create(ctx context.Context, trigger sdktypes.Trigger) (sdktyp
 	}
 
 	trigger = trigger.WithNewID()
+	if schedule, _ := trigger.Data()[sdktypes.ScheduleExpression].ToString(); schedule != "" {
```
itayd: need to check against connection id, not by data

internal/backend/temporalschedule/schedule.go
```diff
@@ -0,0 +1,98 @@
+package temporalschedule
```
itayd: scheduler

ig-ra: done

internal/manifest/plan.go
ig-ra: done

internal/manifest/plan.go
ig-ra: done

sdk/sdktypes/connection.go
ig-ra: done

ig-ra: done

internal/backend/triggers/triggers.go
ig-ra: done

sdk/sdktypes/connection.go
```diff
@@ -28,10 +28,16 @@ func (ConnectionTraits) Validate(m *ConnectionPB) error {
 }
 
 func (ConnectionTraits) StrictValidate(m *ConnectionPB) error {
+	var errProjectID, errIntegrationID error
+	if m.Name != SchedulerConnectionName {
+		errProjectID = mandatory("project_id", m.ProjectId)
```
itayd, line 33: i think project id is always needed?

internal/backend/triggers/triggers.go
```diff
@@ -27,21 +31,75 @@ func (m *triggers) Create(ctx context.Context, trigger sdktypes.Trigger) (sdktyp
 	}
 
 	trigger = trigger.WithNewID()
+	if trigger.ConnectionID() == fixtures.BuiltinSchedulerConnectionID {
```
itayd, line 35: shouldn't this be after the actual trigger creation?

itayd: i guess you worry something will happen between here and line 44. please comment about it.

internal/backend/temporalclient/client.go
```diff
@@ -95,7 +96,8 @@ func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Opti
 	return nil
 }
 
-func (c *impl) Temporal() client.Client { return c.client }
+func (c *impl) Temporal() client.Client               { return c.client }
+func (c *impl) ScheduleClient() client.ScheduleClient { return c.client.ScheduleClient() }
```
itayd: why is this needed? just use Temporal().ScheduleClient()

internal/backend/svc/svc.go
```diff
@@ -158,6 +159,9 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 				})
 			}),
 		),
+		Component(
+			"schedule", configset.Empty, fx.Provide(schedule.New),
+		),
 		Component(
```
itayd:
```suggestion
		Component("schedule", configset.Empty, fx.Provide(schedule.New)),
		Component(
```

internal/backend/projects/projects.go
```diff
@@ -50,6 +51,14 @@ func (ps *Projects) Create(ctx context.Context, project sdktypes.Project) (sdkty
 			return err
 		}
 
+		// create default cron connection
+		if cronConnection, err := tx.GetConnection(ctx, fixtures.BuiltinSchedulerConnectionID); errors.Is(err, sdkerrors.ErrNotFound) {
```
itayd, line 55: interesting. not sure it's needed? if it's built in, no need to display it i think.

internal/backend/fixtures/fixtures.go
```diff
@@ -22,3 +22,5 @@ func NewBuiltinExecutor(xid sdktypes.ExecutorID, opts ...sdkmodule.Optfn) sdkexe
 	vs := kittehs.Must1(mod.Configure(context.TODO(), xid, sdktypes.InvalidConnectionID))
 	return sdkexecutor.NewExecutor(mod, xid, vs)
 }
+
+var BuiltinSchedulerConnectionID = kittehs.Must1(sdktypes.ParseConnectionID("con_3akt5cd1e00000000000000000"))
```
itayd: why this specifc id? is it random?

internal/backend/schedule/schedule.go
```diff
@@ -0,0 +1,98 @@
...
+type temporalScheduleImpl struct {
+	z      *zap.Logger
+	tmprl  temporalclient.Client
+	events sdkservices.Events
+}
+
+type TemporalSchedule interface {
+	CreateScheduledWorkflow(ctx context.Context, scheduleID string, schedule string, triggerID sdktypes.TriggerID) error
+	DeleteSchedulerWorkflow(ctx context.Context, scheduleID string) error
+	UpdateSchedulerWorkflow(ctx context.Context, scheduleID string, schedule string) error
```
itayd:
```suggestion
	Create(ctx context.Context, scheduleID string, schedule string, triggerID sdktypes.TriggerID) error
	Delete(ctx context.Context, scheduleID string) error
	Update(ctx context.Context, scheduleID string, schedule string) error
```

internal/backend/schedule/schedule.go
```diff
@@ -0,0 +1,98 @@
...
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type temporalScheduleImpl struct {
+	z      *zap.Logger
+	tmprl  temporalclient.Client
+	events sdkservices.Events
+}
+
+type TemporalSchedule interface {
```
itayd:
```suggestion
type Schedule interface {
```

and elsewhere. the fact it's temporal is an impl detail.

internal/backend/schedule/schedule.go
```diff
@@ -0,0 +1,98 @@
...
+	CreateScheduledWorkflow(ctx context.Context, scheduleID string, schedule string, triggerID sdktypes.TriggerID) error
```
itayd: why do we need a scheduleID actually? why not just use the triggerID?

internal/backend/schedule/schedule.go
```diff
@@ -0,0 +1,98 @@
...
+	DeleteSchedulerWorkflow(ctx context.Context, scheduleID string) error
+	UpdateSchedulerWorkflow(ctx context.Context, scheduleID string, schedule string) error
+}
+
+func New(z *zap.Logger, events sdkservices.Events, tc temporalclient.Client) TemporalSchedule {
+	return &temporalScheduleImpl{z: z, events: events, tmprl: tc}
+}
+
+func (tsc *temporalScheduleImpl) CreateScheduledWorkflow(ctx context.Context, scheduleID string, schedule string, triggerID sdktypes.TriggerID) error {
+	event := kittehs.Must1(sdktypes.EventFromProto(&sdktypes.EventPB{EventType: sdktypes.SchedulerEventTriggerType}))
```
itayd, line 34: can you add a NewEvent function that does this?

internal/backend/schedule/schedule.go
```diff
@@ -0,0 +1,98 @@
...
+	}
+
+	z := tsc.z.With(zap.String("event_id", eventID.String()))
+	z.Debug("create scheduled event", zap.String("schedule", schedule))
+
+	_, err = tsc.tmprl.ScheduleClient().Create(ctx,
+		client.ScheduleOptions{
+			ID: scheduleID,
+			Spec: client.ScheduleSpec{
+				CronExpressions: []string{schedule},
```
itayd, line 51: do we want to allow multiple schedules on the same trigger as well?

internal/backend/schedule/schedule.go
```diff
@@ -0,0 +1,98 @@
...
+	UpdateSchedulerWorkflow(ctx context.Context, scheduleID string, schedule string) error
+}
+
+func New(z *zap.Logger, events sdkservices.Events, tc temporalclient.Client) TemporalSchedule {
+	return &temporalScheduleImpl{z: z, events: events, tmprl: tc}
+}
+
+func (tsc *temporalScheduleImpl) CreateScheduledWorkflow(ctx context.Context, scheduleID string, schedule string, triggerID sdktypes.TriggerID) error {
+	event := kittehs.Must1(sdktypes.EventFromProto(&sdktypes.EventPB{EventType: sdktypes.SchedulerEventTriggerType}))
+	eventID, err := tsc.events.Save(ctx, event)
```
itayd: we should generate an event id for each tick.

internal/backend/schedule/schedule.go
```diff
@@ -0,0 +1,98 @@
...
+	_, err = tsc.tmprl.ScheduleClient().Create(ctx,
+		client.ScheduleOptions{
+			ID: scheduleID,
+			Spec: client.ScheduleSpec{
+				CronExpressions: []string{schedule},
+			},
+			Action: &client.ScheduleWorkflowAction{
+				ID:        eventID.String(), // workflowID
+				Workflow:  sdktypes.EventsWorkflow,
+				TaskQueue: sdktypes.TaskQueueName,
```
itayd, line 56: why are these in the sdk? this is an impl detail.

internal/backend/dispatcher/dispatcher.go
```diff
@@ -90,8 +91,8 @@ func (d *dispatcher) Redispatch(ctx context.Context, eventID sdktypes.EventID, o
 }
 
 func (d *dispatcher) Start(context.Context) error {
-	w := worker.New(d.temporal.Temporal(), taskQueueName, worker.Options{})
-	w.RegisterWorkflow(d.eventsWorkflow)
+	w := worker.New(d.temporal.Temporal(), sdktypes.TaskQueueName, worker.Options{})
+	w.RegisterWorkflowWithOptions(d.eventsWorkflow, workflow.RegisterOptions{Name: sdktypes.EventsWorkflow})
```
itayd, line 95: i think you should have a new workflow, not reuse the events workflow.

sdk/sdktypes/connection.go
ig-ra: why? in DB we have projectID is optional. moreover cron connection isn't for a project but global one

internal/backend/triggers/triggers.go
ig-ra: no, first I want to create a scheduler workflow 
1. I need to save scheduleID in trigger
2. I want to report an error if creation of schedule workflow failed and not to create a trigger

internal/backend/schedule/schedule.go
ig-ra: scheduleID contains name + triggerID and it's visible in temporalUI. TriggerID is not explanatory enough for the user IMO

internal/backend/temporalclient/client.go
ig-ra: done

internal/backend/schedule/schedule.go
ig-ra: done

internal/backend/schedule/schedule.go
ig-ra: done

internal/backend/fixtures/fixtures.go
ig-ra: hardcoded. I'm using it also in tests and in there I cannot compare with randomID

internal/backend/projects/projects.go
ig-ra: it should be in the DB. maybe we could add additional column that will mark bultin connections, but right now I think this is the simplest way. Do you thought about a better way?

internal/backend/schedule/schedule.go
ig-ra: not for now I think. If there will be a request we could implement it. Meanwhile it supports everything robfig/cron
- standard cron expression
- extended cron expression (up to 7 *, including seconds)
- @monthly, @daily etc shortcuts
- @every <duration>

internal/backend/schedule/schedule.go
ig-ra: did it, but it's a function of 2 lines, I think it's unnecessary actually

internal/backend/schedule/schedule.go
ig-ra: I needed a common place to define consts. plz suggest one?

internal/backend/dispatcher/dispatcher.go
ig-ra: why? IMO it's actually an event workflow which is triggered by temporal and not, e.g. curl. I wanted it to be exactly as normal events workflow. Moreover in temporal UI it make a lot of sense. I could trigger code by curl, or use scheduler as a trigger, or even trigger scheduler workflow manually and all of above will look the same in UI, which is perfect

internal/backend/schedule/schedule.go
ig-ra: let's talk and if we decide that we need it, let's handle this in separate PR.
there is an open question in code whether we need to save cancellation event as well

internal/backend/temporalclient/client.go
ig-ra: done

internal/backend/fixtures/fixtures.go
itayd: 3kth

itayd (approved):
please start working immidietly on generating new events on each tick

cmd/ak/cmd/triggers/create.go
```diff
@@ -8,15 +8,16 @@ import (
...
 	"go.autokitteh.dev/autokitteh/internal/resolver"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-var event, filter, loc, name string
+var event, filter, loc, name, schedule string
 
 var createCmd = common.StandardCommand(&cobra.Command{
-	Use:     "create <--name=...> <--env=...> <--connection=...> <--event=...> <--loc=...>",
+	Use:     `create <--name=...> <--env=...> <--loc=...> { <--connection=...> <--event=...> | <--schedule=...> }`,
```
daabr, line 20: It's a tiny issue, but you shouldn't use `` quotation marks unless there's a specific reasons for it (regex patterns, strings containing "", or multiline strings).

Also, conceptually, the `loc` parameter should come last (after defining the trigger conditions, you define what the trigger will do).

cmd/ak/cmd/triggers/create.go
```diff
@@ -32,30 +33,47 @@ var createCmd = common.StandardCommand(&cobra.Command{
...
 
+		connectionID := ""
+		data := make(map[string]sdktypes.Value)
+		if cmd.Flags().Changed("schedule") {
+			if connection != "" || event != "" {
+				return fmt.Errorf(`flags(s) "connection", "event" are not compatible with "schedule"`)
+			}
+			event = sdktypes.SchedulerEventTriggerType
+			data[sdktypes.ScheduleExpression] = sdktypes.NewStringValue(schedule)
+			connection = sdktypes.SchedulerConnectionName // FIXME: fix resolver. resolving by name isn't working
```
daabr, line 49: Is this new? Is there a ticket with details?

cmd/ak/cmd/triggers/create.go
```diff
@@ -32,30 +33,47 @@ var createCmd = common.StandardCommand(&cobra.Command{
...
 		cl, err := sdktypes.ParseCodeLocation(loc)
 		if err != nil {
 			return fmt.Errorf("invalid entry-point location %q: %w", loc, err)
 		}
 
+		connectionID := ""
+		data := make(map[string]sdktypes.Value)
+		if cmd.Flags().Changed("schedule") {
+			if connection != "" || event != "" {
+				return fmt.Errorf(`flags(s) "connection", "event" are not compatible with "schedule"`)
```
daabr, line 45: As much as possible, you shouldn't write custom logic for this, you should use Cobra's flag modifiers. See `MarkFlagsMutuallyExclusive` and `MarkFlagsRequiredTogether`.

PR #289, opened 2024-05-13
[ENG-834] Fix print using expired context

tebeka:


runtimes/pythonrt/Makefile
```diff
@@ -42,3 +42,11 @@ ngrok:
 
 ngrok-prod:
 	ngrok http --domain autokitteh.ngrok.dev 9980
+
+
+cfg_dir = $(shell ak config where  | grep 'Data home' | awk '{print $$NF}')
+
+clean-dbs:
```
efiShtain, line 49: Technically this only cleans the temporal db file when run as dev server

SQLite file can be in other locations not under data directory, this is controlled by ak_db_Dan Env variable.

And for Postgres, it doesn't do anything

So it won't really work for most cases

efiShtain (approved):

runtimes/pythonrt/Makefile
tebeka: That's Makefile is for my development process, not production.

PR #294, opened 2024-05-13
debugging info

itayd:
- add version to log
- return process signature in http responses and footer
- show uptime in `/uptime`

### DIFF
--- a/internal/backend/fixtures/process.go
+++ b/internal/backend/fixtures/process.go
@@ -22,17 +22,15 @@ func init() {
 	}
 
 	processID = fmt.Sprintf(
-		"%s_%d_%s_%s_%d_%s",
-		strings.Map(func(r rune) rune {
-			if r == '.' || r == '-' {
-				return '_'
-			}
-			return r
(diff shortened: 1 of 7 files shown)


internal/backend/fixtures/process.go
```diff
@@ -22,9 +22,15 @@ func init() {
 	}
 
 	processID = fmt.Sprintf(
-		"%s_%d_%s_%d_%s",
-		strings.ReplaceAll(hostname, ".", "_"),
+		"%s_%d_%s_%s_%d_%s",
+		strings.Map(func(r rune) rune {
```
daabr: Much more readable: https://pkg.go.dev/regexp#Regexp.ReplaceAllString

daabr (approved):
Approving in advance, but please fix my comment. IMO it's unreasonable to define and run an anonymous function **within** a `fmt.Sprintf` call).

internal/backend/fixtures/process.go
itayd: done

PR #300, opened 2024-05-15
github: checks; modules: flat unpack

itayd:
- sdkmodule: add the ability to just specify a struct and unpack will use its fields as destination.
- github: add create and update checks API. this is using the abovementioned feature and a good example.

lmk if you want me to split this pr.

### DIFF
--- a/integrations/github/client.go
+++ b/integrations/github/client.go
@@ -391,13 +391,13 @@ func funcs(i *integration) []sdkmodule.Optfn {
 		sdkmodule.ExportFunction(
 			"create_check_run",
 			i.createCheckRun,
-			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-run"),
+			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/checks/runs?create-a-check-run"),
 			sdkmodule.WithArgs("owner", "repo", "name", "head_sha", "details_url?", "external_url?", "status?", "conclusion?", "output?", "created_at?", "completed_at?", "actions?"),
 		),
 		sdkmodule.ExportFunction(
 			"update_check_run",
 			i.updateCheckRun,
-			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#update-a-check-run"),
+			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/checks/runs?update-a-check-run"),
 			sdkmodule.WithArgs("owner", "repo", "check_run_id", "details_url?", "external_url?", "status?", "conclusion?", "output?", "created_at?", "completed_at?", "actions?"),
 		),
 	}
--- a/integrations/github/client.go
+++ b/integrations/github/client.go
@@ -387,5 +387,18 @@ func funcs(i *integration) []sdkmodule.Optfn {
 			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/actions/workflows#create-a-workflow-dispatch-event"),
 			sdkmodule.WithArgs("owner", "repo", "ref", "workflow_name", "inputs?"),
 		),
+		// Checks
+		sdkmodule.ExportFunction(
+			"create_check_run",
+			i.createCheckRun,
+			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-run"),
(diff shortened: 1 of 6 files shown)


sdk/sdkmodule/unpack.go
```diff
@@ -17,6 +18,11 @@ import (
 // If the parameter name starts with "**", the destination will accept all kwargs as a dict.
 // If the parameter name starts with "*", the destination will aceppt all args as a list.
 //
```
ig-ra, line 22: - plz add examples for the new usage (since this function is documented, I think that documentation should be updated and not be a partial one)
- plz fix the description regarding `=?` (which will be equvalent to kwargs+optional, but it doesn't end with =)

sdk/sdkmodule/unpack.go
```diff
@@ -37,14 +43,64 @@ import (
...
+
+	for i := 0; i < len(dsts); i++ {
+		if _, ok := dsts[i].(string); ok {
+			i++
+			continue
+		}
+
+		t := reflect.TypeOf(dsts[i])
+		if t.Kind() != reflect.Ptr {
+			return fmt.Errorf("dst %d must be name or a pointer to a struct", i)
```
ig-ra, line 61: `name' ?
you skipped all strings in dst. So all remaining dst should be pointers, no?

sdk/sdkmodule/unpack_test.go
```diff
@@ -26,5 +33,38 @@ func TestUnpack(t *testing.T) {
 	}, "i", &i, "s=", &s)) {
 		assert.Equal(t, 64, i)
 		assert.Equal(t, "meow", s)
+		assert.Zero(t, st)
 	}
 }
+
+func TestUnpackFlat(t *testing.T) {
```
ig-ra, line 39: could you/we convert those 2 tests to be parametrized? and cover more cases? (e.g. args, kwargs, missing kwars, optional, more types, etc?)

ig-ra (approved):

ig-ra:
@itayd  - preliminary approve. plz see the comments

sdk/sdkmodule/unpack.go
itayd: done

sdk/sdkmodule/unpack.go
itayd: not sure i follow

integrations/github/client.go
```diff
@@ -327,5 +327,18 @@ func New(vars sdkservices.Vars) sdkservices.Integration {
 			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/actions/workflows#create-a-workflow-dispatch-event"),
 			sdkmodule.WithArgs("owner", "repo", "ref", "workflow_name", "inputs?"),
 		),
+		// Checks
+		sdkmodule.ExportFunction(
+			"create_check_run",
+			i.createCheckRun,
+			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#create-a-check-run"),
```
daabr:
```suggestion
			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/checks/runs#create-a-check-run"),
```

integrations/github/client.go
```diff
@@ -327,5 +327,18 @@ func New(vars sdkservices.Vars) sdkservices.Integration {
...
+			sdkmodule.WithArgs("owner", "repo", "name", "head_sha", "details_url?", "external_url?", "status?", "conclusion?", "output?", "created_at?", "completed_at?", "actions?"),
+		),
+		sdkmodule.ExportFunction(
+			"update_check_run",
+			i.updateCheckRun,
+			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/checks/runs?apiVersion=2022-11-28#update-a-check-run"),
```
daabr:
```suggestion
			sdkmodule.WithFuncDoc("https://docs.github.com/en/rest/checks/runs#update-a-check-run"),
```

daabr (approved):

sdk/sdkmodule/unpack.go
itayd: the things is that the inputs are pairs of strings and anys: n1, &dst1, n2, &dst2. and now one can add something that is not "paired" with a name, such as a struct, where the names come from the fields themselves, see example in doc.

integrations/github/client.go
itayd: done

integrations/github/client.go
itayd: done

sdk/sdkmodule/unpack_test.go
itayd: done

PR #303, opened 2024-05-15
feat: optional connection in trigger

ig-ra:
allow connection to be optional in trigger

### DIFF
--- a/internal/backend/db/dbgorm/triggers.go
+++ b/internal/backend/db/dbgorm/triggers.go
@@ -14,9 +14,11 @@ import (
 
 func triggerToRecord(ctx context.Context, tx *tx, trigger sdktypes.Trigger) (*scheme.Trigger, error) {
 	connID := trigger.ConnectionID()
-	var conn sdktypes.Connection
-	var err error
-	var projID sdktypes.ProjectID
+	var (
+		conn   sdktypes.Connection
+		err    error
+		projID sdktypes.ProjectID
+	)
 
 	if connID.IsValid() {
 		conn, err = tx.GetConnection(ctx, connID)
--- a/internal/backend/db/dbgorm/triggers.go
+++ b/internal/backend/db/dbgorm/triggers.go
@@ -46,7 +46,7 @@ func triggerToRecord(ctx context.Context, tx *tx, trigger sdktypes.Trigger) (*sc
 		TriggerID:    trigger.ID().UUIDValue(),
 		EnvID:        envID.UUIDValue(),
 		ProjectID:    projID.UUIDValue(),
-		ConnectionID: connID.UUIDValue(),
+		ConnectionID: scheme.UUIDOrNil(connID.UUIDValue()),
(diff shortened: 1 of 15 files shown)


internal/backend/db/dbgorm/triggers.go
```diff
@@ -14,24 +14,33 @@ import (
 
 func triggerToRecord(ctx context.Context, tx *tx, trigger sdktypes.Trigger) (*scheme.Trigger, error) {
 	connID := trigger.ConnectionID()
+	var conn sdktypes.Connection
+	var err error
+	var projID sdktypes.ProjectID
```
efiShtain: nit
this is more "go" way
```suggestion
	var (
	       conn sdktypes.Connection
	       err error
	       projID sdktypes.ProjectID
	  )
```

internal/backend/db/dbgorm/triggers.go
```diff
@@ -14,24 +14,33 @@ import (
...
 		}
 
-		if projID != env.ProjectID() {
+		if projID.IsValid() && projID != env.ProjectID() {
 			return nil, fmt.Errorf("env and connection project mismatch: %v != %v", projID, env.ProjectID())
 		}
+		projID = env.ProjectID()
+	}
+
+	if !projID.IsValid() {
```
efiShtain, line 44: move this check after line 26 and fail fast

then you can remove the projID.IsValid() in line 36 as well

proto/autokitteh/triggers/v1/trigger.proto
```diff
@@ -9,7 +9,7 @@ import "buf/validate/validate.proto";
 message Trigger {
   string trigger_id = 1;
   string name = 2;
-  string connection_id = 3;
+  string connection_id = 3; // could be empty
```
efiShtain, line 12: you don't really this comment
all fields are optional in proto by default

sdk/sdktypes/trigger.go
```diff
@@ -89,11 +88,20 @@ func (p Trigger) ToValues() map[string]Value {
 	return map[string]Value{
 		"name": NewStringValue(p.read().Name),
 		"data": kittehs.Must1(NewStructValue(
-			NewStringValue("event_data"),
+			NewStringValue("trigger_data"),
```
efiShtain, line 99: not sure I understand what this is doing
should this be a constant somewhere ?

efiShtain (approved):

sdk/sdktypes/trigger.go
ig-ra: it's just fake name. Probably cut-n-pasted from event code. So I changed the existing typo in code.

proto/autokitteh/triggers/v1/trigger.proto
ig-ra: you are right. But I want to provide addition info why there is no validation check

efiShtain (approved):

PR #305, opened 2024-05-15
ENG-875: upgrade integrations UI

efiShtain:
fun stuff with tailwind to make integration ui a bit better

wdyt?
Screenshot 2024-05-15 at 17 29 51


itayd:
very cool! but do we need to do some extra build step? also @daabr

efiShtain:
we need to run 
```bash
npm run build
```

every time we change the html file with new css from tailwind, 
the output is the updated css file (in the correct directory under web/static)
so assuming you check your changes locally, no extra build step is needed

daabr:
I like it a lot!!

The only comment I have is that this forces us to add NPM to AK's Make rules, CI, linting, and to our list of required build tools in the Readme (similar to our docs repo). We can't rely on people running it manually.

But that's the price of having a modern web UI embedded in an API server, I guess?

I have a few stylistic notes. They should be fixed, but not in this PR, and probably by Ronen:

1. The AK title is waaaaay to big - it's inconsistent with all the other UI pages
2. The squares should have a visual indication for hovering (e.g. slightly different background / slight change in shade?)
3. The gRPC logo is a bit squished horizontally

efiShtain:
> I like it a lot!!
> 
> The only comment I have is that this forces us to add NPM to AK's Make rules, CI, linting, and to our list of required build tools in the Readme (similar to our docs repo). We can't rely on people running it manually.
> 
> But that's the price of having a modern web UI embedded in an API server, I guess?
> 
> I have a few stylistic notes. They should be fixed, but not in this PR, and probably by Ronen:
> 
> 1. The AK title is waaaaay to big - it's inconsistent with all the other UI pages
> 2. The squares should have a visual indication for hovering (e.g. slightly different background / slight change in shade?)
> 3. The gRPC logo is a bit squished horizontally

- you don't need to do anything to build unless you want to change this specific ui page
- I can probably wrap this whole generation of css with docker so no one would have to do anything special
- I assume the "real" UI would be written properly and not by me 
- none of you approved this change, so anyhow I can't merge it

haimzlato:
I love it.
This is roughly how it looks for all applications.
When we have many integrations, we should add search, but we have time :-)

efiShtain:
@itayd 
I've added ```make tailwindcss``` could you check if it works on your machine ? 
probably should install node

daabr (approved):

Makefile
```diff
@@ -158,3 +158,9 @@ generate-migrations:
 	@read -p "Enter migration name: " migration_name; \
 	atlas migrate diff $$migration_name --env sqlite; \
 	atlas migrate diff $$migration_name --env postgres
+
```
itayd, line 161:
```suggestion
```

itayd (approved):

PR #310, opened 2024-05-18
apply: return all effects

itayd:
this returns all effects made (subject id, type and a free message). this can be used by UI to understand which objects where touched and act on it.

one usage is for the UI to check for newly created connections and suggest to initialize them, as it is done in this PR in the cli.


internal/manifest/effect.go
```diff
@@ -0,0 +1,32 @@
+package manifest
+
+import (
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type EffectType string
```
daabr, line 8: I would have called it "action" or "result", but it's entirely not important

internal/kittehs/transforms.go
```diff
@@ -16,6 +16,23 @@ func Transform[A, B any](as []A, f func(A) B) []B {
 	return bs
 }
 
+func TransformFilter[A, B any](as []A, f func(A) *B) []B {
```
daabr, line 19: Self reminder, not affecting this PR: https://linear.app/autokitteh/issue/ENG-881/tlc-for-the-kittehs-package

daabr (approved):

PR #314, opened 2024-05-19
ENG-870: add events order to list events filter

efiShtain:
Notes for reviewers:

- much easier to go commit by commit 
- I fixed a bug in events order + added tests to make this PR work


proto/autokitteh/events/v1/svc.proto
```diff
@@ -40,6 +40,7 @@ message ListRequest {
   string connection_id = 2;
   string event_type = 3;
   uint32 max_results = 4;
+  string order = 5;
```
itayd: add a comment for values or make an enum

sdk/sdkservices/events.go
```diff
@@ -14,8 +14,14 @@ type ListEventsFilter struct {
 	Limit             int
 	CreatedAfter      *time.Time
 	MinSequenceNumber uint64
+	Order             string
 }
 
+const (
+	ListOrderAscending  = "ASC"
+	ListOrderDescending = "DESC"
+)
```
itayd:
```suggestion
	Order             ListOrder
}

type ListOrder string

const (
	ListOrderAscending  ListOrder = "ASC"
	ListOrderDescending ListOrder = "DESC"
)
```

this way it's clear what the expected values are from the type

itayd (approved):

itayd (approved):

itayd (approved):
see comment

proto/autokitteh/events/v1/svc.proto
efiShtain: added comment

sdk/sdkservices/events.go
efiShtain: changed. nice one. I like.

PR #316, opened 2024-05-21
ENG-737: add events filter to workflow signals

efiShtain:


internal/backend/db/db.go
```diff
@@ -152,7 +152,7 @@ type DB interface {
 	SaveSignal(ctx context.Context, signalID string, workflowID string, connectionID sdktypes.ConnectionID, filter string) (string, error)
 	GetSignal(ctx context.Context, signalID string) (scheme.Signal, error)
 	RemoveSignal(ctx context.Context, signalID string) error
-	ListSignalsWaitingOnConnection(ctx context.Context, connectionID sdktypes.ConnectionID, filter string) ([]scheme.Signal, error)
+	ListSignalsWaitingOnConnection(ctx context.Context, connectionID sdktypes.ConnectionID) ([]scheme.Signal, error)
```
itayd, line 155: just fyi - i think returning raw scheme data is wrong, we should wrap it somehow.

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -181,14 +181,36 @@ func (d *dispatcher) signalWorkflows(ctx context.Context, event sdktypes.Event)
...
 	if err != nil {
 		z.Error("could not fetch signals", zap.Error(err))
 		return err
 	}
 
 	z.Debug("found signal candidates", zap.Int("count", len(signals)))
 	for _, signal := range signals {
+		match, err := event.Matches(signal.Filter)
+		if err != nil {
+			z.Error("inavlid signal filter",
```
itayd: i wonder if it's an error, since it's actually a user error, not a backend error. can you add a todo with a ticket to notify the user of these somehow?

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -181,14 +181,36 @@ func (d *dispatcher) signalWorkflows(ctx context.Context, event sdktypes.Event)
...
+				zap.Error(err),
+				zap.String("signal_id", signal.SignalID),
+				zap.String("filter", signal.Filter),
+				zap.String("event_id", event.ID().String()))
```
itayd: you can put all of these in a `With` statement above, then use the logger without these.

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -181,14 +181,36 @@ func (d *dispatcher) signalWorkflows(ctx context.Context, event sdktypes.Event)
...
+				z.Error("failed removing signal with invalid filter", zap.Error(err), zap.String("signal_id", signal.SignalID))
+				continue
+			}
+			z.Debug("signal removed", zap.String("signal_id", signal.SignalID))
+			continue
+		}
+
+		if !match {
+			z.Debug("signal filter not matching event, skipping",
+				zap.String("signal_id", signal.SignalID),
```
itayd: repeated here

sdk/sdktypes/event.go
```diff
@@ -96,6 +96,14 @@ var eventFilterEnv = kittehs.Must1(cel.NewEnv(
 	cel.Variable("data", cel.MapType(cel.StringType, cel.AnyType)),
 ))
 
+func VerifyEventFilter(filter string) (bool, error) {
```
itayd: why not just return error?

internal/backend/db/db.go
efiShtain: I prefer to work with the schema than adding another layer for nothing currently 
especially not another sdk type

but we can discuss that later if you want, I'm no against it

internal/backend/dispatcher/eventsworkflow.go
efiShtain: It's basically server error, since I check the signal filter is valid when I create the signal and notify the user if it's not valid (fail the workflow)
so this check means we have invalid data in the db somehow and it's a server error

sounds ok?

internal/backend/dispatcher/eventsworkflow.go
efiShtain: ok, added

internal/backend/dispatcher/eventsworkflow.go
efiShtain: fixed

sdk/sdktypes/event.go
efiShtain: good point
fixed

internal/backend/db/db.go
itayd: no need for sdktype here - it is not exposed outside.

internal/backend/dispatcher/eventsworkflow.go
itayd: ok

internal/backend/db/db.go
itayd: i still think the scheme should not be exposed - it might contain internal db data and such. can you add a TODO to fix it?

itayd (approved):

internal/backend/db/db.go
efiShtain: added ticket

PR #318, opened 2024-05-21
db: recreate projects with the same name

ig-ra:
allow to recreate project with the same name.
- use hook to check whether non-deleted project exists with the same name
- return duplicated record if so
- allow if there are deleted projects
- change project:name index to be non-unique


daabr (commented):
The code is fine, but I don't see why we should allow this to happen.

This will allow situations where I refer to a project without knowing that it was deleted and replaced by another project.

Other than simple confusion, this is also an opening for abuse and attacks.

itayd:
yes i am not a fan of this approach. will need to brainstorm another solution.

itayd (approved):

PR #322, opened 2024-05-22
fix: revert optional connection in trigger

ig-ra:
undo optional connection in trigger


efiShtain (commented):
I'm not sure you can just remove a migration file 
if someone already ran this migrations they would have a broken migration process
you need to add another migration that re adds the checks

efiShtain (approved):

PR #323, opened 2024-05-22
add sleep/wait w/o condition for test

ig-ra:
allow to sleep in tests for 1-9s

### DIFF
--- a/tests/system/client.go
+++ b/tests/system/client.go
@@ -19,6 +19,11 @@ const (
 	waitInterval = 100 * time.Millisecond
 )
 
+var (
+	sessionStateFinal = regexp.MustCompile(`state:SESSION_STATE_TYPE_(COMPLETED|ERROR|STOPPED)`)
+	sessionStateAll   = regexp.MustCompile(`state:SESSION_STATE_TYPE_(COMPLETED|ERROR|STOPPED|CREATED|RUNNING)`)
+)
+
 func buildClient(t *testing.T) string {
 	wd, err := os.Getwd()
 	if err != nil {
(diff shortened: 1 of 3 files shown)


daabr (commented):
As I mentioned offline, `sleep` is an anti-pattern in system tests:

1. They add unnecessary latency (you wait X seconds even if you got the session you didn't want), and this adds up over time to a lot of wasted cycles in `make test` and CI runs
2. Sleeping for an absolute amount of time instead of a condition with a timeout is flaky - and then the easy fix is to increase the sleep time, which wastes even more cycles

See also "Wait and see" here: https://stackoverflow.com/questions/333682/unit-testing-anti-patterns-catalogue

In general, tests that try to check that something *does't* happen are not reliable tests. In fact, they are an example of [The Halting Problem](https://en.wikipedia.org/wiki/Halting_problem). but I can accept our use-case as an exception, when handled carefully, because it's more important to test for errors.

Instead, how about changing the existing `wait` command?

Instead of `wait <duration> for <session>` we can have `wait <duration> for|unless <session>` - `for` as usual, but the new `unless` alternative will not return an error if we reach the timeout.

Another option: `wait <duration> for <session> [or timeout]`... Or other simple tweaks in the existing syntax?

Either way, don't forget to also update the README.

itayd:
@daabr i agree with you, and i think it would be useful to suggest another way to check cron using system tests. how would you propose doing that?

ig-ra:
> As I mentioned offline, `sleep` is an anti-pattern in system tests:
> 
> 1. They add unnecessary latency (you wait X seconds even if you got the session you didn't want), and this adds up over time to a lot of wasted cycles in `make test` and CI runs
> 2. Sleeping for an absolute amount of time instead of a condition with a timeout is flaky - and then the easy fix is to increase the sleep time, which wastes even more cycles
> 
> See also "Wait and see" here: https://stackoverflow.com/questions/333682/unit-testing-anti-patterns-catalogue
> 
> In general, tests that try to check that something _does't_ happen are not reliable tests. In fact, they are an example of [The Halting Problem](https://en.wikipedia.org/wiki/Halting_problem). but I can accept our use-case as an exception, when handled carefully, because it's more important to test for errors.
> 
> Instead, how about changing the existing `wait` command?
> 
> Instead of `wait <duration> for <session>` we can have `wait <duration> for|unless <session>` - `for` as usual, but the new `unless` alternative will not return an error if we reach the timeout.
> 
> Another option: `wait <duration> for <session> [or timeout]`... Or other simple tweaks in the existing syntax?
> 
> Either way, don't forget to also update the README.

reimplemented as requested.
I'll update README along with approval of coz

tests/system/client.go
```diff
@@ -72,25 +77,41 @@ func waitForSession(akPath, akAddr, step string) (string, error) {
 	if err != nil {
 		return "", fmt.Errorf("invalid duration %q: %w", match[1], err)
 	}
-	id := match[2]
+	isSessionExpected := true // wait .. for .. session
+	if match[2] == "unless" {
+		isSessionExpected = false // wait .. unless .. session
+	}
+	waitType := match[2]
+	id := match[3]
```
daabr: No need for if-else for `isSessionExpected`, and no need to split it from `stateRegex` below.

```suggestion
	id := match[3]
	waitType := match[2]
	isSessionExpected := waitType == "for"
	stateRegex = sessionStateAll
	if isSessionExpected {
		stateRegex = sessionStateFinal
	}
```

tests/system/client.go
```diff
@@ -72,25 +77,41 @@ func waitForSession(akPath, akAddr, step string) (string, error) {
...
 	startTime := time.Now()
 
+	sessionFound := false
 	for time.Since(startTime) < duration {
 		result, err := runClient(akPath, args)
 		if err != nil {
 			return "", fmt.Errorf("failed to get session: %w", err)
 		}
-		if state.MatchString(result.output) {
+		if stateRegex.MatchString(result.output) {
```
daabr: No need for `sessionFound = true` below.

```suggestion
		sessionFound = stateRegex.MatchString(result.output)
		if sessionFound {
```

tests/system/client.go
```diff
@@ -19,6 +19,11 @@ const (
 	waitInterval = 100 * time.Millisecond
 )
 
+var (
+	sessionStateFinal = regexp.MustCompile(`state:SESSION_STATE_TYPE_(COMPLETED|ERROR|STOPPED)`)
+	sessionStateAll   = regexp.MustCompile(`state:SESSION_STATE_TYPE_(COMPLETED|ERROR|STOPPED|CREATED|RUNNING)`)
```
daabr: If we accept all states, we don't need to specify them.
```suggestion
	sessionStateAll   = regexp.MustCompile(`state:SESSION_STATE_TYPE_(\w+)`)
```

daabr (approved):
Approved with a few nits

tests/system/client.go
daabr: Even better: `state:SESSION_STATE_TYPE_` without the suffix, since we don't check/use it in the code

tests/system/client.go
ig-ra: done

tests/system/client.go
ig-ra: done

PR #324, opened 2024-05-22
better args splitting for ak tests

ig-ra:
fix splitting args for ak tests
- previously args were extracted by splitting by whitespace. Which prevents us to pass schedule (e.g. "* * * * *") as an arg
- now args splitting respects quotes. 
- handles
  - single word alphanumeric + [-.:/] args, e.g. `[\w-.:/]+`
  - quoted args, e.g. "* * * * *"
  - var=value args, e.g. `\w+=".*?"` and `\w+=[\w.:]+`

### DIFF
--- a/tests/system/actions.go
+++ b/tests/system/actions.go
@@ -2,6 +2,7 @@ package systest
 
 import (
 	"errors"
+	"regexp"
 	"strings"
 	"testing"
 
@@ -13,13 +14,24 @@ type akResult struct {
 	returnCode int
 }
 
+func splitToArgs(cmdArgs string) []string {
+	re := regexp.MustCompile(`\w+=".*?"|\w+=[\w.:]+|"[^"]+"|[\w-.:/]+`)
+	args := re.FindAllString(cmdArgs, -1)
+	for i, arg := range args {
+		if len(arg) >= 1 && arg[0] == '"' && arg[len(arg)-1] == '"' {
+			args[i] = strings.Trim(arg, "\"")
+		}
+	}
+	return args
+}
+
 func runAction(t *testing.T, akPath, akAddr, step string) (any, error) {
 	match := actions.FindStringSubmatch(step)
 	switch match[1] {
 	case "setenv":
 		return nil, setEnv(strings.TrimSpace(strings.TrimPrefix(step, match[1])))
 	case "ak":
-		args := append(config.ServiceUrlArg(akAddr), strings.Fields(match[3])...)
+		args := append(config.ServiceUrlArg(akAddr), splitToArgs(match[3])...)
 		return runClient(akPath, args)
 	case "http get", "http post":
 		method := strings.ToUpper(match[2])
--- a/tests/system/actions.go
+++ b/tests/system/actions.go
@@ -14,8 +14,16 @@ type akResult struct {
 	returnCode int
 }
 
+// \w+=".*?"       # Match key-value pairs with double quotes
+//
+// \w+=[\w.:]+     # Match key-value pairs without double quotes
+//
+// "[^"]+"         # Match double quoted strings
+//
+// [\w-.:/]+       # Match other words containing alphanumeric, hyphen, dot, colon, or slash characters
+var re = regexp.MustCompile(`\w+=".*?"|\w+=[\w.:]+|"[^"]+"|[\w-.:/]+`)
+
 func splitToArgs(cmdArgs string) []string {
-	re := regexp.MustCompile(`\w+=".*?"|\w+=[\w.:]+|"[^"]+"|[\w-.:/]+`)
 	args := re.FindAllString(cmdArgs, -1)
 	for i, arg := range args {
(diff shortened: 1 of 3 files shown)


tests/system/actions.go
```diff
@@ -13,13 +14,24 @@ type akResult struct {
 	returnCode int
 }
 
+func splitToArgs(cmdArgs string) []string {
+	re := regexp.MustCompile(`\w+=".*?"|\w+=[\w.:]+|"[^"]+"|[\w-.:/]+`)
```
efiShtain: I think in proper regex you have to escape - and /
also, please have examples for this 3 cases 

```suggestion
	re := regexp.MustCompile(`\w+=".*?"|\w+=[\w.:]+|"[^"]+"|[\w\-.:\/]+`)
```

tests/system/actions.go
```diff
@@ -13,13 +14,24 @@ type akResult struct {
...
+}
+
 func runAction(t *testing.T, akPath, akAddr, step string) (any, error) {
 	match := actions.FindStringSubmatch(step)
 	switch match[1] {
 	case "setenv":
 		return nil, setEnv(strings.TrimSpace(strings.TrimPrefix(step, match[1])))
 	case "ak":
-		args := append(config.ServiceUrlArg(akAddr), strings.Fields(match[3])...)
+		args := append(config.ServiceUrlArg(akAddr), splitToArgs(match[3])...)
```
efiShtain, line 32: instead of sophisticated hard to maintain regex you could try strings.FieldsFunc

look at this example
https://stackoverflow.com/questions/47489745/splitting-a-string-at-space-except-inside-quotation-marks

ig-ra: I prefer to use regexp actually, for me it's simpler. but let's hear what @itayd and @daabr think

tests/system/actions.go
tebeka: Move `re.MustCompile` to `init`. We don't want this to panic when running on a bad regexp.
It will also compile once.

Also: Document the regexp with what it does. Something like:
```
// `a "b c" d` -> []string{"a", "b c", "d"}
```
Otherwise in 2 weeks we'll forget what it does.

tests/system/actions.go
```diff
@@ -13,13 +14,24 @@ type akResult struct {
 	returnCode int
 }
 
+func splitToArgs(cmdArgs string) []string {
```
tebeka, line 17: What happens if we have a runaway quote (`a "b c" "d e`), it this an error, if not - how do we handle it?

tests/system/actions.go
tebeka: What about supporting `'` as well?

tebeka: Or maybe https://pkg.go.dev/github.com/google/shlex

tests/system/actions.go
ig-ra: I'm not trying to build generic generic args splitter.  So in runaway quote it will extract `c` `" "` and `d` as arguments and thus will surely fail specific system test where error was introduced

tests/system/actions.go
ig-ra: do we want to use archived unsupported package from 6-10 years go?

ig-ra: same note. It's not supposed to be a generic parser. The old one was splitting args by space, this one is trying to be just a bit more sophisticated since I need to parse a schedule expression

ig-ra: done

ig-ra: @efiShtain I checked with regexp101  and tests and it seems there is no need in additional escaping

tebeka: > do we want to use archived unsupported package from 6-10 years go?

No commits != unsupported. But I'm OK with rolling our own.

tests/system/actions.go
tebeka: OK

tests/system/actions.go
tebeka: OK

tests/system/actions.go
daabr: I love regexes, but I have to admit there are reasons why they're less ideal here:

1. There are standard, popular, and well-tested alternatives (see Efi's link)
2. The explanation of this short regex requires 4 lines of comments - that's a code smell for regex complexity
3. The less we assume/require, the better; in other words, future CLI changes and even small-but-reasonable tweaks in the expected behavior will make it much harder to read/maintain (e.g. supporting more than `\w`, supporting spaces around `=`, specifying URLs with all their complexity in the CLI, etc.)

daabr: That said, it seems to me that https://go.dev/play/p/Ed4IV97L7H is better than `strings.FieldsFunc` - we don't want to preserve `""`

ig-ra: I'll go for csv reader. By default it doesn't allow quotes in parsed fields, but it seems there is LazyQuotes flag, which allows to parse things like `c="meow"` which we are using in our tests

efiShtain (approved):

PR #325, opened 2024-05-22
[ENG-539] Replace Python's sleep

tebeka:
In order to have durable sleep, we need to replace Python's sleep with `ak.syscall("sleep", duration)`

Some unrelated code cleanup changes:
- AST tree recursion the right way
- pytest options in `pyproject.toml`

Tested by adding a call to `sleep(1)` to `testdata/simple/simple.py` and running the workflow.


runtimes/pythonrt/ak_runner.py
```diff
@@ -375,8 +391,8 @@ def module_entries(mod):
 
 
 if __name__ == '__main__':
-    from argparse import ArgumentParser
     import sys
+    from argparse import ArgumentParser
```
daabr, line 334: I feel uncomfortable approving a PR that contains imports not at the top of the module, but I wouldn't have seen this if you didn't just flip the order of existing imports, i.e. it's not material to this PR, so fine... :)

daabr (approved):

runtimes/pythonrt/ak_runner.py
tebeka: I import what's need to run only under `__main__`, old habit.

PR #326, opened 2024-05-22
[ENG-895] Allow setting var from standard input

tebeka:


cmd/ak/cmd/vars/set.go
```diff
@@ -28,7 +31,20 @@ var setCmd = common.StandardCommand(&cobra.Command{
 			return fmt.Errorf("invalid variable name %q: %w", args[0], err)
 		}
 
-		ev := sdktypes.NewVar(n, args[1], secret).WithScopeID(id)
+		var value string
```
itayd, line 34: i think better to have a flag to read from stdin, no?

daabr: e.g. https://github.com/autokitteh/autokitteh/blob/main/cmd/ak/cmd/manifest/validate.go

tebeka: > i think better to have a flag to read from stdin, no?

The common Unix practice is to read from stdin (`grep`, `sed` ...) if no file is specified. Or maybe use `-` (see [here](https://docs.python.org/3/library/argparse.html#argparse.FileType) for example) as the file name. I prefer the former.

There also what `curl` does which is when value is prefixed with `@` it's a name of the file to read from (see `Variables` section [here](https://curl.se/docs/manpage.html)), but it's less flexible than standard output that can also accept a pipe ...

@daabr Your example is like the current behaviour. Am I missing something?

ig-ra (approved):

cmd/ak/cmd/vars/set.go
```diff
@@ -28,7 +31,20 @@ var setCmd = common.StandardCommand(&cobra.Command{
...
+		if len(args) == 1 {
```
daabr: Optional nit: given that `== 2` is the "trivial" case, and it takes only 1 line of code, it's easier to read the code if you reverse the order of the branches.

cmd/ak/cmd/vars/set.go
daabr: Sorry for not being clear, my comment was just intended to provide a concrete example ("e.g.") for how we currently handle files vs stdin.

Personally, I prefer the practice of indicating the usage of stdin with  `-`, but that's not what we usually do in AK.

I'm ok with submitting your PR, but we should use a single convention in all relevant commands.

daabr (approved):

cmd/ak/cmd/vars/set.go
daabr: I opened https://linear.app/autokitteh/issue/ENG-928/cli-standardization-using-stdin-instead-of-filename for follow-up

daabr: I opened https://linear.app/autokitteh/issue/ENG-928/cli-standardization-using-stdin-instead-of-filename for follow-up

cmd/ak/cmd/vars/set.go
tebeka: Done.

PR #327, opened 2024-05-22
[ENG-876] Use TLS in Go<->Python communication (WIP)

tebeka:
User can specify `AK_WORKER_PEM` and `AK_WORKER_KEY` environment variables and then `ak` will use TLS to communicate with the Python process.

@efiShtain I'd love your take on it.


daabr (changes requested):
While the goal of this PR is a positive one, storing PKI-related files in a git repo, much less a public one, is a strict taboo.

tebeka:
> While the goal of this PR is a positive one, storing PKI-related files in a git repo, much less a public one, is a strict taboo.

Strict taboos are [meant to be broken](https://paulgraham.com/say.html) :)
I agree (added `(WIP)` to the PR title). The PKI are for the Python/Go process communication which is external. I'm no security expert but looking for a pragmatic solution.

@efiShtain ?

efiShtain:
> > While the goal of this PR is a positive one, storing PKI-related files in a git repo, much less a public one, is a strict taboo.
> 
> 
> 
> Strict taboos are [meant to be broken](https://paulgraham.com/say.html) :)
> 
> I agree (added `(WIP)` to the PR title). The PKI are for the Python/Go process communication which is external. I'm no security expert but looking for a pragmatic solution.
> 
> 
> 
> @efiShtain ?

Didn't see this one 

I don't mind having this keys here for dev purposes
Any user can choose to keep using them or replace them (they should replace them of course)
We won't use them in our cloud offering 
You should have a way to provide them as Env variables or mounted volumes

tebeka:
> While the goal of this PR is a positive one, storing PKI-related files in a git repo, much less a public one, is a strict taboo.

Changed to pass files in environment, if not specified won't use TLS.

efiShtain:
> > While the goal of this PR is a positive one, storing PKI-related files in a git repo, much less a public one, is a strict taboo.
> 
> Changed to pass files in environment, if not specified won't use TLS.

If you can support passing string value in env variable, it would make life easier in some deployments (where it's harder / not convenient to mount a file)
see what I did with temporal keys in temporal client

```go
if cfg.TLS.Certificate != "" && cfg.TLS.Key != "" {
			cert, err = tls.X509KeyPair([]byte(cfg.TLS.Certificate), []byte(cfg.TLS.Key))
		} else if cfg.TLS.CertFilePath != "" && cfg.TLS.KeyFilePath != "" {
			cert, err = tls.LoadX509KeyPair(cfg.TLS.CertFilePath, cfg.TLS.KeyFilePath)
		} else {
			return nil, errors.New("tls enabled without certificate or key")
		}
```

daabr:
> I don't mind having this keys here for dev purposes Any user can choose to keep using them or replace them (they should replace them of course) We won't use them in our cloud offering You should have a way to provide them as Env variables or mounted volumes

I disagree. The "secure by default" principle means not only that we won't use a leaked key ourselves (which in itself is an optimistic promise and not a guarantee), but also that users won't be able to shoot themselves in the leg and be vulnerable to "Google dorking" just because they don't know any better.

efiShtain:
> > I don't mind having this keys here for dev purposes Any user can choose to keep using them or replace them (they should replace them of course) We won't use them in our cloud offering You should have a way to provide them as Env variables or mounted volumes
> 
> I disagree. The "secure by default" principle means not only that we won't use a leaked key ourselves (which in itself is an optimistic promise and not a guarantee), but also that users won't be able to shoot themselves in the leg and be vulnerable to "Google dorking" just because they don't know any better.

Dev keys are a valid thing
Postgres starts by default without a password, it is documented that they recommend not to do it production environment - but that's for the user to decide and it's the user's responsibility

If it's simpler for us now, I think having these dev keys here is valid

but since miki added the no tls option by default, it doesn't really matter since you don't have to use the keys

tebeka:
I've changed the PR to use environment variables and read files from the OS, no .pem files in the repo.
@efiShtain Can you take a second look?

tebeka:
> The "secure by default" principle means not only that we won't use a leaked key ourselves

Changed the PR. @daabr can you look again?

PR #331, opened 2024-05-23
feat: scheduler workflow

ig-ra:
Brief
1. dispatcher - no actual changes. Code is refactored. Common parts moved to backend/workflows to be used by both dispatcher and scheduler
2. scheduler is written on top of temporal scheduler
3. new scheduler trigger is introduced
4. in order to simplify triggers management new default cron connection was introduced (see fixtures)
5. triggers ops were updated to handle schedule triggers. If create/update/delete they will create/update/delete the relevant temporal schedule
6. each schedule tick is new schedule_tick event


ig-ra:
I moved common dispatcher/scheduler code to separate package. If there is a need I could extract this to separate PR

efiShtain:
Could you please give a quick overview of the changes proposed here ?
not sure I can review it without context

sdk/sdkservices/scheduler.go
```diff
@@ -0,0 +1,13 @@
+package sdkservices
+
+import (
+	"context"
+
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type Scheduler interface {
+	Create(ctx context.Context, scheduleID string, schedule string, triggerID sdktypes.TriggerID) error
```
itayd, line 10: is scheduleID an actual ID, free text? if it's an ID, it should be in sdktypes. This can be done in a separate PR.

sdk/sdktypes/trigger.go
```diff
@@ -88,11 +88,24 @@ func (p Trigger) ToValues() map[string]Value {
...
+			NewStringValue("trigger_data"),
 			kittehs.TransformMapValues(
 				p.read().Data,
 				forceFromProto[Value],
 			),
 		)),
 	}
 }
+
+func (p Trigger) WithUpdatedData(key string, val Value) Trigger {
```
itayd: WithData is more idiomatic

sdk/sdktypes/connection.go
```diff
@@ -28,10 +28,16 @@ func (ConnectionTraits) Validate(m *ConnectionPB) error {
...
+	}
+
 	return errors.Join(
 		mandatory("name", m.Name),
-		mandatory("project_id", m.ProjectId),
-		mandatory("integration_id", m.IntegrationId),
+		errProjectID,
+		errIntegrationID,
 	)
 }
```
itayd: golfing

```suggestion
	var errs []error
	if m.Name != SchedulerConnectionName {
        errs = []error{
		    mandatory("project_id", m.ProjectId),
		    mandatory("integration_id", m.IntegrationId)
        }
	}

	return errors.Join(tmandatory("name", m.Name), errs...)
}
```

sdk/sdktypes/consts.go
```diff
@@ -0,0 +1,9 @@
+package sdktypes
+
+const (
```
itayd: these are all internal consts that belong in fixtures, not here.

internal/backend/triggers/triggers.go
```diff
@@ -28,20 +31,75 @@ func (m *triggers) Create(ctx context.Context, trigger sdktypes.Trigger) (sdktyp
...
 
+	if trigger.ConnectionID() == fixtures.BuiltinSchedulerConnectionID {
+		if schedule, _ := trigger.Data()[sdktypes.ScheduleExpression].ToString(); schedule != "" {
+			scheduleID := newScheduleID(trigger)
+			if err := m.scheduler.Create(ctx, scheduleID, schedule, trigger.ID()); err != nil {
+				return sdktypes.InvalidTriggerID, err
+			}
+			trigger = trigger.WithUpdatedData(sdktypes.ScheduleIDKey, sdktypes.NewStringValue(scheduleID))
+		}
+	}
```
itayd:
```suggestion
	if trigger.ConnectionID() == fixtures.BuiltinSchedulerConnectionID {
		if schedule := trigger.Data()[sdktypes.ScheduleExpression].GetString(); schedule != "" {
			scheduleID := newScheduleID(trigger)
			if err := m.scheduler.Create(ctx, scheduleID, schedule, trigger.ID()); err != nil {
				return sdktypes.InvalidTriggerID, err
			}
			trigger = trigger.WithUpdatedData(sdktypes.ScheduleIDKey, sdktypes.NewStringValue(scheduleID))
		}
	}
```

internal/backend/triggers/triggers.go
```diff
@@ -28,20 +31,75 @@ func (m *triggers) Create(ctx context.Context, trigger sdktypes.Trigger) (sdktyp
...
 func (m *triggers) Update(ctx context.Context, trigger sdktypes.Trigger) error {
+	prevTrigger, err := m.db.GetTrigger(ctx, trigger.ID())
+	if err != nil {
+		return err
+	}
+
+	prevData := prevTrigger.Data()
+	prevScheduleVal, isSchedulerPrevTrigger := prevData[sdktypes.ScheduleExpression]
+	scheduleID, _ := prevData[sdktypes.ScheduleIDKey].ToString()
+	isSchedulerPrevTrigger = isSchedulerPrevTrigger && prevTrigger.ConnectionID() == fixtures.BuiltinSchedulerConnectionID
```
itayd:
```suggestion
	scheduleID := prevData[sdktypes.ScheduleIDKey].GetString()
```

internal/backend/triggers/triggers.go
```diff
@@ -28,20 +31,75 @@ func (m *triggers) Create(ctx context.Context, trigger sdktypes.Trigger) (sdktyp
...
+
+	data := trigger.Data()
+	scheduleVal, isSchedulerTrigger := data[sdktypes.ScheduleExpression]
+	schedule, _ := scheduleVal.ToString()
+	isSchedulerTrigger = isSchedulerTrigger && trigger.ConnectionID() == fixtures.BuiltinSchedulerConnectionID
+
+	if isSchedulerPrevTrigger {
```
itayd, line 72: we need to figure out in a separate PR what happens if we crush here

internal/backend/triggers/triggers.go
```diff
@@ -28,20 +31,75 @@ func (m *triggers) Create(ctx context.Context, trigger sdktypes.Trigger) (sdktyp
...
 
 func (m *triggers) Update(ctx context.Context, trigger sdktypes.Trigger) error {
+	prevTrigger, err := m.db.GetTrigger(ctx, trigger.ID())
+	if err != nil {
+		return err
+	}
+
+	prevData := prevTrigger.Data()
+	prevScheduleVal, isSchedulerPrevTrigger := prevData[sdktypes.ScheduleExpression]
+	scheduleID, _ := prevData[sdktypes.ScheduleIDKey].ToString()
```
itayd: why isn't the schedule id just the trigger id?

internal/backend/svc/svc.go
```diff
@@ -186,6 +187,20 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
...
 
 		Component("healthcheck", configset.Empty, fx.Provide(healthchecker.New)),
+		Component(
+			"schedule",
+			configset.Empty,
+			fx.Provide(schedule.New),
+			// fx.Provide(func(sch schedule.Scheduler) sdkservices.Scheduler { return sch }),
+			// fx.Invoke(func(lc fx.Lifecycle, sch schedule.Scheduler) { HookOnStart(lc, sch.Start) }),
+		),
+		Component(
```
itayd, line 196: why is this a separate component? why is it not included inside schedule?

internal/backend/schedule/scheduleworkflow.go
```diff
@@ -0,0 +1,150 @@
...
+	eventID, err := swf.Services.Events.Save(ctx, event) // create event
+	if err != nil {
+		z.Panic("failed creating schedule tick event")
+	}
+
+	// REVIEW: re-fetching newly created event is costly (db). Right now we are recreating event without CreatedAt.
+	// other solution is to cause events.SaveEvent to return created event and not just ID
+	event = kittehs.Must1(sdktypes.EventFromProto(&sdktypes.EventPB{EventId: eventID.String(), EventType: sdktypes.SchedulerTickEventType}))
+
+	swf.CreateEventRecord(ctx, eventID, sdktypes.EventStateProcessing) // add `processing` event record
```
itayd, line 107: this can fail

internal/backend/schedule/scheduleworkflow.go
```diff
@@ -0,0 +1,150 @@
...
+	return err
+}
+
+func (swf *SchedulerWorkflow) scheduleWorkflow(wfctx workflow.Context, triggerID sdktypes.TriggerID) (*scheduleWorkflowOutput, error) {
+	var err error
+
+	logger := workflow.GetLogger(wfctx)
+	logger.Info("started scheduler workflow", "trigger_id", triggerID.String())
+
+	ctx := context.Background()
```
itayd: you can use `NewWorkflowContextAsGOContext` from https://github.com/autokitteh/autokitteh/blob/main/internal/backend/temporalclient/context.go

internal/backend/schedule/scheduleworkflow.go
```diff
@@ -0,0 +1,150 @@
...
+
+	// REVIEW: re-fetching newly created event is costly (db). Right now we are recreating event without CreatedAt.
+	// other solution is to cause events.SaveEvent to return created event and not just ID
+	event = kittehs.Must1(sdktypes.EventFromProto(&sdktypes.EventPB{EventId: eventID.String(), EventType: sdktypes.SchedulerTickEventType}))
+
+	swf.CreateEventRecord(ctx, eventID, sdktypes.EventStateProcessing) // add `processing` event record
+	return event
+}
+
+func (swf *SchedulerWorkflow) scheduleWorkflowInternal(wfctx workflow.Context, ctx context.Context, triggerID sdktypes.TriggerID, tickEvent *sdktypes.Event) error {
```
itayd: this is used only in scheduleWorkflow, why did you pull this out?

internal/backend/schedule/scheduleworkflow.go
```diff
@@ -0,0 +1,150 @@
...
+func (swf *SchedulerWorkflow) scheduleWorkflow(wfctx workflow.Context, triggerID sdktypes.TriggerID) (*scheduleWorkflowOutput, error) {
+	var err error
+
+	logger := workflow.GetLogger(wfctx)
+	logger.Info("started scheduler workflow", "trigger_id", triggerID.String())
+
+	ctx := context.Background()
+	tickEvent := swf.newScheduleTickEvent(ctx, triggerID) // create tick event and add <processing> state
+
+	state := sdktypes.EventStateCompleted
```
itayd, line 140: shouldn't this be Created?

sdk/sdkservices/scheduler.go
ig-ra: scheduleID is text, it's an ID we are passing to temporal to name schedule. It's not an sdkType and not intended to be

sdk/sdktypes/trigger.go
ig-ra: This is a deliberate naming decision. 
Otherwise it will be unaligned with our naming convention for getters and setters of sdktypes. `WithData` should update all data and receive a map, see `events.WithData()`. The same logic is for getters - `Data()` returns a map and not a single key/value. Therefore, since I wanted to update only a single key I called this method `withUpdatedData(key, value)`.

sdk/sdktypes/connection.go
ig-ra: you cannot mix fixed args and unpacked slices - https://stackoverflow.com/questions/28625546/mixing-exploded-slices-and-regular-parameters-in-variadic-functions
so, I did it differently than suggested

internal/backend/triggers/triggers.go
ig-ra: GetString returns StringValue, from which I need to extract string.
Did you mean `String()` ?

internal/backend/triggers/triggers.go
ig-ra: see my previous comment on similar. GetString returns StringValue, I need a string.

internal/backend/triggers/triggers.go
ig-ra: readability for the user through temporal UI. `daily_trg_xxx` IMO is better then just `trg_xxx` which explains nothing.
If we don't care we could remove this part that creates and saves scheduleID in the data section of the trigger and assume permanently that scheduleID == triggerID

internal/backend/triggers/triggers.go
ig-ra: if we crash where? the comment is on `if` so I'm not sure that I understand

internal/backend/svc/svc.go
ig-ra: otherwise there is an import loop (triggers need temporal scheduler and it needs triggers). Therefore it was splitted this to two components. Scheduler which is responsible for starting/updating/deleting temporal schedules and scheduleWorkflow which is similar to dispatcher::eventsWorkflow. Common workflows code was extracted to workflows/common

internal/backend/schedule/scheduleworkflow.go
ig-ra: CreateEventRecord cannot fail, but it may panic, which preserves the original behavior, which is (see dispatcher, method was extracted to workflows/common) was either panic or returning a nil.

internal/backend/schedule/scheduleworkflow.go
ig-ra: I wanted to create internal function that does all work and logs with zap.Logger and external function that creates event ,updates state and logs with logger. The way how this is structured in dispatcher is much less readable IMO
I could rename this function of coz. `scheduleWorkflowInternal` does all relevant schedule work

internal/backend/schedule/scheduleworkflow.go
ig-ra: could u plz explain more? why should I prefer temporalclient context on regular context in this case?
this context is used only for our services, e.g. create event, create event record, read trigger, deployments etc..

internal/backend/schedule/scheduleworkflow.go
ig-ra: there is no `Created`, but there is `Processing` and then `completed/failure`.
newScheduleTickEvent creates both event and Processing state.
if scheduleWorkflowInternal finishes without error, then state = Completed otherwise Failure.

sdk/sdktypes/consts.go
ig-ra: why fixtures? 
those consts used in manifest plan/exec, triggers, cmd and scheduler. Do you want me to put them in fixture file?

sdk/sdkservices/scheduler.go
itayd: name can change, you need to use trigger id here.

sdk/sdktypes/trigger.go
itayd: WithAdditionalData

internal/backend/svc/svc.go
```diff
@@ -186,6 +187,20 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 		Component("runtimes", configset.Empty, fx.Provide(runtimes.New)),
 
 		Component("healthcheck", configset.Empty, fx.Provide(healthchecker.New)),
+		Component(
+			"schedule",
+			configset.Empty,
+			fx.Provide(schedule.New),
+			// fx.Provide(func(sch schedule.Scheduler) sdkservices.Scheduler { return sch }),
+			// fx.Invoke(func(lc fx.Lifecycle, sch schedule.Scheduler) { HookOnStart(lc, sch.Start) }),
```
itayd: kill

internal/backend/svc/svc.go
```diff
@@ -186,6 +187,20 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
...
+		),
+		Component(
+			"scheduleWorkflow",
+			configset.Empty,
+			fx.Provide(schedule.NewSchedulerWorkflow),
+			// fx.Provide(func(sch schedule.Scheduler) sdkservices.Scheduler { return sch }),
```
itayd: kill

internal/backend/svc/svc.go
itayd: add TODO to consider this

itayd: TODO: consider design where the workflow is an implementation detail of the scheduler and does not need to be exposed.

internal/backend/triggers/triggers.go
```diff
@@ -28,20 +31,75 @@ func (m *triggers) Create(ctx context.Context, trigger sdktypes.Trigger) (sdktyp
 
 	trigger = trigger.WithNewID()
 
+	if trigger.ConnectionID() == fixtures.BuiltinSchedulerConnectionID {
+		if schedule, _ := trigger.Data()[sdktypes.ScheduleExpression].ToString(); schedule != "" {
```
itayd: validate in trigger that schedule is string

internal/backend/triggers/triggers.go
ig-ra: what will happen if we crash in the middle of creating trigger? unrelated to schedule part..

internal/backend/triggers/triggers.go
ig-ra: done

internal/backend/svc/svc.go
ig-ra: added todo.

internal/backend/schedule/scheduleworkflow.go
ig-ra: done

internal/backend/schedule/scheduleworkflow.go
ig-ra: resolving, right?

sdk/sdkservices/scheduler.go
ig-ra: done

sdk/sdktypes/consts.go
ig-ra: 1. I moved builtinSchedulerConnectionID to `sdktypes/const`. I cannot keep it in fixtures, since `sdktypes.Strict()` uses it to check
2. now, considering the fact that anyway I need to keep at least scheduler connection ID in `sdktypes` I suggest to reconsider moving all other consts to fixtures. I think they all need to be in the same place

internal/backend/triggers/triggers.go
ig-ra: actually it's already like this
- if non-empty but invalid schedule passed, then temporal will complain and no trigger will be created/updated
- if none schedule is passed we cannot distinguish between cron trigger and normal trigger, so it will complain about not having connection (since both schedule and connection will be empty in proto), so this is a valid error

internal/backend/triggers/triggers.go
itayd: the scheduler makes this non-atomic, hence it is related

internal/backend/schedule/scheduleworkflow.go
itayd: need to think of a better name

sdk/sdkservices/scheduler.go
itayd: i think you didn't push the latest changes - i still see scheduleID.

internal/backend/triggers/triggers.go
ig-ra: my question is how we preserve atomicity on any other operation, e.g. creating event + event record. Ak could fail after creating the record but before creating the record.
the ak just between workflow creation and saving the trigger, this will mean that we have orphaned workflow that is running without the trigger. So we could ensure this in workflow code and if there is no trigger to remove the workflow. Right now workflow will be triggered by temporal each tick, but workflow function will fail since no such trigger will be found

internal/backend/schedule/scheduleworkflow.go
ig-ra: ScheduleWorkflowEntry and ScheduleWorkflow?
ScheduleWorkflow and ScheduleWorkflowRunSessions?

sdk/sdkservices/scheduler.go
ig-ra: I pushed. We don't save `scheduleID` anymore in trigger data, since scheduleID == triggerID as you wanted (e.g. not prepending a trigger name to it)

we need to pass `ScheduleID` is to create an normal API, same way we are doing for every other object. There are 3 functions in the interface - Create, Update and Delete. Very similar to all other our interfaces. 
- Update and Delete should receive `ScheduleID` (the same way any connection function receives `connectionID` or trigger functions receives `triggerID` etc..).
- Create could receive either `Schedule` object and return scheduleID (if we want to add Schedule to sdktypes). Right now I see no point in doing so, so `Create` also receives `scheduleID`.

ig-ra: I pushed. We don't save `scheduleID` anymore in trigger data, since scheduleID == triggerID as you wanted (e.g. not prepending a trigger name to it)

we need to pass `ScheduleID` is to create an normal API, same way we are doing for every other object. There are 3 functions in the interface - Create, Update and Delete. Very similar to all other our interfaces. 
- Update and Delete should receive `ScheduleID` (the same way any connection function receives `connectionID` or trigger functions receives `triggerID` etc..).
- Create could receive either `Schedule` object and return scheduleID (if we want to add Schedule to sdktypes). Right now I see no point in doing so, so `Create` also receives `scheduleID`.

internal/backend/schedule/scheduleworkflow.go
itayd: scheduleSessions

internal/backend/triggers/triggers.go
itayd: comment that the session will complain if we crash and we need to monitor it

internal/backend/triggers/triggers.go
ig-ra: Added a Note in the code

internal/backend/triggers/triggers.go
ig-ra: added a NOTE in the code

internal/backend/workflows/common.go
```diff
@@ -0,0 +1,141 @@
+package workflows
```
itayd, line 1: `commonsessionworkflow`?

internal/backend/workflows/common.go
```diff
@@ -0,0 +1,141 @@
...
+
+	inputs := map[string]sdktypes.Value{
+		"event": eventStruct,
+		"data":  eventInputs["data"],
+	}
+
+	sessions := make([]*sdktypes.Session, len(sessionsData))
+	for i, sd := range sessionsData {
+		if t := sd.Trigger; t.IsValid() {
+			inputs = maps.Clone(inputs)
```
itayd, line 76: this is a bug - it'll clone and reclone the previous iteration. you want to clone the original inputs in line 68.

internal/backend/dispatcher/eventsworkflow.go
```diff
@@ -266,82 +249,6 @@ func (d *dispatcher) eventsWorkflow(ctx workflow.Context, input eventsWorkflowIn
 		return nil, err
 	}
 	// Set event to Completed
-	if err = d.createEventRecord(context.Background(), input.EventID, sdktypes.EventStateCompleted); err != nil {
-		return nil, err
-	}
-
+	d.CreateEventRecord(context.Background(), input.EventID, sdktypes.EventStateCompleted)
```
itayd, line 252: why no error handling here?

internal/backend/schedule/schedule.go
```diff
@@ -0,0 +1,102 @@
...
+		client.ScheduleOptions{
+			ID: scheduleID,
+			Spec: client.ScheduleSpec{
+				CronExpressions: []string{schedule},
+			},
+			Action: &client.ScheduleWorkflowAction{
+				ID:        triggerID.String(), // workflowID
+				Workflow:  wf.SchedulerWorkflow,
+				TaskQueue: wf.ScheduleTaskQueueName,
+				Args:      []interface{}{triggerID},
```
itayd: s/interface{}/any

internal/backend/workflows/common.go
ig-ra: too long for me and IMO it's reduces readability :(

internal/backend/dispatcher/eventsworkflow.go
ig-ra: createEventRecord returns nothing. It panics. Note that previously this it was returning just nil always.

internal/backend/workflows/common.go
ig-ra: 1. not sure I understand. Could u plz elaborate what's wrong? 
2. Why this function should be different from what is done for eventsworkflow?

internal/backend/workflows/common.go
ig-ra: I would suggest to leave the naming as is - it's just the common functions/things used in all our workflows

itayd (approved):

PR #332, opened 2024-05-24
Add munch to Python requirements

daabr:
Add https://github.com/Infinidat/munch to our Python requirements list.

Munch allows scripts to convert JSON-based dictionaries with arbitrary structure into struct-like objects, i.e. dynamically convert dict keys into object attributes.

See example usage in https://github.com/autokitteh/samples/pull/82

### DIFF
--- a/runtimes/pythonrt/requirements.txt
+++ b/runtimes/pythonrt/requirements.txt
@@ -14,5 +14,6 @@ twilio ~= 9.0
 
 # General
 
+munch ~= 4.0
 PyYAML ~= 6.0
 requests ~= 2.31


tebeka (commented):
`munch` looks cool, but I prefer not to add it.

We'll need to support any package that we add to the default venv indefinitely for backward compatibility. That's why I think we should stick to parity with integration and maybe very selected few that are super popular (currently only PyYAML and requests). The user can install whatever they want to the environment if they want to work.

I don't think `munch` passes the threshold of making it to the default environment.
IMO the right solution is to solve https://linear.app/autokitteh/issue/ENG-538/support-python-dependencies and not beef up the default venv.

@daabr What is the reasoning for adding it to the default environment?

tebeka:
@daabr I'm sold :smile:  See https://github.com/autokitteh/autokitteh/pull/341

PR #334, opened 2024-05-27
ENG-913: compress session and event data

itayd:
compress session and event data in order to save db space. this is backward compatible as it can read non-compress data as well.

compression happens only if payload is bigger than configured threshold. threshold is configured by supplying an extra option in configuration. default is 4k.

### DIFF
--- a/internal/backend/db/dbgorm/dbgorm.go
+++ b/internal/backend/db/dbgorm/dbgorm.go
@@ -1,14 +1,19 @@
 package dbgorm
 
 import (
+	"bytes"
+	"compress/gzip"
 	"context"
 	"database/sql"
+	"encoding/json"
 	"errors"
 	"fmt"
+	"strconv"
 	"sync"
 
 	"github.com/pressly/goose/v3"
 	"go.uber.org/zap"
+	"gorm.io/datatypes"
 	"gorm.io/gorm"
 	"gorm.io/gorm/logger"
 
@@ -25,9 +30,10 @@ import (
 type Config = gormkitteh.Config
 
 type gormdb struct {
-	z   *zap.Logger
-	db  *gorm.DB
-	cfg *Config
+	z                    *zap.Logger
+	db                   *gorm.DB
+	cfg                  *Config
+	compressionThreshold int
 
 	// See https://github.com/mattn/go-sqlite3/issues/274.
 	// Used only for protecting writes when using sqlite.
@@ -49,6 +55,17 @@ func New(z *zap.Logger, cfg *Config) (db.DB, error) {
 
 	if cfg.Type == "sqlite" {
 		db.mu = new(sync.Mutex)
+	}
+
+	opts := cfg.ParseOptions()
+	db.compressionThreshold = 4096
+	if v, found := opts["compression_threshold"]; found {
+		thres, err := strconv.ParseInt(v, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("compression threshold: %w", err)
+		}
+
+		db.compressionThreshold = int(thres)
 	}
 
 	return db, nil
@@ -242,4 +259,27 @@ func delete[T any](db *gorm.DB, ctx context.Context, where string, args ...any) 
 
 func (db *gormdb) client() *sql.DB {
 	return kittehs.Must1(db.db.DB())
+}
+
+func (db *gormdb) compressJSON(data any) (datatypes.JSON, []byte, error) {
+	j, err := json.Marshal(data)
+	if err != nil {
+		return nil, nil, fmt.Errorf("json: %w", err)
+	}
+
+	if len(j) < db.compressionThreshold {
+		return datatypes.JSON(j), nil, nil
+	}
+
+	var b bytes.Buffer
+	w := gzip.NewWriter(&b)
+	if _, err = w.Write(j); err != nil {
+		return nil, nil, fmt.Errorf("compress: %w", err)
+	}
+
+	if err = w.Close(); err != nil {
+		return nil, nil, fmt.Errorf("compress: %w", err)
+	}
+
+	return nil, b.Bytes(), nil
 }
--- a/internal/backend/db/dbgorm/dbgorm_test.go
+++ b/internal/backend/db/dbgorm/dbgorm_test.go
@@ -1,0 +1,29 @@
+package dbgorm
+
+import (
+	"strings"
+	"testing"
+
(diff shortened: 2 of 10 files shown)


internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -51,6 +57,17 @@ func New(z *zap.Logger, cfg *Config) (db.DB, error) {
...
 
+	opts := cfg.ParseOptions()
+	db.compressionThreshold = 4096
+	if v, found := opts["compression_threshold"]; found {
+		thres, err := strconv.ParseInt(v, 10, 64)
+		if err != nil {
+			return nil, fmt.Errorf("compression threshold: %w", err)
+		}
+
+		db.compressionThreshold = int(thres)
```
tebeka, line 68: Validate that it's a positive number?

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -243,3 +260,26 @@ func delete[T any](db *gorm.DB, ctx context.Context, where string, args ...any)
...
+	if err != nil {
+		return nil, nil, fmt.Errorf("json: %w", err)
+	}
+
+	if len(j) < db.compressionThreshold {
+		return datatypes.JSON(j), nil, nil
+	}
+
+	var b bytes.Buffer
+	w := gzip.NewWriter(&b)
```
tebeka, line 275: Why gzip? Why not bzip2 or another format?

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -243,3 +260,26 @@ func delete[T any](db *gorm.DB, ctx context.Context, where string, args ...any)
 func (db *gormdb) client() *sql.DB {
 	return kittehs.Must1(db.db.DB())
 }
+
+func (db *gormdb) compressJSON(data any) (datatypes.JSON, []byte, error) {
```
tebeka, line 264: I not too happy about the function name. It does not always compress.

internal/backend/db/dbgorm/dbgorm_test.go
```diff
@@ -0,0 +1,29 @@
...
+	}
+
+	bs := []byte(strings.Repeat("meow", 512))
+
+	u, c, err = db.compressJSON(bs)
+	if assert.NoError(t, err) {
+		assert.Nil(t, u)
+		assert.NotNil(t, c)
+		t.Logf("compressed: %d -> %d", len(bs), len(c))
+	}
```
tebeka, line 28: Make sure it's a valid gzip data?

internal/backend/db/dbgorm/events.go
```diff
@@ -22,13 +22,19 @@ func (db *gormdb) SaveEvent(ctx context.Context, event sdktypes.Event) error {
...
-		ConnectionID: scheme.UUIDOrNil(cid.UUIDValue()),
-		EventType:    event.Type(),
-		Data:         kittehs.Must1(json.Marshal(event.Data())),
-		Memo:         kittehs.Must1(json.Marshal(event.Memo())),
-		CreatedAt:    event.CreatedAt(),
+		EventID:        event.ID().UUIDValue(),
+		ConnectionID:   scheme.UUIDOrNil(cid.UUIDValue()),
+		EventType:      event.Type(),
+		Data:           data,
+		CompressedData: cdata,
```
tebeka, line 35: Use the same field for `Data` and have a `Compressed` flag.
This can also changed the signature of `compressJSON` to return data and a flag.

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -49,6 +52,21 @@ var Tables = []any{
 	&Trigger{},
 }
 
+func unmarshalData(dst any, j datatypes.JSON, c []byte) error {
```
tebeka, line 55: Same comment on API. Get data and boolean flag weather it's compressed. In general you can even skip this flag and look at first few bytes of data (magic, which for gzip is `1F8B`).

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -347,25 +366,27 @@ func ParseSessionLogRecord(c SessionLogRecord) (spec sdktypes.SessionLogRecord,
...
-	Seq       uint32        `gorm:"uniqueIndex:idx_session_id_seq_attempt,priority:2"`
-	Attempt   uint32        `gorm:"uniqueIndex:idx_session_id_seq_attempt,priority:3"`
-	Start     datatypes.JSON
-	Complete  datatypes.JSON
+	SessionID          sdktypes.UUID `gorm:"uniqueIndex:idx_session_id_seq_attempt,priority:1;type:uuid;not null"`
+	Seq                uint32        `gorm:"uniqueIndex:idx_session_id_seq_attempt,priority:2"`
+	Attempt            uint32        `gorm:"uniqueIndex:idx_session_id_seq_attempt,priority:3"`
+	Start              datatypes.JSON
+	Complete           datatypes.JSON
+	CompressedComplete []byte
```
tebeka, line 389: Use a bool for weather `Complete` is compressed or not. Data type of `Complete` should be `[]byte`

internal/backend/gormkitteh/config.go
```diff
@@ -25,6 +27,18 @@ type Config struct {
 	// If false, the server will fail to start if a migration is required,
 	// and the user has to run 'ak server migrate' explicitly.
 	AutoMigrate bool `koanf:"auto_migrate"`
+
+	Options string `koanf:"options"`
+}
+
+func (c Config) ParseOptions() map[string]string {
```
tebeka, line 34: What is the reason options are in this format (`port=8080, level=INFO`) and not an established format line JSON?

tebeka (commented):
I don't like the added `Compressed*` columns. I prefer to have a single column and a `compress` column which is boolean.

Also, PostgreSQL supports compression in the database, maybe we can work with it instead?

efiShtain:
I missed this PR and commented on the other PR (https://github.com/autokitteh/autokitteh/pull/336)

## the gist
no need for compressed data, just omit the fields we don't need (return them only when specifically required)

db compress stuff for you automatically, it also split big chunks of data to toast table

this compression add another tax on CPU which is part of the problems we wanted to solve with json unmarshalling

we better solve the root issue (way too big json data for no real reason) instead of adding another layer of complexity (compression)

PR #335, opened 2024-05-27
ENG-913: sessions and events list do not return big data

itayd:


itayd:
looks like atlas doesn't know how to deal with this :(

PR #336, opened 2024-05-27
list does not return data

itayd:


internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -54,6 +54,10 @@ var Tables = []any{
 
 func unmarshalData(dst any, j datatypes.JSON, c []byte) error {
```
tebeka, line 55: Check for `len(c) > 0 && len(j) > 0` and return error.

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -54,6 +54,10 @@ var Tables = []any{
...
 	if len(j) == 0 {
+		if len(c) == 0 {
+			return nil
+		}
+
 		r, err := gzip.NewReader(bytes.NewReader(c))
```
tebeka, line 61: This is error prone, people need to pay attention to the scope of `r`.

Maybe something like:

```go
func unmarshalData(dst any, j datatypes.JSON, c []byte) error {
	if len(j) == 0 && len(c) == 0 {
		return nil
	}

	if len(j) > 0 && len(c) > 0 {
		return fmt.Errorf("got both compressed and uncompressed data")
	}

	var r io.Reader
	if len(c) > 0 {
		var err error
		r, err = gzip.NewReader(c)
		if err != nil {
			return err
		}
	} else {
		r = bytes.NewReader(j)
	}

	return json.NewDecoder(r).Decode(dst)
}
```

internal/backend/db/dbgorm/scheme/scheme.go
tebeka, line 55: You can use generics to make sure `dst` is a pointer.

```go
func unmarshalData[T any](dst *T, j datatypes.JSON, c []byte) error {
```

internal/kittehs/lists.go
```diff
@@ -36,6 +36,18 @@ func ContainedIn[T comparable](xs ...T) func(T) bool {
 	}
 }
 
+func NotContainedIn[T comparable](xs ...T) func(T) bool {
```
tebeka, line 39: Why do we need this?

proto/gen/py/autokitteh_pb/events/v1/svc_pb2_grpc.py
tebeka, line 1: Why do we still generate Python or ts bindings?

internal/backend/db/dbgorm/events.go
```diff
@@ -92,7 +92,7 @@ func (db *gormdb) ListEvents(ctx context.Context, filter sdkservices.ListEventsF
 	}
 
 	var es []scheme.Event
-	if err := q.Find(&es).Error; err != nil {
+	if err := q.Omit("data", "compressed_data").Find(&es).Error; err != nil {
```
efiShtain, line 95: no need for compressed_data 
some of the issues are around CPU cycles for unmarshalling big JSON objects, adding decompression won't solve this issue

also, postgres is performming compression where it can

another also, postgres is probably toasting the fields (if they ar too big, they are stored in a different table automatically, omitting them from select should be enough unless we really need the data)


The best option is to solve the root of the problem: storing simpler smaller json without all the starlark stuff (v, string, etc...)

internal/backend/db/dbgorm/sessions.go
```diff
@@ -176,6 +176,7 @@ func (db *gormdb) listSessions(ctx context.Context, f sdkservices.ListSessionsFi
 	if err := q.
 		Order(clause.OrderByColumn{Column: clause.Column{Name: "created_at"}, Desc: true}).
 		Order(clause.OrderByColumn{Column: clause.Column{Name: "session_id"}, Desc: true}).
+		Omit("inputs", "compressed_inputs").
```
efiShtain, line 179: same as events, no need for compressed data
just get the inputs when really required (even on get you don't really need them unless specifically set)

internal/kittehs/lists.go
efiShtain: I agree, we don't need this

if you insist, probably better to return !ContainedIn which is already implemented and just negates the result (for small arrays, the performance is neglectable and it's more readable and less code to maintain)

sdk/sdkservices/events.go
```diff
@@ -31,6 +31,7 @@ type ListEventRecordsFilter struct {
 type Events interface {
 	Save(ctx context.Context, event sdktypes.Event) (sdktypes.EventID, error)
 	Get(ctx context.Context, eventID sdktypes.EventID) (sdktypes.Event, error)
+	// List returns events without their data.
```
efiShtain, line 34: why do we need this comment ?
who's going to look at it anyway ?

PR #339, opened 2024-05-27
[ENG-892] Allow users to configure which Python to use

tebeka:
Currently works with environment variable. I don't know how the Python runtime can access `ak` configuration other than creating a dummy `python` service.

@itayd Ideas?


runtimes/pythonrt/py.go
```diff
@@ -63,7 +63,27 @@ func parsePyVersion(s string) (major, minor int, err error) {
...
+
+	return !info.IsDir()
+}
+
+const exeEnvKey = "AK_PYTHON"
+
 func findPython() (string, error) {
+	exePath := os.Getenv(exeEnvKey)
+	if exePath != "" {
+		if !isFile(exePath) {
```
efiShtain: any chance it is a file but not executable? 
probably an overkill check but in limited environments (like docker) we might block users from doing stuff (the docker user, not any user) and it might be helpful to know at first hand

runtimes/pythonrt/py.go
```diff
@@ -63,7 +63,27 @@ func parsePyVersion(s string) (major, minor int, err error) {
...
+func isFile(path string) bool {
+	info, err := os.Stat(path)
+	if err != nil {
+		return false
+	}
+
+	return !info.IsDir()
+}
+
+const exeEnvKey = "AK_PYTHON"
```
efiShtain: I'd give it a more hierarchical name like AK__WORKERS__PYTHON_PATH
since this env var is irrelevant to AK itself, it is only relevant to ak's worker and specifically to python worker

efiShtain (approved):
approved with some options for better readability, not a must

runtimes/pythonrt/py.go
```diff
@@ -63,7 +63,27 @@ func parsePyVersion(s string) (major, minor int, err error) {
...
+
 func findPython() (string, error) {
+	exePath := os.Getenv(exeEnvKey)
+	if exePath != "" {
+		if !isFile(exePath) {
+			return "", fmt.Errorf("%q (AK_PYTHON) is not a file", exePath)
```
daabr:
```suggestion
			return "", fmt.Errorf("%q (%s) is not a file", exePath, exeEnvKey)
```

runtimes/pythonrt/py.go
```diff
@@ -63,7 +63,27 @@ func parsePyVersion(s string) (major, minor int, err error) {
...
+		return false
+	}
+
+	return !info.IsDir()
+}
+
+const exeEnvKey = "AK_PYTHON"
+
 func findPython() (string, error) {
+	exePath := os.Getenv(exeEnvKey)
```
daabr: 1. I can't comment on the exact line because it's not new/changed, but there's a typo in the error at the bottom of this function:

   `"non of %v found in PATH"` --> `none of...`

2. Please document this function, unlike `isFile` it's not obvious what it should return

runtimes/pythonrt/pythonrt.go
```diff
@@ -97,13 +84,26 @@ func dirExists(path string) bool {
 	return info.IsDir()
 }
 
-func ensureVEnv(log *zap.Logger, pyExe string) error {
+func ensureVEnv(log *zap.Logger) error {
 	if dirExists(venvPath) {
 		return nil
 	}
 
+	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
```
daabr: Please use a const for the timeout, and put it at the top of the file

daabr (approved):
Approved with nits

runtimes/pythonrt/py.go
tebeka: When we detect the python we call `python --version` to get the version, it'll fail there.

runtimes/pythonrt/py.go
tebeka: Changed to `AK_WORKER_PYTHON` (why the double `_`?)

efiShtain: We have a convention we use __ to split env hierarchy 
This is for koanf to parse configurations correctly

I know you don't use koanf here, but we should probably keep the consistency

runtimes/pythonrt/py.go
tebeka: Done.

runtimes/pythonrt/pythonrt.go
tebeka: Done.

tebeka:
@daabr , @efiShtain I change the logic according to discussion in ENG-892. Please re-review.

PR #342, opened 2024-05-28
fix: cross project trigger collision

ig-ra:
trigger filter list was lack of projectID


internal/backend/triggersgrpcsvc/svc.go
```diff
@@ -126,9 +126,15 @@ func (s *server) List(ctx context.Context, req *connect.Request[triggersv1.ListR
...
 
+	pid, err := sdktypes.ParseProjectID(msg.ProjectId)
+	if err != nil {
+		return nil, sdkerrors.AsConnectError(err)
+	}
+
 	filter := sdkservices.ListTriggersFilter{
 		EnvID:        eid,
 		ConnectionID: cid,
+		ProjectID:    pid,
```
efiShtain, line 137: why do you need projectID if you have envID ?

ig-ra: this was a bug, mate. We have filter with projectID, but we are not passing projectID

ig-ra: so when planning the manifest, we were deleting triggers for other projects.
Why? because we failed to list triggers for THIS project to plan

efiShtain: so it's additional option to filter  ?

efiShtain (approved):

PR #343, opened 2024-05-28
fix: cant delete connections due to fk constrain

ig-ra:
event is dependent on connection. Since event is soft-deleted, then connection should be soft-deleted as well

### DIFF
--- a/internal/backend/db/dbgorm/scheme/scheme.go
+++ b/internal/backend/db/dbgorm/scheme/scheme.go
@@ -76,6 +76,8 @@ type Connection struct {
 	StatusCode    int32 `gorm:"index"`
 	StatusMessage string
 
+	DeletedAt gorm.DeletedAt `gorm:"index"`
+
 	// enforce foreign keys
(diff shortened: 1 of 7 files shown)


internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -76,6 +76,8 @@ type Connection struct {
 	StatusCode    int32 `gorm:"index"`
 	StatusMessage string
 
+	DeletedAt gorm.DeletedAt `gorm:"index"`
```
efiShtain, line 79: Don't you need to update queries accordingly to not fetch deleted?

ig-ra: gorm is handling this

ig-ra: there is this fixture - assertDeleted, which tests exactly this - tests that find query will skip deleted objects and raw db access will still find them

efiShtain (approved):
Approved

Please don't mix efforts in future PRs

PR #347, opened 2024-05-30
ENG-894: load connection variables into env

itayd:
Each connection's variables (which also include secrets) are loaded into env with key "{conn_name}__{var_key}". This should allow python code to load them and reuse them when initializing libraries.

### DIFF
--- a/internal/backend/sessions/sessionworkflows/workflow.go
+++ b/internal/backend/sessions/sessionworkflows/workflow.go
@@ -295,7 +295,7 @@ func (w *sessionWorkflow) createEventSubscription(ctx context.Context, connectio
 		return "", fmt.Errorf("get current sequence: %w", err)
 	}
 
-	signalID := uuid.New().String() // fmt.Sprintf("wid_%s_cn_%s_seq_%d", workflowID, connectionName, minSequence)
+	signalID := uuid.New().String()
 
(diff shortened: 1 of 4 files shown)


ig-ra (approved):

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -273,7 +295,7 @@ func (w *sessionWorkflow) createEventSubscription(ctx context.Context, connectio
 		return "", fmt.Errorf("get current sequence: %w", err)
 	}
 
-	signalID := uuid.New().String() //fmt.Sprintf("wid_%s_cn_%s_seq_%d", workflowID, connectionName, minSequence)
+	signalID := uuid.New().String() // fmt.Sprintf("wid_%s_cn_%s_seq_%d", workflowID, connectionName, minSequence)
```
tebeka: Delete commented code.

tebeka (approved):
Approving. I need to check with Python.

Question: Say a connection has a oauth token, what will be it's name.

internal/backend/sessions/sessionworkflows/workflow.go
itayd: done

PR #348, opened 2024-05-30
WIP: ENG-865: automatic retries

itayd:
This adds the ability to specify retry policies as follows:
```
def retry_policy(err, attempt):
  return "1s" if attempt < 3 else None

some_connection.some_method(..., ak_retry=retry_policy)
```

See session tests for more information.

This is marked as WIP as there is no docs yet, and I'd like to discuss this in sync beforehand to settle on exact usage.

### DIFF
--- a/internal/backend/akmodules/ak/module.go
+++ b/internal/backend/akmodules/ak/module.go
@@ -25,19 +25,23 @@ func New(syscall sdkexecutor.Function) sdkexecutor.Executor {
 			"syscall",
 			syscall,
 			sdkmodule.WithFlags(
-				sdktypes.PureFunctionFlag,           // no need to run in an activity, we must have it in the workflow.
-				sdktypes.PrivilidgedFunctionFlag,    // provide workflow context.
-				sdktypes.DisablePollingFunctionFlag, // no polling.
+				sdktypes.PureFunctionFlag,            // no need to run in an activity, we must have it in the workflow.
+				sdktypes.PrivilidgedFunctionFlag,     // provide workflow context.
+				sdktypes.DisablePollingFunctionFlag,  // no polling.
+				sdktypes.DisableRetryingFunctionFlag, // no retrying.
 			),
 		),
 		sdkmodule.ExportFunction(
 			"callopts",
 			callopts,
 			sdkmodule.WithFlags(
-				sdktypes.PureFunctionFlag,           // no need to run in an activity, we must have it in the workflow.
-				sdktypes.DisablePollingFunctionFlag, // no polling.
+				sdktypes.PureFunctionFlag,            // no need to run in an activity, we must have it in the workflow.
+				sdktypes.DisablePollingFunctionFlag,  // no polling.
+				sdktypes.DisableRetryingFunctionFlag, // no retrying.
 			),
 		),
+		// Functions useful for testing.
+		sdkmodule.ExportFunction("meow", meow),
 	)
 }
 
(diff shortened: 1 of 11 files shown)


haimzlato:
The user needs to define a function for retries? I don't like it, room for too many errors. How can we check for errors in compilation? 
I would prefer some kind of structure.

What happens in Python?

efiShtain:
> The user needs to define a function for retries? I don't like it, room for too many errors. How can we check for errors in compilation? I would prefer some kind of structure.
> 
> What happens in Python?

WDYM ?
there should be a default retry policy, implicit
and if the user wants a different policy then the user has to define it somehow
probably in code

internal/backend/akmodules/ak/module.go
```diff
@@ -48,3 +52,17 @@ func callopts(_ context.Context, args []sdktypes.Value, kwargs map[string]sdktyp
 
 	return sdktypes.NewStructValue(sdktypes.NewSymbolValue(CallOptsCtorSymbol), kwargs)
 }
+
+func meow(_ context.Context, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
```
tebeka, line 56: Why is this function required?

internal/backend/akmodules/time/time.go
```diff
@@ -17,27 +17,30 @@ import (
 
 var ExecutorID = sdktypes.NewExecutorID(fixtures.NewBuiltinIntegrationID("time"))
 
-func New() sdkexecutor.Executor {
+type t0Type time.Time
```
tebeka, line 20: Why is it named `t0Type`?

internal/backend/sessions/sessioncalls/call.go
```diff
@@ -17,38 +17,48 @@ import (
...
 
 	// Duration: retry with specified interval.
-	if interval, err = pollRet.ToDuration(); err != nil {
-		err = fmt.Errorf("poller return value must be either a boolean or convertible to duration: %w", err)
+	if interval, err = ret.ToDuration(); err != nil {
+		err = fmt.Errorf("policy return value must be either nothing, a boolean, or convertible to duration: %w", err)
 		return
 	}
 
 	retry = true
```
tebeka, line 50: I don't like named return values used like that and prefer explicit `return true, interval`

internal/backend/sessions/sessioncalls/call.go
```diff
@@ -17,38 +17,48 @@ import (
...
 		return
 	}
 
-	if isBool := pollRet.IsBoolean(); isBool {
-		// True: no retry. False: immediate retry.
-		return !pollRet.GetBoolean().Value(), 0, nil
+	if ret.IsBoolean() {
+		// True: immediate retry. False: no retry.
+		// TODO: invert for poll?
+		retry = ret.GetBoolean().Value()
```
tebeka, line 40: We need system level limit on number of retries, this can blow up the system.

internal/backend/sessions/sessioncalls/call.go
```diff
@@ -181,8 +207,21 @@ func (cs *calls) executeCall(ctx context.Context, sessionID sdktypes.SessionID,
 		}()
 
 		if err := result.GetError(); err != nil {
-			last = true
+			last = !retryer.IsValid()
 			z.Debug("call returned an error", zap.Error(err))
+
+			if !last {
```
tebeka, line 213: I find this logic confusing.

tests/sessions/retry.txtar
```diff
@@ -0,0 +1,47 @@
...
+10s
+
+-- main.star:main --
+def immediate_policy(err, attempt):
+    print(attempt, err)
+    return attempt < 2
+
+def delayed_policy(err, attempt):
+    print(attempt, err)
+    return "1s" if attempt < 2 else None
```
tebeka, line 25: I prefer that instead of a string you'll return a duration (or timedelta as it's called in Python).

tebeka (commented):
I don't like it that we add magic arguments to function calls for retry polity.
IMO the retry policy should be in the manifest or other configuration and not in the code.
In Python this can be done with a decorator, and there are many already written our there.

tebeka:
@itayd Can we close this PR or do we still talk about it?

itayd:
this is WIP, need to modify this according to what we talked about.

PR #349, opened 2024-05-30
feat(env)!: non-optional project

ig-ra:
disallowing project to be nil in Env

BREAKING CHANGE: DBs with standalone ENVs (not connected to any project) cannot be migrated and should be reconnected to some existing project or deleted

### DIFF
--- a/internal/backend/db/dbgorm/deployments_test.go
+++ b/internal/backend/db/dbgorm/deployments_test.go
@@ -82,8 +82,11 @@ func TestCreateDeploymentsForeignKeys(t *testing.T) {
 	d.EnvID = nil
 
 	// test with existing assets
+	p := f.newProject()
 	e := f.newEnv()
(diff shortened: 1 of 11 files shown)


efiShtain (commented):
Did you check what happens on a database which has null env?
Migration would probably fail
What is the proposed solution for that?

ig-ra:
> Did you check what happens on a database which has null env? Migration would probably fail What is the proposed solution for that?

you mean default env? Default env is created while the project creation.
But there could be some old db with standalone envs. I think we should declare this change as a breaking one

efiShtain (approved):
Approve
note this is a breaking change

ig-ra:
in create env CMD it's already defined as required flag

itayd:
we currently don't have "dangling envs", but we might in the future where their parents is either an org or a user.

PR #351, opened 2024-06-02
fix: resolve github vars from secrets

ig-ra:
vars (e.g. PAT) were unresolved

### DIFF
--- a/integrations/github/jwt.go
+++ b/integrations/github/jwt.go
@@ -34,7 +34,12 @@ const (
 )
 
 func (i integration) NewClient(ctx context.Context, user string) (*github.Client, error) {
-	data, err := i.getConnection(ctx)
+	// Extract the connection token from the given context.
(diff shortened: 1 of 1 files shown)


efiShtain (approved):
I guess you checked it and it works now

integrations/github/jwt.go
```diff
@@ -34,7 +34,12 @@ const (
 )
 
 func (i integration) NewClient(ctx context.Context, user string) (*github.Client, error) {
-	data, err := i.getConnection(ctx)
+	// Extract the connection token from the given context.
+	cid, err := sdkmodule.FunctionConnectionIDFromContext(ctx)
+	if err != nil {
+		return nil, err
+	}
+	data, err := i.vars.Reveal(ctx, sdktypes.NewVarScopeID(cid))
```
daabr, line 42: You're replacing a function call to `getConnection` with it's (fixed) code - which is causing a lint error. Either fix the function or remove it.

daabr (approved):

PR #353, opened 2024-06-03
ENG-960: fix count only sessions not returning any results

efiShtain:
count only sessions returned no results due to incorrect error check


ig-ra (approved):

internal/backend/db/dbgorm/sessions_test.go
```diff
@@ -25,7 +25,7 @@ func (f *dbFixture) addSessionLogRecordAndAssert(t *testing.T, logr scheme.Sessi
 	findAndAssertCount[scheme.SessionLogRecord](t, f, expected, "session_id = ?", logr.SessionID)
 }
 
-func (f *dbFixture) listSessionsAndAssert(t *testing.T, expected int) []scheme.Session {
+func (f *dbFixture) listSessionsAndAssert(t *testing.T, expected int64) []scheme.Session {
```
ig-ra, line 28: please add more tests
- with count only
- list session when there is no sessions + count only for the same case
- count in case of pagination (could u define page size = 1? and create 2 sessions?)

efiShtain: Done
PTAL

ig-ra (approved):

PR #354, opened 2024-06-03
fix (sdk): unwrapping embedded time.Time

ig-ra:
embedded anonymous time.Time struct isn't wrapped.
This prevents us to work with github Timestamps

### DIFF
--- a/sdk/sdktypes/value_wrap.go
+++ b/sdk/sdktypes/value_wrap.go
@@ -84,6 +84,7 @@ func (w ValueWrapper) Wrap(v any) (Value, error) {
 
 		for _, vfs := range reflect.VisibleFields(vt) {
 			if !vfs.IsExported() || vfs.Anonymous {
+				// allow to wrap embedded unexported time.Time field
 				if vfs.Type != reflect.TypeOf(time.Time{}) {
 					continue
 				}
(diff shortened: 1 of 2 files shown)


sdk/sdktypes/value_wrap.go
```diff
@@ -84,7 +84,9 @@ func (w ValueWrapper) Wrap(v any) (Value, error) {
 
 		for _, vfs := range reflect.VisibleFields(vt) {
 			if !vfs.IsExported() || vfs.Anonymous {
-				continue
+				if vfs.Type != reflect.TypeOf(time.Time{}) {
```
efiShtain, line 88: could you explain this one ?
not sure I get this if condition

sdk/sdktypes/value_wrap_test.go
```diff
@@ -361,6 +361,10 @@ func TestUnwrapIntoKitchenSink(t *testing.T) {
 		Z string
 	}
 
+	type Timestamp struct {
```
efiShtain: fix static analysis error

sdk/sdktypes/value_wrap.go
ig-ra: github's Timestamp is 
```
type Timestamp struct {
   time.Time
}
```
when we are receiving answers from github and converting them to starlark values we are passing through wrap function. When we arrive to wrapping Timestamp we will skip it, since there is no exported or non-anonymous types. GH Timestamp type is embedding time.Time, thus all it's fields are not-exported and anonymous.
So in starlark all GH Timestamps will be empty as a result. With this fix it's working good. You could see the relevant sample in samples repo

efiShtain (approved):
Please add a comment explaining why you did it
what to do in case we have another type like this in the future

PR #356, opened 2024-06-04
fix: wrap only exported time

ig-ra:
change condition and warp only exported anonymous time.Time

### DIFF
--- a/sdk/sdktypes/value_wrap.go
+++ b/sdk/sdktypes/value_wrap.go
@@ -83,11 +83,13 @@ func (w ValueWrapper) Wrap(v any) (Value, error) {
 		fs := make(map[string]Value)
 
 		for _, vfs := range reflect.VisibleFields(vt) {
-			if !vfs.IsExported() || vfs.Anonymous {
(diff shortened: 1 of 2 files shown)


sdk/sdktypes/value_wrap.go
```diff
@@ -83,11 +83,13 @@ func (w ValueWrapper) Wrap(v any) (Value, error) {
...
-			if !vfs.IsExported() || vfs.Anonymous {
-				// allow to wrap embedded unexported time.Time field
-				if vfs.Type != reflect.TypeOf(time.Time{}) {
-					continue
-				}
+			if !vfs.IsExported() {
+				continue
+			}
+
+			if vfs.Anonymous && vfs.Type != reflect.TypeOf(time.Time{}) {
```
tebeka, line 90: This seems very specific, can you document why we look only at `time.Time`?

tebeka (approved):

sdk/sdktypes/value_wrap.go
ig-ra: added a note

PR #358, opened 2024-06-06
[ENG-978] Install autokitteh to venv instead of injecting it

tebeka:


runtimes/pythonrt/requirements.txt
```diff
@@ -12,6 +12,10 @@ redis ~= 5.0
 slack-sdk ~= 3.27
 twilio ~= 9.0
 
+# AutoKitteh SDK
+
+autokitteh ~= 0.1
```
daabr, line 17: Given that we own this, doesn't it make sense to always get the latest version instead of pinning?

daabr (approved):
Yay! See my question about AK package version pinning.

runtimes/pythonrt/requirements.txt
tebeka: I want to give users the option when to upgrade. If we make breaking change to `autokitteh` and they still want to run on older version - it's a problem.

PR #361, opened 2024-06-07
decouple cvars from env

itayd:
Connection Vars should eventually be an implementation details of the integration. This PR takes a step toward that goal by decoupling the env module vars supplied to scripts from the actual connection vars by adding a function to integrations to returns a configuration, instead of reading directly from the connection vars.


internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -83,15 +88,17 @@ func runWorkflow(
 		signals: make(map[string]uint64),
 	}
 
-	if err = w.initEnvModule(ctx); err != nil {
+	var cinfos map[string]connInfo
```
tebeka, line 91: Map is not initialised, is this intentional?

sdk/sdkservices/integrations.go
```diff
@@ -29,11 +29,13 @@ type Integration interface {
 	// external service that it wraps.
 	Get() sdktypes.Integration
 
-	Configure(ctx context.Context, cid sdktypes.ConnectionID) (map[string]sdktypes.Value, error)
+	Configure(ctx context.Context, cid sdktypes.ConnectionID) (map[string]sdktypes.Value, map[string]string, error)
```
tebeka: Document output parameters.

tebeka (approved):
A lot of changes, but I don't see that many tests for it.

internal/backend/sessions/sessionworkflows/workflow.go
itayd: yup, don't want to overwrite err in the next line

sdk/sdkservices/integrations.go
itayd: done

itayd:
> A lot of changes, but I don't see that many tests for it.

Actually, current tests cover it. This is a refactoring. (and they actually failed! and i fixed it!)

daabr (changes requested):
My concern is that optional function arguments that are actually required (based on subtle and undocumented semantics) bit me hard in the past, in the exact same place actually.

When people write integrations, the **won't** know to add this argument, and the results would be... Hard to debug?

Any way you can improve this experience for integration developers?

Marking this as "request change" just to make sure we discuss this before you merge.

daabr (approved):
Discussed offline

PR #362, opened 2024-06-08
ENG-965: new Jira integration

daabr:
See also stacked PR: #366


integrations/jira/client.go
```diff
@@ -0,0 +1,83 @@
...
+	i := &integration{vars: cvars}
+	return sdkintegrations.NewIntegration(
+		desc,
+		sdkmodule.New( /* No exported functions for Starlark */ ),
+		connStatus(i),
+		connTest(i),
+	)
+}
+
+func connStatus(i *integration) sdkintegrations.OptFn {
```
itayd, line 49: you can just do `func (i *integrtation)`, and supply it as a parameter above to NewIntegration, I think it's nicer.

integrations/jira/client.go
```diff
@@ -0,0 +1,83 @@
...
+			return initReq, nil
+		}
+
+		vs, err := i.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		// TODO(ENG-965): Implement a real check, and reuse in the OAuth handler.
+		// if vs.Has(vars.PAT) {
```
itayd: what's wrong with the commented code?

integrations/jira/client.go
```diff
@@ -0,0 +1,83 @@
...
+		// 	return sdktypes.NewStatus(sdktypes.StatusCodeOK, "using PAT"), nil
+		// }
+
+		n := len(kittehs.Filter(vs, func(v sdktypes.Var) bool {
```
itayd, line 66: you can use kittehs.ContainedIn

integrations/jira/client.go
```diff
@@ -0,0 +1,83 @@
...
+			return strings.HasPrefix(v.Name().String(), "app_id__")
+		}))
+
+		if n == 0 {
+			return initReq, nil
+		}
+
+		return sdktypes.NewStatus(sdktypes.StatusCodeOK, fmt.Sprintf("%d installations", n)), nil
```
itayd: you can have multiple installations per connector?

integrations/jira/client.go
```diff
@@ -0,0 +1,83 @@
...
+	})
+}
+
+func connTest(*integration) sdkintegrations.OptFn {
```
itayd: i would just drop this for now if it's not doing anything.

integrations/jira/oauth.go
```diff
@@ -0,0 +1,68 @@
...
+// Either way, it redirects the user to success or failure webpages.
+func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
+	l := h.logger.With(zap.String("urlPath", r.URL.Path))
+
+	// Handle errors (e.g. the user didn't authorize us) based on:
+	// https://developers.google.com/identity/protocols/oauth2/web-server#handlingresponse
+	e := r.FormValue("error")
+	if e != "" {
+		l.Warn("OAuth redirect request reported an error", zap.Error(errors.New(e)))
+		u := fmt.Sprintf("%serror.html?error=%s", uiPath, url.QueryEscape(e))
```
itayd: just return 400. don't think we need al this fancy error display.

integrations/jira/oauth.go
```diff
@@ -0,0 +1,68 @@
...
+	oauthToken := data.Token
+	if oauthToken == nil {
+		l.Warn("OAuth redirect request without token in data parameter", zap.Error(err))
+		u := uiPath + "error.html?error=" + url.QueryEscape("missing OAuth token")
+		http.Redirect(w, r, u, http.StatusFound)
+		return
+	}
+
+	// TODO(ENG-965):
+	// Create a webhook to receive, parse, and dispatch Jira events,
```
itayd, line 59: this is the next pr?

integrations/jira/client.go
```diff
@@ -0,0 +1,83 @@
...
+var integrationID = sdktypes.NewIntegrationIDFromName("jira")
+
+var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.IntegrationPB{
+	IntegrationId: integrationID.String(),
+	UniqueName:    "jira",
+	DisplayName:   "Atlassian Jira",
+	Description:   "Atlassian Jira is an issue tracking and project management system.",
+	LogoUrl:       "/static/images/jira.svg",
+	UserLinks: map[string]string{
+		"1 REST API":          "https://developer.atlassian.com/cloud/jira/platform/rest/v3/intro/",
```
tebeka, line 26: Why do we need the numbers (the `1`  before `REST API`)?

integrations/jira/client.go
tebeka: -0

integrations/jira/client.go
tebeka: Comment what this code does.

integrations/jira/oauth.go
tebeka: The error did help us when debugging oauth issues with github.

integrations/jira/oauth.go
```diff
@@ -0,0 +1,68 @@
...
+	if err != nil {
+		l.Warn("OAuth redirect request with invalid data parameter", zap.Error(err))
+		u := uiPath + "error.html?error=" + url.QueryEscape("invalid data parameter")
+		http.Redirect(w, r, u, http.StatusFound)
+		return
+	}
+
+	oauthToken := data.Token
+	if oauthToken == nil {
+		l.Warn("OAuth redirect request without token in data parameter", zap.Error(err))
```
tebeka: Create a utility function to redirect to specific error. This code is duplicated 3 times.

integrations/jira/server.go
```diff
@@ -0,0 +1,29 @@
...
+	"go.uber.org/zap"
+
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/web/static"
+)
+
+const (
+	// uiPath is the URL root path of a simple web UI to interact with users at
+	// the beginning and the end of their 3-legged OAuth 2.0 flow with Atlassian.
+	uiPath = "/jira/connect/"
```
tebeka: Can't you use the URL from the description (`ConnectionUrl`)?

internal/backend/integrations/integrations.go
```diff
@@ -59,6 +61,7 @@ func Start(_ context.Context, l *zap.Logger, mux *http.ServeMux, vars sdkservice
 	github.Start(l, mux, vars, o, d)
 	google.Start(l, mux, o, d)
 	httpint.Start(l, mux, d, c, p)
+	jira.Start(l, mux, vars, o, d)
```
tebeka, line 64: Nit. We start web server per integration, even if there's no connection using it. Can we make it lazy?

internal/backend/oauth/oauth.go
```diff
@@ -49,6 +49,19 @@ func New(l *zap.Logger) sdkservices.OAuth {
 		)
 	}
 
+	// TODO(ENG-965): From new-connection form instead of env var.
+	jiraBaseURL := os.Getenv("JIRA_BASE_URL")
```
tebeka, line 54: Configuration?

internal/backend/oauth/oauth.go
```diff
@@ -49,6 +49,19 @@ func New(l *zap.Logger) sdkservices.OAuth {
...
+	if jiraBaseURL == "" {
+		jiraBaseURL = "https://auth.atlassian.com"
+	}
```
tebeka, line 57: Log here which URL we're using, it'll help diagnose issues.

web/static/images/discord.svg
tebeka, line 1: How is this related to the PR?

web/static/jira/connect/styles.css
tebeka, line 1: Do we have an issue to unite all HTML/CSS in integrations?

integrations/jira/client.go
daabr: It's a soft convention for sorting these links in (future) auto-generated reference documentation (these numbers are stripped in the generated content).

integrations/jira/client.go
daabr: I just copied it from an existing integration (GitHub), but anyway I try to avoid large anonymous functions embedded inside other functions. It increases code complexity without benefits.

integrations/jira/client.go
daabr: API keys and PATs are not implemented yet, so this var doesn't exist - but anyway I fixed the function based on the next comment.

integrations/jira/client.go
daabr: Copied from GitHub (where this question is also relevant, but for a different PR). Anyway, fixed.

integrations/jira/client.go
daabr: Simplified the entire function based on the next comment, so added the comment above the function.

integrations/jira/client.go
daabr: No problem, but I'll add it back in #366

integrations/jira/oauth.go
daabr: Good idea. Done.

integrations/jira/oauth.go
daabr: Yes, #366

integrations/jira/server.go
daabr: That's a recent addition, but yes. Good idea.

internal/backend/integrations/integrations.go
daabr: Not really - consider this scenario:

1. User configures a webhook to send AK events
2. User doesn't configure the AK connection yet, or has temporarily deleted the project
3. Service sends events to webhook

Even though the event will be dispatched to 0 AK connections, we still want to appear responsive to the 3rd-party service.

internal/backend/oauth/oauth.go
daabr: Even better - this shouldn't have to be a server-wide detail, I think it's more convenient as a connection setup detail. Either way, the usage of an env var is very temporary. As soon as I unblock Pasha, I'll improve this.

internal/backend/oauth/oauth.go
daabr: Done.

web/static/images/discord.svg
daabr: Ugh, sorry. But it's more annoying to remove it at this point and just keep it here. After all, we will have a Discord integration no matter what (to relay trigger events).

web/static/jira/connect/styles.css
daabr: Yes, they look similar but they're not, and I'm afraid my frontend engineering skills are not enough to merge all of them together correctly and efficiently. :(

integrations/jira/oauth.go
itayd: i mean you still show it in the error content, less dancing around it

itayd (approved):

PR #365, opened 2024-06-10
github: copilot api

itayd:


daabr (commented):
Thanks for picking up this task!

1. Please add documentation links in `ExportFunction` and as comments for the impl functions, so users and maintainers know what they're doing
2. Please avoid "rm" as a shortcut for "remove" - we want to align with the official APIs as closely as possible - in both function and arg names

itayd:
> 1. Please add documentation links in `ExportFunction` and as comments for the impl functions, so users and maintainers know what they're doing

done

> 2. Please avoid "rm" as a shortcut for "remove" - we want to align with the official APIs as closely as possible - in both function and arg names

done

daabr (approved):

PR #367, opened 2024-06-11
[ENG-937] Split ak_runner.py to several files

tebeka:


runtimes/pythonrt/ak_runner/attrdict.py
```diff
@@ -0,0 +1,23 @@
+class AttrDict(dict):
```
daabr, line 1: Any chance to make this class available also in the autokitteh pip package?

runtimes/pythonrt/py_test.go
```diff
@@ -205,3 +205,32 @@ func Test_pyExports(t *testing.T) {
 	require.Equal(t, expected, exports)
 
 }
+
+const testRunnerPath = "/tmp/zzz/ak_runner"
+
+var adjustCases = []struct {
```
daabr, line 211: Table-based tests are a Go idiom rather than a Python one.

Also, I haven't seen `struct` being used for this purpose, usually I'd consider a `namedtupled` or a `dataclass`.

But other than this looking weird, it work, so no action item I guess.

runtimes/pythonrt/tests/comm_test.py
```diff
@@ -0,0 +1,52 @@
+import json
```
daabr, line 1: I'm seeing in this file (or others) some tiny format issues that a formatter like Black would automatically fix (i.e. extra empty lines). Are you using Black, or a different formatter, and is it configure to auto-format? Let's settle on one and require it for all our Python code.

runtimes/pythonrt/ak_runner/call.py
```diff
@@ -0,0 +1,80 @@
...
+            message = self.comm.recv(MessageType.callback, MessageType.response)
+            
+            if message['type'] == MessageType.callback:
+                payload = self.comm.extract_activity(message)
+                fn, args, kw = payload['data']
+                if isinstance(fn, str):
+                    fn = getattr(self.module, fn, None)
+                    if fn is None:
+                        mod_name = self.module.__name__
+                        raise ValueError(f'function {fn} not found in {mod_name}')
```
daabr: Maybe `!r` to quote the names? Up to you.

runtimes/pythonrt/ak_runner/comm.py
```diff
@@ -0,0 +1,109 @@
...
+    done = 'done'
+    log = 'log'
+    module = 'module'
+    response = 'response'
+    run = 'run'
+    sleep = 'sleep'
+    
+
+class Comm:
+    """Comm does communication with ak server, JSON lines over socket."""
```
daabr: 1. I'm not strict on docstrings for classes, but this class needs a more thorough explanation for someone who doesn't know it in advance - the code is simple, but the intent and usage aren't
2. "JSON lines over socket" is vague: not specific, and not explaining anything
3. "Does communication" = "communicates"

runtimes/pythonrt/ak_runner/comm.py
```diff
@@ -0,0 +1,109 @@
...
+        self.rdr = sock.makefile('r')
+
+    def _send(self, message):
+        data = json.dumps(message) + '\n'
+        self.sock.sendall(data.encode('utf-8'))
+
+    def recv(self, *msg_types):
+        data = self.rdr.readline()
+        if not data:
+            raise ValueError('connection closed')
```
daabr: Would it be helpful to be more specific about which connection is closed, in case a user sees this error?

runtimes/pythonrt/py.go
```diff
@@ -242,14 +220,27 @@ func writeTar(rootDir string, data []byte) (string, error) {
 	return tarName, err
 }
 
-func overrideEnv(envMap map[string]string) []string {
+func adjustPYTHONPATH(env []string, runnerPath string) []string {
```
daabr: Nit: `PYTHONPATH` is indeed the name of the env var you're adjusting here, but I think `adjustPythonPath` is still a clearer function name. Up to you.

daabr (commented):
Good PR. I left some comments, but the reason I'm not approving is because I want to decide on auto-formatting (see comment in [runtimes/pythonrt/tests/comm_test.py](https://github.com/autokitteh/autokitteh/pull/367/files/d32f54c251583baf336fb6a7a23367978cb2e96c#diff-1de9e3ea0844b7fcd3ff1202ed8ed1d2236079b9b4df2c3233aa319639a7cc66)). Let me know if you want to do it after merging and not before.

daabr (approved):
Punting on code formatting, filed tracking ticket: https://linear.app/autokitteh/issue/ENG-1021

runtimes/pythonrt/ak_runner/attrdict.py
tebeka: Yes. It's in https://github.com/autokitteh/autokitteh/blob/miki/events/runtimes/pythonrt/py-sdk/autokitteh.py which still at work. If it's urgent I'll make a new PR since I don't know how much time it'll take me to debug the event issues I have.

runtimes/pythonrt/py_test.go
tebeka: This is Go code.

runtimes/pythonrt/tests/comm_test.py
tebeka: We'll take it offline. There's also ENG-860

runtimes/pythonrt/ak_runner/call.py
tebeka: Done.

runtimes/pythonrt/ak_runner/comm.py
tebeka: Enhanced the docs.

runtimes/pythonrt/py.go
tebeka: Done.

runtimes/pythonrt/ak_runner/comm.py
tebeka: Changed to `connection from autokitteh closed`.

PR #368, opened 2024-06-11
ENG-997: Fix trigger CLI commands

daabr:
* Added optional `--project` flag, and made `--env` optional (at least one of them must be specified)
* Added missing `--data` flag (for HTTP URL paths)
* Improved `--event` flag description a bit

Also, important fixes, though not directly requested in ENG-997:

* Fixed a bug in the connection resolver, uncovered by the addition of `--project`
* Using Cobra's framework for flag checking, not custom `if` checks
* `Use` line way too long, split into two lines:

```
Create event trigger

Usage:
  ak trigger create -n name --call file:func [-p project] [-e env] -c connection ...
             create -n name --call file:func [-p project] [-e env] -s "schedule"

Flags:
```

* Renamed `--loc` flag to `--call` for consistency with manifest file, improved description


itayd (approved):
i don't see test for data?

itayd (approved):

PR #369, opened 2024-06-12
add pprof support

itayd:
exposed through port 6060

### DIFF
--- a/internal/backend/svc/svc.go
+++ b/internal/backend/svc/svc.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	_ "net/http/pprof"
 	"os"
 	"strings"
 	"sync/atomic"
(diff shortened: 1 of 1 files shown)


internal/backend/svc/svc.go
```diff
@@ -359,6 +360,13 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 				z.Info("ready", zap.String("version", version.Version), zap.String("id", fixtures.ProcessID()))
 			})
 		}),
+		fx.Invoke(func(z *zap.Logger) {
+			go func() {
+				if err := http.ListenAndServe("localhost:6060", nil); err != nil {
```
daabr: I know 6060 is standard, but isn't it likely that it'll be bound sometimes? Maybe make the port number configurable? Even if it's free, some people may not appreciate that we're hogging it without asking.

itayd: made configurable

daabr (approved):

PR #370, opened 2024-06-12
add current time in utc to dashboard footer

itayd:


daabr (approved):
Approved, but isn't it better to convert all times in the UI to local time? I don't think our users don't need cross-timezone sync/collab.

itayd:
this is for the tech dashboard, i'm not going to invest in that rn :)

PR #372, opened 2024-06-12
github: wrap timestamp as time

itayd:
github lib returns a weird timestamp value. we need to convert it to a more useful timestamp. this introduces a Prewrap function for value wrapper to do that.

### DIFF
--- a/integrations/github/values.go
+++ b/integrations/github/values.go
@@ -1,0 +1,18 @@
+package github
+
+import (
+	"github.com/google/go-github/v60/github"
+
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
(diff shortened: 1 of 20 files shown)


integrations/github/values.go
```diff
@@ -0,0 +1,18 @@
...
+
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var valueWrapper = sdktypes.ValueWrapper{
+	Prewrap: func(v any) (any, error) {
+		switch v := v.(type) {
+		case github.Timestamp:
+			return v.Time, nil
+		}
```
daabr, line 14:
```suggestion
		if v, ok := v.(github.Timestamp); ok {
			return v.Time, nil
		}
```

daabr (approved):

PR #373, opened 2024-06-12
bugfix: log users in http request

ig-ra:
in httpscv and interceptor func, users are not in the request context.
Auth middleware connected after the interceptor, therefore parsed user isn't present in request context before `httpServe`.
After `httpServe` user still won't be present in request context, since request will be override in auth (due to new context).

### DIFF
--- a/internal/backend/auth/authhttpmiddleware/authmiddleware.go
+++ b/internal/backend/auth/authhttpmiddleware/authmiddleware.go
@@ -108,18 +108,31 @@ func New(deps Deps) AuthMiddlewareDecorator {
 
 type AuthHeaderExtractor func(*http.Request) sdktypes.User
 
-// just for logs in interceptor
+// Auth middleware is extracting, parsing and adding user to the request context (see newTokensMiddleware above)
+// Unfortunately the main log is in interceptor, and auth middleware chained several hops after the interceptor
+// handler, where httpRequest is logged. So this function duplicates partly the extraction logic of
+// newTokensMiddleware in order to be applied earlier in chain and log the user passed in the request.
 func AuthorizationHeaderExtractor(deps Deps) AuthHeaderExtractor {
 	return func(r *http.Request) sdktypes.User {
-		if deps.Tokens != nil {
-			if auth := r.Header.Get("Authorization"); auth != "" {
-				if kind, payload, _ := strings.Cut(auth, " "); kind == "Bearer" {
-					if user, err := deps.Tokens.Parse(payload); err == nil {
-						return user
-					}
-				}
-			}
+		if deps.Tokens == nil {
+			return sdktypes.InvalidUser
 		}
-		return sdktypes.InvalidUser
+
+		auth := r.Header.Get("Authorization")
+		if auth == "" {
+			return sdktypes.InvalidUser
+		}
+
+		kind, payload, _ := strings.Cut(auth, " ")
+		if kind != "Bearer" {
+			return sdktypes.InvalidUser
+		}
+
+		user, err := deps.Tokens.Parse(payload)
+		if err != nil {
+			return sdktypes.InvalidUser
+		}
+
+		return user
 	}
 }
--- a/internal/backend/auth/authhttpmiddleware/authmiddleware.go
+++ b/internal/backend/auth/authhttpmiddleware/authmiddleware.go
@@ -11,6 +11,7 @@ import (
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authsessions"
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authtokens"
 	"go.autokitteh.dev/autokitteh/internal/backend/configset"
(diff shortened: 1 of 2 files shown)


internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -104,3 +105,21 @@ func New(deps Deps) AuthMiddlewareDecorator {
 		return f
 	}
 }
+
+type AuthHeaderExtractor func(*http.Request) sdktypes.User
+
+// just for logs in interceptor
+func AuthorizationHeaderExtractor(deps Deps) AuthHeaderExtractor {
+	return func(r *http.Request) sdktypes.User {
+		if deps.Tokens != nil {
```
daabr: Go really frowns upon nesting, especially a chain of `if`s. Instead, inverse the conditions, and return `InvalidUser` every time.

See:
- https://go.dev/wiki/CodeReviewComments#indent-error-flow
- https://testing.googleblog.com/2017/06/code-health-reduce-nesting-reduce.html
- https://github.com/uber-go/guide/blob/master/style.md#reduce-nesting

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -104,3 +105,21 @@ func New(deps Deps) AuthMiddlewareDecorator {
 		return f
 	}
 }
+
+type AuthHeaderExtractor func(*http.Request) sdktypes.User
+
+// just for logs in interceptor
```
daabr: This comment doesn't actually explain what this functions does, or why/when.

internal/backend/svc/svc.go
```diff
@@ -127,7 +126,7 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 		Component("auth", configset.Empty, fx.Provide(authsvc.New)),
 		Component("authjwttokens", authjwttokens.Configs, fx.Provide(authjwttokens.New)),
 		Component("authsessions", authsessions.Configs, fx.Provide(authsessions.New)),
-		Component("authhttmiddleware", authhttpmiddleware.Configs, fx.Provide(authhttpmiddleware.New)),
+		Component("authhttmiddleware", authhttpmiddleware.Configs, fx.Provide(authhttpmiddleware.New), fx.Provide(authhttpmiddleware.AuthorizationHeaderExtractor)),
```
daabr: It's easy to miss your addition to this line - please use line wrapping like in other `Component` calls.

internal/backend/svc/svc.go
```diff
@@ -258,8 +260,10 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 						varsv1connect.VarsServiceName,
 					},
 					[]httpsvc.RequestLogExtractor{
+						// Note: auth middleware is connected after interceptor, so in httpsvc and interceptor there is no user in the context.
+						// Therefore we just extract and parse user from the Authorization header.
 						func(r *http.Request) []zap.Field {
-							if user := authcontext.GetAuthnUser(r.Context()); user.IsValid() {
+							if user := authHdrExtractor(r); user.IsValid() {
```
daabr, line 279: Please add a brief explanation why extracting the user ID from the request is needed, and extracting it from the context isn't, i.e. why the user ID can't be in the context.

Are there situations where you'll want to rely on both?

internal/backend/auth/authhttpmiddleware/authmiddleware.go
ig-ra: I will rewrite it of coz, but I'm strongly disagree here with you.

Code should be clean and facilitate it's understanding and reading. Usually reducing nesting levels is great, but in this case it feels like we are trying to follow the suggestion blindly while forgetting the reasons why go suggests this.

I will explain my point:
In this function there are 2 return paths - happy one and the error one.
In my version (although there are 3 nested ifs, yes), it focuses the reader on understanding the happy path, when it will happen and what are the all conditions needed. So the reader explicitly understands what should be passed to extract user from the header.
In version you suggested - it (1) blows up the code (lines count, so the reader cannot understand at a glance what code is doing) and focuses the user to see the error path, without understanding the happy path (since happy path is everything which is !error ). Thus in order to understand the code  and what should be passed to get a valid user the reader need to go one step further and apply this mental NOT in his head to all error cases.
So in this particular case, mental effort for code understanding is bigger (IMO) in version you suggest.

rewritten

internal/backend/auth/authhttpmiddleware/authmiddleware.go
ig-ra: added more comment lines

internal/backend/svc/svc.go
ig-ra: added the explanation in the function body itself and why we could use only extractor there.

daabr (approved):

PR #374, opened 2024-06-12
ENG-1011: Fix environment variables not displaying in Sheets, Gmail, and Jira

pashafateev:


daabr (approved):
FYI, `sec` is an obsolete reference to `Secrets` objects. Please rename that in all `client.go` files that still have it in their `New()` function to `cvars` - but in a separate PR, because this one is urgent.

PR #375, opened 2024-06-12
Refactor integrations based on recent work and reviews on Jira integration

daabr:
Sorry for lumping these together, but there was no way to split them cleanly.

* Add `ConnectionCapabilities` to integrations that require initialization by the user
* Simplification of content-type check
* `redirectToErrorPage` function, instead of many `fmt.Sprintf` calls
* Replace all instances of `uiPath` consts with `desc.ConnectionURL().Path`
* Add link to Python client API
* GitHub: simplify `connStatus` function, multiple installs for a single connection are no longer possible, since we're not using tokens anymore

New additions:

* GitHub: fix ENG-991
* HTTP mux paths: add Go 1.22's HTTP method specification
* GitHub: replace per-file URL paths with Go 1.22's pattern wildcard
* Move all URL path consts to `server.go` - where their primary usage is
* Reduce multiple calls to `r.Context()` with a helper var `ctx`
* Replace references to `valuesv1.Value` with `sdktypes.ValuePB`
* Construct SDK event objects before, not during, dispatches
* Better code reused in Slack event handlers

### DIFF
--- a/integrations/github/client.go
+++ b/integrations/github/client.go
@@ -28,7 +28,7 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
 		"2 Go client API":     "https://pkg.go.dev/github.com/google/go-github/v57/github",
 		"3 Python client API": "https://pygithub.readthedocs.io/en/stable/",
 	},
-	ConnectionUrl: "/github/connect/",
+	ConnectionUrl: "/github/connect",
 	ConnectionCapabilities: &sdktypes.ConnectionCapabilitiesPB{
 		RequiresConnectionInit: true,
 	},
--- a/integrations/github/client.go
+++ b/integrations/github/client.go
@@ -2,8 +2,6 @@ package github
 
 import (
 	"context"
-	"fmt"
-	"strings"
 
 	"go.autokitteh.dev/autokitteh/integrations/github/internal/vars"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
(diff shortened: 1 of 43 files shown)


integrations/github/client.go
```diff
@@ -68,17 +69,9 @@ func connStatus(i *integration) sdkintegrations.OptFn {
 
 		if vs.Has(vars.PAT) {
 			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "using PAT"), nil
+		} else {
+			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "using OAuth 2.0"), nil
```
itayd, line 73: drop 2.0

integrations/github/webhooks/webhook.go
```diff
@@ -203,38 +185,40 @@ func extractInstallationID(event any) (inst string, err error) {
...
+			return
+		}
 
-		e := kittehs.Must1(sdktypes.EventFromProto(event))
 		eventID, err := d.Dispatch(ctx, e, nil)
 		if err != nil {
-			l.Error("Dispatch failed", zap.Error(err))
+			l.Error("Event dispatch failed",
+				zap.String("eventID", eventID.String()),
+				zap.String("connectionID", cid.String()),
```
itayd: use With

integrations/jira/event.go
```diff
@@ -59,26 +59,30 @@ func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
...
 	if !strings.HasPrefix(contentType, contentTypeJSON) {
+		// This is probably an attack, so no user-friendliness.
 		http.Error(w, "Bad Request", http.StatusBadRequest)
 		return
 	}
 
 	// Parse some of the metadata in the Jira event's JSON content.
 	body, err := io.ReadAll(r.Body)
 	if err != nil {
+		// This is probably an attack, so no user-friendliness.
```
itayd: not an attack

itayd (approved):

integrations/github/webhooks/webhook.go
daabr: Fixed across the PR

integrations/jira/event.go
daabr: Fixed across the PR

PR #379, opened 2024-06-13
bugfix: actually fetch event data in signals

itayd:
also Match can fail on invalid filter, do not use Must on it.

### DIFF
--- a/internal/backend/sessions/sessionworkflows/workflow.go
+++ b/internal/backend/sessions/sessionworkflows/workflow.go
@@ -399,7 +399,7 @@ func (w *sessionWorkflow) getNextEvent(ctx context.Context, signalID string) (ma
 
 			match, err := ev.Matches(signal.Filter)
 			if err != nil {
-				// TODO: inform user.
+				// TODO(ENG-566): inform user.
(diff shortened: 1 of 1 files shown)


internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -390,10 +390,23 @@ func (w *sessionWorkflow) getNextEvent(ctx context.Context, signalID string) (ma
...
+			ev, err := w.ws.svcs.DB.GetEventByID(ctx, evs[0].ID())
+			if err != nil {
+				return sdktypes.InvalidEventID, err
+			}
+
+			filter.MinSequenceNumber = ev.Seq() + 1
+
+			match, err := ev.Matches(signal.Filter)
+			if err != nil {
+				// TODO: inform user.
```
daabr: Ticket number?

daabr (approved):

internal/backend/sessions/sessionworkflows/workflow.go
itayd:
```suggestion
				// TODO(ENG-566): inform user.
```

itayd: done

PR #382, opened 2024-06-14
let script know if its deployment is still active

itayd:
`ak.is_deployment_active()`

if it's not, it must be draining.

### DIFF
--- a/internal/backend/akmodules/ak/module.go
+++ b/internal/backend/akmodules/ak/module.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 
 	"go.autokitteh.dev/autokitteh/internal/backend/fixtures"
+	"go.autokitteh.dev/autokitteh/internal/backend/sessions/sessiondata"
+	"go.autokitteh.dev/autokitteh/internal/backend/sessions/sessionsvcs"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/sdk/sdkexecutor"
 	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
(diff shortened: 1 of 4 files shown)


internal/backend/akmodules/ak/module.go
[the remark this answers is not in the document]
itayd: ENG-1023

tests/system/parse.go
```diff
@@ -24,7 +24,7 @@ var (
 	// wait <duration> for session <session ID>
 	waitAction = regexp.MustCompile(`^wait\s+(.+)\s+(for|unless)\s+session\s+(.+)`)
 
-	// output <equals|equals_json|contains|regex> [file] *
+	// output <equals|equals_json|equals_contains|regex> [file] *
```
daabr: Is this a mistake? I don't see `equals_contains` anywhere else

daabr (approved):

tests/system/parse.go
itayd:
```suggestion
	// output <equals|equals_json|equals|regex> [file] *
```

tests/system/parse.go
```diff
@@ -24,7 +24,7 @@ var (
 	// wait <duration> for session <session ID>
 	waitAction = regexp.MustCompile(`^wait\s+(.+)\s+(for|unless)\s+session\s+(.+)`)
 
-	// output <equals|equals_json|contains|regex> [file] *
+	// output <equals|equals_json|equals|regex> [file] *
```
itayd:
```suggestion
	// output <equals|equals_json|contains|regex> [file] *
```

tests/system/parse.go
itayd: oops

PR #383, opened 2024-06-14
ENG-802: Slack interaction - fix message update

daabr:
Calling `post()` with the Slack-provided URL to update the original message failed, because it failed to get the connection ID, and subsequently the OAuth token for it.

But responding to a Slack-provided temporary URL shouldn't require an OAuth token at all.

The code before eliminating connection tokens simply generated an empty OAuth token and couldn't fail, so I replicated this behavior in the second commit.

### DIFF
--- a/integrations/slack/api/post.go
+++ b/integrations/slack/api/post.go
@@ -208,7 +208,6 @@ func getConnection(ctx context.Context, varsSvc sdkservices.Vars) (*oauth2.Token
 	}
 
 	if varsSvc == nil {
-		// test.
 		return &oauth2.Token{}, nil
 	}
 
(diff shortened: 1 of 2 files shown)


integrations/slack/api/post.go
```diff
@@ -38,9 +38,7 @@ const (
 // slackURL is a var and not a const for unit-testing purposes.
 var slackURL = "https://slack.com/api/"
 
-type ctxKey string
-
-var OAuthTokenContextKey = ctxKey("OAuthTokenContext")
+var OAuthTokenContextKey = struct{}{}
```
itayd, line 41:
```suggestion
var OAuthTokenContextKey struct{}
```

itayd: return to how it was before

daabr: Done.

I researched this as we discussed offline, and I still prefer `struct{}` which AFAIK is the official recommendation of the Go team, and more efficient memory-wise.

Anyway, we all agree that `type ctxKey string` is not as terrible as a regular string, so it's good enough.

itayd: @tebeka?

itayd: probelm is that struct{}{} is equal to any other struct{}{}.

itayd (approved):
approving to unblock, but i think we need to talk about the context thingy.

itayd (approved):

integrations/slack/api/post.go
daabr: You could say the same thing about a type alias for strings: what if I define a different type in a different package, but use the same string as a key?

Anyway, this is interesting topic. I'll follow-up offline with Miki to see his sources, either way I'll learn something new. :)

PR #384, opened 2024-06-14
Client init functions in AK's Python SDK

daabr:
Users shouldn't copy-paste them across samples and projects.


internal/backend/oauth/oauth.go
tebeka, line 1: How is this file related to the PR?

runtimes/pythonrt/py-sdk/autokitteh.py
```diff
@@ -1,11 +1,306 @@
...
+try:
+    from atlassian import Jira
+    from google.auth.transport.requests import Request
+    import google.oauth2.credentials as credentials
+    import google.oauth2.service_account as service_account
+    from googleapiclient.discovery import build
+    from jira import JIRA
+    import slack_sdk
+except ModuleNotFoundError:
+    pass
```
tebeka: From the Python zen:
```
Errors should never pass silently.
Unless explicitly silenced.
```

Issue a warning if fails.

In general, I prefer to break this to a module with `__init__.py` and then `jira.py` and friends. Then each module will import what it needs.

runtimes/pythonrt/py-sdk/autokitteh.py
```diff
@@ -1,11 +1,306 @@
...
+    Code samples:
+    https://github.com/atlassian-api/atlassian-python-api/tree/master/examples/jira
+
+    Args:
+        connection: AutoKitteh connection name.
+
+    Returns:
+        Atlassian-Python-API Jira client.
+    """
+    if not re.fullmatch(r"[A-Za-z_]\w*", connection):
```
tebeka: Refactor to `is_valid_connection` function and use everywhere.

runtimes/pythonrt/py-sdk/autokitteh.py
```diff
@@ -1,11 +1,306 @@
...
+            "token": {
+                "access_token": os.getenv(connection + "__oauth_AccessToken"),
+                "token_type": os.getenv(connection + "__oauth_TokenType"),
+            },
+        },
+        **kwargs,
+    )
+
+
+def pycontribs_jira_client(connection: str, **kwargs):
```
tebeka: Why the `pycontribs_` prefix?

runtimes/pythonrt/py-sdk/autokitteh.py
```diff
@@ -1,11 +1,306 @@
...
+            return Jira(url=base_url, token=secret, **kwargs)
+        return Jira(
+            url=base_url,
+            username=email,
+            password=secret,
+            cloud=True,
+            **kwargs,
+        )
+
+    raise RuntimeError(f'AutoKitteh connection "{connection}" not initialized')
```
tebeka: Create our own exceptions or use an exception from the `jira` module (say https://jira.readthedocs.io/api.html#jira.exceptions.JIRAError)

runtimes/pythonrt/py-sdk/autokitteh.py
```diff
@@ -1,11 +1,306 @@
...
+
+
+def _atlassian_jira_client_cloud_oauth2(connection: str, **kwargs):
+    """Initialize a Jira client for Atlassian Cloud using OAuth 2.0."""
+    expiry = os.getenv(connection + "__oauth_Expiry")
+    if not expiry:
+        raise RuntimeError(f'AutoKitteh connection "{connection}" not initialized')
+
+    iso8601 = re.sub(r"[ A-Z]+$", "", expiry)  # Convert from Go's time string.
```
tebeka: Add a comment with example of before and after.
Or use `datetime.strptime`?

tebeka (changes requested):
We need to discuss as a team the role of the `autokitteh` Python module and if your changes fit there. Other than that, looks good minus some comments.

If you need help converting `autokitteh.py` to a directory LMK.

internal/backend/oauth/oauth.go
daabr: Oops, removed

runtimes/pythonrt/py-sdk/autokitteh.py
daabr: > Unless explicitly silenced.

That's exactly what the try-except is for here. I improved the comment to clarify.

Breaking into sub-modules - yes, it makes sense. WIP.

runtimes/pythonrt/py-sdk/autokitteh.py
daabr: I thought it would be useful to distinguish between the 2 libs, but you're right, removing.

runtimes/pythonrt/py-sdk/autokitteh.py
daabr: I strongly prefer a standard function such as `fromisoformat` over a custom pattern in `strptime`. I though the var name `iso8601` would be enough, but I'll add a specific example in a comment instead.

runtimes/pythonrt/py-sdk/autokitteh.py
daabr: About the try-except - nm, I needed to add integration libs to the Python CI to avoid test failures.

internal/backend/oauth/oauth.go
tebeka, line 1: How are the changes here related to the PR?

runtimes/pythonrt/dev-requirements.txt
```diff
@@ -2,3 +2,12 @@
 
 pytest ~= 8.0
 ruff ~= 0.2
+
+# AutoKitteh integrations
+
+atlassian-python-api ~= 3.41
```
tebeka, line 8: Use `-r ../requirements.txt` ?

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
```diff
@@ -0,0 +1,18 @@
+"""AutoKitteh Python SDK."""
+
+from . import atlassian
+from .attr_dict import AttrDict
+from .decorators import activity
+from . import errors
+from . import google
```
tebeka: Don't import `google` and `slack`. It'll force users to have the corresponding SDK installed.

tebeka (changes requested):
See comments in `__init__.py`
Also need to change `py-sdk/pyproject.toml` to account for a directory:

```
[tool.setuptools.packages]
find={}
```

internal/backend/oauth/oauth.go
daabr: Probably a bad rebase, considering https://github.com/autokitteh/autokitteh/pull/398 was already merged into main? Fixed now

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
daabr: done

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
```diff
@@ -0,0 +1,15 @@
...
+
+
+__all__ = [
+    "AttrDict",
+    "activity",
+    "atlassian",
+    "errors",
+    "google",
+    "slack",
+]
```
tebeka:
```suggestion
__all__ = [
    "AttrDict",
    "activity",
    "errors",
]
```

runtimes/pythonrt/dev-requirements.txt
```diff
@@ -2,3 +2,12 @@
...
+google-api-python-client ~= 2.122
+google-auth-httplib2 ~= 0.2
+google-auth-oauthlib ~= 1.2
+jira ~= 3.8
+slack-sdk ~= 3.27
```
tebeka, line 13: These should be in `requirements.txt` since `ak` uses it to build the virtual environment.
Maybe you can update the comment at the top of `requirements.txt` to be clearer?

tebeka (approved):
Left two comments, please fix them. Pre-approving.

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
daabr: done

runtimes/pythonrt/dev-requirements.txt
daabr: Removed entirely

PR #385, opened 2024-06-16
ENG-623: Allow users to configure zap output

daabr:
Allow users to configure Zap logging style and volume via AK config:
- `zap.level` = -1 (debug) / 0 (info) / ...
- `zap.encoding` = console / json

The default behavior is still the same:
- Default: production settings
- Dev: dev settings
- Logging level: info for both

### DIFF
--- a/internal/backend/logger/logger.go
+++ b/internal/backend/logger/logger.go
@@ -39,10 +39,15 @@ func (onPanicHook) OnWrite(ce *zapcore.CheckedEntry, fs []zapcore.Field) {
 func New(cfg *Config) (*zap.Logger, error) {
 	cfg.Zap.Level.SetLevel(zapcore.Level(cfg.Level)) // Default = 0 = Info.
 
+	// Optional override for the default encoding:
+	// AK default mode = Zap production config = "json" encoding,
+	// AK dev mode = Zap development config = "console" encoding.
 	if cfg.Encoding != "" {
 		cfg.Zap.Encoding = cfg.Encoding
 	}
 
+	// Regardless of AK mode, if the encoding is "console" (whether
+	// as a default or as an override), tweak it for easier reading.
 	if cfg.Zap.Encoding == "console" {
 		cfg.Zap.EncoderConfig.EncodeLevel = zapcore.CapitalColorLevelEncoder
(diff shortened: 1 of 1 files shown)


itayd:
we already allow to set whatever as it is now, we just set some convenient defaults.

daabr:
> we already allow to set whatever as it is now, we just set some convenient defaults.

We expose the Zap config - but users can't really modify it via our config mechanism. So This PR fixes exactly this gap.

The Zap config is unchanged, but users now have the option to inject level and encoding into it.

itayd:
> We expose the Zap config - but users can't really modify it via our config mechanism. So This PR fixes exactly this gap.

Why can't they

internal/backend/logger/logger.go
```diff
@@ -46,6 +37,18 @@ func (onPanicHook) OnWrite(ce *zapcore.CheckedEntry, fs []zapcore.Field) {
 }
 
 func New(cfg *Config) (*zap.Logger, error) {
+	cfg.Zap.Level.SetLevel(zapcore.Level(cfg.Level)) // Default = 0 = Info.
+
+	if cfg.Encoding != "" {
+		cfg.Zap.Encoding = cfg.Encoding
+	}
+
+	if cfg.Zap.Encoding == "console" {
```
itayd, line 51: this is currently the default. iiuc in your change this will active only if we explicitly set encoding to "console"?

daabr: I added these comments to clarify the logic:
```
	// Optional override for the default encoding:
	// AK default mode = Zap production config = "json" encoding,
	// AK dev mode = Zap development config = "console" encoding.
	if cfg.Encoding != "" {
		cfg.Zap.Encoding = cfg.Encoding
	}

	// Regardless of AK mode, if the encoding is "console" (whether
	// as a default or as an override), tweak it for easier reading.
	if cfg.Zap.Encoding == "console" {
		...
	}
```

itayd: oh i see https://github.com/uber-go/zap/blob/master/config.go#L231 now
ok

itayd (approved):

PR #386, opened 2024-06-16
Update Buf & Temporal dep versions + other "go mod tidy" cleanups

daabr:
No functional change in AK, but:

- New Temporal SDK (ENG-820) - includes Miki's https://github.com/temporalio/sdk-go/pull/1452 (follow-up in separate PR)
- Many proto & test changes due to newly-working Buf validation checks in messages and services (which were silently violated until now)
- Removed usage of `shortuuid` (replaced with a TypeID without a prefix)
- Replaced direct usage of YAML v2 parsing with v3

Follow-up in separate PRs:
- ENG-1024
- ENG-1026
- ENG-1027


proto/autokitteh/sessions/v1/session.proto
```diff
@@ -36,7 +36,7 @@ message SessionState {
       (buf.validate.field).map.keys.string.min_len = 1,
       (buf.validate.field).map.values.required = true
     ];
-    values.v1.Value return_value = 3 [(buf.validate.field).required = true];
+    values.v1.Value return_value = 3;
```
ig-ra, line 39: return value not required?

proto/autokitteh/sessions/v1/svc.proto
```diff
@@ -36,18 +36,13 @@ message ListRequest {
...
 
-  // TODO: Pagination.
   // TODO: FieldMask.
 
-  int32 page_size = 20 [(buf.validate.field).cel = {
-    id: "session.list.page_size",
-    message: "Must be between 10 and 100",
-    expression: "this >= 10 && this <= 100"
-  }];
+  int32 page_size = 20; // Invalid values are fixed by the gRPC service.
```
ig-ra: could u please elaborate when/where/how it fixed?

proto/autokitteh/triggers/v1/trigger.proto
```diff
@@ -11,7 +11,7 @@ message Trigger {
   string name = 2;
   string connection_id = 3;
   string env_id = 4; // if empty, applies to all envs.
-  string event_type = 5 [(buf.validate.field).string.min_len = 1];
+  string event_type = 5;
```
ig-ra, line 14: are there triggers with empty event_type?

proto/autokitteh/values/v1/values.proto
```diff
@@ -82,11 +82,11 @@ message Function {
   // executor identifies who owns the execution of the function.
   // essentially, where the function is defined. that entity
   // is responsible of the execution of the function.
-  string executor_id = 1 [(buf.validate.field).string.min_len = 1];
+  string executor_id = 1;
```
ig-ra, line 85: allowed empty?

ig-ra:
Daniel, I suggest for the next time to split such a big PRs covering too many issues.
For example, (i)removing short uuid, (ii)switching to yaml v3 and (iii) using new temporal should be IMO in separate PRs

tests/system/testdata/deployments/state.txtar
```diff
@@ -37,14 +37,16 @@ project:
   triggers:
     - name: http_get_short
       connection: my_connection
-      call: program.star:on_http_get_short
+      event_type: get
```
ig-ra, line 40: is there a way to define unite event, e.g. for both GET and POST?

web/github/connect/index.go
```diff
@@ -15,9 +17,10 @@ var index string
 var tmpl = template.Must(template.New("index").Parse(index))
 
 func ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	random := kittehs.Must1(typeid.WithPrefix(""))
 	data := map[string]string{
 		"address": os.Getenv("WEBHOOK_ADDRESS"),
-		"path":    shortuuid.New(),
+		"path":    random.String(),
```
ig-ra, line 27: is there any additional security consideration here? will random be enough? seed?

ig-ra (changes requested):

proto/autokitteh/sessions/v1/session.proto
daabr: Yes, sessions logs are broken with this check.

proto/autokitteh/sessions/v1/svc.proto
daabr: https://github.com/autokitteh/autokitteh/blob/main/internal/backend/sessionsgrpcsvc/svc.go

Changed the comment to be more specific.

proto/autokitteh/triggers/v1/trigger.proto
daabr: Yes, the convention according to Itay is that either "event_type" or "filter" is required - see the new checks in `svc.proto`.

FYI, the CLI (ak trigger create) requires "event_type" or "filter" or "data" (because the scheduler requires only "data" and automatically sets "event_type").

proto/autokitteh/values/v1/values.proto
daabr: Yes, multiple places create values without a xid and without a desc - according to Itay that's ok and shouldn't change, so the check is outdated.

tests/system/testdata/deployments/state.txtar
daabr: Not in `event_type` - you can either use:
1. Two triggers pointing to the same `call` function
2. Single trigger with a `filter` that accepts both event types

web/github/connect/index.go
daabr: Good question. Added comment.

proto/autokitteh/sessions/v1/session.proto
ig-ra: should session log return value?

proto/autokitteh/sessions/v1/svc.proto
ig-ra: could u plz add TODO in sessionsgrpcsvc to remove magic values to some config/consts?

tests/system/testdata/deployments/state.txtar
ig-ra: so, if I want trigger to be triggered on GET I need to use `event_type`, and if I want it to be triggered on both Get and POST I need to use filter?

proto/autokitteh/sessions/v1/session.proto
daabr: I'll let @itayd decide on that and follow-up in a separate PR (keep? deprecate field? delete field?).

proto/autokitteh/sessions/v1/svc.proto
daabr: Configurability: services are always the deciders when it comes to pagination, based on our understanding of performance. This is not a setting that users should be able to control, so they're always hard-coded and not configurable.

As for consts - that's a good idea, but beyond the scope of this PR. I opened this ticket: https://linear.app/autokitteh/issue/ENG-1033/improvements-in-pagination-in-grpc-services

tests/system/testdata/deployments/state.txtar
daabr: Yes. You can also say that you can always use a filter, but if you want to avoid "complicated" CEL syntax you can use event_type.

proto/autokitteh/sessions/v1/session.proto
itayd: this is odd - this should always be specified. @daabr can you reporoduce?

itayd (approved):

proto/autokitteh/sessions/v1/session.proto
daabr: @itayd yes - if you add this check back, the system test `connections/var.txtar` and the sessions test `activity_global.txtar` both fail with the same problem related to getting session logs.

See this commit's GitHub action failures as an example: https://github.com/autokitteh/autokitteh/actions/runs/9539935684/job/26290934987

ig-ra (approved):

proto/autokitteh/sessions/v1/session.proto
itayd: please create a ticket as a bug on me for this

daabr: https://linear.app/autokitteh/issue/ENG-1040/session-state-return-value-proto-validation-check

PR #387, opened 2024-06-16
upgrade temporal: sdk -> 1.27.0, api -> 1.34.0

itayd:
this brings in miki's fix: https://github.com/temporalio/sdk-go/pull/1452
need to use it.


ig-ra:
1. failed checks
2. we probably need to merge https://github.com/autokitteh/autokitteh/pull/386 first (also cleaning of mod/sum)
3. should we remove our temporal re-implementation of the same feature?

PR #388, opened 2024-06-17
ENG-1027: Redirect Temporal dev server log to file instead of stdout/stderr

daabr:
Redo of #282, after #386


internal/backend/temporalclient/client.go
```diff
@@ -31,6 +35,7 @@ type impl struct {
 	z      *zap.Logger
 	cfg    *Config
 	srv    *testsuite.DevServer
+	log    *os.File
```
itayd: logFile?

internal/backend/temporalclient/client.go
```diff
@@ -76,11 +81,19 @@ func New(cfg *Config, z *zap.Logger) (Client, error) {
...
 }
 
-func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Options) error {
-	cfg.DevServer.ClientOptions = &opts
-
+func (c *impl) startDevServer(ctx context.Context) error {
 	var err error
-	if c.srv, err = testsuite.StartDevServer(ctx, cfg.DevServer); err != nil {
+	logPath := path.Join(xdg.DataHomeDir(), "temporal_dev.log")
+	c.log, err = os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0o644)
```
itayd: this can accumulate pretty quickly, i would drop the append.

itayd (commented):
I was thinking we can just keep sending to stderr, but when the server goes down, close it for temporal.

@tebeka ?

internal/backend/temporalclient/client.go
daabr: Done

internal/backend/temporalclient/client.go
daabr: I thought `O_WRONLY` forces truncation anyway, but you're right, so dropping `O_APPEND`.

See verification here: https://goplay.tools/snippet/lKfAXSidc97

daabr:
> I was thinking we can just keep sending to stderr, but when the server goes down, close it for temporal.

I don't think that would work? The whole point of this change is to hide stdout and stderr instead of spamming the user's terminal with Temporal errors - which is at its worst when AK shuts down.

itayd:
> > I was thinking we can just keep sending to stderr, but when the server goes down, close it for temporal.
> 
> I don't think that would work? The whole point of this change is to hide stdout and stderr instead of spamming the user's terminal with Temporal errors - which is at its worst when AK shuts down.

Yeah but we might not want to make ALL errors disappear, just those that happen when we close down the server. Anyhow, no biggy.

itayd (approved):

internal/backend/temporalclient/client.go
```diff
@@ -76,11 +81,19 @@ func New(cfg *Config, z *zap.Logger) (Client, error) {
...
 }
 
-func (c *impl) startDevServer(ctx context.Context, cfg *Config, opts client.Options) error {
-	cfg.DevServer.ClientOptions = &opts
-
+func (c *impl) startDevServer(ctx context.Context) error {
 	var err error
-	if c.srv, err = testsuite.StartDevServer(ctx, cfg.DevServer); err != nil {
+	logPath := path.Join(xdg.DataHomeDir(), "temporal_dev.log")
+	c.logFile, err = os.OpenFile(logPath, os.O_CREATE|os.O_WRONLY, 0o644)
```
tebeka, line 87: You are overwriting the log file every time, is this intentional?
Also, why not use `os.Create`?

tebeka (approved):

internal/backend/temporalclient/client.go
daabr: Intentional: yes, (1) to keep the file small over time, (2) the logs of previous runs are rarely useful

Why not `os.Create`: to specify file permissions (644 = `rw-r--r--`) instead of the default (666 = `rw-rw-rw-`)

PR #393, opened 2024-06-18
[ENG-983] Create a list of functions/modules that should not become an activity

tebeka:


runtimes/pythonrt/ak_runner/deterministic.py
```diff
@@ -0,0 +1,113 @@
...
+import datetime
+
+
+def is_determinstic(fn):
+    """Return True if fn (callable) can run outside of activity."""
+    return fn in functions or fn.__module__ in modules
+
+
+# Modules are represented as strings func.__module__ is a string
+modules = {
```
daabr, line 18: Allow-list vs. deny-list question: would it be simpler to assume that most builtins are deterministic and exclude known/suspected exceptions, or not? I'm asking because "Common enough to be included" seems too subjective/controversial.

i.e.

```
stdlib_modules = frozenset({m for m in sys.stdlib_module_names if not m.startswith("_")})
modules = stdlib_modules - frozenset({
    ...
})
```

**Update:** I just saw your response in Linear about preferring an allow-list - that's fine. But in that case, I'd ask to add these modules:

types
typing
unittest
uuid
xml
zipfile
zoneinfo

runtimes/pythonrt/ak_runner/deterministic.py
```diff
@@ -0,0 +1,113 @@
+"""List of functions/modules that are deterministic should not run as activities."""
```
daabr:
```suggestion
"""List of functions/modules that are deterministic and should not run as activities."""
```

daabr (approved):

runtimes/pythonrt/ak_runner/deterministic.py
tebeka: I've added types, typing,  xml, zipfile and zoneinfo.

Others:
- unittest is problematic (especially the main function). And IMO most developers don't use code from unittest in their code - what do you want from there?
- uuid is not determinstic
- I've added zoneinfo, but you might get different results if the zone db changes. IMO it's OK

daabr: I already approved the PR, because this list is a suggestion and not a strong requirement, so this is fine, but just to respond:

* `unittest` - I want people to have a clear and familiar way to unit-test their code, so as long as it works that's fine; I see [you already use it](https://github.com/autokitteh/samples/pull/84/files#diff-cc0fd32bb3d003d46c443c67953575c8d1124293552175b91ef784791fc77dce), but I'd like to know more about the user experience, anyway not a high priority
* `UUID` - I see, I agree with you; I'll check if we need to handle it with function granularity or not, like datetime
* `zoneinfo` - ack, I agree

PR #395, opened 2024-06-19
[ENG-1042] Support class definition in user code

tebeka:


daabr (approved):
The motivation for ENG-1042 was empty classes - and this PR fixes that issue, so approving. But that raises a question: do we also support regular/static/class methods defined inside classes?

PR #400, opened 2024-06-20
ENG-1048: Support connection status in Slack

daabr:
Add a `connStatus()` function to the Slack integration.

On the way, simplify Slack's `New()` function by moving all the functions to a separate function, and clean-up cosmetic details in GitHub's `connStatus()` function.

### DIFF
--- a/integrations/slack/client.go
+++ b/integrations/slack/client.go
@@ -40,7 +40,7 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
 func New(vs sdkservices.Vars) sdkservices.Integration {
 	return sdkintegrations.NewIntegration(
 		desc,
-		sdkmodule.New(funcs(vs)...),
+		sdkmodule.New(exportFuncs(vs)...),
 		connStatus(vs),
 		sdkintegrations.WithConnectionConfigFromVars(vs),
 	)
@@ -68,7 +68,7 @@ func connStatus(vs sdkservices.Vars) sdkintegrations.OptFn {
 	})
 }
 
-func funcs(vs sdkservices.Vars) []sdkmodule.Optfn {
+func exportFuncs(vs sdkservices.Vars) []sdkmodule.Optfn {
 	authAPI := auth.API{Vars: vs}
 	bookmarksAPI := bookmarks.API{Vars: vs}
 	botsAPI := bots.API{Vars: vs}
@@ -90,7 +90,7 @@ func funcs(vs sdkservices.Vars) []sdkmodule.Optfn {
 			"bookmarks_add",
 			bookmarksAPI.Add,
 			sdkmodule.WithFuncDoc("https://api.slack.com/methods/bookmarks.add"),
-			sdkmodule.WithArgs("channel_id", "title" /* , "type" */, "link?", "emoji?", "entity_id?", "parent_id?"),
+			sdkmodule.WithArgs("channel_id", "title", "type?", "link?", "emoji?", "entity_id?", "parent_id?"),
 		),
 		sdkmodule.ExportFunction(
 			"bookmarks_edit",
@@ -260,7 +260,7 @@ func funcs(vs sdkservices.Vars) []sdkmodule.Optfn {
 		),
 
 		// Users.
-		// TODO: sdkmodule.ExportFunction(
+		// TODO(ENG-1057): sdkmodule.ExportFunction(
 		// "users_conversations",
 		// 	sdkmodule.WithFuncDoc("https://api.slack.com/methods/users.conversations"),
 		// 	sdkmodule.WithArgs(...TODO...),
--- a/integrations/slack/client.go
+++ b/integrations/slack/client.go
@@ -1,6 +1,8 @@
 package slack
 
 import (
+	"context"
+
 	"go.autokitteh.dev/autokitteh/integrations/slack/api/auth"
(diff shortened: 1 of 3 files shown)


integrations/slack/client.go
```diff
@@ -34,230 +37,257 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
...
-			sdkmodule.ExportFunction(
-				"bookmarks_remove",
-				bookmarksAPI.Remove,
-				sdkmodule.WithFuncDoc("https://api.slack.com/methods/bookmarks.remove"),
-				sdkmodule.WithArgs("bookmark_id", "channel_id", "quip_section_id?"),
-			),
+// connStatus is an optional connection status check provided by the
+// integration to AutoKitteh. The possible results are "init required"
+// (the connection is not usable yet) and "using X" (where "X" is the
+// authentication method: an OAuth v2 app, or a Socket Mode app).
```
tebeka, line 52: There should also be an `error` status to differentiate from regular error which means `I can't get the connection status`. The connection error status can contain reason, say invalid credentials.

integrations/slack/client.go
```diff
@@ -34,230 +37,257 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
...
+func connStatus(vs sdkservices.Vars) sdkintegrations.OptFn {
+	return sdkintegrations.WithConnectionStatus(func(ctx context.Context, cid sdktypes.ConnectionID) (sdktypes.Status, error) {
+		if !cid.IsValid() {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "init required"), nil
```
tebeka, line 56: We need a formal definition of the state machine of connection. "not valid" can mean also bad credentials.

integrations/slack/client.go
```diff
@@ -34,230 +37,257 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
...
-				sdkmodule.WithFuncDoc("convenience wrapper for chat.postMessage"),
-				sdkmodule.WithArgs("target", "text", "thread_ts?", "reply_broadcast?"),
-			),
-			sdkmodule.ExportFunction(
-				"send_approval_message",
-				chatAPI.SendApprovalMessage,
-				sdkmodule.WithFuncDoc("convenience wrapper for chat.postMessage"),
-				sdkmodule.WithArgs("target", "header", "message", "green_button?", "red_button?", "thread_ts?", "reply_broadcast?"),
-			),
+func funcs(vs sdkservices.Vars) []sdkmodule.Optfn {
```
tebeka: I don't like this name. Maybe `initExports` or `exportFuncs`?

integrations/slack/client.go
```diff
@@ -34,230 +37,257 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
...
-				reactionsAPI.Add,
-				sdkmodule.WithFuncDoc("https://api.slack.com/methods/reactions.add"),
-				sdkmodule.WithArgs("channel", "name", "timestamp"),
-			),
+		// Bookmarks.
+		sdkmodule.ExportFunction(
+			"bookmarks_add",
+			bookmarksAPI.Add,
+			sdkmodule.WithFuncDoc("https://api.slack.com/methods/bookmarks.add"),
+			sdkmodule.WithArgs("channel_id", "title" /* , "type" */, "link?", "emoji?", "entity_id?", "parent_id?"),
```
tebeka: Delete comment

integrations/slack/client.go
```diff
@@ -34,230 +37,257 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
...
+		// Reactions.
+		sdkmodule.ExportFunction(
+			"reactions_add",
+			reactionsAPI.Add,
+			sdkmodule.WithFuncDoc("https://api.slack.com/methods/reactions.add"),
+			sdkmodule.WithArgs("channel", "name", "timestamp"),
+		),
+
+		// Users.
+		// TODO: sdkmodule.ExportFunction(
```
tebeka: Is there an issue associated with this TODO?

tebeka (approved):
Some comments, nothing that blocks a merge.

integrations/slack/client.go
daabr: In theory I agree, and it will be done, but for now we don't have a check for stateful connections.

integrations/slack/client.go
daabr: I agree with the need for a formal definition - I'll follow-up with Haim, because this has a strong UX & doc'n aspect more than a technical aspect.

Bad creds: creds are validated for each connection before saving them, so this is not a pressing concern, but you're right that they can still be invalidated later.

When we'll add statefulness to post-creation connection tests and, we will also reuse the existing pre-creation checks code.

integrations/slack/client.go
daabr: Renamed to `exportFuncs` in both integrations

integrations/slack/client.go
daabr: Fixed by uncommenting.

integrations/slack/client.go
daabr: Added to the comment: ENG-1057

PR #401, opened 2024-06-20
[ENG-1051] Don't make built-in methods to activities

tebeka:


runtimes/pythonrt/ak_runner/deterministic.py
```diff
@@ -10,9 +10,33 @@
...
+        return True
+
+    self = getattr(fn, '__self__', None)
+    if self is not None:
+        return self.__class__ in builtin_types
+
+    return False
+
+
+# Please keep the following sorted in alphabetical order.
```
daabr, line 26: ❤️

runtimes/pythonrt/ak_runner/deterministic.py
```diff
@@ -10,9 +10,33 @@
...
-    return fn in functions or fn.__module__ in modules
+    if fn in functions:
+        return True
 
-# Please keep the following modules and functions sorted in alphabetical order.
+    if fn.__module__ in modules:
+        return True
+
+    self = getattr(fn, '__self__', None)
+    if self is not None:
```
daabr, line 20: There's no values of `__self__` that can be evaluated as `False` other than `None`, right?
```suggestion
    if self:
```

runtimes/pythonrt/ak_runner/deterministic.py
```diff
@@ -122,4 +146,4 @@ def is_determinstic(fn):
     datetime.time.utcoffset,
     datetime.time.dst,
     datetime.time.tzname,
-}
+} | builtin_types
```
daabr: Nit: I think it's cleaner and easier to understand that `modules` includes `builtin_types` when you reverse the order, i.e.:

```
modules = builtin_types | {
   ...
}
```

But that's optional.

daabr (approved):

runtimes/pythonrt/ak_runner/deterministic.py
tebeka: Not sure and I don't want to risk it. Since I default to `None` I check for it.

runtimes/pythonrt/ak_runner/deterministic.py
tebeka: Good idea. Done.

PR #402, opened 2024-06-20
ENG-1045: un/marshal Slack blocks without structs

daabr:
This avoids the need for 100% accurate modeling of Slack UI blocks, to support users with advanced Slack needs.


itayd (approved):
make sure this works.

daabr:
> make sure this works.

Already did :)

PR #403, opened 2024-06-20
Jira events: send JSON instead of text

daabr:


integrations/jira/event.go
```diff
@@ -41,63 +41,76 @@ func NewHTTPHandler(l *zap.Logger, o sdkservices.OAuth, v sdkservices.Vars, d sd
...
-		// Attack or network loss, so no need for user-friendliness.
+		l.Warn("Failed to read content of incoming Jira event", zap.Error(err))
 		http.Error(w, "Bad Request", http.StatusBadRequest)
 		return
 	}
 
-	var e event
-	if err := json.Unmarshal(body, &e); err != nil {
-		// Probably an attack, so no need for user-friendliness.
+	var jiraEvent map[string]interface{}
```
itayd:
```suggestion
	var jiraEvent map[string]any
```

integrations/jira/event.go
```diff
@@ -41,63 +41,76 @@ func NewHTTPHandler(l *zap.Logger, o sdkservices.OAuth, v sdkservices.Vars, d sd
...
+	if err := json.Unmarshal(body, &jiraEvent); err != nil {
+		l.Warn("Failed to unmarshal JSON in incoming Jira event", zap.Error(err))
```
itayd, line 82: error?

itayd: and all above

integrations/jira/event.go
```diff
@@ -41,63 +41,76 @@ func NewHTTPHandler(l *zap.Logger, o sdkservices.OAuth, v sdkservices.Vars, d sd
...
+			l.Warn("Invalid webhook ID in Jira event", zap.Any("id", v))
+			return 0
+		}
+		return int(f)
+	})
+
 	ctx := extrazap.AttachLoggerToContext(l, r.Context())
-	key := sdktypes.NewSymbol("webhook_id")
-	for _, id := range e.MatchedWebhookIDs {
+	key := sdktypes.NewSymbol("WebhookID")
```
itayd: move it into var outside the func

integrations/jira/event.go
```diff
@@ -129,11 +142,10 @@ func verifyJWT(l *zap.Logger, authz string) bool {
 	return token.Valid
 }
 
-func constructEvent(l *zap.Logger, body []byte, e event) (sdktypes.Event, error) {
-	l = l.With(zap.Any("event", body))
+func constructEvent(l *zap.Logger, jiraEvent map[string]interface{}) (sdktypes.Event, error) {
```
itayd:
```suggestion
func constructEvent(l *zap.Logger, jiraEvent map[string]any) (sdktypes.Event, error) {
```

itayd (approved):
a few comments

integrations/jira/event.go
daabr: I'm using `Error` logs for things that AK messed up or suffers from, and `Warn` for issues that require attention but are not our fault/problem. In this case all these failures are because of external data (which should be reliable, but is beyond our control) and although they represent event loss and may need to be investigated and improved, it's not a misbehavior of the server.

integrations/jira/event.go
daabr: done

integrations/jira/event.go
daabr: done

PR #404, opened 2024-06-21
Small but important changes in connection pages

daabr:
1. Bug fix: save Jira base URL - without this, AK saves an empty string in Jira connections that use an API key, which causes the Jira Python client to fail
2. Fix the Jira token param name across the index file and connection saving webhook
3. Add our OAuth guide to Google pages
4. Cosmetic tweaks in Jira page

### DIFF
--- a/web/static/jira/connect/index.html
+++ b/web/static/jira/connect/index.html
@@ -109,10 +109,10 @@
         </div>
 
         <div class="form-group">
-          <label for="key">API Token / Personal Access Token (PAT)</label>
+          <label for="token">API Token / Personal Access Token (PAT)</label>
           <input
             type="text"
-            name="key_or_pat"
+            name="token"
             autocomplete="off"
             spellcheck="false"
             required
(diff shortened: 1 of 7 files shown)


web/static/jira/connect/index.html
```diff
@@ -97,20 +92,15 @@ <h1>Atlassian Jira - Create a New Connection</h1>
...
 
       <!-- TODO(ENG-799): remove ID once its unused -->
       <form id="save-form" method="post" action="/jira/save">
         <div class="form-group">
-          <label for="url">Base URL</label>
+          <label for="base_url">Base URL</label>
           <input
             type="url"
-            name="url"
+            name="base_url"
```
RonenMars, line 103: Since this field has a placeholder, I would add a placeholder for the rest of the fields (email and token).

daabr: Discussed offline - added email placeholder (`name@example.com`), but no token placeholder (it's just a random-looking string).

RonenMars (approved):

PR #405, opened 2024-06-23
ENG-1038: Secure tar file extraction

daabr:
Add an explicit data filter when extracting Python code from an AK build's tar file. This is important for security, and removes a future deprecation warning about Python 3.14.

Details:
- https://docs.python.org/3/library/tarfile.html#extraction-filters
- https://docs.python.org/3/library/tarfile.html#tarfile.data_filter

This limits AK to Python version 3.11.4+ instead of 3.11+, but it seems reasonable given that 3.11.4 was released a year ago (2023-06-06).


tebeka (approved):
Love that we're starting to pay attention to security.

PR #408, opened 2024-06-23
[ENG-1055] Split runner & user code directories, delete them after run

tebeka:


runtimes/pythonrt/py.go
```diff
@@ -143,66 +143,88 @@ func pyExeInfo(ctx context.Context, exePath string) (exeInfo, error) {
...
+	runOK := false
+	var ri pyRunInfo
+	var err error
+	defer func() {
+		if !runOK {
+			ri.Cleanup()
+		}
+	}()
+
+	ri.userRootDir, err = os.MkdirTemp("", "ak-")
```
daabr, line 184: Question: do we actually care / need to identify our temp dirs? I'm not strictly against it, just wondering if it has a useful purpose at this point.

runtimes/pythonrt/py.go
```diff
@@ -143,66 +143,88 @@ func pyExeInfo(ctx context.Context, exePath string) (exeInfo, error) {
...
 
 type runOptions struct {
 	log            *zap.Logger
-	pyExe          string
-	tarData        []byte
-	rootPath       string
-	env            map[string]string
+	pyExe          string            // Python executable to use
+	tarData        []byte            // tar data from build stage
+	rootPath       string            // simple.py:greet
```
daabr: `rootPath` is a confusing name. It's associated with directories rather than filenames and function names. It's not a showstopper, but what about `entryPoint` or `call`?

daabr (approved):
Approved with 2 optional nits

runtimes/pythonrt/py.go
tebeka: When debugging it's easy to find these directories.

runtimes/pythonrt/py.go
tebeka: In the `Call` function signature this is called `mainPath`. Changed to `entryPoint`

PR #418, opened 2024-06-26
ENG-955: on conn init, redirect back to vscode if init from there

itayd:


daabr (approved):
My only comment is that instead of `method=vscode` (which has a hard-coded effect on the URL), we want to support generic base URLs (e.g. `origin=<URL-encoded base URL>`).

However, doing that requires careful design to prevent attacks - so actually your approach is better for now.

itayd:
> My only comment is that instead of `method=vscode` (which has a hard-coded effect on the URL), we want to support generic base URLs (e.g. `origin=<URL-encoded base URL>`).
> 
> However, doing that requires careful design to prevent attacks - so actually your approach is better for now.

disagree - i don't want people to inject random URLs, might be a vector for attacks

PR #419, opened 2024-06-26
feat: optional project id in build

ig-ra:
add optional project_id in builds (proto/sdktypes/gorm) in order to connect build and projects for upcoming auth checks

ENG-1078


proto/autokitteh/builds/v1/svc.proto
```diff
@@ -29,7 +29,7 @@ message SaveRequest {
     expression: "has(this.build) && this.build.build_id == ''"
   };
 
-  builds.v1.Build build = 1;
```
itayd, line 32: how did it work before?

sdk/sdktypes/build.go
```diff
@@ -25,13 +25,18 @@ func StrictBuildFromProto(m *BuildPB) (Build, error) { return Strict(BuildFromPr
...
-func (p Build) CreatedAt() time.Time { return p.read().CreatedAt.AsTime() }
+func (p Build) ID() (_ BuildID)          { return kittehs.Must1(ParseBuildID(p.read().BuildId)) }
+func (p Build) ProjectID() (_ ProjectID) { return kittehs.Must1(ParseProjectID(p.read().ProjectId)) }
+func (p Build) CreatedAt() time.Time     { return p.read().CreatedAt.AsTime() }
 
 func (p Build) WithNewID() Build {
 	return Build{p.forceUpdate(func(m *BuildPB) { m.BuildId = NewBuildID().String() })}
 }
 
+func (p Build) WithProjectID(pid ProjectID) Build {
```
itayd, line 42: need to add validation of project_id to Validate() above

proto/autokitteh/builds/v1/svc.proto
ig-ra: which way is preffered? `Build` vs. `builds.v1.Build` ? In our proto code we are using the both, so I'm trying to add more consistency to the files I'm touching

sdk/sdktypes/build.go
ig-ra: done

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -50,15 +50,20 @@ var Tables = []any{
 }
 
 type Build struct {
-	BuildID   sdktypes.UUID `gorm:"primaryKey;type:uuid;not null"`
+	BuildID   sdktypes.UUID  `gorm:"primaryKey;type:uuid;not null"`
+	ProjectID *sdktypes.UUID `gorm:"index;type:uuid"`
```
efiShtain, line 54: I think we can index not null, it will make the index smaller
could you check if it is viable for this case?

efiShtain (approved):
one index question but looks good to me

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: ENG-1093

PR #421, opened 2024-06-27
ENG-1088 Add confluence client

pashafateev:
`__atlassian_confluence_client_cloud_oauth2` and `__atlassian_jira_client_cloud_oauth2` could be combined into one?
But for the sake of speed I just duplicated.

### DIFF
--- a/runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
@@ -4,7 +4,7 @@ from datetime import UTC, datetime
 import re
 import os
 
-from atlassian import Jira
+from atlassian import Confluence, Jira
 from jira import JIRA
 
 from .connections import check_connection_name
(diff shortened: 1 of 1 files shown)


runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
```diff
@@ -87,6 +87,82 @@ def __atlassian_jira_client_cloud_oauth2(connection: str, **kwargs):
...
+    raise ConnectionInitError(connection)
+
+
+def __atlassian_confluence_client_cloud_oauth2(connection: str, **kwargs):
+    """Initialize a Confluence client for Atlassian Cloud using OAuth 2.0."""
+    expiry = os.getenv(connection + "__oauth_Expiry")
+    if not expiry:
+        raise ConnectionInitError(connection)
+
+    # Convert Go's time string (e.g. "2024-06-20 19:18:17 -0700 PDT") to
```
tebeka, line 139: Why not use `datetime.strptime`?

tebeka (approved):
Approved, one question.

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
pashafateev: I copy and pasted existing code. There wasn't much additional thought behind it other than it's what we're already using.

PR #426, opened 2024-06-27
ENG-1081: Confluence integration

daabr:
Turns out Confluence and Jira are not at all alike as one would assume, and that Confluence's APIs are... Not as mature as one would expect.

Specific notes:

- Webhook creation via REST in Confluence **Cloud** is hacky
- Works only with API tokens and not OAuth 2.0 (again, only in Cloud)
- These webhooks don't support secrets for callback signature verification
- No event type ID in callbacks, need to use a separate webhook URL per event category (created, updated, removed, etc.) - focusing only on "created" for now


integrations/atlassian/confluence/event.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// Parse some of the metadata in the Atlassian event's JSON content.
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		l.Warn("Failed to read content of incoming Atlassian event", zap.Error(err))
+		http.Error(w, "Bad Request", http.StatusBadRequest)
+		return
+	}
+
+	var atlassianEvent map[string]any
```
pashafateev, line 72: Why is this `map[string]any` and not something more specific? Is that due to not having much information about the events to begin with or is this common practice?

pashafateev (approved):

integrations/atlassian/confluence/event.go
daabr: Exactly. `map[string]any` is the most generic way to unmarshal JSON text into a Go data structure, without knowing anything about it (literally "a mapping of string keys to whatever"). In the case of Atlassian events - their schema varies so wildly that it doesn't make sense to enforce any structure on them between the point of receiving them (our webhook) and dispatching them to session workflows.

integrations/atlassian/confluence/vars.go
```diff
@@ -0,0 +1,21 @@
...
+import (
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var (
+	baseURL = sdktypes.NewSymbol("BaseURL")
+	token   = sdktypes.NewSymbol("Token")
+	email   = sdktypes.NewSymbol("Email")
+
+	oauthAccessToken = sdktypes.NewSymbol("oauth_AccessToken")
```
itayd, line 12: i think using the struct approach would have been nicer here

integrations/atlassian/confluence/webhooks.go
```diff
@@ -0,0 +1,225 @@
...
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		l.Warn("Failed to read Confluence webhooks list response", zap.Error(err))
+		return 0, false
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		l.Warn("Unexpected response to Confluence webhooks list request",
+			zap.Int("status", resp.StatusCode),
+			zap.ByteString("body", body),
```
itayd, line 78: you sure you want to spit out the body here?

integrations/atlassian/confluence/webhooks.go
```diff
@@ -0,0 +1,225 @@
...
+		)
+		return 0, false
+	}
+
+	var list []webhook
+	if err := json.Unmarshal(body, &list); err != nil {
```
itayd, line 84: you could use a decoder and skip the ReadAll

itayd (commented):
only nits

PR #431, opened 2024-06-28
Add BeautifulSoup to AK's Python venv

daabr:
We already support BeautifulSoup in Starlark, and we recently encountered a need to parse HTML in Python (Confluence page content is returned as HTML).

On the other hand, we want to be judicious about adding new libraries.

WDYT?


itayd:
very reasonable

itayd (approved):

tebeka (approved):
LGTM

PR #444, opened 2024-07-03
ENG-1119: allow to supply pure json inputs to session

itayd:


internal/backend/sessionsgrpcsvc/svc.go
```diff
@@ -39,6 +41,29 @@ func (s *server) Start(ctx context.Context, req *connect.Request[sessionsv1.Star
...
+		if err := d.Decode(&x); err != nil {
+			err = sdkerrors.NewInvalidArgumentError(`json_inputs["%s"]: %w`, k, err)
+			return nil, sdkerrors.AsConnectError(err)
+		}
+
+		if msg.Session.Inputs == nil {
+			msg.Session.Inputs = make(map[string]*sdktypes.ValuePB)
+		}
+
+		v, err := sdktypes.WrapValue(x)
```
ig-ra, line 58: do we need to warp it to unwrap later?

proto/autokitteh/sessions/v1/svc.proto
```diff
@@ -13,6 +13,9 @@ message StartRequest {
   };
 
   Session session = 1 [(buf.validate.field).required = true];
+
+  // Helper: if set, overwrites the session's inputs.
```
ig-ra: if set, *merge and possibly override* the session inputs?

internal/backend/sessionsgrpcsvc/svc.go
itayd: it's coming in unwrapped. we need to wrap it.

proto/autokitteh/sessions/v1/svc.proto
itayd: done

daabr (approved):

PR #445, opened 2024-07-04
fix(vars): deleting scope vars if scope is deleted

ig-ra:
- eliminates vars for deleted scopes by deleting them if scope (env or connection) is deleted
- fixes connection deletion (and relevant vars) from several places

fixes ENG-1049

### DIFF
--- a/internal/backend/db/dbgorm/connections.go
+++ b/internal/backend/db/dbgorm/connections.go
@@ -2,6 +2,7 @@ package dbgorm
 
 import (
 	"context"
+	"fmt"
 
 	"github.com/google/uuid"
 	"gorm.io/gorm"
@@ -22,13 +23,13 @@ func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection
 	return gdb.createEntityWithOwnership(ctx, createFunc, conn, conn.ProjectID)
 }
 
-func (gdb *gormdb) deleteConnectionsAndVars(where string, id sdktypes.UUID) error {
+func (gdb *gormdb) deleteConnectionsAndVars(what string, id sdktypes.UUID) error {
 	// should be transactional with context already applied
 
 	var ids []sdktypes.UUID
 	q := gdb.db.Model(&scheme.Connection{})
 	q = q.Clauses(clause.Returning{Columns: []clause.Column{{Name: "connection_id"}}})
-	if err := q.Delete(&ids, where, id).Error; err != nil {
+	if err := q.Delete(&ids, fmt.Sprintf("%s = ?", what), id).Error; err != nil {
 		return err
 		// REVIEW: proceed to vars deletion if there are any?
 	}
@@ -45,7 +46,7 @@ func (gdb *gormdb) deleteConnection(ctx context.Context, id sdktypes.UUID) error
 			return err
 		}
 		// delete connection and associated vars
-		return tx.deleteConnectionsAndVars("connection_id = ?", id)
+		return tx.deleteConnectionsAndVars("connection_id", id)
 	})
 }
 
--- a/internal/backend/db/dbgorm/connections.go
+++ b/internal/backend/db/dbgorm/connections.go
@@ -5,6 +5,7 @@ import (
 
 	"github.com/google/uuid"
 	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
 
 	"go.autokitteh.dev/autokitteh/internal/backend/db/dbgorm/scheme"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
@@ -21,19 +22,30 @@ func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection
 	return gdb.createEntityWithOwnership(ctx, createFunc, conn, conn.ProjectID)
 }
 
+func (gdb *gormdb) deleteConnectionsAndVars(where string, args ...interface{}) error {
+	// should be transactional with context already applied
+
+	var ids []sdktypes.UUID
+	q := gdb.db.Model(&scheme.Connection{})
+	q = q.Clauses(clause.Returning{Columns: []clause.Column{{Name: "connection_id"}}})
+	if err := q.Delete(&ids, where, args).Error; err != nil {
+		return err
+		// REVIEW: proceed to vars deletion if there are any?
+	}
+
+	if len(ids) > 0 {
+		return gdb.db.Where("var_id IN (?)", ids).Delete(&scheme.Var{}).Error
+	}
+	return nil
+}
+
 func (gdb *gormdb) deleteConnection(ctx context.Context, id sdktypes.UUID) error {
 	return gdb.transaction(ctx, func(tx *tx) error {
 		if err := tx.isCtxUserEntity(ctx, id); err != nil {
 			return err
 		}
-
-		// delete connection
-		if err := tx.db.Delete(&scheme.Connection{ConnectionID: id}).Error; err != nil {
-			return err
-		}
-
-		// and associated vars
-		return tx.db.Where("var_id = ?", id).Delete(&scheme.Var{}).Error
+		// delete connection and associated vars
+		return tx.deleteConnectionsAndVars("connection_id = ?", id)
 	})
 }
 
--- a/internal/backend/db/dbgorm/connections.go
+++ b/internal/backend/db/dbgorm/connections.go
@@ -22,7 +22,7 @@ func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection
 	return gdb.createEntityWithOwnership(ctx, createFunc, conn, conn.ProjectID)
 }
 
-func (gdb *gormdb) deleteConnectionsAndVars(where string, args ...interface{}) error {
+func (gdb *gormdb) deleteConnectionsAndVars(where string, args ...any) error {
 	// should be transactional with context already applied
 
 	var ids []sdktypes.UUID
--- a/internal/backend/db/dbgorm/connections.go
+++ b/internal/backend/db/dbgorm/connections.go
@@ -26,7 +26,14 @@ func (gdb *gormdb) deleteConnection(ctx context.Context, id sdktypes.UUID) error
 		if err := tx.isCtxUserEntity(ctx, id); err != nil {
 			return err
 		}
-		return tx.db.Delete(&scheme.Connection{ConnectionID: id}).Error
+
+		// delete connection
+		if err := tx.db.Delete(&scheme.Connection{ConnectionID: id}).Error; err != nil {
+			return err
+		}
+
+		// and associated vars
+		return tx.db.Where("var_id = ?", id).Delete(&scheme.Var{}).Error
 	})
 }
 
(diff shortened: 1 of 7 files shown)


internal/backend/db/dbgorm/connections.go
```diff
@@ -21,12 +22,30 @@ func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection
 	return gdb.createEntityWithOwnership(ctx, createFunc, conn, conn.ProjectID)
 }
 
+func (gdb *gormdb) deleteConnectionsAndVars(where string, args ...interface{}) error {
```
efiShtain: Why args in interface?

internal/backend/db/dbgorm/connections.go
```diff
@@ -21,12 +22,30 @@ func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection
...
+	var ids []sdktypes.UUID
+	q := gdb.db.Model(&scheme.Connection{})
+	q = q.Clauses(clause.Returning{Columns: []clause.Column{{Name: "connection_id"}}})
+	if err := q.Delete(&ids, where, args).Error; err != nil {
+		return err
+		// REVIEW: proceed to vars deletion if there are any?
+	}
+
+	if len(ids) > 0 {
+		return gdb.db.Where("var_id IN (?)", ids).Delete(&scheme.Var{}).Error
```
efiShtain, line 38: Can you do it in one query with the connection delete ?

internal/backend/db/dbgorm/connections.go
ig-ra: it's called from several places. 
- with where connection_id =?
- with where project_id = ?

ig-ra: btw, replaced interface{} by any. We are doing the same in `dbgorm.go`

efiShtain: but what is args ?
it's a delete statement without a proper signature
feels a bit scary

internal/backend/db/dbgorm/events.go
```diff
@@ -88,7 +89,7 @@ func (db *gormdb) SaveEvent(ctx context.Context, event sdktypes.Event) error {
 	if connectionID.IsValid() { // only if exists
 		conn, err := db.GetConnection(ctx, connectionID)
 		if err != nil {
-			return err
+			return fmt.Errorf("connection: %w", err)
```
efiShtain, line 92: better to have ```invalid connection: ``` instead of just ```connection```

internal/backend/db/dbgorm/projects.go
```diff
@@ -49,9 +49,7 @@ func (gdb *gormdb) deleteProject(ctx context.Context, projectID sdktypes.UUID) e
 
 // delete project, its envs, deployments, sessions and build
 func (gdb *gormdb) deleteProjectAndDependents(ctx context.Context, projectID sdktypes.UUID) error {
-	// NOTE: should be transactional
-
-	db := gdb.db.WithContext(ctx)
+	// NOTE: should be transactional and with context applied
```
efiShtain, line 52: if it has to be in a transaction, why not start a transaction here ?

internal/backend/db/dbgorm/events.go
ig-ra: it's a wrapped error. it will be
"save event: connection: not found"

internal/backend/db/dbgorm/projects.go
ig-ra: it's called from transactional context everywhere anyway.
Since we are not that good at maintaining recursive transaction in our gormdb, in all methods where transactional context expected there is a note clarifying this

internal/backend/db/dbgorm/connections.go
ig-ra: delete from 2 different schemas? I don't think so. 
maybe it's possible with assocications, but varID isn't a foreign key since it could be wither endID or connectionID. moreover gorm cannot cascade on delete from soft-deleted entries (e.g. env or connection)

internal/backend/db/dbgorm/connections.go
ig-ra: it's an internal generic method. And we have this in gormdb in generic delete forexample which accepts exactly the same params, e.g. where + args.

efiShtain: we have it doesn't mean it is a great thing

WDYT about something like this? still generic, but not total open to anything
It's not a full example, no error handling and stuff, just a different way to implement

```go
type queryAugmenter func(q *gorm.DB) *gorm.DB

func (gdb *gormdb) deleteProjectConnections() {
	gdb.deleteConnectionsAndVars(func(q *gorm.DB) *gorm.DB {
		return q.Where("project_id = ?", "my_porject")
	})
}

func (gdb *gormdb) deleteConnectionsAndVars(where queryAugmenter) {
	q := gdb.db
	q = where(q)
	q.Delete(&scheme.Connection{})
}

```

internal/backend/db/dbgorm/connections.go
efiShtain: I think there is a way to do polymorphic foreign keys, probably over kill for this case
but we should probably wrap it in transaction if it is mandatory to be in transaction (it's okay to have tx in tx)

ig-ra: I already dealt with this problem while implementing users. There is no way in gorm to enforce polymorphic foreign keys unfortunately.
regarding transaction - it's an internal method which is called in transaction already. I don't want to add overhead of recursive transactions. I tested this while implementing users and got to the conclusion that I will use internal methods and single transaction where possible. Recursive transactions is uneeded overhead IMO for now

internal/backend/db/dbgorm/connections.go
ig-ra: I'm sorry Efi, but I'm probably missing the point. Could u explain me why this is better? I'm passing the same "project_id = ?" and projectID as a parameter. And in the second case "connection_id =?" and connection_id. I could add those scope functions, and use it with but I very puzzled, since I see no benefits neither in usability nor in security. 
- In our version you are passing the func with hardcoded `Where` parameters. 
- In my case I'm passing hardcoded `parameters` and doing `Where` in place. 

Could you plz once again explain why you think it's a problem to do exactly what we are doing in other places of our code?

internal/backend/db/dbgorm/envs.go
```diff
@@ -43,7 +43,12 @@ func (gdb *gormdb) deleteEnvs(ctx context.Context, ids []sdktypes.UUID) error {
 		return err
 	}
 
-	return db.Where("env_id IN ?", ids).Delete(&scheme.Env{}).Error
+	// delete envs and associated vars
+	if err := db.Where("env_id IN ?", ids).Delete(&scheme.Env{}).Error; err != nil {
+		return err
+	}
+
+	return db.Where("var_id IN ?", ids).Delete(&scheme.Var{}).Error
```
daabr, line 51: Can you combine both deletions into a single transaction?

internal/backend/db/dbgorm/vars.go
```diff
@@ -89,8 +89,10 @@ func (gdb *gormdb) deleteVars(ctx context.Context, scopeID sdktypes.UUID, names
 func (gdb *gormdb) listVars(ctx context.Context, scopeID sdktypes.UUID, names ...string) ([]scheme.Var, error) {
 	db := varsCommonQuery(gdb.withUserVars(ctx), scopeID, names)
 
+	// Note(s):
+	// - will skip not user owned vars
```
daabr: I see you're just moving this comment here and below, but I don't understand it - what is "not" referring to? Users, ownership, or vars?

internal/backend/db/dbgorm/connections.go
daabr: I think the point Efi is making is that there's a difference between the lack of safety in Gorm's `Delete` (which is by design for maximum flexibility at the lowest level), and the lack of safety in this higher-level function.

In our case we don't want to offer maximum flexibility, we want to ensure the caller knows exactly what they're doing, and can't do what we didn't plan to allow them. Type safety is a way to achieve that.

If you don't want separate functions like Efi suggested, you can also use a single function with a variadic option pattern - `opts` is type-safe, but allows a variable number of customizations. For example:

https://github.com/autokitteh/autokitteh/blob/main/sdk/sdkintegrations/integration.go#L59

...But personally I think it's more complicated and less useful for just 2 options.

internal/backend/db/dbgorm/vars.go
ig-ra: not is referring to check of whethr scope is valid or not, e.g. deleted. 
now scope (env or connection) deletion will result in marking scope as deleted (soft-deleted) and hard-deleting scope vars

internal/backend/db/dbgorm/connections.go
ig-ra: I changed type from interface{} to sdktypes.UUID

internal/backend/db/dbgorm/projects.go
ig-ra: because we are introducing one more transactional layer in our gormdb (to lock mutex) on top of gorm.DB.
Anyhow - it's an internal method, called from several transactions, previously it was cut-n-pasted logic, now extracted to a method. I think it's unnecessarily to introduce more transactions and locking where unneeded.

internal/backend/db/dbgorm/envs.go
ig-ra: you mean via cascade delete?

internal/backend/db/dbgorm/connections.go
```diff
@@ -21,12 +22,30 @@ func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection
 	return gdb.createEntityWithOwnership(ctx, createFunc, conn, conn.ProjectID)
 }
 
+func (gdb *gormdb) deleteConnectionsAndVars(where string, id sdktypes.UUID) error {
+	// should be transactional with context already applied
+
+	var ids []sdktypes.UUID
+	q := gdb.db.Model(&scheme.Connection{})
+	q = q.Clauses(clause.Returning{Columns: []clause.Column{{Name: "connection_id"}}})
```
efiShtain, line 31: try something like this please where = fmt.Sprintf("%x = ?", where)

efiShtain (approved):

internal/backend/db/dbgorm/connections.go
ig-ra: done

internal/backend/db/dbgorm/envs.go
ig-ra: 1. I don't think I could delete from two different tables in one statement
2. we also cannot use on delete cascade - (i) gorm not supporting it on soft-delete (env) and (ii) env isn't a foreign key and cannot be since it's a scope id - either var or connection

daabr: No, I meant this: https://gorm.io/docs/transactions.html

This allows you to run multiple queries as a single unit, so if one fails they all fail together. This is done to avoid inconsistent states and having to implement custom rollback logic.

Anyway, if this makes sense, you can do it in a separate PR.

daabr (approved):

PR #447, opened 2024-07-05
ENG-1106 (part 1/2): improve connection init flow + fix ENG-1153

daabr:
- Small misc improvements in connection init flow: code documentation, small readability tweaks
- Rename connection init flow's "method" to "origin" (method is ambiguous)
- Generalize connection init flow: all origins are stated explicitly:
  - `/connections/<CID>/init[?method=...]` --> `/connections/<CID>/init/{origin}`
- Use HTTP status "302 Found" for redirects instead of "301 Moved Permanently"
- Fix bug: ENG-1153 (VSCode and SaaS redirection didn't work for OAuth connections)

Splitting this task into 2 PRs for easier reviewing.

The next PR (part 2) will implement the different output of connection init flows with `"web"` origins, as described in ENG-1106.

PR #448, opened 2024-07-06
ENG-1106 (part 2/2): 

daabr:
Continuation of https://github.com/autokitteh/autokitteh/pull/447

All integrations now report connection init successes and failures in a standard way, and the display of this status depends on the init request's origin.

- `/connections/CID/init/vscode` --> VS Code URL
- `/connections/CID/init/web` --> JSON message with "status" (200, 400, 500, etc.) and optional "error" string
- Everything else (CLI, web dashboard) - as usual

The core of this PR is in these files:

- internal/backend/dashboardsvc/connections.go
- internal/backend/oauth/webhook.go
- proto/autokitteh/oauth/v1/svc.proto
- sdk/sdkintegrations/initconn.go

### DIFF
--- a/sdk/sdkintegrations/initconn.go
+++ b/sdk/sdkintegrations/initconn.go
@@ -50,7 +50,8 @@ func (c ConnectionInit) Abort(err string) {
 // and redirects the user to the last HTTP response, based on its
 // origin (local AK server / local VS Code extension / SaaS web UI).
 func (c ConnectionInit) AbortWithStatus(status int, err string) {
-	switch c.Request.FormValue("origin") {
+	origin := c.Request.FormValue("origin")
+	switch origin {
 	case "vscode":
 		u := "vscode://autokitteh.autokitteh?cid=%s&status=%d&error=%s"
 		u = fmt.Sprintf(u, c.ConnectionID, status, url.QueryEscape(err))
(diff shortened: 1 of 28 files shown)


sdk/sdkintegrations/initconn.go
```diff
@@ -5,37 +5,86 @@ import (
...
+	case "vscode":
+		u := "vscode://autokitteh.autokitteh?cid=%s&status=%d&error=%s"
+		u = fmt.Sprintf(u, c.ConnectionID, status, url.QueryEscape(err))
+		http.Redirect(c.Writer, c.Request, u, http.StatusFound)
+	case "web":
+		u := "/connections/%s/result?status=%d&error=%s"
+		u = fmt.Sprintf(u, c.ConnectionID, status, url.QueryEscape(err))
+		http.Redirect(c.Writer, c.Request, u, http.StatusFound)
+	default: // Local server ("cli", "dash", etc.)
+		path := c.Integration.ConnectionURL().Path + "/error.html"
```
itayd: You need to put the right status code here, not just found. If there is an error, the right error code should be returned. 
Also - how do you make Ronen parse the error?

daabr: The behavior that Ronen asked for was to always get HTTP 200, and parse the status from a JSON body - but you're right. I confirmed with Ronen that HTTP status + error message in the HTTP response's body is in fact better.

itayd: No, we need to return proper status codes. I also told this to Ronen.

sdk/sdkintegrations/initconn.go
```diff
@@ -5,37 +5,84 @@ import (
...
+}
+
+// Abort aborts the connection initialization flow due to
+// a runtime error. It encodes the error status and message,
+// and redirects the user to the last HTTP response, based on its
+// origin (local AK server / local VS Code extension / SaaS web UI).
+func (c ConnectionInit) AbortWithStatus(status int, err string) {
+	switch c.Request.FormValue("origin") {
+	case "vscode":
+		u := "vscode://autokitteh.autokitteh?cid=%s&status=%d&error=%s"
```
itayd, line 56: did you coordinate with ronen re this? not sure though if we shouldn't just error out like before here - does it matter if it's in vscode or not?

sdk/sdkintegrations/initconn.go
daabr: Yes - already done, see line 59

sdk/sdkintegrations/initconn.go
daabr: Yes, discussed this specific detail this morning

itayd (approved):

PR #450, opened 2024-07-07
Fix error handling in Confluence connection init

daabr:
Now, if we fail to create/update the connection's webhook (if the connection details are wrong), we abort the connection with an informative error message.

### DIFF
--- a/integrations/atlassian/confluence/webhooks.go
+++ b/integrations/atlassian/confluence/webhooks.go
@@ -3,6 +3,7 @@ package confluence
 import (
 	"bytes"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -138,6 +139,7 @@ func registerWebhook(l *zap.Logger, base, user, key string) (int, string, error)
 			zap.Any("request", r),
 			zap.Error(err),
 		)
+		return 0, "", err
 	}
 
 	jsonReader := bytes.NewReader(body)
@@ -169,7 +171,7 @@ func registerWebhook(l *zap.Logger, base, user, key string) (int, string, error)
 			zap.Int("status", resp.StatusCode),
 			zap.ByteString("body", body),
 		)
-		return 0, "", err
+		return 0, "", errors.New(string(body))
 	}
 
 	var reg webhook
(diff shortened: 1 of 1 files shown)


integrations/atlassian/confluence/webhooks.go
```diff
@@ -169,7 +171,7 @@ func registerWebhook(l *zap.Logger, base, user, key string) (int, string, error)
 			zap.Int("status", resp.StatusCode),
 			zap.ByteString("body", body),
 		)
-		return 0, "", err
+		return 0, "", errors.New(string(body))
```
efiShtain: body might be empty, are you sure this is informative ?
maybe we should at least add additional data?

ig-ra: 1. why we are logging with Warn and not Error, if this is an error? (see other places as well)
2. I think if we are logging the body, there is no need to wrap the whole body as en error. Body could be not only empty but huge. 
3. Could we extact the error message from the body (and even limit its size)? then IMO it would be logical both to log it and send as error. Otherwise I would suggest only to log and create short descriptive error

itayd: 1. this is not an infra error
2.

daabr: @efiShtain (empty body): the body is guaranteed to be a JSON status object, since this is the response to an API call. For example (in case of an error):

```json
{"message":"Current user not permitted to use Confluence","statusCode":403}
```

And even if it is empty, the user will see this error text (`save.go:66`):

```
failed to register webhook:
```

But I added a check anyway.

@ig-ra 2 (huge body): the body is guaranteed to be small for the same reason as above, i.e. `status` int + short human-readable `message` string. But I added a 256-char limit, just in case.

@ig-ra 3 (parse JSON): Given the small size of the JSON object, and the fact that the entire object is useful, I prefer not to parse it (unnecessary assumptions about a third-party can potentially break).

@ig-ra 1 (warn instead of fail): like @itayd wrote, I'm warning that something failed because of bad user input, which may be investigated, but isn't an AK problem that requires fixing.

ig-ra (approved):

PR #451, opened 2024-07-07
[ENG-1110] Install Python & create venv on docker image

tebeka:


Dockerfile
```diff
@@ -57,6 +57,13 @@ USER appuser
 # Copy the executable from the "build" stage.
 COPY --from=build /bin/ak /bin/
 
+# Create initial venv
+RUN python3 -m venv ~/.local/share/autokitteh/venv
```
efiShtain, line 58: Any chance of doing this is a different stage and copy the data in the final stage ?
It will make stuff more clear and allow parallel execution and better caching (changes to go files won't require re build of this venv)

tebeka: I'll bump it up before the `COPY` command, but virtual environment are not relocatable :)
I need it to be after the `USER` command so it'll create a venv for the user running ak.

efiShtain (approved):
It's ok for now 
We might need to optimize it a bit later

PR #453, opened 2024-07-07
Install GCC on ARM

tebeka:


itayd (commented):
why is this needed?

efiShtain:
> why is this needed?

near future we would like to run on arm machines in aws, it's more performance for the same price (or lower price with same performance, depends on what you want)

it also helps with building on local m1 machines

itayd (approved):

PR #455, opened 2024-07-07
fix: pass context in resolver

ig-ra:
- pass existing context to resolver instead of using a new one
- stop filtering ErrNotFound on server/API side and let clients handle it

fixes ENG-1157

### DIFF
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -21,6 +21,7 @@ type Resolver struct {
 	Client sdkservices.DBServices
 }
 
+// FIXME: move to sdkerrors
 type NotFoundError struct {
 	Type, Name string
 }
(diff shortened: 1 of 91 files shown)


efiShtain (approved):
Please add tests at least for the problem that caused this change

internal/resolver/resolver.go
```diff
@@ -57,9 +57,6 @@ func (r Resolver) BuildID(id string) (b sdktypes.Build, bid sdktypes.BuildID, er
 		return
 	}
 
-	ctx, cancel := limitedContext()
```
daabr, line 60: Are the `limitedContext()` function in line 39, and the `timeout` const in line 18, needed anymore? If not, delete them

daabr (approved):
Approved with 1 comment about possibly-dead-code in `resolver.go`

PR #456, opened 2024-07-08
ENG-1161: publish latest image on merge to main, publish stable image on release

efiShtain:
So we won't require a release just for testing purposes

latest images can be broken 
stable images are only on releases

### DIFF
--- a/.github/workflows/ci-go.yml
+++ b/.github/workflows/ci-go.yml
@@ -31,7 +31,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
-          python-version: '3.11'  # Should be in sync with runtimes/pythonrt/pythonrt.go:minPyVersion
+          python-version: "3.11" # Should be in sync with runtimes/pythonrt/pythonrt.go:minPyVersion
       - name: Test
         run: make test-unit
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci-go.yml
```diff
@@ -104,3 +104,44 @@ jobs:
 
       - name: Session tests
         run: make bin/ak test-sessions
+
+  publish_docker_image:
+    needs:
```
itayd, line 109: do we need really these here? i'm good either way.

itayd (approved):

.github/workflows/ci-go.yml
efiShtain: It just make sure we don't publish an image if tests aren't passing

It won't start the publish job until all build and test jobs finish

daabr: The confusing part for me is that this file runs for all PRs as well as merges into main, and this job has a specific filter: `if: github.ref == 'refs/heads/main'`. Also, this ignores non-Golang tests.

It's better to just put it in a separate file, with this trigger:

```
on:
  push:
    branches:
      - main
```

i.e. without pull requests, and even without depending on tests, since our PRs are required to have linear history and all test passing before merging.

.github/workflows/ci-go.yml
```diff
@@ -104,3 +104,44 @@ jobs:
...
+          aws-region: us-east-1
+      - name: Login to Amazon ECR
+        id: login-ecr
+        uses: aws-actions/amazon-ecr-login@v2
+      - name: Build And Push
+        uses: docker/build-push-action@v4
+        with:
+          context: .
+          platforms: linux/amd64
+          tags: ${{ steps.login-ecr.outputs.registry }}/autokitteh:latest
```
daabr, line 145: What's the effect if `latest` here but `stable` in `release.yml`?

efiShtain: basically it means every push to main generate a latest version
but once you release, stable tag is updated to the latest release (assuming release is tested)
release also get a tagged version

so push to main generate new ```autokitteh:latest``` image
release tag x.y.z generate ```autokitteh:x.y.z``` and update ```autokitteh:stable``` to match ```autokitteh:x.y.z```

since we can release not the latest commit, latest tag is not updated when we release

PR #457, opened 2024-07-08
ENG-1065 Add connection status to gmail, google and sheets clients

pashafateev:
Originally `utils.go` was part of google package but that causes a circular dependency with the sheets client so I decided to put it in a separate directory.

**Question:**
Aren't we using connection status in all the other integrations? Would it make sense to have this function be in `/integrations`?

### DIFF
--- a/integrations/google/calendar/client.go
+++ b/integrations/google/calendar/client.go
@@ -1,7 +1,7 @@
 package calendar
 
 import (
-	"go.autokitteh.dev/autokitteh/integrations/google"
+	"go.autokitteh.dev/autokitteh/integrations/google/utils"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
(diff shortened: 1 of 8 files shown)


daabr (commented):
This is a good PR, the only thing I would change is `utils`: this is a common patterns, but not a good one. Especially in Go, try to minimize package/file names such as "utils", "common", etc. - prefer more specific grouping categories, if they're needed at all.

In this case, you have to have a separate sibling package to avoid circular dependency, but a better name might be "connections" maybe? (And we'll add there a common connection testing function later too).

As for your question:

> Aren't we using connection status in all the other integrations? Would it make sense to have this function be in /integrations?

Maybe we could (and if so feel free to do it), but these functions are not identical, because they reflect different connection modes, var names, etc. that different integrations use.

daabr (approved):

PR #459, opened 2024-07-09
cfg where to allow only printing directory alone

itayd:
allows to
```
cd "$(ak cfg where -c)"
```

### DIFF
--- a/cmd/ak/cmd/configuration/where.go
+++ b/cmd/ak/cmd/configuration/where.go
@@ -11,6 +11,8 @@ import (
 	"go.autokitteh.dev/autokitteh/internal/xdg"
 )
 
+var onlyCfg, onlyData bool
+
 var whereCmd = common.StandardCommand(&cobra.Command{
 	Use:     "where",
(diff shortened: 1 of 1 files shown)


cmd/ak/cmd/configuration/where.go
```diff
@@ -19,11 +21,23 @@ var whereCmd = common.StandardCommand(&cobra.Command{
 
 	RunE: func(cmd *cobra.Command, args []string) error {
 		cfg := xdg.ConfigHomeDir()
+
+		if onlyCfg {
+			fmt.Fprintln(cmd.OutOrStdout(), cfg)
+			return nil
+		}
+
```
daabr, line 29: Move your block under the `if strings.Contains(cfg, " ")` - so if `cfg` contains spaces you'll print `"path"` instead of `path`. Same with `data` below.

daabr (approved):
Actually, if you run `cd "$(ak cfg where -c)"` as noted in the PR description, this should be fine so approving.

PR #462, opened 2024-07-09
py: `ak` module can be missing, add tests

itayd:


runtimes/pythonrt/pythonrt.go
```diff
@@ -238,7 +236,7 @@ func (py *pySvc) handleLog(msg Message) error {
 func (py *pySvc) loadSyscall(values map[string]sdktypes.Value) error {
 	ak, ok := values["ak"]
 	if !ok {
-		return fmt.Errorf("`ak` not found")
+		return nil
```
tebeka, line 262: Log a warning

runtimes/pythonrt/py.go
```diff
@@ -331,6 +333,8 @@ func pyExports(ctx context.Context, pyExe string, fsys fs.FS) ([]Export, error)
 		return nil, fmt.Errorf("inspect: %w.\nPython output: %s", err, buf.String())
 	}
 
+	sdklogger.Warn(ctx, "inspect output", zap.String("output", buf.String()))
```
tebeka: I prefer to pass the logger from pythonrt

tebeka (approved):
Two comment,  not blocking merge.

runtimes/pythonrt/py.go
itayd: no idea why i even added this warn

runtimes/pythonrt/pythonrt.go
itayd: i don't think this needs a warning, it's completely legit not to have it if not running in a session.

PR #464, opened 2024-07-09
ENG-1178: fix Confluence event type extraction

daabr:
1. Recognize attachment events correctly
2. Don't create events if their type is unknown

### DIFF
--- a/integrations/atlassian/confluence/event.go
+++ b/integrations/atlassian/confluence/event.go
@@ -81,6 +81,7 @@ func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
 	eventType, ok := extractEntityType(l, atlassianEvent, r.PathValue("category"))
 	if !ok {
 		http.Error(w, "Bad Request", http.StatusBadRequest)
+		return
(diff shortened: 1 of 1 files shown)


integrations/atlassian/confluence/event.go
```diff
@@ -126,7 +127,10 @@ func extractEntityType(l *zap.Logger, atlassianEvent map[string]any, category st
...
 	}
 
-	// Some "relation_*" events look a little different.
+	// Some "attachment_*" and "relation_*" events look a little different.
+	if _, ok := atlassianEvent["attachments"]; ok {
+		return fmt.Sprintf("attachment_%s", category), true
+	}
 	if _, ok := atlassianEvent["relationData"]; ok {
 		return fmt.Sprintf("relation_%s", category), true
 	}
```
pashafateev, line 136: So the event can only have either `attachments` or `relationData`?

daabr: Explained offline.

pashafateev (approved):

PR #468, opened 2024-07-10
ENG-1187: session log record duplicate key on insert

efiShtain:
log records has a primary key of session_id,sequence
if we try to insert the same sequence twice it will fail
on duplicate and we need to retry
sequence is calculated again on each insertion so it has
to succeed in the end
we try for 10 times instead of indefinite but we might
want to consider it in the future

### DIFF
--- a/internal/backend/db/dbgorm/sessions.go
+++ b/internal/backend/db/dbgorm/sessions.go
@@ -3,6 +3,7 @@ package dbgorm
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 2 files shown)


internal/backend/db/dbgorm/sessions.go
```diff
@@ -128,11 +129,31 @@ func createLogRecord(db *gorm.DB, logr *scheme.SessionLogRecord) error {
 	// This is a hack in gorm to use subquery when creating an object.
 	// The goal is to have a sequence number per session.  The insert uses a subquery to get
 	// the current max sequence for this session_id and insert the next value in one query
+	// The loop is to handle optimistic locking, since concurrent writes can happen
+	// we have to retry on failure to commit
+	// We will try fo 10 times which should be enough retries
+	// This convention does not guarantee ordering of the log records
+	// Since inheritnly in the records them selfs, two records can happen in the same exact time
 	data["Seq"] = clause.Expr{SQL: "(select COALESCE(MAX(seq), 0)+1 from session_log_records where session_id = ?)", Vars: []interface{}{logr.SessionID}}
```
tebeka, line 137: I think the right solution is to use an `AUTOINCREMENT` field in the database.

efiShtain: we can't use it in this case since we want to keep using sqlite/postgres and gorm
sqlite does not support auto increment column on a non primary key
and we actually want a sequence per session and not a generic sequence for all

we never going to query by sequence, only by (session_id,sequence) or just session_id

itayd:
hey hold your horses here this is not the right solution. the reason that we try to do it multiplie times sometimes is not because we actually need to, but because it's not encapsulated properly in an activity. need to think this more deeply, i am going to engage with this soon.

efiShtain:
> hey hold your horses here this is not the right solution. the reason that we try to do it multiplie times sometimes is not because we actually need to, but because it's not encapsulated properly in an activity. need to think this more deeply, i am going to engage with this soon.

It has nothing to do with activities in the case

we need to retry to write for few reasons
1. writes can fail, like anything, sometimes you need to retry
2. specifically in this case, in cases we try to write simultaneously we get a duplicate error, which will be solved on the next retry (we don't retry on other options)



the activity fix you're talking about (which I'm not aware of) is orthogonal to this one, both should be solved

itayd:
> 1. writes can fail, like anything, sometimes you need to retry

So why auto increment? We need to keep the same number.

> 2. specifically in this case, in cases we try to write simultaneously we get a duplicate error, which will be solved on the next retry (we don't retry on other options)

This shouldn't happen - the workflow needs to be running in a single instance at any one time. Are you talking about a split brain situation?

efiShtain:
Daniel had a problem with duplicate keys, 
not sure how it happened, but I reproduced it easily on a side project
this solves the issue as a valid fix

we can drop it later in case we think we don't need it, but I think this is a valid solution for it (maybe for every db access we do)

even from one worker, if we issue multiple activities in parallel (which should be supported), two logs can be tried to be written to db at the same time (or close enough)

internal/backend/db/dbgorm/sessions.go
tebeka: OK then.

efiShtain:
I'm closing this since it didn't resolve the issue and a bigger change is required
working on a new PR handling the errors

PR #470, opened 2024-07-10
[ENG-1170] Do not print empty output

tebeka:


itayd:
i really don't understand this pr... i know what it's trying to do, just not sure how.

tebeka:
> i really don't understand this pr... i know what it's trying to do, just not sure how.

This is the empty `[stdout] ` lines in the session log.

daabr (approved):
Oddly, I think this PR is also a workaround like #474 (even though I think the removal of the panic is important anyway).

PR #472, opened 2024-07-10
Fix: retain CID and origin when retrying connection init

daabr:
I wish we could reuse a single copy of these files - but Go file embedding doesn't support symbolic links.

We might use in the future a single dynamic `error.html` file - single template but different titles and logos.

Also, there's a tiny fix and auto-formatting by Prettier in `web/static/github/connect/error.html`

### DIFF
--- a/web/static/chatgpt/connect/error.js
+++ b/web/static/chatgpt/connect/error.js
@@ -1,8 +1,14 @@
-// Display the error message in the page too, not just as a URL parameter.
-
-const param = new URLSearchParams(window.location.search).get("error");
+// Display the error message in the page, not just as a URL parameter.
+const err = new URLSearchParams(window.location.search).get("error");
 const elem = document.getElementById("error");
 
-if (param) {
-  elem.textContent = param;
+if (err) {
+  elem.textContent = err;
 }
+
+// Allow the user to retry initializing the connection.
+document.getElementById("retry").onclick = function () {
+  const id = new URLSearchParams(window.location.search).get("cid");
+  const origin = new URLSearchParams(window.location.search).get("origin");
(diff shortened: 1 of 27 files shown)


web/static/chatgpt/connect/error.js
```diff
@@ -1,8 +1,14 @@
...
-  elem.textContent = param;
+if (err) {
+  elem.textContent = err;
 }
+
+// Allow the user to retry initializing the connection.
+document.getElementById("retry").onclick = function () {
+  const id = new URLSearchParams(window.location.search).get("cid");
+  const origin = new URLSearchParams(window.location.search).get("origin");
+  window.location.href = "./?cid=".concat(id, "&origin=", origin);
```
RonenMars, line 13: What are the odds that we won't have id and origin?
Would you consider a check like this?
```
  if (id && origin) {
    window.location.href = `./?cid=${id}&origin=${origin}`;
  }
```

And what do you think about splitting that code into 2 functions and make it a bit more readable?
```
function displayErrorMessage() {
  const err = new URLSearchParams(window.location.search).get("error");
  if (err) {
    const elem = document.getElementById("error");
    if (elem) {
      elem.textContent = err;
    }
  }
}

function retryConnection() {
  const params = new URLSearchParams(window.location.search);
  const id = params.get("cid");
  const origin = params.get("origin");
  if (id && origin) {
    window.location.href = `./?cid=${id}&origin=${origin}`;
  }
}

document.getElementById("retry")?.addEventListener("click", retryConnection);

displayErrorMessage();
```

RonenMars (approved):
LGTM, added a NIT

PR #474, opened 2024-07-11
Workaround for ENG-1018

daabr:


internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -488,7 +488,7 @@ func (w *sessionWorkflow) run(wctx workflow.Context) (prints []string, err error
 			}
 
 			if err != nil {
-				w.z.Panic("add print", zap.Error(err))
+				w.z.Error("failed to add print session record", zap.String("run_id", runID.String()), zap.String("text", text))
```
pashafateev, line 491: did the meaning change or are you clarifying/expanding details?

runtimes/pythonrt/pythonrt.go
```diff
@@ -402,7 +400,7 @@ func (py *pySvc) Close() {
 // We split it from Call since Call is also used to execute activities.
 func (py *pySvc) initialCall(ctx context.Context, funcName string, event map[string]any) (sdktypes.Value, error) {
 	defer func() {
-		py.log.Info("python done, killing")
+		py.log.Info("Python subprocess cleanup after initial call is done")
```
pashafateev, line 403: Is "killing" implied then?

pashafateev (approved):

runtimes/pythonrt/pythonrt.go
daabr: Yes, but it's more than just killing the process, it's also disposing of the input/output streams communicating with it (lines 405-415).

PR #476, opened 2024-07-11
feat: cleanup cli code due to notFoundError

ig-ra:


tests/system/testdata/connections/list.txtar
```diff
@@ -50,8 +50,8 @@ return code == 0
 output equals ''
 
 ak connection list --integration meow -J
-return code == 1
-output equals 'Error: integration "meow" not found'
+return code == 0
```
daabr, line 53: This is good, but please also add a test of the previous error when using `--fail`, for completeness.

daabr (approved):

tests/system/testdata/connections/list.txtar
ig-ra: added

PR #478, opened 2024-07-12
Rename Google auth field

daabr:
"json" to "json_key" - to be more accurate.

On the way, remove a couple of redundant `connStatus` functions from Google integrrations that were not includes in #457.

This PR is unrelated to - but a foundation for - ENG-1103 (receive events from Google Forms).

### DIFF
--- a/integrations/google/creds.go
+++ b/integrations/google/creds.go
@@ -34,5 +34,5 @@ func (h handler) HandleCreds(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	c.Finalize(sdktypes.EncodeVars(&vars.Vars{JSON: r.Form.Get("json")}))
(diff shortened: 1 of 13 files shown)


integrations/google/creds.go
```diff
@@ -34,5 +34,5 @@ func (h handler) HandleCreds(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	c.Finalize(sdktypes.EncodeVars(&vars.Vars{JSON: r.Form.Get("json")}))
+	c.Finalize(sdktypes.EncodeVars(&vars.Vars{JSONKey: r.Form.Get("json_key")}))
```
itayd, line 37: wouldn't this screw up current installations?

itayd: nsure

PR #480, opened 2024-07-14
ENG-1103 Part 2: Google Forms events

daabr:
Create and renew event watch when connection are initialized.

Also receive, parse, enrich, and dispatch Google Forms events.

Also tiny fixes in other integrations that are not worth separate PRs:

- `jira/event.go`: `continue` --> `break`
- `integrations/google/drive/client.go`: remove obsolete function (like all other integrations)
- `integrations/google/sheets/client.go`: rename `sec` to `cvars` (like all other integrations)


tebeka (approved):
The code looks OK, but I'm not that familiar with the logic and have no way of testing.

PR #485, opened 2024-07-16
support timeout on next_event

itayd:
based on https://github.com/autokitteh/autokitteh/pull/561.

add a `timeout` argument to `next_event`.

eventually we'll need a more wholistic solution, see ENG-1209.

### DIFF
--- a/internal/backend/sessions/sessionworkflows/syscalls.go
+++ b/internal/backend/sessions/sessionworkflows/syscalls.go
@@ -226,14 +226,14 @@ func (w *sessionWorkflow) nextEvent(ctx context.Context, args []sdktypes.Value, 
 
 	wctx := sessioncontext.GetWorkflowContext(ctx)
 
-	var f workflow.Future
+	var timeoutFuture workflow.Future
 	if timeout != 0 {
-		f = workflow.NewTimer(wctx, timeout)
+		timeoutFuture = workflow.NewTimer(wctx, timeout)
 	}
 
 	for {
 		// no event, wait for first signal
-		signalID, err := w.waitOnFirstSignal(wctx, signals, f)
+		signalID, err := w.waitOnFirstSignal(wctx, signals, timeoutFuture)
 		if err != nil {
 			return sdktypes.InvalidValue, err
 		}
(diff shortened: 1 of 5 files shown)


daabr (approved):

internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -198,8 +198,9 @@ Get the next event on this signal (has to exists since we got a signal on it)
 return this event
 */
 func (w *sessionWorkflow) nextEvent(ctx context.Context, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
-	if len(kwargs) > 0 {
-		return sdktypes.InvalidValue, errors.New("unexpected keyword arguments")
+	var timeout time.Duration
```
efiShtain, line 201: time.Duration is int64, it can be negative
I think we should fail on negative value or max it to 0
```go
timeout = max(0, timeout)
```

itayd: i think we should error out if < 0. added check.

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -349,21 +349,28 @@ func (w *sessionWorkflow) createEventSubscription(ctx context.Context, connectio
 	return signalID, nil
 }
 
-func (w *sessionWorkflow) waitOnFirstSignal(ctx context.Context, signals []string) (string, error) {
-	wctx := sessioncontext.GetWorkflowContext(ctx)
+// Returns "", nil on timeout.
+func (w *sessionWorkflow) waitOnFirstSignal(wctx workflow.Context, signals []string, f workflow.Future) (string, error) {
 	selector := workflow.NewSelector(wctx)
 
+	if f != nil {
```
efiShtain, line 356: this is specifically for timeouts, please rename f to timeout for readability

efiShtain (approved):

internal/backend/sessions/sessionworkflows/workflow.go
itayd: here it's more generic, but i changed it everywhere else.

PR #489, opened 2024-07-16
python: Support binOp function calls.

tebeka:
Happend when user has an expression like (datetime.now() - start).total_second()

Fixes ENG-1213

### DIFF
--- a/runtimes/pythonrt/ak_runner/loader.py
+++ b/runtimes/pythonrt/ak_runner/loader.py
@@ -5,6 +5,22 @@ from types import ModuleType
 
 from . import log
 
+bin_ops = {
+    ast.Add: "+",
+    ast.Sub: "-",
+    ast.Mult: "*",
+    ast.MatMult: "@",
+    ast.Div: "/",
+    ast.Mod: "%",
+    ast.Pow: "**",
+    ast.LShift: "<<",
+    ast.RShift: ">>",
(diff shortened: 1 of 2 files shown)


runtimes/pythonrt/ak_runner/loader.py
```diff
@@ -26,6 +42,12 @@ def name_of(node):
     if isinstance(node, ast.Name):
         return node.id
 
+    if isinstance(node, ast.BinOp):
+        left = name_of(node.left)
+        right = name_of(node.right)
+        op = bin_ops.get(type(node.op), "?")
```
itayd, line 48: shouldn't this be an error?

itayd (commented):
can you explain the fix?

runtimes/pythonrt/ak_runner/loader.py
tebeka: The code that caused it is `(datetime.now() - start).total_seconds()`, not a bug.

tebeka:
> can you explain the fix?

Support code like: `(datetime.now() - start).total_seconds()`

daabr (approved):

itayd:
> > can you explain the fix?
> 
> Support code like: `(datetime.now() - start).total_seconds()`

you're funny. i mean how does it do that? not clear to me.

PR #491, opened 2024-07-16
python: Don't run user code when building

tebeka:
Users can run code on module level, say start connections. Changed inspect to use the ast instead of importing the user module.

Fixes ENG-1211


itayd (approved):
I'm so happy you got this figured out. I feared it would take a month to do it.

PR #492, opened 2024-07-16
fix(vars): use user checker in CreateVar

ig-ra:
CreateVar ignores user checker and does this manually

fixes ENG-1220

### DIFF
--- a/internal/backend/db/dbgorm/ownership.go
+++ b/internal/backend/db/dbgorm/ownership.go
@@ -156,7 +156,7 @@ func verifyOwnerships(uid string, ownerships []scheme.Ownership) error {
 func ensureUserAccessToEntitiesWithOwnerships(db *gorm.DB, uid string, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
 	ownerships, err := getOwnershipsForEntities(db, ids...)
 	if err != nil {
-		return ownerships, err
+		return nil, err
 	}
 
 	if len(ownerships) < len(ids) { // should be equal
--- a/internal/backend/db/dbgorm/ownership.go
+++ b/internal/backend/db/dbgorm/ownership.go
@@ -131,7 +131,7 @@ func (gdb *gormdb) createEntityWithOwnership(
 	})
 }
 
-func getOwnerships(db *gorm.DB, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
+func getOwnershipsForEntities(db *gorm.DB, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
 	if len(ids) == 0 {
 		return nil, nil
 	}
@@ -154,7 +154,7 @@ func verifyOwnerships(uid string, ownerships []scheme.Ownership) error {
 
 // fetches ownerships for given ids and verifies that user have access to all of them
 func ensureUserAccessToEntitiesWithOwnerships(db *gorm.DB, uid string, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
-	ownerships, err := getOwnerships(db, ids...)
+	ownerships, err := getOwnershipsForEntities(db, ids...)
 	if err != nil {
 		return ownerships, err
 	}
--- a/internal/backend/db/dbgorm/ownership.go
+++ b/internal/backend/db/dbgorm/ownership.go
@@ -131,26 +131,38 @@ func (gdb *gormdb) createEntityWithOwnership(
 	})
 }
 
-func EnsureUserAccessToEntities(db *gorm.DB, uid string, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
+func getOwnerships(db *gorm.DB, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
 	if len(ids) == 0 {
 		return nil, nil
 	}
 
-	var oo []scheme.Ownership
-	if err := db.Model(&scheme.Ownership{}).Where("entity_id IN ?", ids).Select("user_id", "entity_type").Find(&oo).Error; err != nil {
+	var ownerships []scheme.Ownership
+	if err := db.Model(&scheme.Ownership{}).Where("entity_id IN ?", ids).Select("user_id", "entity_type").Find(&ownerships).Error; err != nil {
 		return nil, err
 	}
+	return ownerships, nil
+}
 
-	if len(oo) < len(ids) {
-		return oo, gorm.ErrRecordNotFound
-	}
-
-	for _, o := range oo {
+func verifyOwnerships(uid string, ownerships []scheme.Ownership) error {
+	for _, o := range ownerships {
 		if o.UserID != uid {
-			return oo, sdkerrors.ErrUnauthorized
+			return sdkerrors.ErrUnauthorized
 		}
 	}
-	return oo, nil
+	return nil
+}
+
+// fetches ownerships for given ids and verifies that user have access to all of them
+func ensureUserAccessToEntitiesWithOwnerships(db *gorm.DB, uid string, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
+	ownerships, err := getOwnerships(db, ids...)
+	if err != nil {
+		return ownerships, err
+	}
+
+	if len(ownerships) < len(ids) { // should be equal
+		return ownerships, gorm.ErrRecordNotFound
+	}
+	return ownerships, verifyOwnerships(uid, ownerships)
 }
 
 func (gdb *gormdb) isUserEntity(uid string, ids ...sdktypes.UUID) error {
@@ -199,7 +211,7 @@ func (c *UsersOwnershipChecker) EnsureUserAccessToEntitiesWithOwnership(
 	db *gorm.DB, uid string, ids ...sdktypes.UUID,
 ) ([]scheme.Ownership, error) {
 	c.z.Debug("isUserEntity", zap.Any("entityIDs", ids), zap.Any("uid", uid))
-	return EnsureUserAccessToEntities(db, uid, ids...)
+	return ensureUserAccessToEntitiesWithOwnerships(db, uid, ids...)
 }
 
 func (c *UsersOwnershipChecker) EnsureUserAccessToEntities(db *gorm.DB, uid string, ids ...sdktypes.UUID) error {
@@ -219,7 +231,7 @@ type PermissiveOwnershipChecker struct {
 func (c *PermissiveOwnershipChecker) EnsureUserAccessToEntitiesWithOwnership(
 	db *gorm.DB, uid string, ids ...sdktypes.UUID,
 ) ([]scheme.Ownership, error) {
-	oo, err := EnsureUserAccessToEntities(db, uid, ids...)
+	oo, err := ensureUserAccessToEntitiesWithOwnerships(db, uid, ids...)
 	if err != nil && errors.Is(err, sdkerrors.ErrUnauthorized) {
 		err = nil // ignore not authorized, but keep all other (e.g. NotFound) - see var
 	}
--- a/internal/backend/db/dbgorm/ownership.go
+++ b/internal/backend/db/dbgorm/ownership.go
@@ -2,6 +2,7 @@ package dbgorm
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"go.uber.org/zap"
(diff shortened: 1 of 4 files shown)


internal/backend/db/dbgorm/ownership.go
```diff
@@ -120,40 +121,40 @@ func (gdb *gormdb) createEntityWithOwnership(
...
+		if err := create(tx.db, uid); err != nil { // create
 			return err
 		}
-		return saveOwnershipForEntities(tx.db, user, ownerships...) // ensure user and add ownerships
+		return saveOwnershipForEntities(tx.db, uid, ownerships...) // ensure user and add ownerships
 	})
 }
 
-func isUserEntity(db *gorm.DB, uid string, ids ...sdktypes.UUID) error {
+func EnsureUserAccessToEntities(db *gorm.DB, uid string, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
```
efiShtain: is this suppose to be exported ?

naming wise, why ensure something returns a list of objects ?

internal/backend/db/dbgorm/ownership.go
```diff
@@ -120,40 +121,40 @@ func (gdb *gormdb) createEntityWithOwnership(
...
 	var oo []scheme.Ownership
-	if err := db.Model(&scheme.Ownership{}).Where("entity_id IN ?", ids).Select("user_id").Find(&oo).Error; err != nil {
-		return err
+	if err := db.Model(&scheme.Ownership{}).Where("entity_id IN ?", ids).Select("user_id", "entity_type").Find(&oo).Error; err != nil {
+		return nil, err
 	}
 
 	if len(oo) < len(ids) {
-		return gorm.ErrRecordNotFound
+		return oo, gorm.ErrRecordNotFound
```
efiShtain: what is oo?

internal/backend/db/dbgorm/ownership.go
```diff
@@ -185,17 +186,25 @@ func (gdb *gormdb) withUserEntity(ctx context.Context, entity string) *gorm.DB {
 }
 
 type OwnershipChecker interface {
-	IsUserEntity(db *gorm.DB, user string, ids ...sdktypes.UUID) error
+	EnsureUserAccessToEntities(db *gorm.DB, user string, ids ...sdktypes.UUID) error
```
efiShtain, line 201: what does this do? 
if it has no ownership (comparing to WithOwnership), so any user has access, am I wrong?

efiShtain (commented):
Not sure I understand the changes

could you please add the design decisions with high level functions name and how they should be used to solve each case you are trying to solve ?
maybe we should also document it in a readme or something

internal/backend/db/dbgorm/ownership.go
ig-ra: `o` is ownerships, `oo` is ownershipS :) , e.g. many

internal/backend/db/dbgorm/ownership.go
ig-ra: to ensure I need to fetch ownerships from DB. 
So regular Ensure (in all places expect vars) just returns error/nil. Thus permissive checker does nothing.
EnsureWithOwnership returns both ownerships fetched and error. Because in Vars I need those ownerships to emulate foreign key check and verify that scope wasn't deleted

internal/backend/db/dbgorm/ownership.go
ig-ra: Ensure returns err
EnsureWithOwnerships returns ownerships, err

ig-ra: 2. Checker is an interface, so methods should be visible, no?

ig-ra: 3. There is a checker interface, which exports functions used for user checks/authorization.
Normal checker and Permissive checker. Permissive checker allows all users

internal/backend/db/dbgorm/ownership.go
tebeka: Maybe `ows`

internal/backend/db/dbgorm/ownership.go
```diff
@@ -207,7 +216,17 @@ type PermissiveOwnershipChecker struct {
 	z *zap.Logger
 }
 
-func (c *PermissiveOwnershipChecker) IsUserEntity(db *gorm.DB, uid string, ids ...sdktypes.UUID) error {
+func (c *PermissiveOwnershipChecker) EnsureUserAccessToEntitiesWithOwnership(
+	db *gorm.DB, uid string, ids ...sdktypes.UUID,
+) ([]scheme.Ownership, error) {
+	oo, err := EnsureUserAccessToEntities(db, uid, ids...)
+	if err != nil && errors.Is(err, sdkerrors.ErrUnauthorized) {
+		err = nil // ignore not authorized, but keep all other (e.g. NotFound) - see var
```
tebeka, line 236: Do we want to log here?

internal/backend/db/dbgorm/ownership_permissive_checker_test.go
```diff
@@ -5,12 +5,14 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.uber.org/zap"
 	"gorm.io/gorm"
 )
 
 func preOwnershipTestPermissive(t *testing.T) *dbFixture {
 	f := preOwnershipTest(t)
 	f.gormdb.owner = &PermissiveOwnershipChecker{f.gormdb.z}
+	f.gormdb.z.Info("patched ownership checker", zap.String("type", "permissive"))
```
tebeka, line 15: Why "patched"?

tebeka:
On a general note, the fact that this bug was not caught by type systems or tests after the changes @itayd did is an issue by itself.

internal/backend/db/dbgorm/ownership.go
efiShtain: please change to something meaningful like ownerships

internal/backend/db/dbgorm/ownership.go
ig-ra: fixed

internal/backend/db/dbgorm/ownership.go
ig-ra: fixed

internal/backend/db/dbgorm/ownership.go
ig-ra: there is nothing to log in permissive checker - we allow access to all users. 
On the other hand in UserChecker we are logging of coz

internal/backend/db/dbgorm/ownership_permissive_checker_test.go
ig-ra: gorm will be initialized by default with user checker. Tests are not using the same infrastructure as gormkitteh, thus I cannot pass which checker to use. So just printing in logs that we are patching to make it clear when you read test logs
```
{"level":"info","msg":"ownership","checker":"users"}
2024/07/16 16:05:01 OK   20240704121927_baseline.sql (3.9ms)
2024/07/16 16:05:01 OK   20240714051159_no-db-user.sql (2.85ms)
2024/07/16 16:05:01 goose: successfully migrated database to version: 20240714051159
=== RUN   TestSetVarWithOwnershipP
{"level":"info","msg":"patched ownership checker","type":"permissive"}
```

ig-ra:
> On a general note, the fact that this bug was not caught by type systems or tests after the changes @itayd did is an issue by itself.

@tebeka this bug is unrelated to what itay did. Bug is due to the fact that Vars are connected to either Conneciton or Env and I missed to apply checker via interface on CreateVar since it's uses non-standard logic as in all other entities.
Thus additional test was added in this PR to test exactly this

internal/backend/db/dbgorm/vars.go
```diff
@@ -29,32 +26,29 @@ func (gdb *gormdb) setVar(ctx context.Context, vr *scheme.Var) error {
...
 		db := tx.db.WithContext(ctx)
 
-		// fetch user_id and entity_type for provided scope_id
-		if err := db.Model(&scheme.Ownership{}).
-			Select("user_id", "entity_type").
-			Where("entity_id = ? AND entity_type IN ('Connection', 'Env')", vr.ScopeID).
-			First(&ownership).Error; err != nil {
-			// if no records were found then fail with foreign keys validation (#1), since there should be one
+		// if no records were found then fail with foreign keys validation (#1), since there should be one
+		oo, err := tx.owner.EnsureUserAccessToEntitiesWithOwnership(db, uid, vr.ScopeID)
```
efiShtain, line 32: please rename oo to meaningful name

ig-ra: done

internal/backend/db/dbgorm/ownership.go
```diff
@@ -185,17 +198,25 @@ func (gdb *gormdb) withUserEntity(ctx context.Context, entity string) *gorm.DB {
...
 }
 
 type UsersOwnershipChecker struct {
 	z *zap.Logger
 }
 
-func (c *UsersOwnershipChecker) IsUserEntity(db *gorm.DB, uid string, ids ...sdktypes.UUID) error {
+func (c *UsersOwnershipChecker) EnsureUserAccessToEntitiesWithOwnership(
+	db *gorm.DB, uid string, ids ...sdktypes.UUID,
+) ([]scheme.Ownership, error) {
```
efiShtain, line 212: you return here ```[]scheme.Ownership``` but in actual you return only part of the objects, since you select only user_id and event_type, missing the entity_id. This can easily lead to mistakes next time someone uses it. 

wdyt ?

ig-ra: fixed. returning full model now

internal/backend/db/dbgorm/ownership.go
```diff
@@ -120,40 +121,52 @@ func (gdb *gormdb) createEntityWithOwnership(
...
+		if err := create(tx.db, uid); err != nil { // create
 			return err
 		}
-		return saveOwnershipForEntities(tx.db, user, ownerships...) // ensure user and add ownerships
+		return saveOwnershipForEntities(tx.db, uid, ownerships...) // ensure user and add ownerships
 	})
 }
 
-func isUserEntity(db *gorm.DB, uid string, ids ...sdktypes.UUID) error {
+func getOwnerships(db *gorm.DB, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
```
efiShtain: maybe getEntitiesOwnership is better name
since you can query by entity_id or by user_id

ig-ra: getOwnerships is good enough IMO since we are passing ids of the entities to get ownerships for

ig-ra: getOwnerships not getting any user_id, just a list of entities ids

efiShtain: you are passing ids as a general
userIDs and entityIDs in this case are the same

internal/backend/db/dbgorm/ownership.go
```diff
@@ -120,40 +121,52 @@ func (gdb *gormdb) createEntityWithOwnership(
...
+func ensureUserAccessToEntitiesWithOwnerships(db *gorm.DB, uid string, ids ...sdktypes.UUID) ([]scheme.Ownership, error) {
+	ownerships, err := getOwnerships(db, ids...)
+	if err != nil {
+		return ownerships, err
+	}
+
+	if len(ownerships) < len(ids) { // should be equal
+		return ownerships, gorm.ErrRecordNotFound
+	}
+	return ownerships, verifyOwnerships(uid, ownerships)
```
efiShtain, line 165: this is error prone if you return an error and someone forget to check it, 
they would still have a list of ownerships which looks like a valid thing, 
probably should return null if there is an error

ig-ra: Disagree actually. according to the go conventions you should check the error. 
The reason for this that error could be notFOund, e.g. we fetch less ownerships then requested. But maybe it's ok for any specific case, so I would like to return an error and return data fetched.

internal/backend/db/dbgorm/ownership.go
ig-ra: but the query is entity_id =?

ig-ra: I could rename ids to entitiesIDs if this makes it better

ig-ra: renamed  to `getOwnershipForEntities`

internal/backend/db/dbgorm/ownership.go
ig-ra: rewritten. 
Now it returns nil in all cases when I cannot guarantee that results are valid.
Only if case when there are less ownerships than requested it will return NotFound error. But in this case it's guaranteed that returned results are VALID. So it's caller decision how and when use results in case of NotFound error. 
So I think it's really OK - to return valid results and non-nil error and allow the called to handle this case.

Note that this is similar to what Reader.Read is doing - when some bytes are read, but less than requested. So you could know both that there was some problem and you could know how many bytes read.

ig-ra: @tebeka , could u plz advise on this one?
is it OK to return an error along valid, but less expected result?

efiShtain: acking this one for now, 
@tebeka will have a look later and in case majority thinks current solution is fine, we're done, otherwise we would need to fix it in a different PR
not resolving this issue but approving the PR

efiShtain (approved):

internal/backend/db/dbgorm/ownership.go
ig-ra: fixed

PR #494, opened 2024-07-16
python: Temporary way to install packages

tebeka:
Allow users to install packages from workflow code. This is a workaround until we have a way to create custom Python environments per workflow.

See ENG-1219

### DIFF
--- a/runtimes/pythonrt/py-sdk/autokitteh/packages.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/packages.py
@@ -40,7 +40,7 @@ def _install_package(specifier, import_name):
         )
 
 
-def install(packages):
+def install(*packages):
     """Install Python packages using pip.
 
     A package can be either a package requirement specifier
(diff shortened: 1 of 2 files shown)


tebeka:
@haimzlato This add a way the user can write in the workflow code something like:

```from autokitteh import packages

packages.install(['numpy', 'pandas'])
```

And it'll be installed on the worker. Second call to `install` on the worker will do nothing.
Thoughts?

daabr (approved):
On one hand, this is mitigating the issue of users not being able to use their preferred packages quickly and easily - which is great!

On the other hand, I'm worried about the indirect implication of lack of protection against disk writing and shell execution - we have to use strict jailing before we make AK more public.

itayd:
interesting, but when should it be executed? in global scope?

tebeka:
> interesting, but when should it be executed? in global scope?

Yes, users will write at top of file
```python
from autokitteh import packages
packages.install(['numpy', 'matplotlib'])
```

I wonder if the API will be better with *args
```python
packages.install('numpy', 'matplotlib')
```

PR #496, opened 2024-07-17
Refactoring: switch to reusable connection result web pages for SaaS

daabr:
Before: "web"-originating connection initialization flows ended in HTTP responses with a relevant HTTP states, and optionally a body with a relevant error message.

Now: `success` and `error` webhooks which are reused across all integrations, instead of copies of static HTML files in each integration.

I tried to make the success page auto-close, but even StackOverflow claims that this is no longer possible in modern browsers.

Refs: ENG-1222


web/connect/result/handler.go
```diff
@@ -0,0 +1,60 @@
...
+func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request, t *template.Template) {
+	// Read the connection ID from the request path.
+	cid, err := sdktypes.StrictParseConnectionID(r.PathValue("id"))
+	if err != nil {
+		http.Error(w, "Bad connection ID", http.StatusBadRequest)
+		return
+	}
+
+	// Fetch the connection details.
+	conn, err := h.connections.Get(r.Context(), cid)
```
ig-ra: I think you need to apply sdk ignoreNotFound. 
Get will return notFound and ignoreNotFound will convert error to nil and object to invalid

daabr: I need to return an error if the connection doesn't exist. But if I understand you correctly, I can delete the `conn.IsValid()` check because it will always be valid if `err == nil`?

ig-ra: I meant that if you want to return two distinct errors - internal error and connection not found one, then the way to do it is to use sdkerrors.ignoreNotFound. Then notFound error will be converted to nil and obj to invalid and your code will be valid

daabr: Done

sdk/sdkintegrations/initconn.go
```diff
@@ -56,11 +56,9 @@ func (c ConnectionInit) AbortWithStatus(status int, err string) {
...
 		u = fmt.Sprintf(u, c.ConnectionID, status, url.QueryEscape(err))
 		http.Redirect(c.Writer, c.Request, u, http.StatusFound)
-	case "web":
-		http.Error(c.Writer, err, status)
 	default: // Local server ("cli", "dash", etc.)
-		u := c.Integration.ConnectionURL().Path + "/error.html?cid=%s&origin=%s&error=%s"
-		u = fmt.Sprintf(u, c.ConnectionID, origin, url.QueryEscape(err))
+		u := "/connections/%s/error?cid=%s&origin=%s&error=%s"
+		u = fmt.Sprintf(u, c.ConnectionID, c.ConnectionID, origin, url.QueryEscape(err))
 		http.Redirect(c.Writer, c.Request, u, http.StatusFound)
```
itayd, line 64: this will return a statusfound code which i don't think it's right. we should render html with the correct status code.

daabr: 302 Found is the standard HTTP status for redirection, but the target of this redirection is a regular HTML error page (with dynamic content based on the integration type). See `/web/connect/result/error.html`.

daabr: As for the status of that error page - done. I also deleted all `error.js` files on the way.

web/connect/result/error.html
```diff
@@ -20,9 +20,12 @@
       href="/static/favicon-16x16.png"
     />
     <link rel="manifest" href="/static/site.webmanifest" />
-    <link rel="stylesheet" type="text/css" href="styles.css" />
+    <link
```
efiShtain, line 23: do we really need a different css for each integration ?
aren't we plan to move all this UI code to the standard UI ?

web/connect/result/handler.go
```diff
@@ -0,0 +1,82 @@
...
+		return
+	}
+
+	if !conn.IsValid() {
+		http.Error(w, "Connection not found", http.StatusNotFound)
+		return
+	}
+
+	// Fetch the connection's integration details.
+	integ, err := h.integrations.GetByID(r.Context(), conn.IntegrationID())
```
efiShtain, line 45: don't we have getIntegrationByConnectionID? do we really need to fetch the connection just to get the integration?

sdk/sdkintegrations/initconn.go
```diff
@@ -56,11 +56,9 @@ func (c ConnectionInit) AbortWithStatus(status int, err string) {
 		u := "vscode://autokitteh.autokitteh?cid=%s&status=%d&error=%s"
 		u = fmt.Sprintf(u, c.ConnectionID, status, url.QueryEscape(err))
 		http.Redirect(c.Writer, c.Request, u, http.StatusFound)
-	case "web":
-		http.Error(c.Writer, err, status)
```
RonenMars: @daabr we need those lines, otherwise, I don't get the error correctly on our SaaS. It sends back the whole HTML of the error page redirected to.

web/connect/result/error.html
daabr: I already merged all `error.html` and `error.js` files. It's possible to do the same for the CSS files, but it requires more manual testing because they're not identical.

web/connect/result/handler.go
daabr: We don't, afaik. Also, I'm not sure it would belongs in the integration service, because it couples it to the connections service. But it's an idea to think about in a separate PR.

sdk/sdkintegrations/initconn.go
daabr: Ack. Removing these 2 lines simply means that `"web"` is handled by the `default` block below.

Specifically, it will return an HTML page of the error, but as requested by @itay it will also return the right HTTP status code - see `web/connect/result/handler.go` line 77.

RonenMars: Yeah, but this way I don't get the error, I can't take it from the HTML file.

web/connect/result/handler.go
efiShtain: opened a ticket ENG-1242
resolving this for now

sdk/sdkintegrations/initconn.go
daabr: Summary of offline discussion: bringing back these 2 lines. They are useful whenever the user is creating a non-OAuth connection via SaaS.

The difference is that non-OAuth connections don't open a new browser tab, the SaaS FE bypasses the init page, goes directly to "save", and does expect a machine-readable response. For OAuth connections, the SaaS has to open the init page in a separate browser tab (it requires user interaction) and lose any ties to it.

This means that AK will need to distinguish between OAuth and non-OAuth in "web" origin requests, and treat "web" origins for OAuth connection as non-"web", i.e. result in a user-friendly HTML page.

itay: > Ack. Removing these 2 lines simply means that `"web"` is handled by the `default` block below.
> 
> 
> 
> Specifically, it will return an HTML page of the error, but as requested by @itay it will also return the right HTTP status code - see `web/connect/result/handler.go` line 77.

FYI I think you meant a different Itay 😀

daabr: Oops! Sorry for the typo, indeed I meant @itayd. But this is a friendly neighborhood, feel free to hang around. 😄

daabr: FYI for @itayd - as a final step (in a separate PR), @RonenMars will change the FE's OAuth connection flow to redirect instead of open a new tab, and I will change AK to redirect back to the SaaS instead of AK (which currently results in responding with an HTML page).

sdk/sdkintegrations/initconn.go
itayd: where is it done?

itayd (approved):

PR #497, opened 2024-07-17
refactor(db): schema to SDK types conversion

ig-ra:


internal/backend/db/dbgorm/utils.go
```diff
@@ -0,0 +1,21 @@
+package dbgorm
+
+import (
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func schemaToSDK[M any, O sdktypes.Object](model *M, err error, parseFunc func(model M) (O, error)) (O, error) {
```
efiShtain, line 8: I think it is weird that we pass an error to this function just so it can check it and return the translated error

internal/backend/db/dbgorm/builds.go
```diff
@@ -84,11 +80,8 @@ func (db *gormdb) GetBuildData(ctx context.Context, buildID sdktypes.BuildID) ([
 }
 
 func (db *gormdb) ListBuilds(ctx context.Context, filter sdkservices.ListBuildsFilter) ([]sdktypes.Build, error) {
-	builds, err := db.listBuilds(ctx, filter)
-	if builds == nil || err != nil {
-		return nil, translateError(err)
-	}
-
+	bs, err := db.listBuilds(ctx, filter)
+	builds, err := schemasToSDK(bs, err, scheme.ParseBuild)
```
efiShtain, line 84: I think the a bit more verbose but direct approach that we currently have is better, it's more readable and very simple
schemasToSDK hides error handling that is irrelevant to the conversion from schema to sdk

internal/backend/db/dbgorm/utils.go
ig-ra: This refactoring just removes duplicated code present in almost every file in gorm - which consists of checking for model/list and error, then returning invalid object or running conversion. 4 duplicated lines in 2 places * 10 files.
Moreover in some places we were checking only for error and in some we were checking objects as well - so I think in terms of uniform handling of things it's important to remove duplicated code. Functions became more clear and easy to read.

internal/backend/db/dbgorm/builds.go
ig-ra: IMO the opposite. It's not hiding the error handling, but clarifies the function and ensures that errors and results are handled in uniform way. The same we(or actually, I) did in `toConnectError`, for example - where errors and messages were handled non-uniformly resulting in different user errors produced by the same flow but for the different objects.
So for me it's not about verbosity, but about uniformity and clarity.

tebeka: I agree that passing error to a function, other than logging or wrapping it feels "off".
The version that checks for errors is more verbose but clearer IMO:

```go
func (db *gormdb) ListBuilds(ctx context.Context, filter sdkservices.ListBuildsFilter) ([]sdktypes.Build, error) {
	bs, err := db.listBuilds(ctx, filter)
	if err != nil {
		return nil, err
	}

	builds, err := schemasToSDK(bs, scheme.ParseBuild)
	if err != nil {
		return nil, err
	}

	slices.Reverse(builds)
	return builds, nil
}
```

This quotes comes to mind:
GSiQThWXwAAnz_v

PR #498, opened 2024-07-17
fix: duplicate log records

efiShtain:
this fix assumes all log records are sequentials
and that time.micro is fine grained enough

closes ENG-1187 
might require a new design later

### DIFF
--- a/internal/backend/db/dbgorm/sessions_test.go
+++ b/internal/backend/db/dbgorm/sessions_test.go
@@ -411,7 +411,7 @@ func TestSessionLogRecordNextPageTokenNotEmpty(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, len(res.Log.Records()), 2)
 	assert.Equal(t, res.TotalCount, int64(3))
-	assert.Equal(t, res.PaginationResult.NextPageToken, "2")
+	assert.NotEmpty(t, res.NextPageToken)
 
 	// Get Next Batch to exhaust next page token
 	res, err = f.gormdb.GetSessionLog(context.Background(),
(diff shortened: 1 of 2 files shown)


itayd (approved):
comment this is a temporary fix, and we should solve it in another way with regular seq number.

the underlying issue currently is the print concurrency issue.

internal/backend/db/dbgorm/sessions_test.go
```diff
@@ -411,7 +411,7 @@ func TestSessionLogRecordNextPageTokenNotEmpty(t *testing.T) {
 	assert.NoError(t, err)
 	assert.Equal(t, len(res.Log.Records()), 2)
 	assert.Equal(t, res.TotalCount, int64(3))
-	assert.Equal(t, res.PaginationResult.NextPageToken, "2")
+	assert.NotEmpty(t, res.NextPageToken)
```
daabr, line 414: Better, but not critical for this PR: https://stackoverflow.com/questions/18970265/is-there-an-easy-way-to-stub-out-time-now-globally-during-test

daabr (approved):
Using timestamps as a monotonously-increasing counter isn't perfect in high-scale scenarios, but this is much better than before.

As far as I'm concerned it doesn't need to be considered as temporary until we actually need to shard AK or sync across multiple regions.

internal/backend/db/dbgorm/sessions_test.go
efiShtain: It's not perfect for highly concurrent scale, 
but since we limit it to a specific session, 
concurrency is basically not an option (except for the current bug we have with python prints)
so it should be ok

also, this is related to session log records, and we are not a high scale system in the workflow sense, so it's not an issue and we can take this assumption for now

PR #508, opened 2024-07-18
fix(tests): get event system test

ig-ra:


daabr (approved):
Please use correct case: https://go.dev/wiki/CodeReviewComments#initialisms

PR #509, opened 2024-07-18
feat(workflows): pass ctx

ig-ra:
- pass ctx in session- and event- workflows
- fix passing ctx in resolver
- cleanup/refactoring

PR #511, opened 2024-07-20
feat(workflow): access DB via session services

ig-ra:


ig-ra:
extracted from https://github.com/autokitteh/autokitteh/pull/509

from @daabr 

> The main thing that this PR is doing is to replace service calls with direct DB access. Unless I'm not aware of a fundamental design change, you should always prefer access via services to direct DB access.
> 
> This allows services to manage access security and safety, instead trusting the clients, and implement synchronization logic centrally that clients shouldn't be aware of.
> 
> Direct DB access violates design principles like Separation of Concerns, and usually leads to "spaghetti code", and to clashes between clients that require access safeties and other clients that ignore them.
> 
> Let's discuss this as a group, to clarify the right approach.

efiShtain:
@daabr I'm not aware of such design decision, could you clarify ?

anyhow, we can't use the services as they are bounded to sdk currently which makes everything hard to use and we can't add methods to the services layer without implementing them on the client side as well

if we want to take such a decision to go only through the services layer, we have to have internal services layer which is not exposed to the sdk (which I think we should defiantly do). We should also stop working with sdktypes internally which makes everything super hard and cumbersome. 

currently we have places that access the db layer directly **but not the DB itself** and some through the services, and most of places the consumer services get the db as well injected by FX, so this change is not something that breaks any current status

We need to open this again for discussion and decide what to do
it rises up again and again on different times so clearly what we currently have it not good enough (which is fine, we should re iterate now that we know the pains better)

itayd:
i see no reason for this change, please close.

PR #512, opened 2024-07-21
Feat(google): add support for Gmail events

daabr:
This PR generalized the implementation of Google Forms push notifications to implement Gmail push notifications as well.

Both of these integrations share the same code for event handling, but have integration-specific functions to construct `sdktypes.Event` objects and to create/update API watches.

Note: #521 is a fix stacked over this PR

Refs: implements ENG-388


integrations/google/creds.go
```diff
@@ -63,19 +64,31 @@ func (h handler) handleCreds(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
+	// Sanity check: the connection ID is valid.
+	cid, err := sdktypes.StrictParseConnectionID(c.ConnectionID)
+	if err != nil {
+		l.Warn("Invalid connection ID", zap.Error(err))
+		c.Abort("invalid connection ID")
```
itayd, line 71: can you merge warn and abort?

integrations/google/forms/events.go
```diff
@@ -0,0 +1,79 @@
...
+				// Don't abort, dispatch the event without this data.
+			}
+			formsEvent["response"] = lastResponse(responses)
+		}
+	}
+
+	// Convert the raw data to an AutoKitteh event.
+	wrapped, err := sdktypes.DefaultValueWrapper.Wrap(formsEvent)
+	if err != nil {
+		l.Error("Failed to wrap Google Forms event", zap.Error(err))
```
itayd, line 44: I think we should either log or return an error. If we return an error the caller need to log it if needed. wdyt?
See https://dave.cheney.net/tag/logging

internal/backend/httpsvc/config.go
```diff
@@ -80,6 +80,7 @@ var Configs = configset.Set[Config]{
 			UnimportantRegexes: []string{
 				`^/autokitteh.+/(Get|List)$`, // gRPC Get and List methods
 				`/(healthz|readyz)$`,         // Kubernetes health checks
+				`^/oauth/|/oauth$|/save$`,    // Connection initialization
```
itayd, line 83: shouldn't we see those in logs?

daabr: Yes - with the debug level, it's redundant as info, considering all the other intercepted requests in the init flow

integrations/google/creds.go
daabr: No, if I use the logger inside `Abort` we lose the location of the problem in the logs.

integrations/google/forms/events.go
daabr: I'm assuming you're referring to [Let’s talk about logging](https://dave.cheney.net/2015/11/05/lets-talk-about-logging)? I disagree with many of the recommendations there. They do not apply to modern software systems, even when they were written, a decade ago. :)

In this function I think it's more useful to log the specific problem that would help a developer contextualize the function's failure during debugging, instead of burying it in the stack trace of a wrapped error object that isn't visible without double-clicking on the log entry.

To be clear, this is a case-by-case decision. Other functions may not benefit from this pin-pointing, or it may be too granular during debugging - in which case a single error message - as you suggest - is more appropriate.

internal/backend/httpsvc/config.go
daabr: I would even go further and say that some of the intercepted requests in this flow shouldn't be logged at all, so we don't leak sensitive user data, but that's a task for another time - see ENG-1249.

integrations/google/forms/events.go
ig-ra: I agree with @daabr . I also support logging the problem where it happened instead/in addition to the caller

integrations/google/gmail/events.go
```diff
@@ -0,0 +1,42 @@
...
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func ConstructEvent(ctx context.Context, vars sdkservices.Vars, gmailEvent map[string]any, cids []sdktypes.ConnectionID) (sdktypes.Event, error) {
+	l := extrazap.ExtractLoggerFromContext(ctx)
+
+	// TODO(ENG-1235): Enrich the event with relevant data, using API calls.
+
+	// Convert the raw data to an AutoKitteh event.
+	wrapped, err := sdktypes.DefaultValueWrapper.Wrap(gmailEvent)
```
itayd, line 20: sdktypes.WrapValue

integrations/google/notifs.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+type gmailNotifBody struct {
+	EmailAddress string `json:"emailAddress"`
+	HistoryID    int    `json:"historyId"`
+}
+
+// handlNotif receives and dispatches asynchronous Gmail
+// notifications from a push subscription to a GCP Cloud Pub/Sub topic.
+func (h handler) handleGmailNotif(w http.ResponseWriter, r *http.Request) {
```
itayd: Notification

itayd (approved):

integrations/google/gmail/events.go
daabr: Doing this across the repo in a separate PR

integrations/google/notifs.go
daabr: done

PR #517, opened 2024-07-23
feat(users): system-orginator ownership tests

ig-ra:


internal/backend/db/dbgorm/ownership_test.go
```diff
@@ -461,6 +471,13 @@ func TestGetEnvWithOwnership(t *testing.T) {
 
 	_, err = f.gormdb.getEnvByName(f.ctx, p.ProjectID, e.Name)
 	assert.Error(t, err, sdkerrors.ErrUnauthorized)
+
+	// as system orginator
+	f.ctx = akCtx.WithRequestOrginator(f.ctx, akCtx.Dispatcher)
```
efiShtain: You are not truly testing what you mean

You only test that dispatcher origin works, it's not system user and it's not all origin options, so the test is missing cases and the comment is wrong

ig-ra: since we are checking in code that origin == user, then any other origin is testing the other option

efiShtain: But that's white box testing, you should do black box in this case

You don't know if someone would change the internal logic and your tests would still pass which is bad

ig-ra: Please suggest the way you think it should be done?

ig-ra: For example I could create a function that will randomly choose non-user originator for each test

efiShtain: I think you should exhaust all options since it's a small subset of options and it would just require few more lines to get full confident 

Otherwise we might get 100% successful tests but errors in production 

Your call

efiShtain (approved):

internal/backend/db/dbgorm/ownership_test.go
ig-ra: done

PR #518, opened 2024-07-23
feat(users): default user

ig-ra:
- move default user to auth middleware
- add config and enable default user in dev and test modes
- fix tests

fixes ENG-1254

### DIFF
--- a/internal/backend/auth/authhttpmiddleware/authmiddleware.go
+++ b/internal/backend/auth/authhttpmiddleware/authmiddleware.go
@@ -38,13 +38,19 @@ type Deps struct {
 
 func newTokensMiddleware(next http.Handler, deps Deps) http.HandlerFunc {
 	return func(w http.ResponseWriter, r *http.Request) {
-		ctx := r.Context()
-		ctx = akCtx.WithRequestOrginator(ctx, akCtx.User)
+		ctx := akCtx.WithRequestOrginator(r.Context(), akCtx.User)
+		user := authcontext.GetAuthnUser(ctx)
+		authHdr := r.Header.Get("Authorization")
 
-		if user := authcontext.GetAuthnUser(ctx); !user.IsValid() {
-			if auth := r.Header.Get("Authorization"); auth != "" {
-				kind, payload, _ := strings.Cut(auth, " ")
-
+		if deps.Cfg.AllowDefaultUser {
+			if user.IsValid() || authHdr != "" {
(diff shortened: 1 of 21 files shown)


daabr (approved):

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -34,19 +36,25 @@ type Deps struct {
...
-
-		if user := authcontext.GetAuthnUser(ctx); !user.IsValid() {
-			if auth := r.Header.Get("Authorization"); auth != "" {
-				kind, payload, _ := strings.Cut(auth, " ")
+		ctx := akCtx.WithRequestOrginator(r.Context(), akCtx.User)
+		user := authcontext.GetAuthnUser(ctx)
+		authHdr := r.Header.Get("Authorization")
 
+		if deps.Cfg.AllowDefaultUser {
+			if user.IsValid() || authHdr != "" {
```
efiShtain, line 46: Is it possible that the user you got from context is the default user ?
If so, this check would not allow the default user to work

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -34,19 +36,25 @@ type Deps struct {
...
+				http.Error(w, "only default user is allowed", http.StatusUnauthorized)
+				return
+			}
+			ctx = authcontext.SetAuthnUser(ctx, sdktypes.DefaultUser)
+		} else {
+			if !user.IsValid() && authHdr != "" {
```
efiShtain, line 52: If the user is not valid and the authHeader is empty, I think you can pass the test and continue.

Isn't it problematic?

Auth header has to be there if users are enabled (default user if off)

internal/backend/auth/authhttpmiddleware/authmiddleware.go
ig-ra: in my view - default user is not a user that could have token or could be provided externally.
We may consider change it of coz (for example create default credentials file thus providing default user in cli), but meanwhile I see no reason - and backend is handling lack of the user by attaching default user to the request

internal/backend/auth/authhttpmiddleware/authmiddleware.go
ig-ra: https://linear.app/autokitteh/issue/ENG-1266/auth-middleware-and-no-user

PR #521, opened 2024-07-25
Fix(google): save auth data before setting event watches

daabr:
Originally, the watch initialization was done after calling `c.Finalize`. However, this doesn't actually work: the effect of `c.Finalize` (a redirect to `/postinit` which saves the connection's vars) happens only after the handler functions ends.

The solution is to save the vars explicitly before the watch initialization.

This PR is stacked over #512 

Refs: ENG-1236

### DIFF
--- a/integrations/google/creds.go
+++ b/integrations/google/creds.go
@@ -70,7 +70,7 @@ func (h handler) handleCreds(w http.ResponseWriter, r *http.Request) {
 	case "", "json":
 		ctx := extrazap.AttachLoggerToContext(l, r.Context())
 		vs := sdktypes.EncodeVars(&vars.Vars{JSON: r.PostFormValue("json"), FormID: formID})
-		h.finalizeJSON(ctx, c, vs)
+		h.finalize(ctx, c, vs)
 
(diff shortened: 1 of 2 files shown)


integrations/google/creds.go
```diff
@@ -64,27 +65,12 @@ func (h handler) handleCreds(w http.ResponseWriter, r *http.Request) {
...
-		if err := forms.UpdateWatches(ctx, h.vars, cid); err != nil {
-			l.Error("Google form watches creation error", zap.Error(err))
-		}
-
-		if err := gmail.UpdateWatch(ctx, h.vars, cid); err != nil {
-			l.Error("Gmail watch creation error", zap.Error(err))
-		}
+		ctx := extrazap.AttachLoggerToContext(l, r.Context())
+		vs := sdktypes.EncodeVars(&vars.Vars{JSON: r.PostFormValue("json"), FormID: formID})
+		h.finalizeJSON(ctx, c, vs)
```
itayd: finalize

itayd (approved):

integrations/google/creds.go
daabr: done

PR #522, opened 2024-07-25
Refactor: improve connection-init abort errors

daabr:
Instead of `Abort` and `AbortWithStatusCode`, use these clearer function names: `AbortBadRequest` and `AbortISE` (Internal Server Error).

We still have `AbortWithStatusCode` in case integrations want to use other less-common HTTP status codes.

This PR is stacked over #521, but only because it touches some of the same code. It's not strictly related. Anyway, merge this PR only after #521 is merged.

### DIFF
--- a/integrations/atlassian/confluence/save.go
+++ b/integrations/atlassian/confluence/save.go
@@ -36,14 +36,14 @@ func (h handler) handleSave(w http.ResponseWriter, r *http.Request) {
 	// Check "Content-Type" header.
 	contentType := r.Header.Get(headerContentType)
 	if !strings.HasPrefix(contentType, contentTypeForm) {
-		c.Abort("unexpected content type")
+		c.AbortBadRequest("unexpected content type")
 		return
 	}
 
 	// Read and parse POST request body.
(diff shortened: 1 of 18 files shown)


integrations/atlassian/confluence/save.go
```diff
@@ -76,17 +76,17 @@ func (h handler) handleSave(w http.ResponseWriter, r *http.Request) {
 			// initData = initData.Set(webhookID, fmt.Sprintf("%d", id), false)
 
 			if err := deleteWebhook(l, b, e, t, id); err != nil {
-				l.Warn("Failed to delete existing webhook", zap.Error(err))
-				c.Abort("failed to delete existing webhook")
+				l.Error("Failed to delete existing webhook", zap.Error(err))
+				c.AbortISE("failed to delete existing webhook")
```
itayd: `AbortInternalServerError`

itayd (approved):

integrations/atlassian/confluence/save.go
daabr: done

PR #523, opened 2024-07-25
Refactor: log only intercepted HTTP responses, not requests

daabr:
This merges two log lines into one.

The diff looks confusing for no reason, the best way to review this PR is to compare the entire function before and after this PR:

Before: https://github.com/autokitteh/autokitteh/blob/main/internal/backend/httpsvc/intercept.go#L62-L106

After: `https://github.com/autokitteh/autokitteh/blob/5d3d36b40043f38d737d3ef970bfd3e29e6b1018/internal/backend/httpsvc/intercept.go#L63-L101`

### DIFF
--- a/internal/backend/httpsvc/intercept.go
+++ b/internal/backend/httpsvc/intercept.go
@@ -15,14 +15,15 @@ import (
 
 type ctxKey string
 
-var t0CtxKey = ctxKey("t0")
+var startTimeCtxKey = ctxKey("t0")
 
-func GetT0(ctx context.Context) time.Time {
-	if t0, ok := ctx.Value(t0CtxKey).(time.Time); ok {
-		return t0
+func GetStartTime(ctx context.Context) time.Time {
+	if startTime, ok := ctx.Value(startTimeCtxKey).(time.Time); ok {
+		return startTime
 	}
 
-	return time.Time{}
+	// No start time? Duration = 0, not start time = 0.
+	return time.Now()
 }
 
 type responseInterceptor struct {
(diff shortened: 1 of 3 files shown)


internal/backend/httpsvc/intercept.go
```diff
@@ -15,14 +15,15 @@ import (
 
 type ctxKey string
 
-var t0CtxKey = ctxKey("t0")
+var startTimeCtxKey = ctxKey("t0")
```
efiShtain, line 18: Thanks for the meaningful name :)

internal/backend/httpsvc/intercept.go
```diff
@@ -66,42 +67,36 @@ func intercept(z *zap.Logger, cfg *LoggerConfig, extractors []RequestLogExtracto
...
+		startTime := time.Now()
+		r = r.WithContext(context.WithValue(r.Context(), startTimeCtxKey, startTime))
 
 		next.ServeHTTP(rwi, r)
-		d := time.Since(t0)
+		d := time.Since(startTime)
 
 		w.Header().Add("X-AutoKitteh-Duration", d.Truncate(time.Microsecond).String())
 
+		level := cfg.ImportantLevel.Level()
```
efiShtain, line 83: Not sure what are you trying to do here

Could you elaborate?

daabr: The diff looks confusing for no reason, the best way to review this PR is to compare the entire function before and after - and then you see that I basically cancelled the first log entry, and moved its fields to the second log entry.

daabr: I updated the PR description with relevant links

internal/backend/httpsvc/intercept.go
itayd: you people suck

itayd (approved):

PR #526, opened 2024-07-25
feat(db): update project name

ig-ra:


internal/backend/db/dbgorm/projects.go
```diff
@@ -110,7 +110,10 @@ func (gdb *gormdb) updateProject(ctx context.Context, p *scheme.Project) error {
 		if err := tx.isCtxUserEntity(ctx, p.ProjectID); err != nil {
 			return err
 		}
-		return tx.db.Updates(p).Error
+
+		data := map[string]any{"Name": p.Name}
+		allowedFields := []string{"Name"}
+		return tx.db.Model(&scheme.Project{ProjectID: p.ProjectID}).Select(allowedFields).Updates(data).Error
```
efiShtain, line 116: Could you explain what are you trying to do here?
Why not just regular update(column,value) or update(object) ?

efiShtain:
## TL;DR

DB layer should not enforce any business logic 

## Full Details

The **db**  can and should enforce relations / uniqueness using the indexing/constraints systems of Postgres
The **db layer**  should not enforce any business logic like restricting which field it allows to update (unless the fields are managed by the db layer itself for data consistency or something like that)

The decision on which fields to update are business logic, they belong in the application layer (aka services in our case, projects service specifically)

### Problem with the suggested approach
- The projects layer is meaningless, since even if the project layer wants/need to perform an update on which ever field, the db layer takes a proactive approach and limit this update which kind of beats the purpose of separation of layers

@daabr / @itayd  / @tebeka you're take on it?

ig-ra:
> ## TL;DR
> DB layer should not enforce any business logic
> 
> ## Full Details
> The **db** can and should enforce relations / uniqueness using the indexing/constraints systems of Postgres The **db layer** should not enforce any business logic like restricting which field it allows to update (unless the fields are managed by the db layer itself for data consistency or something like that)
> 
> The decision on which fields to update are business logic, they belong in the application layer (aka services in our case, projects service specifically)
> 
> ### Problem with the suggested approach
> * The projects layer is meaningless, since even if the project layer wants/need to perform an update on which ever field, the db layer takes a proactive approach and limit this update which kind of beats the purpose of separation of layers
> 
> @daabr / @itayd / @tebeka you're take on it?

The project model defined as follows
```
type Project struct {
	ProjectID sdktypes.UUID `gorm:"primaryKey;type:uuid;not null"`
	Name      string        `gorm:"index"`
	RootURL   string
	Resources []byte
	DeletedAt gorm.DeletedAt `gorm:"index"`
	Ownership *Ownership     `gorm:"polymorphic:Entity;"`
}
```
1. Obviously the DB layer should ensure that neither ProjectID, DeletedAt and Ownership could be updated. Therefore they should be banned. This is an internal function of db layer and it should preserve DB level consistency, it receives scheme (with all fields including all those that should be banned from updating) and should do the right thing, e.g. allow only fields that could be updated. 
2. So it should allow updating of only Name, RootURL and Resources
3. But to update resources we have different API - i.e. SetResources, so UpdateProject API should allow only updating of the Name and RootUrl and resources should be banned from updating as well in this function. Otherwise we have different double logic to update resources.
4. RPC message for update Project contain only Project msg struct, which is `project_id` and `name`. Thus Update project API could update only name and not `root url`. That's why I allowed only name, but I could allow rootUrl as well, since this definitely shouldn't be enforced on DB layer and if someone changed RPC/message code it would work. 
5. Explicit is better than implicit - therefore I just marked it clearly in code that only name could be updated

efiShtain:
I agree about explicit, I just don't agree that this should be at the db layer

it's the responsibility of the project's service to not override data in the database, it it wants to wipe it out - it should be allowed to do it

you can block updating the deletedAt/updatedAt/createdAt field since the db manage them, that's fine

finally Db layer shouldn't have any knowledge on the outside, so it doesn't matter what we have on RPC or other layers, the db layer should be agnostic to it

ig-ra:
> I agree about explicit, I just don't agree that this should be at the db layer
> 
> it's the responsibility of the project's service to not override data in the database, it it wants to wipe it out - it should be allowed to do it
> 
> you can block updating the deletedAt/updatedAt/createdAt field since the db manage them, that's fine
> 
> finally Db layer shouldn't have any knowledge on the outside, so it doesn't matter what we have on RPC or other layers, the db layer should be agnostic to it

in order to wipe data, e.g. delete project or it's resources - there should be one clear path to do this - and we have an DB layer API (e.g. SetResources and DeleteProject). So there shouldn't be any hidden tricks to do this not via API, that's why updateProject should disallow it.

efiShtain:
The only way to delete projects and its dependencies should be from the projects service, not from the the db layer 

Db layer exposes delete project
Delete env
Delete sessions
And so on

Each service is responsible for calling this db capabilities in the order it knows it should 

The services we have are not APIs (although they have strong coupling to the api), the grpcsvc are APIs 


It's the service's responsibility to decide if it should change the user input or do more stuff or limit updates 

It's the db layer responsibility to translate service request to storage request, add db only logic like auto increment or manage soft delete, manage connection pooling, reconnection, multiple databases, etc

Currently it uses gorm in our case but it's just an abstraction of gorm, not another layer above gorm. Gorm is not the actual db layer, it's just a convenient instead of writing raw sql. Technically we can remove it, use sql directly and it should work the same.

We should be able to switch the db layer to another db as long as the api stays the same and it should work the same. That's the purpose of the db layer abstraction, abstract how data is stored and retrieved and provide this capabilities to store the read.

Exactly like the grpc layer shouldn't decide to call any service that it wants, it should translate grpc to service call and service response to grpc

App logic should be in one place which is the services layer

I think the main issue we have is the current meaning we have for each layer, I'll show you some examples and let's see if we can align on it better

ig-ra:
> The only way to delete projects and its dependencies should be from the projects service, not from the the db layer
> 
> Db layer exposes delete project Delete env Delete sessions And so on
> 
> Each service is responsible for calling this db capabilities in the order it knows it should
> 
> The services we have are not APIs (although they have strong coupling to the api), the grpcsvc are APIs
> 
> It's the service's responsibility to decide if it should change the user input or do more stuff or limit updates
> 
> It's the db layer responsibility to translate service request to storage request, add db only logic like auto increment or manage soft delete, manage connection pooling, reconnection, multiple databases, etc
> 
> Currently it uses gorm in our case but it's just an abstraction of gorm, not another layer above gorm. Gorm is not the actual db layer, it's just a convenient instead of writing raw sql. Technically we can remove it, use sql directly and it should work the same.
> 
> We should be able to switch the db layer to another db as long as the api stays the same and it should work the same. That's the purpose of the db layer abstraction, abstract how data is stored and retrieved and provide this capabilities to store the read.
> 
> Exactly like the grpc layer shouldn't decide to call any service that it wants, it should translate grpc to service call and service response to grpc
> 
> App logic should be in one place which is the services layer
> 
> I think the main issue we have is the current meaning we have for each layer, I'll show you some examples and let's see if we can align on it better

we are slipping to discussing totally different thing, Efi :) We had this discussion yesterday and agreed that we should aim to create and maintain the meaning of each layer, since right now service layer is meaningless for most. 

But it's beyond the scope of this particular PR I believe. let's assume for simplicity that `deleteProject()` in gorm layer is deleting the project only and we already reached our goal for splitting gorm/services layer, ok?
So whether updateProject should allow updating DeletedAt? e.g. could we delete or undelete project via `updateProject` API? I definitely think that no, there is `deleteProject` API for this. Therefore, gorm layer in its internal functions should keep its own consistency, means `updateProject` shouldn't allow updating DeletedAt field, for example. In other words, `updateProject` should allow updating of updatable fields without introducing any user- or app-related logic. Which ends up in decision to allow update to user- or app- updatable fields, since other fields are not user/app accessible.

itayd:
the db layer should care only about what its schema allows. it should not allow to do "illegal" stuff, like changing the id of the object.

sdktypes should enforce what a legal object is.

anything higher level than that should be done in the service level.

itayd:
i'm not saying this possible right now, but that's what we should aspire to.

itayd (approved):

PR #527, opened 2024-07-25
ENG-1253 remove redundant jira client

pashafateev:


daabr (commented):
Also need to update these files:
- `runtimes/pythonrt/py-sdk/pyproject.toml`
- `runtimes/pythonrt/py-sdk/docs/conf.py` (SDK version)
- `runtimes/pythonrt/py-sdk/docs/index.rst`
- `runtimes/pythonrt/py-sdk/pyproject.toml` (lib list + SDK version)
- `runtimes/pythonrt/dev-requirements.txt`
- `runtimes/pythonrt/requirements.txt`

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
```diff
@@ -5,7 +5,6 @@
 import os
 
 from atlassian import Confluence, Jira
-from jira import JIRA
```
tebeka, line 8: What's the different between `Jira` and `JIRA`? Who added it in the first time and why?

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
```diff
@@ -15,7 +14,7 @@
 __TOKEN_URL = "https://auth.atlassian.com/oauth/token"
 
 
-def atlassian_jira_client(connection: str, **kwargs) -> Jira:
+def jira_client(connection: str, **kwargs) -> Jira:
```
tebeka, line 17: This is an API breaking change. 
Do we know if there's a current client who's using it?
If so, keep the old function as well, something like:

```python
import warnings

def atlassian_jira_client(connection: str, **kwargs) -> Jira:
    warnings.warn("`attlasian_jira_client` is deprecated, use `jira_client`", DeprecationWarning)
    return jira_client(connection, **kwargs)

```

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
```diff
@@ -15,7 +14,7 @@
...
     """Initialize an Atlassian Jira client, based on an AutoKitteh connection.
 
     API reference:
```
tebeka, line 20: Missing docstring for the arguments, see the original docstring.

tebeka (changes requested):

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
pashafateev: @daabr I don't think anyone but us is using this correct?

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
daabr: I added 2 Atlassian client libs that seemed to be the most popular ones - one only for Jira (client is `jira.JIRA`) and the other is a lib for multiple Atlassian products (specifically Confluence, but also Jira - `atlassian.Jira` and `atlassian.Confluence` respectively).

After experiments and actual usage, we learned that the Jira-only lib is less suited for our auth needs, so this PR removes it in favor of using only the atlassian lib for Jira too, not just Confluence.

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
daabr: We have one customer who requested the Jira integration, but they don't have any projects that use it yet. The only places that reference this are our projects in the `samples` and `kittehub` repos - and there are PRs to fix them, which will be merged as soon as this PR is merged.

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
pashafateev: I don't understand what you mean. Missing docstring in `jira_client` ?
Do you mean this?

>    Args:
        connection: AutoKitteh connection name.

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
tebeka: OK

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
tebeka: Go ahead then.

runtimes/pythonrt/py-sdk/autokitteh/atlassian.py
tebeka: Ah, github diff clipped this. Please document `kwargs`

tebeka (approved):

PR #529, opened 2024-07-27
feat(db): project: not nil name

ig-ra:


tebeka (approved):
LGTM, would love to see a test.

ig-ra:
> LGTM, would love to see a test.

@tebeka which test? that we cannot create project with null name? 
I'm unsure that we should test the constrain (e.g. `not nil`) is implemented correctly gorm. 
I think that we should check only our logic on top of gorm in test and assume that gorm handles it's own things properly.
WDYT?

just to make it clear - it's not a problem to write a test. The question is whether we should or we could assume that gorm is OK

PR #530, opened 2024-07-27
chore: don't log some HTTP requests, unless there's an error

daabr:
We're logging about 720 `/healthz` requests per hour. This is both unhelpful and expensive - so don't log them at all.

However, obviously we do want to log them if there's an error.

### DIFF
--- a/internal/backend/httpsvc/intercept.go
+++ b/internal/backend/httpsvc/intercept.go
@@ -60,11 +60,6 @@ func intercept(z *zap.Logger, cfg *LoggerConfig, extractors []RequestLogExtracto
 	unlogged := regexp.MustCompile(strings.Join(cfg.UnloggedRegexes, `|`))
 
 	return func(w http.ResponseWriter, r *http.Request) {
-		if unlogged.MatchString(r.URL.Path) {
-			next.ServeHTTP(w, r)
-			return
-		}
-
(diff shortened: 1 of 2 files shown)


internal/backend/httpsvc/intercept.go
```diff
@@ -60,26 +60,29 @@ func intercept(z *zap.Logger, cfg *LoggerConfig, extractors []RequestLogExtracto
...
 		w.Header().Set("Trailer", "X-AutoKitteh-Duration")
 
+		// Call the next handler in the chain, and calculate the duration.
 		startTime := time.Now()
 		r = r.WithContext(context.WithValue(r.Context(), startTimeCtxKey, startTime))
 
 		next.ServeHTTP(rwi, r)
-		d := time.Since(startTime)
 
+		d := time.Since(startTime)
```
efiShtain: What's d? Duration?
Better to have endTime

efiShtain (approved):

internal/backend/httpsvc/intercept.go
daabr: We're logging the duration, not the end time, so having a variable for that is an unnecessary extra step.

The return type of `time.Since()` is `time.Duration` so `d` is a standard variable name. But I don't mind renaming to `duration` - done.

PR #532, opened 2024-07-28
feat(logger)!: configure min logging level by name, not ID

daabr:
The default minimum logging level is [Info](https://pkg.go.dev/go.uber.org/zap#pkg-constants). To change this, set the following config value to the lower-case or all-caps name of the level you want.

For example:

- CLI: `ak config set logger.level debug`

- Config YAML file:

  ```yaml
  logger:
    level: debug
  ```

- Environment variable: `AK_LOGGER__LEVEL=DEBUG`

Per @itayd and according to the principle of least surprise, invalid names result in an error.

BREAKING CHANGE: Expected values are now strings instead of integers! This PR still accepts old numeric values (e.g. -1 = debug), but this is temporary and will be removed after all existing servers migrate.

### DIFF
--- a/internal/backend/logger/logger.go
+++ b/internal/backend/logger/logger.go
@@ -1,6 +1,7 @@
 package logger
 
 import (
+	"strconv"
 	"time"
 
 	"go.uber.org/zap"
@@ -37,11 +38,17 @@ func (onPanicHook) OnWrite(ce *zapcore.CheckedEntry, fs []zapcore.Field) {
 }
 
 func New(cfg *Config) (*zap.Logger, error) {
-	// Optional override for the default level (info).
+	// Optional override for the default level (0 = info).
 	if cfg.Level != "" {
+		// Accept lower-case or all-caps level names, as defined by Zap.
 		level, err := zapcore.ParseLevel(cfg.Level)
 		if err != nil {
-			return nil, err
+			// Temporary: numeric level IDs, as used internally by Zap.
+			if n, e := strconv.Atoi(cfg.Level); e == nil {
+				level = zapcore.Level(n)
+			} else {
+				return nil, err
+			}
 		}
 		cfg.Zap.Level.SetLevel(level)
 	}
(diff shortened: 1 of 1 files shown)


internal/backend/logger/logger.go
```diff
@@ -13,7 +13,7 @@ import (
 
 type Config struct {
 	Zap      zap.Config `koanf:"zap"`
-	Level    int        `koanf:"level"`    // -1 = Debug, 0 = Info, etc.
+	Level    string     `koanf:"level"`    // Lower-case or all caps, default = "info".
```
tebeka, line 17: Are we OK with breaking schema change? If not maybe call it `LevelName` and keep `Level` around.

internal/backend/logger/logger.go
```diff
@@ -37,7 +37,14 @@ func (onPanicHook) OnWrite(ce *zapcore.CheckedEntry, fs []zapcore.Field) {
 }
 
 func New(cfg *Config) (*zap.Logger, error) {
-	cfg.Zap.Level.SetLevel(zapcore.Level(cfg.Level)) // Default = 0 = Info.
+	// Optional override for the default level (info).
```
tebeka: Where is the default level defined? By zap?

tebeka (approved):
Two minor comments, feel free to merge.

internal/backend/logger/logger.go
daabr: Excellent point. I forgot to mention this in the PR title & description.

Originally I thought about this as a breaking change, but I can also accept numbers (as before) as a temporary fallback.

internal/backend/logger/logger.go
daabr: Yes. Zap uses numbers (https://pkg.go.dev/go.uber.org/zapcore#InfoLevel), and if the field isn't set then the default value is 0, i.e. InfoLevel.

PR #533, opened 2024-07-28
feat(python): next_event support several subscriptions.

tebeka:
Fixes ENG-1209

### DIFF
--- a/runtimes/pythonrt/py-sdk/autokitteh/events.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/events.py
@@ -17,7 +17,7 @@ def unsubscribe(subscription_id: str) -> None:
     pass
 
 
-def next_event(subscription_id: str) -> AttrDict:
-    """Get the next event from the subscription."""
+def next_event(subscription_id: str, *subscription_ids: list[str]) -> AttrDict:
+    """Get the next event from the subscription(s)."""
     # Dummy implementation for local development.
     return AttrDict()


daabr:
Can you explain how this change will be used by users? Not this function itself, which is a dummy for local testing, but the API that it interacts with - what's the difference between `sub_id` and `sub_ids`? Or is this just a halfway to something else?

tebeka:
> Can you explain how this change will be used by users? Not this function itself, which is a dummy for local testing, but the API that it interacts with - what's the difference between `sub_id` and `sub_ids`? Or is this just a halfway to something else?

The user can either do `ak.next_event('sig1')` or `ak.next_event('sig1', 'sig2')`
The way Python sends the call to `ak` is to pack all arguments so there's no need to change the code `pythonrt.go`.
`sub_ids` are the rest of IDs, I want to force the user to specify at least one signal.

daabr:
> The user can either do `ak.next_event('sig1')` or `ak.next_event('sig1', 'sig2')` The way Python sends the call to `ak` is to pack all arguments so there's no need to change the code `pythonrt.go`. `sub_ids` are the rest of IDs, I want to force the user to specify at least one signal.

In that case, maybe call them `required_sub_id` and `optional_sub_ids` as a form of self-documentation? Or add per-arg descriptions in the docstring?

It's not a blocker, so I'm approving, but given that it's visible to users I expect it to be understandable by any user without our help.

daabr (approved):

itayd:
@daabr - this is how it works in starlark as well.

PR #534, opened 2024-07-28
bug(python): node error in name_of

tebeka:
Rewrote `name_of` to user source code, these errors should be gone.

Fixes ENG-1260

### DIFF
--- a/runtimes/pythonrt/tests/loader_test.py
+++ b/runtimes/pythonrt/tests/loader_test.py
@@ -72,23 +72,6 @@ def test_load_simple():
     ak_runner.load_code(root_path, action_fn, "simple")
 
 
-name_of_cases = [
-    # code, name
-    ("print(1)", "print"),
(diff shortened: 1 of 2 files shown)


runtimes/pythonrt/tests/loader_test.py
```diff
@@ -72,23 +72,6 @@ def action_fn(*args, **kw):
     ak_runner.load_code(root_path, action_fn, "simple")
 
 
-name_of_cases = [
```
ig-ra, line 75: just to ensure - you don't need this test, since this covered in test_transform, right?

ig-ra (approved):

runtimes/pythonrt/tests/loader_test.py
tebeka: Actually no, `name_of` is shown only in logging. However the new version returns the code `as-is` and I thought the test was redundant.

PR #535, opened 2024-07-29
docs: proposal for specific logging guidelines

daabr:
This does not replace or modify the logging RFC, it just focuses on practical implementation details - e.g. for future code reviews.

WDYT?

Waiting for feedback from all reviewers, not merging after the first approval.

### DIFF
--- a/docs/logging.md
+++ b/docs/logging.md
@@ -1,0 +1,204 @@
+# Contributor Guidelines - Logging
+
+## Framework
+
+We use structured logging, and specifically we use [Zap](https://github.com/uber-go/zap)
+as our logging framework.
+
+Why not something else?
+
+- [Zerolog](https://github.com/rs/zerolog) is more performant and better
+  looking, but its syntax is more error-prone (no logging in case you forget
+  to call `Msg("...")` at the end)
+- Go's [Slog](https://pkg.go.dev/log/slog) is new, slower (according to Zap
+  benchmarks), and doesn't provide useful benefits
+
+## Default Level
+
+The default minimum logging level is [Info](https://pkg.go.dev/go.uber.org/zap/zapcore#InfoLevel).
+To change this, set the following AutoKitteh config value to the lower-case or
+all-caps name of the level you want. For example:
+
+- CLI: `ak config set logger.level debug`
+
+- Config YAML file:
+
+  ```yaml
+  logger:
+    level: debug
+  ```
+
+- Environment variable: `AK_LOGGER__LEVEL=DEBUG`
+
+## Logger Naming
+
+- Local variables: `l` is preferable when its meaning is clear, otherwise
+  `logger` is acceptable too
+
+  - `z` is discouraged - we use Zap, but we don’t want the code to be tied to
+    that assumption
+  - Other combinations of `l` and `logger` are possible for code readability,
+    but not recommended in general
+
+- Struct fields: `logger` or `Logger`
+
+## Logger Initialization
+
+> [!NOTE]
+> TODO: Refactor `extrazap`
+>
+> TODO: Godoc link below
+
+Pass a logger reference (`*zap.Logger`) as a function parameter, or use
+[extrazap](/TODO) to initialize a logger instance:
+
+```go
+func one(l *zap.Logger, ...) {
+    l.Info("...")
+
+    ctx := extrazap.AttachLoggerToContext(context.Background(), l)
+    two(ctx, ...)
+}
+
+func two(ctx context.Context, ...) {
+    l := extrazap.ExtractLoggerFromContext(ctx)
+    l.Info("...")
+}
+```
+
+You can use `extrazap` even if your function doesn't have a context with an
+embedded logger:
+
+```go
+func three() {
+    l := extrazap.ExtractLoggerFromContext(context.Background())
+    l.Info("...")
+}
+```
+
+In other words, you **should** attach a logger to a context before passing it
+to a function that uses `extrazap`, but it will work even if you don't.
+
+Because of that, there's no need to call [zap.L()](https://pkg.go.dev/go.uber.org/zap#L)
+directly.
+
+Anyway, do not pass both a context and a logger as function parameters, this
+is wasteful.
+
+## Log Messages
+
+1. Should be short
+   - Avoid complex or lengthy sentences
+   - Prefer adding details as fields, not in the message string
+2. Should be informative - if the message is too generic or vague, add vital
+   details to it (in addition to adding them as fields)
+3. Should begin with a capital letter
+   - This is unlike Go error strings, but for the
+     [same reason](https://go.dev/wiki/CodeReviewComments#error-strings)
+4. Shouldn’t end with punctuation
+   - This is like Go error strings, but only to support (1) above
+
+## Adding Log Fields to the Logger
+
+As noted above, we always use loggers by reference (`*zap.Logger`), not by
+value (`zap.Logger`).
+
+Because of that, when you call [Logger.With()](https://pkg.go.dev/go.uber.org/zap#Logger.With)
+in a nested code block, redeclare the logger with the `:=` operator:
+
+```go
+l := l.With(...)
+```
+
+Instead of just reassigning it with the `=` operator:
+
+```go
+l = l.With(...)
+```
+
+Whenever possible, this ensures that the added fields don't remain in the
+logger after leaving exiting the code block.
+
+Example:
+
+```go
+func foo(l *zap.Logger) {
+	l.Info("...", zap.String("aaa", "111"))
+
+    // Good: "bbb" doesn't remain after the loop's block.
+	for _, s := range []string{"ccc"} {
+		l := l.With(zap.String("bbb", "222"))
+		l.Info("...", zap.String(s, "333"))
+	}
+
+    // Bad: "ddd" remains after the condition's block.
+    if ... {
+		l := l.With(zap.String("ddd", "444"))
+		l.Info("...", zap.String("eee", "555"))
+    }
+
+	l.Info("...", zap.String("fff", "666"))
+}
+```
+
+## Other Recommendations
+
+Usually, you should write logs as close as possible to the reason for logging.
+
+For example, this:
+
+```go
+func foo() {
+    // ...
+    if err := bar(); err != nil {
+        return
+    }
+    // ...
+}
+
+func bar() error {
+    if problem1 {
+        l.Error("Boom")
+        return errors.New("problem 1")
+    }
+
+    if err := problem2(); err != nil {
+        l.Error("Pow", zap.Error(err))
+        return fmt.Errorf("problem 2: %w", err)
+    }
+
+    l.Info("Successful bar")
+    return nil
+}
+```
+
+Is better than this:
+
+```go
+func foo() {
+    // ...
+    if err := bar(); err != nil {
+        l.Error("Bar failed", zap.Error(err))
+        return
+    }
+    l.Info("Successful bar")
+    // ...
+}
+
+func bar() error {
+    if problem1 {
+        return errors.New("problem 1")
+    }
+
+    if err := problem2(); err != nil {
+        return fmt.Errorf("problem 2: %w", err)
+    }
+
+    return nil
+}
+```
+
+Unless you actually want all the different error conditions inside `bar()` to
+result in the same common log entry, to keep `bar()` as a "black box".
+
+Anyway, use your judgement.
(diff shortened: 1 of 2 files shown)


docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+Whenever possible, this ensures that the added fields don't remain in the
+logger after leaving exiting the code block.
+
+Example:
+
+```go
+func foo(l *zap.Logger) {
+	l.Info("...", zap.String("aaa", "111"))
```
itayd, line 118:
```
l.Info(fmt.Sprintf("...", a), zap.String("a", a))
```

->

```
l.Info("... {a} ...", zap.Any("a", a))
```

?

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+
+We use structured logging, and specifically we use [Zap](https://github.com/uber-go/zap)
+as our logging framework.
+
+Why not something else?
+
+- [Zerolog](https://github.com/rs/zerolog) is more performant and better
+  looking, but its syntax is more error-prone (no logging in case you forget
+  to call `Msg("...")` at the end)
+- Go's [Slog](https://pkg.go.dev/log/slog) is new, slower (according to Zap
```
tebeka: The biggest benefit is that's in the standard library.

docs/logging.md
```diff
@@ -0,0 +1,204 @@
+# Contributor Guidelines - Logging
+
+## Framework
+
+We use structured logging, and specifically we use [Zap](https://github.com/uber-go/zap)
+as our logging framework.
+
+Why not something else?
```
tebeka: There's also logrus.
I don't think we need to justify which library we use, maybe down in a FAQ. IMO you can remove the `Why not something else?` section.

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+
+- [Zerolog](https://github.com/rs/zerolog) is more performant and better
+  looking, but its syntax is more error-prone (no logging in case you forget
+  to call `Msg("...")` at the end)
+- Go's [Slog](https://pkg.go.dev/log/slog) is new, slower (according to Zap
+  benchmarks), and doesn't provide useful benefits
+
+## Default Level
+
+The default minimum logging level is [Info](https://pkg.go.dev/go.uber.org/zap/zapcore#InfoLevel).
```
tebeka:
```suggestion
The default logging level is [Info](https://pkg.go.dev/go.uber.org/zap/zapcore#InfoLevel).
```

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+To change this, set the following AutoKitteh config value to the lower-case or
+all-caps name of the level you want. For example:
```
tebeka: Why can't we accept any case?

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+  ```yaml
+  logger:
+    level: debug
+  ```
+
+- Environment variable: `AK_LOGGER__LEVEL=DEBUG`
+
+## Logger Naming
+
+- Local variables: `l` is preferable when its meaning is clear, otherwise
```
tebeka, line 27: I don't like `l`, it is easily confused with `I` or `1`. See [Lower Case l Looks a Lot Like the Digit 1](https://cs.fit.edu/~kgallagher/Schtick/How%20To%20Write%20Unmaintainable%20Code.html)

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+  `logger` is acceptable too
+
+  - `z` is discouraged - we use Zap, but we don’t want the code to be tied to
```
tebeka, line 30: This will require a lot of code changes. I counted 63 places in the code that use `z`.

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+> [!NOTE]
+> TODO: Refactor `extrazap`
+>
+> TODO: Godoc link below
+
+Pass a logger reference (`*zap.Logger`) as a function parameter, or use
+[extrazap](/TODO) to initialize a logger instance:
+
+```go
+func one(l *zap.Logger, ...) {
```
tebeka, line 48: Find more meaningful function names.  Same for `three` below.

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+func three() {
+    l := extrazap.ExtractLoggerFromContext(context.Background())
+    l.Info("...")
+}
+```
+
+In other words, you **should** attach a logger to a context before passing it
+to a function that uses `extrazap`, but it will work even if you don't.
+
+Because of that, there's no need to call [zap.L()](https://pkg.go.dev/go.uber.org/zap#L)
```
tebeka:
```suggestion
Because of that, you shouldn't to call [zap.L()](https://pkg.go.dev/go.uber.org/zap#L)
```

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+
+Anyway, do not pass both a context and a logger as function parameters, this
+is wasteful.
+
+## Log Messages
+
+1. Should be short
+   - Avoid complex or lengthy sentences
+   - Prefer adding details as fields, not in the message string
+2. Should be informative - if the message is too generic or vague, add vital
```
tebeka, line 85: "If logs don't help you during development, they won't help you at 4am in production." - Bill Kennedy

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+4. Shouldn’t end with punctuation
+   - This is like Go error strings, but only to support (1) above
+
+## Adding Log Fields to the Logger
+
+As noted above, we always use loggers by reference (`*zap.Logger`), not by
+value (`zap.Logger`).
+
+Because of that, when you call [Logger.With()](https://pkg.go.dev/go.uber.org/zap#Logger.With)
+in a nested code block, redeclare the logger with the `:=` operator:
```
tebeka, line 99: Show a full example. IMO this doesn't happen a lot, most cases the same logger will be use throughout the function.

docs/logging.md
tebeka, line 118:
```suggestion
	l.Info("...", zap.String("user", "joe"))
```

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+
+    // Good: "bbb" doesn't remain after the loop's block.
+	for _, s := range []string{"ccc"} {
+		l := l.With(zap.String("bbb", "222"))
+		l.Info("...", zap.String(s, "333"))
+	}
+
+    // Bad: "ddd" remains after the condition's block.
+    if ... {
+		l := l.With(zap.String("ddd", "444"))
```
tebeka: You mean `l = l.With...`? Otherwise what you write here is wrong.

docs/logging.md
```diff
@@ -0,0 +1,204 @@
...
+```
+
+## Other Recommendations
+
+Usually, you should write logs as close as possible to the reason for logging.
+
+For example, this:
+
+```go
+func foo() {
```
tebeka, line 148: Realistic examples please.

tebeka (commented):
See my comments. Other than that I like Bill Kennedy's approach where only the application and business level log and the foundation layer is now allowed to log.

docs/logging.md
daabr: I thought these specific 2 deserve mentioning, but yeah, I'll remove this paragraph.

docs/logging.md
daabr: I'll remove this paragraph per the previous comment, but here's what I think about Slog:

We both agree that this is the biggest benefit that slog has to offer. :)
My point is that it's not a useful benefit in our case.

A couple of quotes from https://go.dev/blog/slog:
- "replacing existing code that works well is rarely a good use of a developer’s time"
- "existing logging packages can talk to a common backend, so the packages that use them can interoperate without having to be rewritten"

So yeah, if we use more than a single logging framework, or when we need to switch from Zap, we should switch to Slog and rewrite all our log lines once and never again - but I prefer to delay that as much as possible.

Last but not least, I have to admit I was put off by the design process story in https://go.dev/blog/slog:

> Many felt strongly that this was a bad idea. They found it hard to read and easy to get wrong
> [...]
> But we felt that the lighter syntax was important to keeping Go easy and fun to use
> [...]
> We added a [vet check](https://pkg.go.dev/golang.org/x/tools/go/analysis/passes/slog) to catch common mistakes, but did not change the design.

docs/logging.md
daabr: That's what Zap expects, but I fixed that now

docs/logging.md
daabr: Not with modern serif fonts. Or is it just my OCD? 😄

Anyway, personally I don't mind using only `logger`, but I see many cases where Go's strong preference for short var names makes sense:

- https://go.dev/wiki/CodeReviewComments#variable-names
- https://go.dev/talks/2014/names.slide#1 - slides 6 and 9

docs/logging.md
daabr: Ack. We can either agree to enforce this only going forward, or pay this debt now as an early part of a future migration to Slog. Either way works for me.

FYI, in my experience, large-scale refactors are easy when you can use a simple search term for them, like this: https://github.com/search?q=repo%3Aautokitteh%2Fautokitteh+%2Fz%5Cs%2B%5C*zap%2F&type=code

I'm not searching for mentions of "z", only declarations, because renaming a symbol in VS Code fixes all instances.

docs/logging.md
daabr: Totally! But I prefer writing a more specific criterion here (which is usable in code reviews) over the abstract principle behind it.

docs/logging.md
daabr: @itayd can you clarify what you mean? There's no `Sprintf` for the message in this code snippet, only Zap fields.

Also, I didn't mention it in this doc, but users may consider Zap's `SugaredLogger` if/when they do want to use string formatting in log messages (which is relevant for previous section (`## Log Messages`).

docs/logging.md
daabr: Oops, thanks

docs/logging.md
itayd: yeah sorry. this was a suggestion i made during sync to allow adding "f-string" capability to logging. you can ignore for now.

docs/logging.md
itayd: I traditionally use `z` for Zap, but I can see why some people are irked by that. I think `logger` is too long. `l` makes sense since it always written as `l.Info` or some such, so not so easily confused with 1 or I.

docs/logging.md
itayd: I think having `z` is a good marker for what we haven't overviewed yet :) WIll use `l` from now on.

docs/logging.md
itayd: I preffer using cat motifs.

docs/logging.md
```diff
@@ -0,0 +1,201 @@
...
+is wasteful.
+
+## Log Messages
+
+1. Should be short
+   - Avoid complex or lengthy sentences
+   - Prefer adding details as fields, not in the message string
+2. Should be informative - if the message is too generic or vague, add vital
+   details to it (in addition to adding them as fields)
+3. Should begin with a capital letter
```
itayd, line 87: I don't like this

itayd:
We should add guidelines for field names. I think we should go snake style

itayd:
we should also consider using sugared loggers - seems it have `Logf` methods.

itayd:
when should we use temporal's workflow logger and when not. can we combine the two?

docs/logging.md
daabr: "this" = 3? Or something else? If 3 - what do you have to say about https://go.dev/wiki/CodeReviewComments#error-strings ?

daabr:
> We should add guidelines for field names. I think we should go snake style

Good point. Personally I've been methodical is using "camelCase" - we can discuss this F2F.

efiShtain:
I vote to remove all the fine details and stick with a simple approach:
- logs should output json
- log line should be readable and can have attached attributes, but the log line should be self describing
```json
{
"msg": "this is a log message for sessionId 'id', something happened",
"session_id": id,
"workflow_id":wid,
... more attributes if needed
}
```
- Error for real application errors 
- Warn for "errorish" errors, like not found or slow query and such
- Info for most of the logs
- Debug for low level debug logs - will be disabled in prod most of the times
- Log what happened not future possible cases

in my opinion, we won't get much value from too many strict conventions right now
the holistic approach is readability and basic level (if it's a critical error, debug is not the level)

tebeka:
> I vote to remove all the fine details and stick with a simple approach:
Agree

PR #537, opened 2024-07-30
feat(ENG-1274): add `required` to vars

itayd:
Add an optional `required` field to vars. This allows to specify in the manifest that a variable is required, and if a deploy attempt is made without this variable being set, it will fail. This is helpful when we want to specify required variables in manifest in kittehub, but without giving any default value. This will force the user to explicitly set these to deploy.

This also renames `is_secret` to `secret` for said vars, for consistency.


itayd:
btw, I discussed this with @haimzlato and @daabr previously and they seem supportive. If any objection, please lmk.

cmd/ak/cmd/vars/set.go
```diff
@@ -44,7 +44,7 @@ var setCmd = common.StandardCommand(&cobra.Command{
 			value = string(data)
 		}
 
-		ev := sdktypes.NewVar(n, value, secret).WithScopeID(id)
+		ev := sdktypes.NewVar(n).SetValue(value).SetSecret(secret).SetRequired(required).WithScopeID(id)
```
tebeka, line 47: I don't like this style of coding, prefer the former way.

cmd/ak/cmd/vars/set.go
```diff
@@ -62,4 +62,5 @@ var setCmd = common.StandardCommand(&cobra.Command{
 func init() {
 	// Command-specific flags.
 	setCmd.Flags().BoolVarP(&secret, "secret", "s", false, "this value is secret")
+	setCmd.Flags().BoolVarP(&secret, "required", "r", false, "this value is required")
```
tebeka, line 65: Why does `ak set` cares if a variable is required or not?

internal/backend/dashboardsvc/vars.go
```diff
@@ -84,9 +85,15 @@ func (s Svc) genVarsList(w http.ResponseWriter, r *http.Request, sid sdktypes.Va
 			if cv.IsSecret() {
 				v = "🤐"
 			}
+
+			n := cv.Name().String()
+			if cv.IsRequired() {
+				n = "<em>" + n + "</em>"
```
tebeka, line 91: This seems hacky, should be a template for the vars and inside the template you should check `IsRequired`

sdk/sdktypes/var.go
```diff
@@ -23,11 +23,24 @@ func (p Var) ScopeID() VarScopeID { return kittehs.Must1(ParseVarScopeID(p.read(
...
 func (p Var) IsSecret() bool      { return p.read().IsSecret }
+func (p Var) IsRequired() bool    { return p.read().IsRequired }
 
 func (p Var) WithScopeID(id VarScopeID) Var {
 	return Var{p.forceUpdate(func(pb *VarPB) { pb.ScopeId = id.String() })}
 }
 
-func NewVar(n Symbol, v string, isSecret bool) Var {
-	return kittehs.Must1(StrictVarFromProto(&VarPB{Name: n.String(), Value: v, IsSecret: isSecret}))
+func (p Var) SetSecret(s bool) Var {
```
tebeka, line 32: This works, but that fact that it works on a value receiver is confusing.

tebeka: Also, I don't like this builder pattern :)

tests/system/testdata/deployments/required_vars.txtar
```diff
@@ -0,0 +1,24 @@
+ak manifest apply project.yaml
+return code == 0
+
+ak deploy --project my_project --file main.star
+output contains 'Error: activate deployment: invalid_argument: required vars not set: [req]'
```
tebeka, line 5: Nice!

tebeka (approved):
It works. I have sever comments the main one is that I'm not sure it's required feature.
Also the fact you have to change 29 files to add this show something about the way our code is structured.

cmd/ak/cmd/vars/set.go
itayd: it's actually make it easier to do stuff - not all are secrets, not all have values, etc.

cmd/ak/cmd/vars/set.go
itayd: it "declares" if the var is required or not. value is optional.

internal/backend/dashboardsvc/vars.go
itayd: of course it is hacky, it's the ugly dashboard.

sdk/sdktypes/var.go
itayd: i know you hate builders :)

PR #538, opened 2024-07-30
feat(db): unique connection name per project

ig-ra:


itayd (approved):

internal/backend/db/dbgorm/connections.go
```diff
@@ -19,7 +19,20 @@ func (gdb *gormdb) withUserConnections(ctx context.Context) *gorm.DB {
 }
 
 func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection) error {
-	createFunc := func(tx *gorm.DB, uid string) error { return tx.Create(conn).Error }
+	createFunc := func(tx *gorm.DB, uid string) error {
```
efiShtain, line 22: I think you can just try to create and fail on duplicate error instead of checking and then creating. Assuming it is enforced in the database with a unique constraint. 

Theoretically, if two transactions starts at the same time, you might get count = 0 in both, and both would try to write and you'll still fail one on duplicate. 
It's only theoretical cause no one would try to save / update multiple connections on the same project at the same time.

wdyt ?

ig-ra: 1. the check is inside transaction. So second transaction cannot get count without first to be finished.
2. there is also an issue with defining composite inded on project_id and connection_name, project_id is optional and could be nil. And composite index on nil not working. 
3. therefore due to (1) and (2) this is the solution proposed. 
4. we may decide to redefine cron connection as a project one instead of being global and then get rid of optional project_id in connection and then go for the composite index.

ig-ra: meanhiwle I will open a ticket for redefining cron as project connection and not a global and will add there things and notes. But I think this one could wait meanwhile

ig-ra: see https://linear.app/autokitteh/issue/ENG-1352/cron-as-a-project-connection

efiShtain: transactions in postgres run in parallel, there is no blocking between writers / readers or any ordering
the first to commit is the first to succeed
so as I said, if both transactions start at the same time, due to Postgres's architecture (mvcc), both would get the same count and both will try to write, and both might succeed.

you can test it manually using two terminals and psql, just start two transactions, don't commit them, and you'll see you can write duplicate values if the db does not enforce the uniqueness

projectID is null just for cron, so there would be only one like this
and it should be supported as primary key with null value

PR #539, opened 2024-07-30
fix(oauth): google oauth init fails due to unknown db user 

ig-ra:
google oauth fails due to recent changes when it tries to save vars via vars service.
This happens since there is no db user in context (nether via token or session).
Solution - access oauth via authMux, thus passing via auth middleware and adding user to the context

### DIFF
--- a/integrations/google/server.go
+++ b/integrations/google/server.go
@@ -24,35 +24,35 @@ const (
 	gmailWebhookPath = "/gmail/notif"
 )
 
-func Start(l *zap.Logger, mux *http.ServeMux, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.Dispatcher) {
+func Start(l *zap.Logger, muxNoAuth *http.ServeMux, muxAuth *http.ServeMux, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.Dispatcher) {
 	uiPath := "GET " + desc.ConnectionURL().Path + "/"
 
 	// New connection UIs + handlers.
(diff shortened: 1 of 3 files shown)


itayd:
are you sure re authMux? i don't think that oauth redirect have the token we need. @daabr ?
did you test it? i mean manually, end to end?

ig-ra:
> are you sure re authMux? i don't think that oauth redirect have the token we need. @daabr ? did you test it? i mean manually, end to end?

1. Yes, I tested this - previously it failed on google auth. With this fix it's working
2. since we redirected via browser we have our cookie, and thus the user. That's the reason to pass it through authMux - authMiddleware will extract the authenticated user from cookie and put it in the context

integrations/google/server.go
```diff
@@ -24,35 +24,35 @@ const (
 	gmailWebhookPath = "/gmail/notif"
 )
 
-func Start(l *zap.Logger, mux *http.ServeMux, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.Dispatcher) {
+func Start(l *zap.Logger, muxNoAuth *http.ServeMux, muxAuth *http.ServeMux, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.Dispatcher) {
```
daabr, line 27: Please add a function documentation comment for `Start` to explain that it's currently the only integration that uses `muxAuth`, what `muxAuth` is (in comparison to the usual `muxNoAuth`), and why it's needed at all. In other words, the context and description of this PR as a comment here, to help future maintainers of this code.

daabr (approved):
Approved, but before merging please add a Go function comment to `Start` - see my PR comment there.

PR #541, opened 2024-07-31
feat(events): add event_type to cel filter

efiShtain:
type is a keyword in cel, so we can't use it to filter events

This change adds event_type variable to the cel program and maps event.Type to event_type for comparison

closes #ENG-1536

```python
s = autokitteh.subscribe("slack_connection", 'event_type == "interaction"')
```


tebeka (approved):
Can you also add the rest of the event fields:

- ConnectionId 
- Memo
- CreatedAt 
- Seq

itayd (approved):

PR #542, opened 2024-07-31
fix(github): handle user-webhook events correctly

daabr:
"Correctly" means the following:

- If there are no relevant connections for the event, respond with HTTP 200 and do nothing, don't respond with an HTTP error
- If there are multiple PAT & webhook connections, and we succeed in validating and parsing an incoming event with one of these connection, no need to repeat this with more connections

On the way, use new Go mux patterns instead of parsing webhook URL paths.

Refs: ENG-1224

### DIFF
--- a/integrations/github/webhooks/webhook.go
+++ b/integrations/github/webhooks/webhook.go
@@ -68,11 +68,11 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	} else {
 		// This event is for a user-defined webhook, not an app.
-		path := strings.Split(r.URL.Path, "/")
-		suffix := path[len(path)-1]
-		ctx := r.Context()
+		webhookID := r.PathValue("id")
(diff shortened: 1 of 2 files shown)


integrations/github/webhooks/webhook.go
```diff
@@ -97,10 +95,22 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 				return
 			}
+
+			if payload != nil {
+				userCID = cid
+				break // Successful validation with non-empty payload - no need to repeat.
+			}
+		}
+
+		if payload == nil {
```
tebeka, line 110: Please comment that you couldn't find a connection for the event and that github requires 200 on successful delivery.

tebeka (approved):

integrations/github/webhooks/webhook.go
daabr: done

PR #544, opened 2024-08-03
fix: session log record describe properly print call return values

itayd:
need to specify `SafeForJSON` so JSON encoder could serialize the data properly, otherwise it tries to serialize `map[any]any` and fails.


itayd:
this is trivial and affects only dashboard, so I'm going to just merge it.

PR #546, opened 2024-08-03
fix: next_event can now process multiple sequential events

itayd:
there was a problem where next_event skipped events. this change the name of the last event read in sessions to be more explicit, as well as the condition in the db to be more appropriate with the filter meaning. also increase the seq number in the filter given to List as needed.

### DIFF
--- a/internal/backend/sessions/sessionworkflows/workflow.go
+++ b/internal/backend/sessions/sessionworkflows/workflow.go
@@ -332,17 +332,14 @@ func (w *sessionWorkflow) createEventSubscription(ctx context.Context, connectio
 		return "", fmt.Errorf("connection %q not found", connectionName)
 	}
 
+	// must be set before signal is saved, otherwise the signal might reach the workflow before
+	// the map is read.
+	w.lastReadEventSeqForSignal[signalID] = minSequence
+
 	cid := connection.ID()
 	if err := workflow.ExecuteLocalActivity(wctx, w.ws.svcs.DB.SaveSignal, signalID, workflowID, cid, filter).Get(wctx, nil); err != nil {
 		return "", fmt.Errorf("save signal: %w", err)
 	}
-
-	var c sdktypes.Connection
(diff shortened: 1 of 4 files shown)


internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -111,7 +111,7 @@ func runWorkflow(
 
 func (w *sessionWorkflow) cleanupSignals(ctx workflow.Context) {
 	goCtx := temporalclient.NewWorkflowContextAsGOContext(ctx)
-	for signalID := range w.signals {
+	for signalID := range w.lastReadEventSeqForSignal {
```
efiShtain, line 114: I like the rename for readability, but it makes this part is now less readable

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -378,16 +378,16 @@ func (w *sessionWorkflow) getNextEvent(ctx context.Context, signalID string) (ma
...
-		w.z.Panic("signal not found")
+		return nil, fmt.Errorf("no such subscription %q", signalID)
 	}
 
 	filter := sdkservices.ListEventsFilter{
 		IntegrationID:     iid,
 		ConnectionID:      cid,
 		Limit:             1,
-		MinSequenceNumber: minSequenceNumber,
+		MinSequenceNumber: minSequenceNumber + 1,
```
efiShtain, line 387: I'm not sure why +1 here and >= in the db layer makes a difference 

it's suppose to be the same as minSequenceNumber and >

efiShtain (approved):
not sure how it works, and a little name that bugs me, but looks ok

efiShtain:
just fix the bugs before you try to merge please

internal/backend/sessions/sessionworkflows/workflow.go
itayd: if it's min AND >, it'll not include the min. I want it to include the min.

internal/backend/sessions/sessionworkflows/workflow.go
itayd: that's the most meaningful name i could come up with.

PR #551, opened 2024-08-04
feat(login): support client side only logout

efiShtain:
we add a non http only cookie named ak_logged_in
so frontend can remove it when user logs out
we always validate both cookies exists
in this way we keep the authenticated cookie secured and allow the frontend to logout without additional api on the backend side

### DIFF
--- a/internal/backend/auth/authhttpmiddleware/authmiddleware.go
+++ b/internal/backend/auth/authhttpmiddleware/authmiddleware.go
@@ -83,6 +83,11 @@ func newSessionsMiddleware(next http.Handler, sessions authsessions.Store) http.
 				return
 			}
 
+			if session == nil {
+				http.Error(w, "invalid session", http.StatusUnauthorized)
+				return
+			}
+
 			if session != nil {
 				r = r.WithContext(authcontext.SetAuthnUser(ctx, session.User))
 			}
--- a/internal/backend/auth/authhttpmiddleware/authmiddleware.go
+++ b/internal/backend/auth/authhttpmiddleware/authmiddleware.go
@@ -79,7 +79,7 @@ func newSessionsMiddleware(next http.Handler, sessions authsessions.Store) http.
 		if user := authcontext.GetAuthnUser(ctx); !user.IsValid() {
 			session, err := sessions.Get(r)
 			if err != nil {
-				http.Redirect(w, r, "/login", http.StatusFound)
+				http.Error(w, "invalid session", http.StatusUnauthorized)
 				return
 			}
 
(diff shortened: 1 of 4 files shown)


internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -83,6 +83,11 @@ func newSessionsMiddleware(next http.Handler, sessions authsessions.Store) http.
 				return
 			}
 
+			if session == nil {
```
itayd: why is this needed? if we error out here it wont get to the tokens middleware.

efiShtain: we have a problematic case

both session and tokens are optional

```go
	Sessions authsessions.Store `optional:"true"`
	Tokens   authtokens.Tokens  `optional:"true"`
```

so if tokens are optional, sessions has to validate or fail if session is invalid

it also makes an order dependency which is not explicit

we should refactor both checks to one function and make it super explicit and readable, no need for optional checks

```go
if session == nil && token == nil {
   return 401
}
```

what do you want to do ? fix it on this PR ? merge it and open another PR ?

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -83,6 +83,11 @@ func newSessionsMiddleware(next http.Handler, sessions authsessions.Store) http.
...
+				http.Error(w, "invalid session", http.StatusUnauthorized)
+				return
+			}
+
 			if session != nil {
```
tebeka: Due to the `if` above, you don't need this one.

tebeka (approved):
Minor comment, also missing tests.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
efiShtain: my bad, it should work, I'll recheck required config

efiShtain: seems legit on  Itay's debug dashboard as well

internal/backend/auth/authhttpmiddleware/authmiddleware.go
efiShtain: removed this section from code

efiShtain:
> Minor comment, also missing tests.

@tebeka 
I've added tests, not happy with them, but at least we have them
could you please review and give your thoughts ? 

not table tests since I've used lots of different mocks and it's hard with table testing
I'm thinking about refactoring stuff in the future to make stuff more clear and testable without so much mocks

probably injecting the internal session store instead of init it in the new function

tebeka (commented):
LGTM

PR #554, opened 2024-08-05
feat(metrics): session count

ig-ra:


internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,120 @@
...
+	return metricName
+}
+
+func Init(z *zap.Logger) {
+	// REVIEW: insecure? port? maybe over GRPC?
+	// REVIEW: config? maybe env: OTEL_EXPORTER_OTLP_ENDPOINT or OTEL_EXPORTER_OTLP_METRICS_ENDPOINT
+	exporter, err := otlpmetrichttp.New(
+		context.Background(),
+		otlpmetrichttp.WithInsecure(),
+		otlpmetrichttp.WithEndpoint("localhost:4318"),
```
efiShtain: export to env variable, might be on other addresses in some deployments

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,120 @@
...
+		// otlpmetrichttp.WithURLPath("/"), // send to /v1/metrics
+	)
+	if err != nil {
+		z.Error("failed to create metric exporter: %v", zap.Error(err))
+	}
+
+	schemaURL := "https://opentelemetry.io/schemas/1.1.0"
+	resourceAttrs := resource.NewWithAttributes(
+		schemaURL,
+		semconv.ServiceNameKey.String("ak-dev"),
```
efiShtain: why ak-dev ?
it should be from an env variable

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,120 @@
...
+func Init(z *zap.Logger) {
+	// REVIEW: insecure? port? maybe over GRPC?
+	// REVIEW: config? maybe env: OTEL_EXPORTER_OTLP_ENDPOINT or OTEL_EXPORTER_OTLP_METRICS_ENDPOINT
+	exporter, err := otlpmetrichttp.New(
+		context.Background(),
+		otlpmetrichttp.WithInsecure(),
+		otlpmetrichttp.WithEndpoint("localhost:4318"),
+		// otlpmetrichttp.WithURLPath("/"), // send to /v1/metrics
+	)
+	if err != nil {
```
efiShtain, line 73: what's the purpose of continuing if you failed to create the exporter ?

should be with a flag indicating enabled / disabled and we might want to break if enabled: true and we get an error (I'm not sure about it)

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,120 @@
...
+	for i := 0; i < metricsValue.NumField(); i++ {
+		field := metricsValue.Field(i)
+		fieldType := metricsType.Field(i)
+
+		metricName := metricName(field, fieldType)
+		description := api.WithDescription(metricName)
+
+		var err error
+		switch field.Interface().(type) {
+		case Counter:
```
efiShtain: what's the purpose of this initialization ?
is it local initialization ? does it do something remotely ?

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -178,17 +180,23 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 	//                blow us up due to non determinism.
 
 	var prints []string
+	// attrs := map[string]string{"session_id": sessionID}
 
 	data, err := ws.getSessionData(wctx, params.SessionID)
 	if err == nil {
+		// FIXME: session_id? environment?
+		telemetry.Metrics.Sessions.Update(1, map[string]string{"session_id": sessionID, "type": "start"})
```
efiShtain: I think a better way here is to harden the possibilities

telemetry should expose a simple interface with all our possible metrics and each component should just call a simple method 

```go
telemetry.IncSessionCount()
telemetry.IncRunningSessionCount()
telemetry.DecRunningSessionCount()
```

in this way, we have one interface for all metrics we expose
we also control the cardinality of the dimensions 
and we make stuff less error prone and more readable

it can be multiple interfaces for each component instead of one that does everything 

we can expose the underlying sdk if in some cases we need a raw level access (like db interface vs gorm vs raw sql query)

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,120 @@
...
+	counter api.Int64Counter
+}
+
+type UpDownCounter struct {
+	counter api.Int64UpDownCounter
+}
+
+func (m Counter) Update(value int64, attrs map[string]string) {
+	// REVIEW: should we check/report if we get a negative value
+	m.counter.Add(context.Background(), value, api.WithAttributes(toMetricAttrs(attrs)...))
```
efiShtain: why background context instead of passing context ?

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,120 @@
...
+type Counter struct {
+	counter api.Int64Counter
+}
+
+type UpDownCounter struct {
+	counter api.Int64UpDownCounter
+}
+
+func (m Counter) Update(value int64, attrs map[string]string) {
+	// REVIEW: should we check/report if we get a negative value
```
efiShtain: see my other comment, 
expose a nicer simple API to the user and handle all the ugly stuff here
if you don't want to allow negative value in counter (I think that's a valid idea), don't allow it

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,120 @@
...
+	switch field.Interface().(type) {
+	case Counter:
+		metricName = metricName + ".counter"
+	case UpDownCounter:
+		metricName = metricName + ".gauge"
+	}
+	return metricName
+}
+
+func Init(z *zap.Logger) {
```
efiShtain: create a config object, 
at least with the endpoint and enabled flag, default should be disabled

internal/backend/telemetry/telemetry.go
ig-ra: done

internal/backend/telemetry/telemetry.go
ig-ra: done

internal/backend/telemetry/telemetry.go
ig-ra: I see this as non-fatal error - e.g. should be reported and we should continue.
please elaborate on your suggestion?

internal/backend/telemetry/telemetry.go
ig-ra: which context you want to pass and why?
we are initializing object of opentelemetry and I don't think we should pass there any context except Background

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: that's actually what I did, IMO
- you have metrcis file in which you define metric and it's type. Very simple.
- I wrap opentelemetry metrics and methods  to provide one simple interface and method - e.g. Update(). All other things passed are optional tags.
- Specificaly sessions defined as UpDOwnCounter which is similar to Gauge. You could find through it both overall sessions count per env, amount/ratio of failed/stopped and even duration of each session.

What you are suggesting - if I understand you right - requires much more work for developer while adding new metric. In my version, it requires literally nothing - just add name and type to the metric file.

internal/backend/telemetry/telemetry.go
ig-ra: see my comments above

internal/backend/telemetry/telemetry.go
ig-ra: created config

internal/backend/telemetry/telemetry.go
efiShtain: There is always a context from the user or the workflow, no reason not passing it

internal/backend/telemetry/telemetry.go
efiShtain: I agree it's not a fatal error, but if you already have an error, why don't you just return from the function ?

internal/backend/sessions/sessionworkflows/workflows.go
efiShtain: You can provide an internal implementation that every new metric can use (like this update for example), but the telemetry api should expose a very opinionated and strict api.

So developers can add metrics and use metrics easily.

internal/backend/telemetry/telemetry.go
ig-ra: there is no reason to pass it as well, unless we want to add tags from the context. 
So meanwhile the simplest option is skip passing context, I think.

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: agree - that's why I think what I suggest is the simplest version.
We are mimicing opentelemetry and even simplifying it more.
Open telemetry API have Add() method for Counter, another Add method for UpDownCounter, Record() for histogram etc. Having one Update() method is simpler and clearer

internal/backend/telemetry/telemetry.go
ig-ra: and added enable/disable

internal/backend/telemetry/telemetry.go
ig-ra: we are initializing opentelemetry structs

ig-ra: no, it does nothing remotely, till updtae is called and batch is created

internal/backend/telemetry/telemetry.go
ig-ra: because we need to initialize no-op/dummy metrics

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: metrics usage example from code:
```
telemetry.Metrics.Sessions.Update(-1)
```
1. access strictly-typed struct and not fetching metric from the registry with string, which helps in metric definition in metrics file, and simplifies fetching it and preventing errors
2. you just need one line to define metric in metric file
3. and one line to use it in code.

internal/backend/telemetry/telemetry.go
itayd: If there is an init error, we must fail. I think that `New` fails only if there is config error.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,143 @@
...
+	return metricName
+}
+
+func Init(z *zap.Logger, cfg *Config) {
+	enabled = cfg.Enabled
+	if !enabled {
+		z.Warn("metrics are disabled")
+	}
+
+	// REVIEW: insecure? port? maybe over GRPC?
```
itayd: All of these need to be configurable.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,143 @@
...
+		// otlpmetrichttp.WithURLPath("/"), // /v1/metrics by default, unless set
+	)
+	if err != nil {
+		z.Error("failed to create metric exporter: %v", zap.Error(err))
+	}
+
+	schemaURL := "https://opentelemetry.io/schemas/1.1.0"
+	resourceAttrs := resource.NewWithAttributes(
+		schemaURL,
+		semconv.ServiceNameKey.String(cfg.ServiceName),
```
itayd: Please stop using Println. If this is not ready for review, indicate that.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,143 @@
...
+	meterProvider := metric.NewMeterProvider(
+		metric.WithReader(metric.NewPeriodicReader(exporter)),
+		metric.WithResource(resourceAttrs),
+	)
+
+	otel.SetMeterProvider(meterProvider) // set global meter provider
+	meter := otel.GetMeterProvider().Meter("ak")
+
+	metricsValue := reflect.ValueOf(&Metrics).Elem()
+	metricsType := metricsValue.Type()
```
itayd: I do not understand the purpose of this code.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,143 @@
...
+
+var enabled = false
+
+func metricName(field reflect.Value, fieldType reflect.StructField) string {
+	metricName := fieldType.Name
+	metricName = strings.ReplaceAll(metricName, "_", ".")
+	metricName = "ak." + strings.ToLower(metricName)
+	switch field.Interface().(type) {
+	case Counter:
+		metricName = metricName + ".counter"
```
itayd: why do we need to add these suffixes?

internal/backend/sessions/sessionworkflows/workflows.go
itayd: `telemetry.Metrics.Sessions.Update(-1)` is really not clear. Either do what @efiShtain suggested, or imho better yet just expose the actual metric object from otel. Adding another abstraction layer here just complicates stuff.

ig-ra: @itayd - please see metrics RFC.
- I don't want to expose opentelemetry objects directly. I think we need a simple abstraction layer in order to be able to replace/adjust/improve it later
- It's not easy to pass context tags in open telemetry. This also need to be abtracted and this is done in this simple abstraction layer
- I think Efi's suggestion is complicated - it will require to add both metrics (e.g. sessions) and metrics functions (e.g. IncSessionsCounter, IncFailedSessionCounter, etc..). My proposition is simpler and easier to use for developer while preserving the almost the same API that exposed by opentelemetry
- Open telemetry Counters' update method is `Add`. And Gaude's update method is `Set`. My Idea was to introduce  single API to all metrics, but I could change methods naming according to openTelemetry one

internal/backend/telemetry/telemetry.go
ig-ra: of coz. What is marked for review - should be discussed.

internal/backend/telemetry/telemetry.go
ig-ra: sorry, I just forgot debug print while checking config added. removed.

internal/backend/telemetry/telemetry.go
ig-ra: the purpose of this code is to initialize all metrics defined by the user.

internal/backend/telemetry/telemetry.go
ig-ra: plz see datadog and open telemetry metric naming conventions
https://opentelemetry.io/docs/specs/semconv/general/metrics/
usually it's `<service>.<module?>.<metric>.<type>`, e.g. in this case it will be `ak.sessions.gaude`
This is the name that this metric will appear in datadog:
see https://app.datadoghq.com/dashboard/3ei-gvz-72h/igor?fromUser=false&refresh_mode=sliding&view=spans&from_ts=1722949552420&to_ts=1722953152420&live=true

internal/backend/sessions/sessionworkflows/workflows.go
itayd: See what I commented in the RFC. IMO the overhead of maintaining another abstraction layer on top of opentel is too much. For now I think it's best just to expose the actual otel object, or implement the metrics functions specifically in each service.

internal/backend/telemetry/telemetry.go
itayd: We need to pass a proper context, never `Background()`. I understand why you wouldn't want to pass a context from the workflow, but you could at least limit the context passed to this function. By limiting, I mean doing `WithTimeout` with a small value (1s?). If there is any error - log. That way this will not be stuck for a meaningful amount of time.

internal/backend/telemetry/telemetry.go
itayd: that was the discussion. please add it.

internal/backend/telemetry/telemetry.go
itayd: i can live with that.

internal/backend/telemetry/telemetry.go
itayd: i am very wary of adding this code with reflection here. i have no idea how it works. i have no idea why you even use reflection here. either add very clear comments about why it does that and how it does that, or find a better way.

itayd: what i mean is that you need to justify the use of reflection here. and I'm not sure it's possible.

internal/backend/telemetry/telemetry.go
ig-ra: The context passed to the metric used mostly for metadata propogation and correlation with traces. If I understand right this context is unrelated to timeouts/cancellation

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,128 @@
...
+	serviceName string
+}
+
+func New(z *zap.Logger, cfg *Config) *Telemetry {
+	cfg = verifyConfig(cfg)
+
+	telemetry := &Telemetry{z: z, enabled: cfg.Enabled, serviceName: cfg.ServiceName}
+
+	if !telemetry.enabled {
+		z.Warn("metrics are disabled")
```
efiShtain: Info is better, it's not error to disable metrics

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,128 @@
...
+	meterProvider := metric.NewMeterProvider(
+		metric.WithReader(metric.NewPeriodicReader(exporter)),
+		metric.WithResource(resourceAttrs),
+	)
+
+	otel.SetMeterProvider(meterProvider) // set global meter provider
+	return telemetry
+}
+
+type NoOpCounter struct {
```
efiShtain: what's the purpose of this no op counter ?

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,128 @@
...
+	api.Int64UpDownCounter
+}
+
+func (NoOpCounter) Add(context.Context, int64, ...api.AddOption) {}
+func (NoOpCounter) int64UpDownCounter()                          {}
+
+func (t *Telemetry) NewOtelUpDownCounter(name string, description string) api.Int64UpDownCounter {
```
efiShtain: I don't think we need to have Otel in the names fo this functions

it doesn't provide any value and it can only make future refactor hard if we need to (we probably won't, still the otel part is irrelevant)

internal/backend/sessions/sessionworkflows/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+
+	"go.autokitteh.dev/autokitteh/internal/backend/telemetry"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+	"go.opentelemetry.io/otel/metric"
+)
+
+var sessionsCounter metric.Int64UpDownCounter
+
+func initMetrics(t *telemetry.Telemetry) {
+	sessionsCounter = t.NewOtelUpDownCounter("sessions.gaude", "The session counter")
```
efiShtain: gaude is not a word I think, unless I miss something

the type of the counter is gauge

it's not really a part of the name of the metric
the metric name should be something like autokitteh.active_sessions (which is what I understand from the up down counter)
otherwise autokitteh.total_sessions

internal/backend/sessions/sessionworkflows/metrics.go
```diff
@@ -0,0 +1,35 @@
...
+}
+
+func updateSessionCounter(sessionID string, sessionState sdktypes.SessionStateType) {
+	labels := telemetry.Labels{"session_id": sessionID}
```
efiShtain: sessions_id is ultra high cardinality dimension, and it has no use in metrics

you would never use any arithmetic calculation based on session_id

internal/backend/telemetry/telemetry.go
ig-ra: done

internal/backend/telemetry/telemetry.go
ig-ra: you need one if you disabled the metrics. metrics code still will be invoked, so it should do nothing

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	ServiceName string `koanf:"service_name"`
+	Endpoint    string `koanf:"endpoint"`
+}
+
+var Configs = configset.Set[Config]{
+	Default: &Config{Enabled: true, ServiceName: "ak", Endpoint: "localhost:4318"},
+	Dev:     &Config{Enabled: false, ServiceName: "ak", Endpoint: "localhost:4318"},
+}
+
+func verifyConfig(cfg *Config) *Config {
```
itayd: this has a side effect of actually changing the input.

you should make this a method of Config, then either rename it to `fixConfig` or some such, or make it return a new Config object.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	if cfg.ServiceName == "" {
+		cfg.ServiceName = Configs.Default.ServiceName
+	}
+	if cfg.Endpoint == "" {
+		cfg.Endpoint = Configs.Default.Endpoint
+	}
+	return cfg
+}
+
+type Labels map[string]string
```
itayd: do we really need this?

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+
+func WithLabels(labels Labels) api.MeasurementOption {
```
itayd: i think if we go and do a `WithLabels` function, it would be more useful to make it just receive a list of strings:

```
func WithLabels(args ...string) api.MeasurementOption {
    var attrs []string
    for len(args) > 1; args = args[2:] {
        attrs = append(attrs, attribute.String(args[0], args[1]))
    }
    return api.WithAttributes(attrs...)
}
```

Much easier to use.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	var attrs []attribute.KeyValue
+	for k, v := range labels {
+		attrs = append(attrs, attribute.String(k, v))
+	}
+	return api.WithAttributes(attrs...)
+}
+
+type Telemetry struct {
+	z           *zap.Logger
```
itayd: we're going with `l` or `sl` for logger nowadays.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	enabled     bool
+	serviceName string
+}
+
+func New(z *zap.Logger, cfg *Config) *Telemetry {
+	cfg = verifyConfig(cfg)
+
+	telemetry := &Telemetry{z: z, enabled: cfg.Enabled, serviceName: cfg.ServiceName}
```
itayd: personally I like just doing something like `Telemetry{l: l, cfg: cfg}` and then I know I have the config stored and no need to pass each field individually.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+
+	if !telemetry.enabled {
+		z.Info("metrics are disabled")
+		return telemetry
+	}
+
+	// REVIEW: encryption? GRPC?
```
itayd: make this configurable or add a TODO here to make these configurable

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	// REVIEW: should we use AK env or maybe standard OTEL onses?
+	//         - OTEL_EXPORTER_OTLP_ENDPOINT or OTEL_EXPORTER_OTLP_METRICS_ENDPOINT
+	exporter, err := otlpmetrichttp.New(
+		context.Background(),
+		otlpmetrichttp.WithInsecure(),
+		otlpmetrichttp.WithEndpoint(cfg.Endpoint),
+		// otlpmetrichttp.WithURLPath("/"), // /v1/metrics by default, unless set
```
itayd: ?

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+
+	telemetry := &Telemetry{z: z, enabled: cfg.Enabled, serviceName: cfg.ServiceName}
+
+	if !telemetry.enabled {
+		z.Info("metrics are disabled")
+		return telemetry
+	}
+
+	// REVIEW: encryption? GRPC?
+	// REVIEW: should we use AK env or maybe standard OTEL onses?
```
itayd: i'm fine with OTEL's.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	//         - OTEL_EXPORTER_OTLP_ENDPOINT or OTEL_EXPORTER_OTLP_METRICS_ENDPOINT
+	exporter, err := otlpmetrichttp.New(
+		context.Background(),
+		otlpmetrichttp.WithInsecure(),
+		otlpmetrichttp.WithEndpoint(cfg.Endpoint),
+		// otlpmetrichttp.WithURLPath("/"), // /v1/metrics by default, unless set
+	)
+	if err != nil {
+		z.Error("failed to create metric exporter: %v", zap.Error(err))
```
itayd, line 74: i think i've written this before - error out, don't return an object.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+		telemetry.enabled = false
+		return telemetry
+	}
+
+	schemaURL := "https://opentelemetry.io/schemas/1.1.0"
```
itayd: make it a const.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	resourceAttrs := resource.NewWithAttributes(
+		schemaURL,
+		semconv.ServiceNameKey.String(telemetry.serviceName),
+	)
+
+	// REVIEW: consider using controller?
```
itayd: not sure what is a controller here?

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	createFunc func(meter api.Meter, name string, description string) (T, error),
+) T {
+	meter := otel.GetMeterProvider().Meter(t.serviceName)
+	if !strings.HasPrefix(name, t.serviceName) {
+		name = fmt.Sprintf("%s.%s", t.serviceName, name)
+	}
+	metric, err := createFunc(meter, name, description)
+	if err != nil {
+		t.z.Error("failed to create metric", zap.String("name", name), zap.Error(err))
+		// REVIEW: should we panic? kittehs.Must?
```
itayd: return an error

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+	return metric
+}
+
+func (t *Telemetry) NewUpDownCounter(name string, description string) api.Int64UpDownCounter {
+	if !t.enabled {
+		return NoOpMetric{}
```
itayd: if we don't have real telemetry installed (meaning no telemetry initialized), what will `Int64UpDownCounter` do? Maybe it's just a no-op without special help?

itayd (commented):
mostly ok - just a few minor comments.

internal/backend/telemetry/telemetry.go
ig-ra: done

internal/backend/sessions/sessionworkflows/metrics.go
ig-ra: rewritten to use 4 counters

internal/backend/sessions/sessionworkflows/metrics.go
ig-ra: agree. removed

internal/backend/telemetry/telemetry.go
ig-ra: staying with AK var meanwhile.
the question was a bit broader - say, user want's to run ak localy and have OTEL collector and thus OTEL env variables (think about XDG_HOME/DADA but for OTEL). Do we want to react to this and send metrics to it? I think - there is no need, right?

internal/backend/telemetry/telemetry.go
ig-ra: removed

internal/backend/telemetry/telemetry.go
ig-ra: added TODO

internal/backend/telemetry/telemetry.go
ig-ra: done

internal/backend/telemetry/telemetry.go
ig-ra: controller is another way to get more fine grained control on batching and sending from OTEL. Added a comment

internal/backend/telemetry/telemetry.go
ig-ra: made it function of Config

internal/backend/telemetry/telemetry.go
ig-ra: accepted your suggestion

internal/backend/telemetry/telemetry.go
ig-ra: removed

internal/backend/telemetry/telemetry.go
ig-ra: done. storing config

internal/backend/telemetry/telemetry.go
ig-ra: rewritten to use opentelemetry no-ops.

ig-ra: rewritten to use opentelemetry no-ops.
if telemetry disabled then all counters will be nitialized to no-ops

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+	if cfg.Endpoint == "" {
+		cfg.Endpoint = Configs.Default.Endpoint
+	}
+}
+
+func WithLabels(args ...string) metric.MeasurementOption {
+	var attrs []attribute.KeyValue
+	if len(args)%2 != 0 {
+		args = args[:len(args)-1] // strip the last one. TODO: log?
```
itayd: you should DPanic in this case, same as done in zap.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,128 @@
...
+	cfg.fixConfig() // just ensure that endpoint and service name are set
+
+	telemetry := &Telemetry{l: z, cfg: *cfg}
+
+	if !telemetry.cfg.Enabled {
+		z.Info("metrics are disabled")
+		return telemetry
+	}
+
+	// TODO: [ENG-1445] GRPC?
```
itayd: I think the proper format is `// TODO(ENG-1445): gRPC?`.

internal/backend/telemetry/telemetry.go
itayd: This is important. Please return an error.

internal/backend/telemetry/telemetry.go
```diff
@@ -0,0 +1,128 @@
...
+	return metric.WithAttributes(attrs...)
+}
+
+type Telemetry struct {
+	l   *zap.Logger
+	cfg Config
+}
+
+func New(z *zap.Logger, cfg *Config) *Telemetry {
+	cfg.fixConfig() // just ensure that endpoint and service name are set
```
itayd: this mutates `cfg`, which is owned by the caller. either copy `cfg` beforehand or make `fixConfig` return a new object.

itayd (approved):
approved with some issues. if you don't agree to all - do not merge.

internal/backend/telemetry/telemetry.go
itayd: sdklogger.DPanic

ig-ra: done

internal/backend/telemetry/telemetry.go
ig-ra: done

internal/backend/telemetry/telemetry.go
ig-ra: done

internal/backend/telemetry/telemetry.go
ig-ra: done

PR #555, opened 2024-08-06
fix system tests and optional var detection

itayd:


itayd:
merging, as systems tests not running in ci. this will reenable them.

PR #560, opened 2024-08-10
fix(ENG-1407): run next_event uuid generation in a side effect

itayd:
so it looks like under some circumctances that involve signals, temporal decides to rerun a workflow function even after it's completed in replay mode.

when in replay, the next_event generates a new `signalID` though it should have used the previous one. Properly putting it in a SideEffect eliminates the problem.

also added an "optimization" to just not run the replay if the workflow is already done.


internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -175,7 +175,15 @@ func (ws *workflows) getSessionDebugData(data *sessiondata.Data, prints []string
 
 func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkflowParams) (debug any, _ error) {
 	sessionID := params.SessionID.String()
-	z := ws.z.With(zap.String("session_id", sessionID))
+	si := workflow.GetInfo(wctx)
```
efiShtain: What's si?

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -190,6 +198,16 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 
 	data, err := ws.getSessionData(wctx, params.SessionID)
 	if err == nil {
+		if workflow.IsReplaying(wctx) && data.Session.State().IsFinal() {
+			// HACK: If we somehow get a signal to this workflow after it completed,
+			//       in certain delicate timing, the workflow rekicks in replay.
+			//       Though replay would work just fine, and result in a no-op,
+			//       it's nice to not have to run through its entirely. Temporal
+			//       just seems to ignore it...
+			z.Warn("already completed")
```
efiShtain: It's not a warn, the system would keep working as expected with this fix
Info at best

efiShtain (approved):

internal/backend/sessions/sessionworkflows/workflows.go
itayd: session info. should be `wi` i guess.

internal/backend/sessions/sessionworkflows/workflows.go
itayd: yeah... it's still odd and i'd like to get rid of it. but you're right, I'll change to info.

PR #562, opened 2024-08-12
feat: return json data instead of wrapped data in sessions and event grpc endpoints

itayd:
caller (usually ui) need to specify `json_values` in request.

Refs: ENG-1403


sdk/sdktypes/value.go
```diff
@@ -163,3 +164,28 @@ func (v Value) ToStringValuesMap() (map[string]Value, error) {
...
+	}
+
+	v, err := ValueFromProto(pb)
+	if err != nil {
+		return nil, fmt.Errorf("decode: %w", err)
+	}
+
+	uw := ValueWrapper{SafeForJSON: true}
+
+	u, err := uw.Unwrap(v)
```
daabr, line 180:
```suggestion
	u, err := ValueWrapper{SafeForJSON: true}.Unwrap(v)
```

proto/autokitteh/events/v1/svc.proto
```diff
@@ -29,6 +29,10 @@ message SaveResponse {
 
 message GetRequest {
   string event_id = 1 [(buf.validate.field).string.min_len = 1];
+
+  // If set, all values returned will be string values
+  // that contained the native values in JSON format.
```
daabr: 1. s/contained/contain/
2. Explain also what happens if the bool is `false`

proto/autokitteh/events/v1/svc.proto
```diff
@@ -41,6 +45,10 @@ message ListRequest {
   string event_type = 3;
   uint32 max_results = 4;
   string order = 5; // ASC or DESC values are supported
+
+  // If set, all values returned will be string values
+  // that contained the native values in JSON format.
```
daabr: Same as above

proto/autokitteh/sessions/v1/svc.proto
```diff
@@ -74,6 +74,13 @@ message GetResponse {
 
 message GetLogRequest {
   string session_id = 1 [(buf.validate.field).string.min_len = 1];
+
+  // If set, all values returned will be string values
+  // that contained the native values in JSON format.
```
daabr: Same as in `proto/autokitteh/events/v1/svc.proto`

daabr (approved):
If you need to specify a bool in the request that affects the "mode" of returned value, and there's no obvious way to understand the response without that bool - this is a strong sign of unnecessary complexity and a code smell.

It's not blocking this PR, but it needs to be improved.

Before merging, see the minor suggestions.

itayd:
> If you need to specify a bool in the request that affects the "mode" of returned value, and there's no obvious way to understand the response without that bool - this is a strong sign of unnecessary complexity and a code smell.

Yup - we are doing this as a shortcut to achieve a very specific goal. In the future we'll think of something better.

proto/autokitteh/events/v1/svc.proto
itayd: done

proto/autokitteh/sessions/v1/svc.proto
itayd: refed

proto/autokitteh/events/v1/svc.proto
itayd: refed

PR #563, opened 2024-08-14
refactor: make syscalls explicit

itayd:
instead of having a single syscall function that exposes various functions, have these functions explicitly exported by the ak module. this also moves the ak module et al to be an internal sessions package.

this is a prep work for getting rid of syscalls and having them more explicit.


internal/backend/fixtures/fixtures.go
tebeka, line 1: Why is this called fixtures? It's a named usually associated with testing.

internal/backend/sessions/sessionworkflows/module.go
```diff
@@ -0,0 +1,61 @@
...
+
+func callopts(_ context.Context, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
+	if len(args) > 0 {
+		return sdktypes.InvalidValue, fmt.Errorf("expecting only key value arguments")
+	}
+
+	return sdktypes.NewStructValue(sdktypes.NewSymbolValue(fixtures.CallOptsCtorSymbol), kwargs)
+}
+
+func (w *sessionWorkflow) isDeploymentActive(ctx context.Context, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
```
tebeka, line 44: I'd still like to call "native" Go function

```
if cbs.IsDeploymentActive() {
    // ...
}
```

And not do the boxing/unboxing dance.

internal/backend/sessions/sessionworkflows/modules/akmodules.go
```diff
@@ -0,0 +1,18 @@
...
+	"go.autokitteh.dev/autokitteh/internal/backend/sessions/sessionworkflows/modules/time"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func IsAKModuleExecutorID(xid sdktypes.ExecutorID) bool {
+	switch xid {
+	case store.ExecutorID, time.ExecutorID, os.ExecutorID, fixtures.ModuleExecutorID:
+		return true
+	default:
+		return false
```
tebeka, line 16: Can be outside the `switch`

internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -57,7 +57,7 @@ func (w *sessionWorkflow) syscall(ctx context.Context, args []sdktypes.Value, kw
 	}
 }
 
-func (w *sessionWorkflow) fake(args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
+func (w *sessionWorkflow) fake(_ context.Context, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
```
tebeka, line 60: What is `fake` for? What does it do?

tebeka (changes requested):
I'm approving this, don't see much value over what was before.

Also (and that's why I'm not approving), I don't see changes in `pythonrt.go` and how it handles these calls. See `loadSyscall` and `handleCall` there. If you want, I can make the changes in this branch.

internal/backend/fixtures/fixtures.go
itayd: yeah should probably rename it to `consts` or something like that. not in the scope of this PR though.

internal/backend/sessions/sessionworkflows/module.go
itayd: this will happen in a following pr. this is just a prep for that.

internal/backend/sessions/sessionworkflows/modules/akmodules.go
itayd: not sure what you mean - the `return false`? what for?

internal/backend/sessions/sessionworkflows/syscalls.go
itayd: i did it a long while ago as am exploration of testing. it allows to tell autokitteh to run specific activities with predetermined values instead of actually invoking the activity.

tebeka (approved):

PR #564, opened 2024-08-14
test: allow to capture jq expressions from output

itayd:
this is a proposal that can lead to getting rid of "test mode" in ak. this allows to capture results from ak commands using jq queries, and then use them later. the same can be done with regexps.

### DIFF
--- a/tests/system/ak_test.go
+++ b/tests/system/ak_test.go
@@ -44,10 +44,6 @@ func TestSuite(t *testing.T) {
 			return nil // Skip directories and non-test files.
 		}
 
-		if !strings.HasSuffix(path, "build.txtar") {
-			return nil
-		}
-
 		// Each .txtar file is a test-case, with potentially
 		// multiple actions, checks, and embedded files.
 		t.Run(strings.TrimPrefix(path, rootDir), func(t *testing.T) {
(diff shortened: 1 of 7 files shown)


tebeka (approved):
I'm good with dropping test mode.
I feel like too much effort in developing yet another testing language instead of implementing some utility test functions in Go (e.g `RunServer`, `Deploy`, `Build`, `Logs` ...)
I'd also get rid of txtar and use directories and files.

itayd:
> I'm good with dropping test mode.

That's the idea

> I feel like too much effort in developing yet another testing language instead of implementing some utility test functions in Go (e.g `RunServer`, `Deploy`, `Build`, `Logs` ...) I'd also get rid of txtar and use directories and files.

I actually like the txtar way, very straight to the point.

tests/system/ak_test.go
```diff
@@ -44,6 +44,10 @@ func TestSuite(t *testing.T) {
 			return nil // Skip directories and non-test files.
 		}
 
+		if !strings.HasSuffix(path, "build.txtar") {
+			return nil
+		}
+
```
itayd:
```suggestion
```

PR #566, opened 2024-08-14
bug(python): trim logs

tebeka:
Currently the logs from Python are too spammy, mostly since they emit the messages running between Python & Go. This PR reduces the amount of logging.

Fixes ENG-1443

### DIFF
--- a/runtimes/pythonrt/pythonrt.go
+++ b/runtimes/pythonrt/pythonrt.go
@@ -15,6 +15,7 @@ import (
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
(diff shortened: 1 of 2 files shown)


runtimes/pythonrt/pythonrt.go
```diff
@@ -15,6 +15,7 @@ import (
 
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
+	"golang.org/x/exp/maps"
```
itayd, line 18: doesn't golang has first class maps package now?

itayd (approved):
see comment

PR #567, opened 2024-08-15
add a domain to both ui cookies

efiShtain:
backend and frontend might be served from different subdomains, 
we need to set the cookie for the top level domain (.autokitteh.cloud)
but it might be another in on prem installations

this sets some defaults and allow setting the values

### DIFF
--- a/internal/backend/auth/authsessions/config.go
+++ b/internal/backend/auth/authsessions/config.go
@@ -22,6 +22,5 @@ var Configs = configset.Set[Config]{
 		Secure:     false,
 		SameSite:   http.SameSiteLaxMode,
 		CookieKeys: "0000000000000000000000000000000000000000000000000000000000000000,0000000000000000000000000000000000000000000000000000000000000000",
-		Domain:     "localhost",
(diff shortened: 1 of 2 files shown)


internal/backend/auth/authsessions/config.go
```diff
@@ -3,26 +3,26 @@ package authsessions
...
 var Configs = configset.Set[Config]{
 	Default: &Config{
-		Cookie: func() *sessions.CookieConfig {
-			c := sessions.DefaultCookieConfig
-			c.SameSite = http.SameSiteNoneMode
-			return c
-		}(),
+		SameSite: http.SameSiteNoneMode,
+		Secure:   true,
+		Domain:   "autokitteh.cloud",
```
itayd: i don't think this should appear is OSS. we should set it ourselves in terraform.

efiShtain: It's just the default
It can be overridden by env variables

itayd (approved):
just remove the default for prod

PR #569, opened 2024-08-16
DRAFT: feat: remote runtimes working end to end

itayd:
per @tebeka 's request for a concrete example, I improved and implemented the runtimes RPC interface. now it works end to end.

there are several ways to test it, i will show you when i meet.

i implemented it as a streaming call for simplicity (it takes care of a lot of state management), but I'm not sure we should keep it. even connect folks recommend against it unless you get good lbs and such (https://connectrpc.com/docs/go/streaming/). as such, this is more of a well made POC, though it can be used for now to parallelize efforts.

there are some ancillary changes to make it testable, so you should focus only on the proto and the SDK. everything else is accessories.

to test locally:
```
$ # this will start ak in runtime only mode, only exposing the runtime interface on port 9981.
$ AK_COMPONENTS="-,http,auth,authhttpmiddleware,healthcheck,temporalclient,db,runtimes,runtimessvc" ./bin/ak up -m dev -C "http.addr=localhost:9981"

$ # in another window, run autokitteh configured to talk to 9981 for the runtimes themselves:
$ BIDI_RUN=1 ./bin/ak up -m dev -C "runtimes.runtimes=*=http://localhost:9981"

$ # in yet another window, run the sessions and run tests:
$ AK_ADDR="localhost:9980" TESTS=python_simple.txtar PYTHON=only make test-sessions
```

note the `BIDI_RUN` as it enables the new RPC.


backend/runtimes/runtimes.go
```diff
@@ -1,18 +1,160 @@
...
+	"go.autokitteh.dev/autokitteh/sdk/sdkclients/sdkruntimesclient"
 	"go.autokitteh.dev/autokitteh/sdk/sdkruntimes"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-func New() sdkservices.Runtimes {
-	return kittehs.Must1(sdkruntimes.New([]*sdkruntimes.Runtime{
+type Config struct {
+	// Command separated list of runtimes and their addresses. If no address is supplied,
```
tebeka, line 23: Why a string and not a slice?

backend/runtimes/runtimes.go
```diff
@@ -1,18 +1,160 @@
...
+	builtinsMap = kittehs.ListToMap(
+		builtins,
+		func(rt *sdkruntimes.Runtime) (string, *sdkruntimes.Runtime) { return rt.Desc.Name().String(), rt },
+	)
+)
+
+func New(sl *zap.SugaredLogger, cfg *Config) (sdkservices.Runtimes, error) {
+	ctx := context.Background()
+
+	if sl == nil {
```
tebeka, line 53: Shouldn't this be an error? We'll end up logging to the wrong place.

backend/runtimes/runtimes.go
```diff
@@ -1,18 +1,160 @@
...
+
+	if cfg == nil {
+		cfg = &Config{}
+	}
+
+	if cfg.Runtimes == "" {
+		cfg.Runtimes = "*"
+	}
+
+	remotes := kittehs.Transform(strings.Split(cfg.Runtimes, ","), strings.TrimSpace)
```
tebeka, line 68: This should be in an external function.

backend/runtimes/runtimes.go
```diff
@@ -1,18 +1,160 @@
...
+	for _, remote := range remotes {
+		add := func(rtsvc sdkservices.Runtimes, name, addr string) error {
```
tebeka, line 70: Move this to be a "regular" function and not embedded one?  Pass `names` as a parameter and return `rts`

backend/runtimes/runtimes.go
```diff
@@ -1,18 +1,160 @@
...
 )
 
-func New() sdkservices.Runtimes {
-	return kittehs.Must1(sdkruntimes.New([]*sdkruntimes.Runtime{
+type Config struct {
+	// Command separated list of runtimes and their addresses. If no address is supplied,
+	// the runtime is builtin.
+	// Examples:
+	// - select runtimes: python=localhost:9980,starlark=localhost:9981,config
+	// - all builtins: *
```
tebeka, line 27: What is the reason we allow users not to select a builtin runtime? They are shipped with `ak` anyway.

backend/runtimes/runtimes.go
```diff
@@ -1,18 +1,160 @@
...
+				if rt := builtinsMap[name]; rt != nil {
+					sl.Infof("using builtin runtime: %q", name)
+					rts = append(rts, rt)
+					return nil
+				}
+
+				return fmt.Errorf("invalid builtin runtime")
+			}
+
+			sym, err := sdktypes.ParseSymbol(name)
```
tebeka, line 87: Why is runtime name required to be a valid symbol?

backend/runtimes/runtimes.go
```diff
@@ -1,18 +1,160 @@
...
+
+			rts = append(rts, &sdkruntimes.Runtime{
+				Desc: rt.Get(),
+				New:  func() (sdkservices.Runtime, error) { return rt, nil },
+			})
+
+			return nil
+		}
+
+		name, addr, _ := strings.Cut(remote, "=")
```
tebeka, line 107: Add a comment with some remote examples.

backend/runtimes/runtimes.go
tebeka: Also, I prefer to have the user write something more explicit, this will reduce the size of the code parsing the configuration.

```yaml
remote_runtimes:
    - name: rpython
      kind: python
      address: localhost:9980
```

What happens when there's a project with `.py` files? Do we send it to the remote or the local one? IMO we should specify the runtime name in the manifest.

backend/runtimes/runtimes.go
```diff
@@ -1,18 +1,160 @@
...
+			}
+
+			continue
+		}
+
+		// only remotes from this point.
+
+		rtsvc = sdkruntimesclient.New(sdkclient.Params{URL: addr, L: sl.Desugar()})
+
+		if name != "*" {
```
tebeka, line 135: How can remote name be `*`?

backend/runtimes/runtimes.go
tebeka: Also also: What @efiShtain thought was that remote runtimes will connect to `ak` and register themselves, not a fixed list of runtimes. The address of a new runtime in isolation is not known in advance.

cmd/ak/cmd/runtimes/run.go
```diff
@@ -54,6 +56,11 @@ var runCmd = common.StandardCommand(&cobra.Command{
 				return fmt.Errorf("empty txtar archive")
 			}
 
+			if entryPoint == "" {
+				entryPoint = strings.ReplaceAll(a.Files[0].Name, " ", ":")
```
tebeka, line 60: What is that and where it is documented? I don't understand this code that replaces a ` ` with `:` and then ignores everything up to the first `:`, why do we do this?

tebeka: In general, this function is too complex with a lot of `if`. Need to refactor.

cmd/ak/cmd/runtimes/run.go
```diff
@@ -115,9 +126,10 @@ func init() {
 	runCmd.Flags().BoolVar(&txtarFile, "txtar", false, "input file is a txtar archive containing program")
 	runCmd.Flags().BoolVarP(&emitValues, "values", "v", false, "emit result values")
 	runCmd.Flags().BoolVarP(&quiet, "quiet", "q", false, "do not print anything but errors")
+	runCmd.Flags().StringVar(&entryPoint, "entrypoint", "", `entry point ("file:function")`)
 }
 
-func run(ctx context.Context, b *sdkbuildfile.BuildFile, path string) (map[string]sdktypes.Value, []string, error) {
+func run(ctx context.Context, b *sdkbuildfile.BuildFile, ep string) (map[string]sdktypes.Value, []string, error) {
```
tebeka, line 132:
```suggestion
func run(ctx context.Context, b *sdkbuildfile.BuildFile, entryPoint string) (map[string]sdktypes.Value, []string, error) {
```

cmd/ak/cmd/runtimes/runtime.go
```diff
@@ -37,7 +38,17 @@ func init() {
 
 func runtimes() sdkservices.Runtimes {
 	if local {
-		return backendRuntimes.New()
+		var cfg backendRuntimes.Config
+
+		if _, err := common.Config().Get("runtimes", &cfg.Runtimes); err != nil {
+			sdklogger.Panic(err)
```
tebeka, line 44: Why should we panic here?

cmd/ak/cmd/runtimes/runtime.go
```diff
@@ -37,7 +38,17 @@ func init() {
 
 func runtimes() sdkservices.Runtimes {
 	if local {
```
tebeka, line 40: Why do we need this flag? We can indicate local/remote in the output

internal/backend/svc/fx.go
```diff
@@ -3,15 +3,48 @@ package svc
...
 	"go.uber.org/fx"
 	"go.uber.org/fx/fxevent"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 
 	"go.autokitteh.dev/autokitteh/internal/backend/configset"
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
 )
 
+var components = kittehs.FilterZeroes(strings.Split(os.Getenv("AK_COMPONENTS"), ","))
```
tebeka, line 18: What are `AK_COMPONENTS`? Where's the documentation?

internal/backend/svc/fx.go
```diff
@@ -3,15 +3,48 @@ package svc
...
+
+func isComponentEnabled(name string) bool {
+	if len(components) == 0 {
+		return true
+	}
+
+	enabled := false
+
+	for _, c := range components {
+		if c == "" {
```
tebeka, line 28: How can it be `""` if you use `FilterZeros` above?

internal/backend/svc/fx.go
tebeka: I don't understand why we define another mini-language inside a string instead of using a data structure.

internal/backend/svc/svc.go
```diff
@@ -207,7 +207,7 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 		),
 		// TODO: consider design where the workflow is an implementation detail of the scheduler and does not need to be exposed.
 		Component(
-			"scheduleWorkflow",
+			"scheduleworkflow",
```
tebeka, line 210: Is this rename going to cause issues? Why did you rename it?

internal/backend/svc/fx.go
tebeka: How is this related to the PR?

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -30,7 +31,19 @@ message BuildRequest {
 }
 
 message BuildResponse {
-  bytes artifact = 1;
+  bytes build_file = 1;
```
tebeka, line 34: IMO `artifact` is a better name. In Python it's a tar file with everything in it.

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -30,7 +31,19 @@ message BuildRequest {
...
+  program.v1.Error error = 2;
+}
+
+message Build1Request {
+  map<string, bytes> resources = 1;
+  repeated string symbols = 2 [(buf.validate.field).repeated.items.string.min_len = 1];
```
tebeka, line 40: Why do we need symbols in build request? What do they represent?

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -30,7 +31,19 @@ message BuildRequest {
...
+  string path = 3 [(buf.validate.field).string.min_len = 1];
```
tebeka, line 41: Is this the entry point?

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -51,6 +64,96 @@ message RunResponse {
   map<string, values.v1.Value> result = 3;
 }
 
+message BidiRunCall {
+  values.v1.Value value = 1;
```
tebeka, line 68: Why is it a value and not an entry point? (e.g. `program.py:on_slack_command`)

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -51,6 +64,96 @@ message RunResponse {
...
+  repeated values.v1.Value args = 2 [(buf.validate.field).repeated.items.required = true];
+  map<string, values.v1.Value> kwargs = 4 [
+    (buf.validate.field).map.keys.string.min_len = 1,
+    (buf.validate.field).map.values.required = true
+  ];
+}
+
+message BidiRunCallReturn {
+  oneof result {
```
tebeka, line 77: IMO `oneof` just makes it harder to work with. Have a `result` and and `error` field. protobuf doesn't encode zero values anyway.

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -30,7 +31,19 @@ message BuildRequest {
 }
 
 message BuildResponse {
-  bytes artifact = 1;
+  bytes build_file = 1;
+  program.v1.Error error = 2;
+}
+
+message Build1Request {
```
tebeka, line 38: What's the `Bidi` prefix for?

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -51,6 +64,96 @@ message RunResponse {
...
+}
+
+message BidiRunCallReturn {
+  oneof result {
+    values.v1.Value value = 1;
+    program.v1.Error error = 2;
+  }
+}
+
+message BidiRunLoadReturn {
```
tebeka, line 83: On the bigger picture. I don't think `Load` should return exported values. The `Call` should fail if the user tries to call a non existing symbol.

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -51,6 +64,96 @@ message RunResponse {
...
+  map<string, values.v1.Value> values = 1 [
+    (buf.validate.field).map.keys.string.min_len = 1,
+    (buf.validate.field).map.values.required = true
+  ];
+
+  program.v1.Error error = 2;
+}
+
+message BidiRunRequest {
```
tebeka, line 92: I don't understand why run can be several things? We have distinct steps for running.

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -51,6 +64,96 @@ message RunResponse {
...
+  message StartData {
+    string run_id = 1 [(buf.validate.field).string.min_len = 1];
+    map<string, values.v1.Value> globals = 2 [
+      (buf.validate.field).map.keys.string.min_len = 1,
+      (buf.validate.field).map.values.required = true
+    ];
+    string path = 3;
```
tebeka, line 99: `entry_point`?

proto/autokitteh/runtimes/v1/svc.proto
```diff
@@ -51,6 +64,96 @@ message RunResponse {
...
+  }
+
+  message Start {
+    bytes build_file = 1;
+    StartData data = 2 [(buf.validate.field).required = true];
+  }
+
+  message Start1 {
```
tebeka, line 107: Find a better name.

sdk/internal/loggedstream/logged_stream.go
```diff
@@ -0,0 +1,53 @@
...
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+)
+
+type Stream[Rx, Tx any] interface {
+	Send(*Tx) error
+	Receive() (*Rx, error)
+}
+
+type LoggedStream[Rx, Tx any] struct {
```
tebeka, line 16: This is better done with interceptor.

sdk/internal/loggedstream/logged_stream.go
```diff
@@ -0,0 +1,53 @@
...
+	SL     *zap.SugaredLogger
```
tebeka, line 17: `SL` -> `Logger`

sdk/internal/loggedstream/logged_stream.go
```diff
@@ -0,0 +1,53 @@
...
+	S      Stream[Rx, Tx]
+	DescRx func(*Rx) string
+	DescTx func(*Tx) string
+	Level  zapcore.Level
```
tebeka, line 21: Why is `Level` here and not configured in `SL`?

sdk/sdkclients/internal/newclient.go
```diff
@@ -19,5 +24,24 @@ func New[
...
-	return f(p.HTTPClient, p.URL, p.Options...)
+	httpClient := p.HTTPClient
+
+	if httpClient == nil || http.DefaultClient == httpClient {
+		// see https://connectrpc.com/docs/go/deployment#h2c.
+
+		httpClient = &http.Client{
+			Transport: &http2.Transport{
+				AllowHTTP: true,
+				DialTLS: func(network, addr string, _ *tls.Config) (net.Conn, error) {
```
tebeka, line 35: IMO we'll want to encrypt traffic, so there should be an option to get key & cert files.
@efiShtain ?

sdk/sdkclients/sdkclient/sdkclient.go
```diff
@@ -22,11 +23,16 @@ type Params struct {
 	URL        string
 	Options    []connect.ClientOption
 	AuthToken  string
+	L          *zap.Logger
 }
 
 func (p Params) Safe() Params {
 	if p.HTTPClient == nil {
-		p.HTTPClient = &http.Client{}
+		p.HTTPClient = http.DefaultClient
```
tebeka, line 31: Why?

sdk/sdkclients/sdkruntimesclient/bidi_run.go
```diff
@@ -0,0 +1,285 @@
...
+				Args:   kittehs.Transform(args, sdktypes.ToProto),
+				Kwargs: kittehs.TransformMapValues(kwargs, sdktypes.ToProto),
+			},
+		},
+	}); err != nil {
+		return sdktypes.InvalidValue, fmt.Errorf("send call: %w", err)
+	}
+
+	var callret *runtimesv1.BidiRunCallReturn
+	for {
```
tebeka, line 47: There are other messages that might come in: Logs, Prints ...

sdk/sdkclients/sdkruntimesclient/bidi_run.go
```diff
@@ -0,0 +1,285 @@
...
+	return sdktypes.StrictValueFromProto(callret.GetValue())
+}
+
+func (r *bidiRun) handleCallback(ctx context.Context, msg *runtimesv1.BidiRunResponse) error {
+	switch msg := msg.Response.(type) {
+	case *runtimesv1.BidiRunResponse_Print_:
+		r.cbs.SafePrint(ctx, r.rid, msg.Print.Text)
+		return nil
+
+	case *runtimesv1.BidiRunResponse_NewRunId:
```
tebeka, line 82: Why do we need new run ID? Every new event spins a new runtime for now.

sdk/sdkclients/sdkruntimesclient/bidi_run.go
```diff
@@ -0,0 +1,285 @@
...
+		if err != nil {
+			return fmt.Errorf("decode args: %w", err)
+		}
+
+		kwargs, err := kittehs.TransformMapValuesError(msg.Call.Kwargs, sdktypes.StrictValueFromProto)
+		if err != nil {
+			return fmt.Errorf("decode kwargs: %w", err)
+		}
+
+		ret, err := r.cbs.SafeCall(ctx, r.rid, v, args, kwargs)
```
tebeka, line 113: Are these calls to `sleep` and friends? We wanted to make the explicit.

sdk/sdkclients/sdkruntimesclient/run.go
```diff
@@ -25,3 +31,57 @@ func (r *run) Call(context.Context, sdktypes.Value, []sdktypes.Value, map[string
...
+func (c *client) run(
+	ctx context.Context,
+	rid sdktypes.RunID,
+	path string,
+	build *sdkbuildfile.BuildFile,
+	globals map[string]sdktypes.Value,
+	cbs *sdkservices.RunCallbacks,
+) (sdkservices.Run, error) {
+	var a bytes.Buffer
+	if err := build.Write(&a); err != nil {
```
tebeka, line 44: General comment: If `BuildFile` was implementing `io.Reader` you can use `io.Copy`

sdk/sdkclients/sdkruntimesclient/run.go
```diff
@@ -25,3 +31,57 @@ func (r *run) Call(context.Context, sdktypes.Value, []sdktypes.Value, map[string
...
+
+		if msg.Result != nil {
+			result, err = kittehs.TransformMapValuesError(msg.Result, sdktypes.StrictValueFromProto)
+			if err != nil {
+				stream.Close()
+				return nil, fmt.Errorf("invalid result: %w", err)
+			}
+		}
+	}
+
```
tebeka, line 83: Need to check `stream.Err` here

sdk/sdkclients/sdkruntimesclient/runtime.go
```diff
@@ -21,7 +28,36 @@ func (r *runtime) Build(
 	path string,
 	symbols []sdktypes.Symbol,
 ) (sdktypes.BuildArtifact, error) {
-	return sdktypes.InvalidBuildArtifact, sdkerrors.ErrNotImplemented
+	resources, err := kittehs.FSToMap(fs)
```
tebeka, line 31: General note: Pass either `fs` or even the artifact `bytes` to `Build1`

tebeka (changes requested):
I have way too many comments, probably missing something. We need to discuss face to face.

backend/runtimes/runtimes.go
itayd: koanf doesn't support that

backend/runtimes/runtimes.go
itayd: if it's not getting the logger, it logs to NopLogger. Did not think this was arguable :P

backend/runtimes/runtimes.go
itayd: not sure what you mean here

PR #240, opened 2024-05-05
Bye bye tokens

itayd:

PR #570, opened 2024-08-16
fix(github)!: remove owner from OAuth connection var names

daabr:
Remove the GitHub app installation's owner (org/user) from the names of OAuth connection vars.

This change spans both the AK server and the Python SDK - both need to be released and deployed together!

Core files in this PR (the rest are just small adjustments):

- integrations/github
  - internal/vars/vars.go
  - jwt.go
  - oauth.go
- runtimes/pythonrt/py-sdk/autokitteh/github.py

Tested manually with this Python code snippet:

```python
import random

from autokitteh.github import github_client

# https://docs.github.com/en/rest/reactions/reactions#about-reactions
REACTIONS = ["+1", "-1", "laugh", "confused", "heart", "hooray", "rocket", "eyes"]

def on_github_issue_comment(event):
    github = github_client("github_conn")
    repo = github.get_repo(event.data.repo.id)
    issue = repo.get_issue(event.data.issue.number)
    comment = issue.get_comment(event.data.comment.id)
    reaction = random.choice(REACTIONS)
    comment.create_reaction(reaction)
```

Refs: ENG-1422, ENG-1445

BREAKING CHANGE: OAuth connection vars in existing GitHub connections need to be renamed.

### DIFF
--- a/integrations/github/actions.go
+++ b/integrations/github/actions.go
@@ -32,7 +32,7 @@ func (i integration) listWorkflowRuns(ctx context.Context, args []sdktypes.Value
 	}
 
 	// Invoke the API method.
-	gh, err := i.NewClient(ctx, owner)
+	gh, err := i.NewClient(ctx)
 	if err != nil {
 		return sdktypes.InvalidValue, err
 	}
--- a/integrations/github/checks.go
+++ b/integrations/github/checks.go
@@ -24,7 +24,7 @@ func (i integration) createCheckRun(ctx context.Context, args []sdktypes.Value, 
 	}
 
 	// Invoke the API method.
-	gh, err := i.NewClient(ctx, owner)
+	gh, err := i.NewClient(ctx)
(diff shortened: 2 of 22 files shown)


pashafateev:
What was the original intent behind having installation's owner (org/user) in the names of OAuth connection vars?
To know when user is part of org?

daabr:
> What was the original intent behind having installation's owner (org/user) in the names of OAuth connection vars? To know when user is part of org?

This was added in #240. I missed this aspect due to the PR's size.

pashafateev (approved):

daabr (commented):
I confirmed with Haim that the product expectation is that connection re-inits will overwrite each other, not accumulate.

Otherwise, we don't have a user-friendly way to manage (i.e. remove) them.

If need need to aggregate/combine/alias multiple connections as one in the future - we'll need to treat that as a new feature rather than a hack.

PR #571, opened 2024-08-17
fix: gracefully error out on apply with objects that are not found

itayd:
... instead of segfaulting.

### DIFF
--- a/internal/backend/connections/connections.go
+++ b/internal/backend/connections/connections.go
@@ -30,6 +30,10 @@ func (c *Connections) Create(ctx context.Context, conn sdktypes.Connection) (sdk
 		return sdktypes.InvalidConnectionID, err
 	}
 
+	if intg == nil {
+		return sdktypes.InvalidConnectionID, sdkerrors.ErrNotFound
+	}
+
 	status := intg.Get().InitialConnectionStatus()
 	if !status.IsValid() {
 		// get the connection status of a new connection.
--- a/internal/manifest/exec.go
+++ b/internal/manifest/exec.go
@@ -59,11 +59,15 @@ func executeAction(ctx context.Context, action actions.Action, execContext *exec
 		pid, err := execContext.resolveProjectID(ctx, action.ProjectKey)
 		if err != nil {
 			return nil, err
+		} else if !pid.IsValid() {
+			return nil, fmt.Errorf("project %q not found", action.ProjectKey)
 		}
 
 		iid, err := execContext.resolveIntegrationID(ctx, action.IntegrationKey)
 		if err != nil {
 			return nil, err
+		} else if !iid.IsValid() {
+			return nil, fmt.Errorf("integration %q not found", action.IntegrationKey)
 		}
 
 		conn := action.Connection.WithProjectID(pid).WithIntegrationID(iid)
@@ -97,6 +101,8 @@ func executeAction(ctx context.Context, action actions.Action, execContext *exec
 		pid, err := execContext.resolveProjectID(ctx, action.ProjectKey)
 		if err != nil {
 			return nil, err
+		} else if !pid.IsValid() {
+			return nil, fmt.Errorf("project %q not found", action.ProjectKey)
 		}
 
 		env := action.Env.WithProjectID(pid)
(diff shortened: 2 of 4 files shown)


examples/testprograms/http/autokitteh.yaml
```diff
@@ -15,6 +15,8 @@ project:
           value: "moo"
         - name: PIG
           value: "oink"
+    - name: gemini2
```
ig-ra: just to ensure that this addition is related to this PR and should be in examples?

internal/manifest/exec.go
```diff
@@ -59,11 +59,15 @@ func executeAction(ctx context.Context, action actions.Action, execContext *exec
 		pid, err := execContext.resolveProjectID(ctx, action.ProjectKey)
 		if err != nil {
 			return nil, err
+		} else if !pid.IsValid() {
```
ig-ra, line 62: maybe I'm missing something, but what's the additional value of handling both error and invalid object cases here (and all other places below) instead of handling them in execContext.resolve* and returning the proper nonFound error instead of returning invalid placeholders?

in resolveIntegration in case of any error `InvalidIntegraiton, err` will be returned, which means that `invalidIntegration => not found` only if `err != nil`. IMO, it's not very obvious for anyone who is using the code. I think that `NotFound` should be the same level error as any other error and it shouldn't be deduced using isValid check, but returned as a proper error.

ig-ra: execContext is using resolver and I recall that I handled many of those cases in resolver recently and for many objects resolver is already returning NotFound. But this shoudl be extended for integrations, since integrations handled differently in resolver - see `resolver.translateError`

internal/backend/connections/connections.go
```diff
@@ -30,6 +30,10 @@ func (c *Connections) Create(ctx context.Context, conn sdktypes.Connection) (sdk
 		return sdktypes.InvalidConnectionID, err
 	}
 
+	if intg == nil {
```
ig-ra, line 33: maybe it's better to handle this in integration interface? e.g. return sdk.ErrNotFound along with nil integration?

examples/testprograms/http/autokitteh.yaml
tebeka: The name "examples" is bad IMO, this code is mostly for tests.

internal/manifest/exec.go
tebeka: Agree with @ig-ra that only error checking. More than that `IsValid` only checks that it's not the zero value, it does not check the database that there is such object.

tests/system/testdata/manifest/apply.txtar
```diff
@@ -68,6 +74,27 @@ Error: invalid YAML input: yaml: unmarshal errors:
...
 version: v1
 
+-- wrong_integration.yaml --
+version: v1
+
+project:
+  name: my_telepathy_project
+  connections:
+    - name: my_connection
+      integration: telepathy
```
tebeka, line 85: Add a comment here that this is an unknown integration

examples/testprograms/http/autokitteh.yaml
itayd: yeah, i will drop this change. it's more for manual testing. maybe it shouldn't belong in the repo at all.

tebeka (approved):
Approving since @itayd says he'll behave :)

examples/testprograms/http/autokitteh.yaml
itayd: done

tests/system/testdata/manifest/apply.txtar
itayd: done

internal/backend/connections/connections.go
itayd: there are some stuff that rely on it returning nil/invalid. we need to make a pass over all of these and settle on a convention.

internal/manifest/exec.go
itayd: there are some stuff that rely on it returning nil/invalid. we need to make a pass over all of these and settle on a convention.

PR #572, opened 2024-08-18
feat(metrics): session duration

ig-ra:
add session duraiton metric

### DIFF
--- a/internal/backend/sessions/sessionworkflows/metrics.go
+++ b/internal/backend/sessions/sessionworkflows/metrics.go
@@ -10,6 +10,9 @@ var (
 	sessionsCompletedCounter metric.Int64Counter
 	sessionsErroredCounter   metric.Int64Counter
 	sessionsStoppedCounter   metric.Int64Counter
+
+	// NOTE: this is session time from AK perspective. Maybe we could ask temporal for better metrics?
+	sessionDurationHistogram metric.Int64Histogram
 )
 
 func initMetrics(t *telemetry.Telemetry) {
@@ -17,4 +20,5 @@ func initMetrics(t *telemetry.Telemetry) {
 	sessionsCompletedCounter, _ = t.NewCounter("sessions.completed", "Completed sessions counter")
 	sessionsErroredCounter, _ = t.NewCounter("sessions.errored", "Errored sessions counter")
 	sessionsStoppedCounter, _ = t.NewCounter("sessions.stopped", "Stopped sessions counter")
+	sessionDurationHistogram, _ = t.NewHistogram("sessions.duration", "Session duration histogram")
 }
--- a/internal/backend/sessions/sessionworkflows/workflows.go
+++ b/internal/backend/sessions/sessionworkflows/workflows.go
@@ -220,15 +220,16 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 		prints, err = runWorkflow(wctx, l, ws, data, params.Debug)
 		duration = time.Since(startTime)
 		sessionDurationHistogram.Record(ctx, duration.Milliseconds())
+		l = l.With(zap.Duration("duration", duration))
 	}
 
 	if err != nil {
-		sl := l.With(zap.Error(err))
+		l := l.With(zap.Error(err))
 
 		if errors.Is(err, workflow.ErrCanceled) || errors.Is(wctx.Err(), workflow.ErrCanceled) {
 			sessionsStoppedCounter.Add(ctx, 1)
 
-			sl.Info("session workflow: canceled", zap.Duration("duration", duration))
+			l.Info("session workflow: canceled")
 			ws.stopped(ctx, params.SessionID)
 		} else {
 			sessionsErroredCounter.Add(ctx, 1)
@@ -238,14 +239,14 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 				// User level error, no need to indicate the workflow as errored.
 				err = nil
 
-				sl.Info("session workflow: program error", zap.Duration("duration", duration))
+				l.Info("session workflow: program error")
 			} else {
-				sl.Error("session workflow: error", zap.Duration("duration", duration))
+				l.Error("session workflow: error")
 			}
 		}
 	} else {
 		sessionsCompletedCounter.Add(ctx, 1)
-		l.Info("session workflow: completed", zap.Duration("duration", duration))
+		l.Info("session workflow: completed")
 	}
 
 	if data != nil {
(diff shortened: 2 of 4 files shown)


internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -204,23 +203,26 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
 
 		sessionsCreatedCounter.Add(ctx, 1)
 
-		z.Info("session workflow: started")
-		prints, err = runWorkflow(wctx, z, ws, data, params.Debug)
+		l.Info("session workflow: started", zap.Int32("attempt", wi.Attempt))
+		startTime := time.Now()
+		prints, err = runWorkflow(wctx, l, ws, data, params.Debug)
+		duration = time.Since(startTime)
+		sessionDurationHistogram.Record(ctx, duration.Milliseconds())
```
efiShtain, line 222: Does this take into account replay? 
Failure of code and restart of the service?

internal/backend/sessions/sessionworkflows/metrics.go
```diff
@@ -10,11 +10,15 @@ var (
 	sessionsCompletedCounter metric.Int64Counter
 	sessionsErroredCounter   metric.Int64Counter
 	sessionsStoppedCounter   metric.Int64Counter
+
+	// NOTE: this is session time from AK perspective. Maybe we could ask temporal for better metrics?
+	sessionDurationHistogram metric.Int64Histogram
```
efiShtain, line 15: That's a valid point
Maybe we should report when session workflow is done and not the actual session code

Wdyt?

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: could u plz give me exact use case that you think we are not handling?

internal/backend/sessions/sessionworkflows/metrics.go
ig-ra: That's ther reason why I added a NOTE :) 
I think that for now it's OK, but we should think how to get a better metric.
btw, we are measuring the session workflow (in AK). what I would like to measure is the actual time spent in workers to run the session.

so, IMO it's good for now. I won't try to invent anything better before we are trying to work with a dashboard and see what's missing

internal/backend/sessions/sessionworkflows/workflows.go
itayd: efi means that when the workflow is being replayed, this will count the execution twice. you _could_ ignore replays, just wrap with a `workflow.IsReplay(wctx)`, but I'm not sure that's the right approach.

we'll need to discuss this in sync.

ig-ra: when workflow is replayed it will be different session, no?

ig-ra: rewritten - now for the sessions we are replaying, the startTime will be set to one we have in DB

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -195,32 +193,42 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
 	}
 
 	if err != nil {
-		z := z.With(zap.Error(err))
+		sl := l.With(zap.Error(err))
 
 		if errors.Is(err, workflow.ErrCanceled) || errors.Is(wctx.Err(), workflow.ErrCanceled) {
 			sessionsStoppedCounter.Add(ctx, 1)
```
efiShtain, line 230: we might need a cancelled counter as well
most of the times there should be no cancelling at all

internal/backend/sessions/sessionworkflows/workflows.go
efiShtain: maybe we need to add a dimention on the final state
otherwise canclled/errored sessions might give us outliers on actual completed runs

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -195,32 +193,42 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
+			fields = append(fields, zap.Int32("attempt", wi.Attempt))
+		} else {
+			sessionsCreatedCounter.Add(ctx, 1)
+		}
 
-		z.Info("session workflow: started")
-		prints, err = runWorkflow(wctx, z, ws, data, params.Debug)
+		startTime = data.Session.CreatedAt()
+		l.Info("session workflow: started", fields...)
+		prints, err = runWorkflow(wctx, l, ws, data, params.Debug)
```
efiShtain, line 220: I'm not sure, 
do you know that if we pass this line, the session is for sure done?
if error is returned there would be no retry or something?
and what about if there server crashed before updating the session state to a final state in db?

i prefer to lost a metric than having an incorrect metric

could you verify this please?

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -195,32 +193,42 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
+		duration = time.Since(startTime)
+		sessionDurationHistogram.Record(ctx, duration.Milliseconds())
 	}
 
 	if err != nil {
-		z := z.With(zap.Error(err))
+		sl := l.With(zap.Error(err))
```
efiShtain: you can just add the duration here since you are creating a new logger anyway, instead of adding the duration all over below
or even before the error so it works for the success cases as well

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: we have cancelled sessions counter

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: talked

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: fixed

internal/backend/sessions/sessionworkflows/metrics.go
efiShtain: we decided to use AK's time, so I'm resolving this

internal/backend/sessions/sessionworkflows/workflows.go
efiShtain: we take the latest time of a final state and compare to start time
this is basically full session from AK's perspective, not temporal, which is better for us
resolving

efiShtain (approved):

PR #575, opened 2024-08-20
feat(ENG-1440): db layer for value storage

itayd:
value storage for projects will use the regular db instead of redis. this supplies the db layer which a service will be built upon to allow the script to manipulate environment wide variables.

### DIFF
--- a/internal/backend/db/db.go
+++ b/internal/backend/db/db.go
@@ -155,4 +155,8 @@ type DB interface {
 
 	// -----------------------------------------------------------------------
 	GetOwnership(ctx context.Context, entityID sdktypes.UUID) (string, error)
+
+	SetValue(ctx context.Context, pid sdktypes.ProjectID, key string, v sdktypes.Value) error
+	GetValue(ctx context.Context, pid sdktypes.ProjectID, key string) (sdktypes.Value, error)
+	ListValues(ctx context.Context, pid sdktypes.ProjectID) (map[string]sdktypes.Value, error)
 }
--- a/internal/backend/db/dbgorm/scheme/scheme.go
+++ b/internal/backend/db/dbgorm/scheme/scheme.go
@@ -33,7 +33,6 @@ func UUIDOrNil(uuid sdktypes.UUID) *sdktypes.UUID {
 var Tables = []any{
 	&Build{},
 	&Connection{},
-	&Var{},
 	&Deployment{},
 	&Env{},
 	&Event{},
@@ -46,6 +45,8 @@ var Tables = []any{
 	&SessionLogRecord{},
 	&Signal{},
 	&Trigger{},
+	&Value{},
+	&Var{},
 }
 
 type Build struct {
@@ -532,4 +533,15 @@ type Ownership struct {
 
 	// TODO: Polymorphic associations won't enforce foreign key to entities in different tables
 	// we might need to have separated ownership tables for each entity.
+}
+
+type Value struct {
+	ProjectID sdktypes.UUID `gorm:"primaryKey;type:uuid;not null"`
+	Key       string        `gorm:"primaryKey;not null"`
+	Value     []byte
+
+	UpdatedAt time.Time
+
+	// foreign key
+	Project *Project
 }
--- a/internal/backend/db/dbgorm/values.go
+++ b/internal/backend/db/dbgorm/values.go
@@ -1,0 +1,87 @@
+package dbgorm
+
+import (
+	"context"
+	"time"
+
+	"google.golang.org/protobuf/proto"
+	"gorm.io/gorm"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/db/dbgorm/scheme"
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+const maxValueSize = 64 * 1024
+
+func (db *gormdb) SetValue(ctx context.Context, pid sdktypes.ProjectID, key string, v sdktypes.Value) error {
+	if !pid.IsValid() {
+		return sdkerrors.NewInvalidArgumentError("invalid project id")
+	}
+
+	if v.ProtoSize() > maxValueSize {
+		return sdkerrors.NewInvalidArgumentError("value too large > %d bytes", maxValueSize)
+	}
+
+	uid, err := userIDFromContext(ctx)
+	if err != nil {
+		return err
+	}
+
+	return translateError(db.transaction(ctx, func(tx *tx) error {
+		db := tx.db
+
+		oo, err := tx.owner.EnsureUserAccessToEntitiesWithOwnership(tx.ctx, db, uid, pid.UUIDValue())
+		if err != nil {
+			return err
+		}
+		if len(oo) != 1 {
+			return gorm.ErrForeignKeyViolated
+		}
+
+		bs, err := proto.Marshal(v.ToProto())
+		if err != nil {
+			return err
+		}
+
+		return tx.db.Save(&scheme.Value{ProjectID: pid.UUIDValue(), Key: key, Value: bs, UpdatedAt: time.Now().UTC()}).Error
+	}))
+}
+
+func (db *gormdb) GetValue(ctx context.Context, pid sdktypes.ProjectID, key string) (sdktypes.Value, error) {
+	r, err := getOne[scheme.Value](db.withUserEnvs(ctx), "project_id = ? AND key = ?", pid.UUIDValue(), key)
+	if err != nil {
+		return sdktypes.InvalidValue, translateError(err)
+	}
+
+	var pb sdktypes.ValuePB
+
+	if err := proto.Unmarshal(r.Value, &pb); err != nil {
+		return sdktypes.InvalidValue, err
+	}
+
+	return sdktypes.ValueFromProto(&pb)
+}
+
+func (db *gormdb) ListValues(ctx context.Context, pid sdktypes.ProjectID) (map[string]sdktypes.Value, error) {
+	var rs []*scheme.Value
+	if err := db.withUserEnvs(ctx).Where("project_id = ?", pid.UUIDValue()).Find(&rs).Error; err != nil {
+		return nil, translateError(err)
+	}
+
+	return kittehs.ListToMapError(rs, func(r *scheme.Value) (string, sdktypes.Value, error) {
+		var pb sdktypes.ValuePB
+
+		if err := proto.Unmarshal(r.Value, &pb); err != nil {
+			return "", sdktypes.InvalidValue, err
+		}
+
+		v, err := sdktypes.ValueFromProto(&pb)
+		if err != nil {
+			return "", sdktypes.InvalidValue, err
+		}
+
+		return r.Key, v, err
+	})
+}
--- a/go.mod
+++ b/go.mod
@@ -92,6 +92,7 @@ require (
 	google.golang.org/api v0.197.0
 	google.golang.org/grpc v1.66.2
 	google.golang.org/protobuf v1.34.2
+	gopkg.in/yaml.v2 v2.4.0
 	gopkg.in/yaml.v3 v3.0.1
 	gorm.io/datatypes v1.2.2
 	gorm.io/driver/postgres v1.5.9
@@ -269,7 +270,6 @@ require (
 	golang.org/x/time v0.6.0 // indirect
 	google.golang.org/genproto/googleapis/api v0.0.0-20240903143218-8af14fe29dc1 // indirect
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20240903143218-8af14fe29dc1 // indirect
-	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gorm.io/driver/mysql v1.5.7 // indirect
 	gorm.io/driver/sqlserver v1.5.3 // indirect
 	gotest.tools/v3 v3.5.1 // indirect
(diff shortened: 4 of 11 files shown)


migrations/postgres/20240820035603_db-values.sql
```diff
@@ -0,0 +1,16 @@
+-- +goose Up
+-- create "values" table
+CREATE TABLE "values" (
+  "env_id" uuid NOT NULL,
+  "key" text NOT NULL,
```
efiShtain: I guess you need a primary key on (env_id, key) to make it a unique tuple
you can't have two keys with the same name under the same env_id

am I missing something?

internal/backend/db/dbgorm/values.go
```diff
@@ -0,0 +1,73 @@
...
+		return sdkerrors.NewInvalidArgumentError("invalid env_id")
+	}
+
+	return db.transaction(ctx, func(tx *tx) error {
+		v, err := getValue(ctx, &tx.gormdb, envID, key)
+		if err != nil && !errors.Is(err, gorm.ErrRecordNotFound) {
+			return err
+		}
+
+		if v, err = f(v); err != nil {
```
efiShtain: why is this a callback based update instead of just getting the value and setting it?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -553,3 +554,9 @@ type Ownership struct {
 	// TODO: Polymorphic associations won't enforce foreign key to entities in different tables
 	// we might need to have separated ownership tables for each entity.
 }
+
+type Value struct {
+	EnvID sdktypes.UUID `gorm:"primaryKey;type:uuid;not null"`
```
efiShtain: do we need an expiry flag ? 
what about a limit on size ? 
what about a limit on number of entities per env ?

internal/backend/db/db.go
```diff
@@ -161,4 +161,9 @@ type DB interface {
 
 	// -----------------------------------------------------------------------
 	GetOwnership(ctx context.Context, entityID sdktypes.UUID) (string, error)
+
+	// -----------------------------------------------------------------------
+	UpdateValue(ctx context.Context, envID sdktypes.EnvID, key string, f func(sdktypes.Value) (sdktypes.Value, error)) error
```
efiShtain: maybe we should prefix it with some string to indicate what are this values
UpdateValue is just too generic

itayd: i don't follow - you got the `key` there. it's all in the context of a specific key.

internal/backend/db/dbgorm/scheme/scheme.go
itayd: all limits should be done in the service layer i think?

re expiry - i agree, but next step?

internal/backend/db/dbgorm/values.go
itayd: atomic operations inside a transaction.

migrations/postgres/20240820035603_db-values.sql
itayd: true. i also want to support scanning all keys for a specific env. how do you suggest i do it?

efiShtain: depends on design, 
if there are no key duplicates then
key is string and value is json, so just get the value of this key and you have a the full list

I think you can paginate json array using ```jsonb_array_elements```

if there could be multiple key with the same string, then just standard select

internal/backend/db/db.go
efiShtain: I mean something like UpdateKeyValValue or something that group this functions as a logical group

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: we can do both I guess
but I'm ok with service layer only for now

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -46,6 +45,8 @@ var Tables = []any{
 	&SessionLogRecord{},
 	&Signal{},
 	&Trigger{},
+	&Value{},
```
efiShtain, line 48: Maybe KeyValue is more suitable than just value in this case

internal/backend/db/dbgorm/values.go
```diff
@@ -0,0 +1,59 @@
...
+	j, err := v.MarshalJSON()
+	if err != nil {
+		return err
+	}
+
+	return translateError(db.db.Save(&scheme.Value{EnvID: envID.UUIDValue(), Key: key, Value: j}).Error)
+}
+
+func getValue(ctx context.Context, db *gormdb, envID sdktypes.EnvID, key string) (sdktypes.Value, error) {
+	r, err := getOne[scheme.Value](db.withUserEnvs(ctx), "env_id = ? AND name = ?", envID.String(), key)
```
efiShtain: what is name ?
should it be key ?

internal/backend/db/dbgorm/values.go
```diff
@@ -0,0 +1,59 @@
...
+
+func (db *gormdb) GetValue(ctx context.Context, envID sdktypes.EnvID, key string) (sdktypes.Value, error) {
+	return getValue(ctx, db, envID, key)
+}
+
+func (db *gormdb) ListValues(ctx context.Context, envID sdktypes.EnvID) (map[string]sdktypes.Value, error) {
+	// TODO: authz.
+
+	var rs []*scheme.Value
+	if err := db.db.WithContext(ctx).Where("env_id = ?", envID.UUIDValue()).Find(&rs).Error; err != nil {
```
efiShtain: this will work for now, but you need to consider paginating this somehow if you plan to support more than 10-20 values

efiShtain: security issue

you don't check the user is allowed to access this field

internal/backend/db/dbgorm/values.go
```diff
@@ -0,0 +1,59 @@
...
+	if !envID.IsValid() {
+		return sdkerrors.NewInvalidArgumentError("invalid env_id")
+	}
+
+	j, err := v.MarshalJSON()
+	if err != nil {
+		return err
+	}
+
+	return translateError(db.db.Save(&scheme.Value{EnvID: envID.UUIDValue(), Key: key, Value: j}).Error)
```
efiShtain: I think we should find a way to limit the size of it

1. we are not a storage service
2. parsing JSON is expensive for large objects
3. We have to have an expiry otherwise we would collect stuff forever


2 is the most important that I'm concerned about
assume a workflow fetches a huge json from somewhere and try to store it and restore it, it can easily block the process with it

internal/backend/db/dbgorm/values.go
```diff
@@ -0,0 +1,59 @@
...
+	"context"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/db/dbgorm/scheme"
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func (db *gormdb) SetValue(ctx context.Context, envID sdktypes.EnvID, key string, v sdktypes.Value) error {
+	if !envID.IsValid() {
```
efiShtain: you don't verify the user has access to this envID

internal/backend/db/dbgorm/values.go
itayd: fixed

internal/backend/db/dbgorm/values.go
itayd: true. for now let's keep it though.

itayd: fixed

internal/backend/db/dbgorm/values.go
itayd: we'll do limits in the service layer, not including expiry though.

internal/backend/db/dbgorm/values.go
itayd: fixed

internal/backend/db/dbgorm/scheme/scheme.go
itayd: i think `Value` is more consistent.

internal/backend/db/dbgorm/values.go
itayd: also - changed serialization to proto
no need to do json

itayd: aaaand added size limit
hard coded for now

internal/backend/db/dbgorm/values.go
```diff
@@ -0,0 +1,90 @@
...
+		return sdkerrors.NewInvalidArgumentError("invalid env_id")
+	}
+
+	if v.ProtoSize() > maxValueSize {
+		return sdkerrors.NewInvalidArgumentError("value too large > %d bytes", maxValueSize)
+	}
+
+	uid, err := userIDFromContext(ctx)
```
ig-ra, line 27: @itayd
please add tests using the convension you could see in other dbgorm-related files, e.g. split `SetValue` to 2 functions, one of which, i.e. `setValue` should work with schema args and add tests to it.

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -497,3 +498,9 @@ type Ownership struct {
 	// TODO: Polymorphic associations won't enforce foreign key to entities in different tables
 	// we might need to have separated ownership tables for each entity.
 }
+
+type Value struct {
+	EnvID sdktypes.UUID `gorm:"primaryKey;type:uuid;not null"`
+	Key   string        `gorm:"primaryKey;not null"`
+	Value []byte
```
ig-ra: foreign key to envID

internal/backend/db/dbgorm/values.go
```diff
@@ -0,0 +1,90 @@
...
+	if err != nil {
+		return err
+	}
+
+	return translateError(db.transaction(ctx, func(tx *tx) error {
+		db := tx.db
+
+		oo, err := tx.owner.EnsureUserAccessToEntitiesWithOwnership(tx.ctx, db, uid, envID.UUIDValue())
+		if err != nil {
+			if errors.Is(err, gorm.ErrRecordNotFound) {
```
ig-ra: I think you don't need this and this block could be deleted. It feels that you copied the code from `Var`, but there  it's implemented this way since scopeID is either `Env` or `Connection`, means we cannot set foreign key constraint. In `Value` struct it could be done simpler, with fk constraint, so I'd prefer to enforce this in DB and not in function logic

internal/backend/db/dbgorm/values.go
```diff
@@ -0,0 +1,90 @@
...
+	if err := proto.Unmarshal(r.Value, &pb); err != nil {
+		return sdktypes.InvalidValue, err
+	}
+
+	return sdktypes.ValueFromProto(&pb)
+}
+
+func (db *gormdb) ListValues(ctx context.Context, envID sdktypes.EnvID) (map[string]sdktypes.Value, error) {
+	var rs []*scheme.Value
+	if err := db.withUserEnvs(ctx).WithContext(ctx).Where("env_id = ?", envID.UUIDValue()).Find(&rs).Error; err != nil {
```
ig-ra: there is no need for `WithContext` after `withUserEnvs`

ig-ra (changes requested):
please add tests.

internal/backend/db/dbgorm/values.go
efiShtain: fetching too much stuff from the database bothers me more than sending stuff to the user

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: I think it says nothing 
but ok

internal/backend/db/dbgorm/scheme/scheme.go
itayd: done

internal/backend/db/dbgorm/values.go
itayd: removed

internal/backend/db/dbgorm/values.go
itayd: removed

itayd:
> please add tests.

done, black box style.

internal/backend/db/dbgorm/values.go
itayd: done black box testing.

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -533,3 +534,12 @@ type Ownership struct {
 	// TODO: Polymorphic associations won't enforce foreign key to entities in different tables
 	// we might need to have separated ownership tables for each entity.
 }
+
+type Value struct {
+	EnvID sdktypes.UUID `gorm:"primaryKey;type:uuid;not null"`
```
efiShtain: Let's link it to the project directly instead of env

I think we should refactor to remove envs as a whole and adding more stuff to it right now won't help

What do you think?

migrations/postgres/20241029052632_values.sql
```diff
@@ -0,0 +1,13 @@
+-- +goose Up
+-- create "values" table
+CREATE TABLE "values" (
+  "env_id" uuid NOT NULL,
```
efiShtain: I think we should add created_at and updated_at columns

We should also think about deleted_at for soft delete
Cause users might delete this value while a workflow is still using them and can replay

internal/backend/db/dbgorm/scheme/scheme.go
itayd: i think we should keep env for now. we'll probably kill them, but i want it to be somewhat consistent.

migrations/postgres/20241029052632_values.sql
itayd: adding these for debugging/auditing reasons makes sense, but it doesn't matter for replayability - these things will run in activities which will save the result and will not run multiple times.

internal/backend/db/dbgorm/scheme/scheme.go
itayd: ok, i will change to project.

efiShtain (approved):

internal/backend/db/dbgorm/scheme/scheme.go
itayd: done

migrations/postgres/20241029052632_values.sql
itayd: added updated_at

PR #577, opened 2024-08-20
feat(db): return old deployment state on update

ig-ra:


internal/backend/db/dbgorm/deployments.go
```diff
@@ -53,20 +54,43 @@ func (gdb *gormdb) deleteDeploymentsAndDependents(ctx context.Context, depIDs []
...
+		if err := tx.db.Model(&d).Update("state", int32(state.ToProto())).Error; err != nil {
+			return err
 		}
 		return nil
-	})
+	}); err != nil {
+		return sdktypes.DeploymentStateUnspecified, err
+	}
+	state, err := sdktypes.DeploymentStateFromProto(deploymentsv1.DeploymentState(oldState))
+	if err != nil {
```
efiShtain: so what about the transaction in this case ?
it already happened, 
what's the system state now ?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -470,6 +470,13 @@ type Deployment struct {
 	Ownership *Ownership `gorm:"polymorphic:Entity;"`
 }
 
+func (d *Deployment) BeforeUpdate(tx *gorm.DB) (err error) {
+	if tx.Statement.Changed() { // if any fields changed
+		tx.Statement.SetColumn("UpdatedAt", time.Now())
```
efiShtain, line 475: wasn't this working already at the db it self before ?

ig-ra: nop, we were updating it manually in `updateDeploymentState`

internal/backend/db/dbgorm/deployments.go
ig-ra: plz explain? I don't understand

efiShtain: I think if you might have an invalid state in the db (this is what you check here)
then you should fail the transaction instead of committing it and then return error on the old state

ig-ra: I could change it to kittehs.must1.
we cannot have invalid state in DB, it's converted from sdktypes.DeploymentState which defined well and cannot be some unknown value

ig-ra: changed to kittehs.Must1

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: ok, so we probably need to add it to other tables as well or add a trigger in db
anyhow resolving this

efiShtain (approved):

PR #580, opened 2024-08-21
fix(google): repeat #539 for JSON

daabr:
Saving Google JSON keys requires an authenticated user identity, like OAuth.

~Also, because we save secret vars twice, make sure their values are not changed after the first save.~

False alarm about that 2nd part: `postInit()` already adds scope IDs to the vars, so we can `Finalize(vs)` instead of `Finalize(vsl)` (which was modified by `h.vars.Set`).

Refs: ENG-1470

### DIFF
--- a/integrations/google/creds.go
+++ b/integrations/google/creds.go
@@ -123,11 +123,6 @@ func (h handler) finalize(ctx context.Context, c sdkintegrations.ConnectionInit,
 		return
 	}
 
-	// Reset the list of vars because "h.vars.Set()" modifies the values of secret vars.
-	vsl = kittehs.TransformMapToList(vs.ToMap(), func(_ sdktypes.Symbol, v sdktypes.Var) sdktypes.Var {
-		return v.WithScopeID(sdktypes.NewVarScopeID(cid))
-	})
-
 	if err := forms.UpdateWatches(ctx, h.vars, cid); err != nil {
 		l.Error("Google Forms watches creation error", zap.Error(err))
 		c.AbortServerError("form watches creation error")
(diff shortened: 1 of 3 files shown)


integrations/google/creds.go
```diff
@@ -123,6 +123,11 @@ func (h handler) finalize(ctx context.Context, c sdkintegrations.ConnectionInit,
 		return
 	}
```
ig-ra: @daabr
- is there any other integration that could potentially modify the list of vars via Set?
- could we check this in c.Finalize and error to prevent/report errors?

ig-ra (approved):

ig-ra:
plz (before merge) check whether we could ensure that finalize is receiving secret vars as key:value and not as key: uuid ref to secret storage

integrations/google/creds.go
```diff
@@ -123,6 +123,11 @@ func (h handler) finalize(ctx context.Context, c sdkintegrations.ConnectionInit,
...
 
+	// Reset the list of vars because "h.vars.Set()" modifies the values of secret vars.
```
itayd: wait what, Set modified the inputs?!

itayd: this doesn't make sense as all sdktypes are immutable

integrations/google/creds.go
daabr: Discussed offline

integrations/google/creds.go
daabr: https://github.com/autokitteh/autokitteh/blob/main/internal/backend/vars/vars.go#L43

itayd: oh got it.

PR #584, opened 2024-08-22
fix: reduce image size and harden permissions and make sure it builds on m1 Mac

efiShtain:
moved python build to a separate stage - rebuild images works faster, image size reduced

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -28,7 +28,7 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
     CGO_ENABLED=0 go build -o /bin/ak -ldflags="${LDFLAGS}" ./cmd/ak
 EOF
 
-FROM python:3.12 AS python_deps
+FROM python:3.12-bookworm AS python_deps
 
 WORKDIR /app
 
@@ -55,7 +55,7 @@ RUN /app/env/bin/python3 -m pip install --no-cache-dir -r requirements.txt
 # Create a new stage for running the application that contains the minimal
 # runtime dependencies for the application.
 
-FROM python:3.12-alpine3.20 AS final
+FROM python:3.12-slim-bookworm AS final
 
 
 # Create a non-privileged user 
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,12 +21,35 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
 # source code into the container.
 RUN --mount=type=cache,target=/go/pkg/mod/ \
     --mount=type=bind,target=. \
-<<EOF
+    <<EOF
     export VERSION_PKG_PATH="go.autokitteh.dev/autokitteh/internal/version"
     export TIMESTAMP="$(date -u "+%Y-%m-%dT%H:%MZ")"
-    export LDFLAGS="-X "${VERSION_PKG_PATH}.Version=$(cat .version || echo)" -X "${VERSION_PKG_PATH}.Time=${TIMESTAMP}" -X "${VERSION_PKG_PATH}.Commit=$(cat .commit || echo)""
+    export LDFLAGS="-s -w -X "${VERSION_PKG_PATH}.Version=$(cat .version || echo)" -X "${VERSION_PKG_PATH}.Time=${TIMESTAMP}" -X "${VERSION_PKG_PATH}.Commit=$(cat .commit || echo)""
     CGO_ENABLED=0 go build -o /bin/ak -ldflags="${LDFLAGS}" ./cmd/ak
 EOF
+
+FROM python:3.12 AS python_deps
+
+WORKDIR /app
+
+# RUN --mount=type=cache,target=/var/cache/apk \
+#     apt-get \
+#     g++ libstdc++ \
+#     ca-certificates \
+#     tzdata \
+#     && \
+#     update-ca-certificates
+
+COPY ./scripts/gcc-on-arm.sh /tmp
+RUN /tmp/gcc-on-arm.sh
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -21,53 +21,62 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
...
-
-# Create initial venv
-RUN python3 -m venv ~/.local/share/autokitteh/venv
-COPY --chown=appuser ./runtimes/pythonrt/requirements.txt /tmp/requirements.txt
-RUN ~/.local/share/autokitteh/venv/bin/python -m pip install --no-cache-dir -r /tmp/requirements.txt
-RUN rm /tmp/requirements.txt
 
+WORKDIR /usr/src/autokitteh
 # Copy the executable from the "build" stage.
+COPY --chown=appuser:appuser --from=python_deps /app/env /usr/src/autokitteh/.local/share/autokitteh/venv
```
ig-ra, line 74: I'm unsure, but maybe `/usr/src` isn't the right place for this. I would go for usual `/home` or `/opt`

Dockerfile
```diff
@@ -21,53 +21,62 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
...
+
+
 # Create a non-privileged user 
 # See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#user
 ARG UID=10001
 RUN adduser \
     --disabled-password \
     --gecos "" \
-    --home "/nonexistent" \
+    --home "/usr/src/autokitteh" \
```
ig-ra, line 67: why not in regular place for user's home? e.g. /home/someuser?

ig-ra (changes requested):

Dockerfile
```diff
@@ -21,53 +21,62 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
 # source code into the container.
 RUN --mount=type=cache,target=/go/pkg/mod/ \
     --mount=type=bind,target=. \
-<<EOF
+    <<EOF
     export VERSION_PKG_PATH="go.autokitteh.dev/autokitteh/internal/version"
     export TIMESTAMP="$(date -u "+%Y-%m-%dT%H:%MZ")"
-    export LDFLAGS="-X "${VERSION_PKG_PATH}.Version=$(cat .version || echo)" -X "${VERSION_PKG_PATH}.Time=${TIMESTAMP}" -X "${VERSION_PKG_PATH}.Commit=$(cat .commit || echo)""
+    export LDFLAGS="-s -w -X "${VERSION_PKG_PATH}.Version=$(cat .version || echo)" -X "${VERSION_PKG_PATH}.Time=${TIMESTAMP}" -X "${VERSION_PKG_PATH}.Commit=$(cat .commit || echo)""
```
tebeka, line 27: On my machine this reduces `ak` size from 158MB to 114MB. Do we care that much about 44MB? Does it worth losing debug information over it?

Dockerfile
```diff
@@ -21,53 +21,62 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
...
     CGO_ENABLED=0 go build -o /bin/ak -ldflags="${LDFLAGS}" ./cmd/ak
 EOF
 
-################################################################################
-# Create a new stage for running the application that contains the minimal
-# runtime dependencies for the application.
+FROM python:3.12 AS python_deps
```
tebeka: `python:3.12` is based on debian and you copy it over to an alpine based Docker. IMO use the python:3.12-alpine3.20

Dockerfile
```diff
@@ -21,53 +21,62 @@ RUN --mount=type=cache,target=/go/pkg/mod/ \
...
+#     tzdata \
+#     && \
+#     update-ca-certificates
 
 COPY ./scripts/gcc-on-arm.sh /tmp
 RUN /tmp/gcc-on-arm.sh
 RUN rm /tmp/gcc-on-arm.sh
 
+
+RUN python3 -m venv env
```
tebeka, line 48: I'm not sure copying `venv` is the right thing, you can do it but it has risks. Why not skip this step and install on the final image? I don't think we have any dependencies that require compilation.

Dockerfile
efiShtain: lower image size allows us to spin more instances, loads them faster to memory, cost reduction on moving bytes in AWS and so on...

we can have a debug and final image if we want

Dockerfile
efiShtain: for building or as the final image ?

Dockerfile
efiShtain: it makes stuff goes much faster if you don't reinstall the packages on every go file change
and it reduces the final image size since you don't manage all the build tools required

Dockerfile
efiShtain: it's not a real user and it has no shell
so I'm not sure it matters

but let me know if its critical to you and I'll try and change it

Dockerfile
efiShtain: python code expects to find it in a specific place currently so I kept the same place
I just reorganized the layers a bit and make sure it builds on m1 Macs

Dockerfile
efiShtain: I can probably go with bookworm and bookworm-slim to make it compatible

tebeka: Up to you. But the Python build and final should be same base image. IMO debian will have less issues.

Dockerfile
tebeka: OK

tebeka (approved):

PR #586, opened 2024-08-22
fix(sdk/trigger): strict validate mandatory connection

ig-ra:


efiShtain (approved):
approved as I don't see any problem with the change

but is it really needed?
what's changed? what does it solve ?

PR #587, opened 2024-08-22
fix(cli/trigger): disallow empty connection on create

ig-ra:
- ability to threat empty args as not-provided ones
- report not found types in errors
- disallow empty connection in trigger create

### DIFF
--- a/cmd/ak/common/flags.go
+++ b/cmd/ak/common/flags.go
@@ -6,6 +6,7 @@ import (
 	"strings"
 
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/internal/resolver"
@@ -67,4 +68,20 @@ func ToExitCodeWithSkipNotFoundFlag(cmd *cobra.Command, err error, whats ...stri
 		}
 	}
 	return exitErr
+}
+
+// maybe we could just run this check for all required flags instead of providing explicit ones?
+func UnsetEmptyFlags(cmd *cobra.Command, flags ...string) {
+	flagsMap := make(map[string]struct{}, len(flags))
+	for _, flag := range flags {
+		flagsMap[flag] = struct{}{}
+	}
+
+	cmd.Flags().Visit(func(flag *pflag.Flag) {
+		if _, ok := flagsMap[flag.Name]; ok {
+			if flag.Value.Type() == "string" && flag.Changed && flag.Value.String() == "" {
+				flag.Changed = false
+			}
+		}
+	})
 }
--- a/cmd/ak/cmd/triggers/create.go
+++ b/cmd/ak/cmd/triggers/create.go
@@ -1,7 +1,6 @@
 package triggers
 
 import (
-	"errors"
 	"fmt"
 
 	"github.com/spf13/cobra"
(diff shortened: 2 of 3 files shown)


cmd/ak/common/flags.go
```diff
@@ -68,3 +69,19 @@ func ToExitCodeWithSkipNotFoundFlag(cmd *cobra.Command, err error, whats ...stri
 	}
 	return exitErr
 }
+
+// maybe we could just run this check for all required flags instead of providing explicit ones?
```
itayd, line 73: give it a shot

cmd/ak/common/flags.go
```diff
@@ -68,3 +69,19 @@ func ToExitCodeWithSkipNotFoundFlag(cmd *cobra.Command, err error, whats ...stri
...
+func UnsetEmptyFlags(cmd *cobra.Command, flags ...string) {
+	flagsMap := make(map[string]struct{}, len(flags))
+	for _, flag := range flags {
+		flagsMap[flag] = struct{}{}
+	}
+
```
itayd, line 79: why is this needed?

cmd/ak/common/flags.go
```diff
@@ -68,3 +69,19 @@ func ToExitCodeWithSkipNotFoundFlag(cmd *cobra.Command, err error, whats ...stri
 	}
 	return exitErr
 }
+
+// maybe we could just run this check for all required flags instead of providing explicit ones?
+func UnsetEmptyFlags(cmd *cobra.Command, flags ...string) {
+	flagsMap := make(map[string]struct{}, len(flags))
```
itayd, line 75: it's easier just to `make(map[string]bool, len(flags)`. then check if `if flagsMap[whatever] {`.

cmd/ak/common/flags.go
```diff
@@ -68,3 +69,19 @@ func ToExitCodeWithSkipNotFoundFlag(cmd *cobra.Command, err error, whats ...stri
...
+	for _, flag := range flags {
+		flagsMap[flag] = struct{}{}
+	}
+
+	cmd.Flags().Visit(func(flag *pflag.Flag) {
+		if _, ok := flagsMap[flag.Name]; ok {
+			if flag.Value.Type() == "string" && flag.Changed && flag.Value.String() == "" {
+				flag.Changed = false
```
itayd, line 83: don't we want to reset the value as well?

cmd/ak/common/flags.go
ig-ra: I'm afraid to apply this to all required arguments actually :) That's why 
So I prefer (meanwhile) to explicitly specify commands and which arguments cannot be empty

cmd/ak/common/flags.go
ig-ra: sorry, I failed to understand your comment.
The idea is to pass in `flags` arguments that should be unset if they are empty. This will allow cobra to run all other validation logic as if they were not passed by the user.
Therefore I want to convert them to map to easily check later in callback called from `Visit` function on each arg set by user

cmd/ak/common/flags.go
ig-ra: I don't want to unset all empty args passed by user, but only those I asked explicitly in `flags`. So in `Visit` which will visit all user-set args I need to check whether the args is one from the list requested

cmd/ak/common/flags.go
ig-ra: no, because the value in both cases is `""` e.g. empty string.
But in one it's a default value for argument and in the other it's the value passed by the user

PR #589, opened 2024-08-22
fix(ENG-1456): events flaky test to not be flaky

itayd:
two fixes:
- test condition was indeterministic - didn't know which subscription would catch first.
- another indeterminism was that ingest of event was slower than running the ak watch, so now it can wait for session to be created.


itayd:
Merging as this is blocking. please review anyway.

PR #590, opened 2024-08-22
kill `examples/` and clitest

itayd:
- cli tests are pretty much covered in system tests
- examples/ is just confusing, we got ample of these in samples


daabr (commented):
I agree with this PR, and it will save us time and CI cycles. But I want to make sure we're not throwing the baby out with the bathwater, i.e. that none of the (important) checks in `clitest` are lost.

tebeka:
> But I want to make sure we're not throwing the baby out with the bathwater, i.e. that none of the (important) checks in `clitest` are lost.

I agree with @daabr, are there tests in cli that are not in system?

daabr (approved):

PR #592, opened 2024-08-23
feat(integrations): add connStatus with authType to aws, chatgpt, discord and twilio

pashafateev:
Manually tested connections using Python samples

**Refs**:
- #583 
- eng-1481


integrations/twilio/webhooks/auth.go
```diff
@@ -43,17 +44,23 @@ func (h handler) HandleAuth(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	at := ""
```
itayd:
```suggestion
	auth := ""
```

integrations/twilio/client.go
```diff
@@ -40,6 +44,49 @@ func New(vars sdkservices.Vars) sdkservices.Integration {
...
+			return sdktypes.InvalidStatus, err
+		}
+
+		at := vs.Get(s)
+		if !at.IsValid() || at.Value() == "" {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "init required"), nil
+		}
+
+		// Align with:
+		// https://github.com/autokitteh/web-platform/blob/main/src/enums/connections/connectionTypes.enum.ts
```
itayd: why ts and not the proto?

integrations/twilio/client.go
```diff
@@ -40,6 +44,49 @@ func New(vars sdkservices.Vars) sdkservices.Integration {
...
+
+		vs, err := i.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		var decodedVars webhooks.Vars
+		vs.Decode(&decodedVars)
+
+		s, err := sdktypes.ParseSymbol("AuthType")
```
itayd: i would put this in the global vars, using NewSymbol. this is effectively a constant.

integrations/twilio/client.go
```diff
@@ -40,6 +44,49 @@ func New(vars sdkservices.Vars) sdkservices.Integration {
...
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		at := vs.Get(s)
```
itayd: auth?

integrations/aws/client.go
```diff
@@ -86,10 +92,44 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
...
+		connStatus(i),
+		sdkintegrations.WithConnectionConfigFromVars(cvars),
 	)
 }
+
+// connStatus is an optional connection status check provided by the
+// integration with AutoKitteh. The possible results are "init required"
+// (indicating the connection is not yet usable), "using X" (indicating
+// one of multiple available authentication methods is in use), or
+// "initialized" when only one authentication method is available.
```
daabr: Per #583 :

```suggestion
// connStatus is an optional connection status check provided by
// the integration to AutoKitteh. The possible results are "init
// required" (the connection is not usable yet) and "using X".
```

integrations/aws/client.go
```diff
@@ -86,10 +92,44 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
...
+		}
+
+		at := vs.Get(authType)
+		if !at.IsValid() || at.Value() == "" {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "init required"), nil
+		}
+
+		// Align with:
+		// https://github.com/autokitteh/web-platform/blob/main/src/enums/connections/connectionTypes.enum.ts
+		switch at.Value() {
```
daabr: See #595:
1. Remove comment
2. Use a const

Unrelated to #595 :
3. If there's only one option by definition, then use `if-else` instead of `switch`
4. Why not a generic `"initialized"` instead of an AWS-only `"awsConfig"`? Then other integrations could reuse this const

integrations/chatgpt/client.go
```diff
@@ -40,6 +46,39 @@ func New(vars sdkservices.Vars) sdkservices.Integration {
...
 		sdkintegrations.WithConnectionConfigFromVars(vars),
 	)
 }
+
+// connStatus is an optional connection status check provided by the
+// integration with AutoKitteh. The possible results are "init required"
+// (indicating the connection is not yet usable), "using X" (indicating
+// one of multiple available authentication methods is in use), or
+// "initialized" when only one authentication method is available.
+func connStatus(i *integration) sdkintegrations.OptFn {
```
daabr, line 58: See related comments in AWS

integrations/discord/client.go
```diff
@@ -28,10 +36,43 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
...
 		sdkintegrations.WithConnectionConfigFromVars(cvars),
 	)
 }
+
+// connStatus is an optional connection status check provided by the
+// integration with AutoKitteh. The possible results are "init required"
+// (indicating the connection is not yet usable), "using X" (indicating
+// one of multiple available authentication methods is in use), or
+// "initialized" when only one authentication method is available.
+func connStatus(i *integration) sdkintegrations.OptFn {
```
daabr, line 52: See related comments in AWS

integrations/twilio/client.go
```diff
@@ -1,6 +1,10 @@
 package twilio
 
 import (
+	"context"
+
+	// "go.autokitteh.dev/autokitteh/integrations/twilio/webhooks"
+	"go.autokitteh.dev/autokitteh/integrations/twilio/webhooks"
```
daabr: Did you forget to delete this comment?

integrations/twilio/client.go
daabr: Also, why `AuthType` and not `authType`?

integrations/twilio/client.go
daabr: `at` is taken from #583. I prefer this acronym over "auth" because I don't want to create ambiguity between "auth type" and "auth data".

integrations/twilio/client.go
daabr: AFAIK this is not related to any proto...?

integrations/twilio/webhooks/auth.go
```diff
@@ -20,6 +20,7 @@ const (
 )
 
 type Vars struct {
+	AuthType   string
```
daabr: Note that `EncodeVars` uses the field's name, i.e. `AuthType`, but it should be `authType`. What you did in AWS should work here (although I would argue that might want to remove `EncodeVars` entirely, but that's beyond the scope of this PR).

integrations/aws/client.go
pashafateev: Yes, I took that and added the piece about "initialized". If @itayd wants "initialized" to be used, I think the doc comment should reflect that.

integrations/twilio/client.go
pashafateev: Yes :)

integrations/twilio/client.go
pashafateev: I don't understand what's happening here

integrations/aws/client.go
pashafateev: 1. done
2. added a TODO for when we resolve the name below

3. The decision to use switch was based on the idea that we're going to add more ways to connect soon.
4. I thought it had to match one of [these](https://github.com/autokitteh/web-platform/blob/main/src/enums/connections/connectionTypes.enum.ts)

integrations/twilio/client.go
pashafateev: @itayd this is happening through EncodeVars. I tried removing EncodeVars, but it breaks a lot of the starlark stuff so I decided not to touch it. I could do a separate vars, but I thought it looked cleaner this way.

integrations/twilio/webhooks/auth.go
pashafateev: I started removing EncodeVars but it broke a lot of stuff that had nothing to do with this PR, so I left it. Should I open a new PR for that?

integrations/twilio/client.go
pashafateev: @daabr because
> Note that EncodeVars uses the field's name, i.e. AuthType

**i.e.**

`auth.go`:
```
type Vars struct {
        // line below
	AuthType   string
	AccountSID string
	Username   string `vars:"secret"`
	Password   string `vars:"secret"`
}
```

integrations/aws/client.go
pashafateev: remove "using x"

integrations/aws/client.go
pashafateev: Done

integrations/aws/client.go
pashafateev: Done

integrations/chatgpt/client.go
pashafateev: Done

integrations/discord/client.go
pashafateev: Done

integrations/twilio/client.go
pashafateev: Done

integrations/twilio/webhooks/auth.go
pashafateev: Done

integrations/aws/client.go
```diff
@@ -86,10 +93,39 @@ var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.Integrati
...
+		}
+
+		at := vs.Get(authType)
+		if !at.IsValid() || at.Value() == "" {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "init required"), nil
+		}
+
+		if at.Value() == integrations.Init {
+			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "initialized"), nil
+		} else {
```
daabr: According Go idiomatic style, no need for `else` when both branches simply `return` - use this:

```go
if ... {
    return ...
}
return ...
```

...And in other files too

integrations/twilio/client.go
```diff
@@ -40,6 +44,40 @@ func New(vars sdkservices.Vars) sdkservices.Integration {
...
+		vs.Decode(&decodedVars)
+
+		at := vs.Get(webhooks.AuthType)
+		if !at.IsValid() || at.Value() == "" {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "init required"), nil
+		}
+
+		switch at.Value() {
+		case integrations.APIKey:
+			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "using api key"), nil
```
daabr:
```suggestion
			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "using API key"), nil
```

integrations/twilio/webhooks/auth.go
```diff
@@ -43,12 +46,17 @@ func (h handler) HandleAuth(w http.ResponseWriter, r *http.Request) {
...
+
 	accountSID := r.Form.Get("account_sid")
 	username := accountSID
 	password := r.Form.Get("auth_token")
 	if password == "" {
 		username = r.Form.Get("api_key")
 		password = r.Form.Get("api_secret")
+		at = integrations.APIKey
+	} else {
+		at = integrations.APIToken
```
daabr: More idiomatic Go style: remove the `else` branch completely, and assign `at := integrations.APIToken` in line 49.

daabr (approved):
Approved with a few cosmetic style comments

integrations/aws/client.go
pashafateev: Done

integrations/twilio/client.go
pashafateev: done

integrations/twilio/webhooks/auth.go
pashafateev: Done

PR #593, opened 2024-08-24
feat: add var description

itayd:
could be useful to have description as a first class thing in a var. we could show it in the UI as well.


cmd/ak/cmd/vars/set.go
```diff
@@ -44,7 +47,7 @@ var setCmd = common.StandardCommand(&cobra.Command{
 			value = string(data)
 		}
 
-		ev := sdktypes.NewVar(n).SetValue(value).SetSecret(secret).SetOptional(optional).WithScopeID(id)
+		ev := sdktypes.NewVar(n).SetValue(value).SetSecret(secret).SetOptional(optional).WithScopeID(id).SetDescription(description)
 		if err != nil {
```
daabr, line 51: This error check is meaningless, nothing is setting `err` here.

cmd/ak/cmd/vars/set.go
```diff
@@ -11,10 +11,13 @@ import (
...
 
-var secret, optional bool
+var (
+	secret, optional bool
+	description      string
+)
 
 var setCmd = common.StandardCommand(&cobra.Command{
-	Use:     "set <key> [<value>] [--secret] [--optional] <--env=.. | --connection=....> [--project=...]",
+	Use:     "set <key> [<value>] [--secret] [--description <DESC>] [--optional] <--env=.. | --connection=....> [--project=...]",
```
daabr, line 20: 1. I honestly don't believe that anyone will ever use this flag in the CLI
2. This allows anyone to override an existing description of a var, which makes it even less useful as far as documentation goes

cmd/ak/cmd/vars/set.go
```diff
@@ -44,7 +47,7 @@ var setCmd = common.StandardCommand(&cobra.Command{
 			value = string(data)
 		}
 
-		ev := sdktypes.NewVar(n).SetValue(value).SetSecret(secret).SetOptional(optional).WithScopeID(id)
+		ev := sdktypes.NewVar(n).SetValue(value).SetSecret(secret).SetOptional(optional).WithScopeID(id).SetDescription(description)
```
daabr, line 50: Using a builder-like pattern when you set all the values is an incorrect usage of the builder pattern. This was not added in the PR, but this PR makes it even clearer that it's not the right choice.

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -110,12 +110,13 @@ func ParseConnection(c Connection) (sdktypes.Connection, error) {
...
-	Value      string
-	IsSecret   bool
-	IsOptional bool
+	VarID       sdktypes.UUID `gorm:"primaryKey;index;type:uuid;not null"`
+	ScopeID     sdktypes.UUID `gorm:"-"`
+	Name        string        `gorm:"primaryKey;index;not null"`
+	Value       string
+	Description string
+	IsSecret    bool
+	IsOptional  bool
```
daabr, line 119: Throughout this PR you define/use these parameters in a different order every time.

daabr (commented):
While this PR technically works, I think its (incremental) complication outweighs its value. AFAIK this wasn't discussed or planned, and does not improve anything important or urgent in the product.

PR #594, opened 2024-08-24
feat(ENG-1474): kill reveal var

itayd:
no need for it in the api level. better do it only in the UI level.


cmd/ak/cmd/vars/get.go
```diff
@@ -33,12 +33,7 @@ var getCmd = common.StandardCommand(&cobra.Command{
 		ctx, cancel := common.LimitedContext()
 		defer cancel()
 
-		get := vars().Get
-		if reveal {
-			get = vars().Reveal
-		}
-
-		vs, err := get(ctx, id, ks...)
+		vs, err := vars().Get(ctx, id, ks...)
```
daabr, line 36: Please also remove the reveal flag - it's not useful for CLI users, only (maybe) in web UIs.

daabr (approved):
Approved, but please remove the `--reveal` flag from the CLI `vars get` command, it's not useful for CLI users

cmd/ak/cmd/vars/get.go
itayd: discussed offline: cli is ui, and we don't want to accidently reveal secrets.

PR #595, opened 2024-08-24
chore(integrations): auth mode consts for connStatus funcs

daabr:


RonenMars (approved):
LGTM and I'll be glad you expose the `integrations/auth.go` file using gRPC so that we can match and validate our values with the frontend, that way, we will protect ourselves (not 100%) from having different **authTypes** in use on frontend and the backend.

integrations/auth.go
```diff
@@ -0,0 +1,11 @@
+package integrations
+
+// Align these with:
+// https://github.com/autokitteh/web-platform/blob/main/src/enums/connections/connectionTypes.enum.ts
+const (
+	APIToken   = "apiToken"
+	JSONKey    = "jsonKey" // Google integrations only
+	OAuth      = "oauth"
+	PAT        = "pat"
+	SocketMode = "socketMode" // Slack integration only
```
pashafateev, line 10: I noticed that not all types found [here](https://github.com/autokitteh/web-platform/blob/main/src/enums/connections/connectionTypes.enum.ts) are represented. Is this intentional?

pashafateev (approved):

daabr:
> LGTM and I'll be glad you expose the `integrations/auth.go` file using gRPC so that we can match and validate our values with the frontend, that way, we will protect ourselves (not 100%) from having different **authTypes** in use on frontend and the backend.

Let's discuss offline, I don't think a gRPC endpoint is valuable.

integrations/auth.go
daabr: Yes, the list in this PR is (or rather will be) the authoritative one. The TS file is tentative.

PR #597, opened 2024-08-25
fix(cli/trigger): disallow empty connection

ig-ra:


internal/backend/db/dbgorm/triggers.go
```diff
@@ -126,7 +126,7 @@ func (db *gormdb) triggerToRecord(ctx context.Context, trigger sdktypes.Trigger)
 }
 
 func (db *gormdb) CreateTrigger(ctx context.Context, trigger sdktypes.Trigger) error {
-	if err := trigger.Strict(); err != nil {
+	if err := trigger.Strict(); err != nil { // name, connection, env but not project
```
daabr, line 129: This is not a proper place for a comment - either above `if` or inside the block. More importantly, I don't understand what this comment means - please write more.

cmd/ak/cmd/triggers/create.go
```diff
@@ -15,8 +14,7 @@ import (
 
 var (
 	call, event, filter, name, project, schedule string
-
-	data map[string]string
+	data                                         map[string]string
```
daabr, line 17: This is a style issue, but in my opinion there are too many whitespaces here, so it's worse not better than before - that's why there was an empty line.

cmd/ak/cmd/triggers/create.go
```diff
@@ -38,37 +36,28 @@ var createCmd = common.StandardCommand(&cobra.Command{
...
 		// Project and/or environment are required.
-		p, _, err := r.ProjectNameOrID(ctx, project)
-		if err != nil {
-			return err
-		}
-		if project != "" && !p.IsValid() {
-			err = resolver.NotFoundError{Type: "project", Name: project}
-			return common.NewExitCodeError(common.NotFoundExitCode, err)
+		if project != "" {
+			p, _, err := r.ProjectNameOrID(ctx, project)
```
daabr, line 40: `r.ProjectNameOrID(ctx, "")` returns an invalid project, which would cause `common.AddNotFoundErrIfCond(err, p.IsValid())` to return an error - as expected.

So `if project != ""` is not only redundant, but actually harmful, because it passes the error handling of an empty project to `r.EnvNameOrID(ctx, env, project)` below.

Same issue with `if env != ""` below, I think.

cmd/ak/common/types.go
```diff
@@ -0,0 +1,24 @@
+package common
+
+import "errors"
+
+type nonEmptyStringValue string
```
daabr, line 5: You need to document what is the problem that this type solves, and add demonstrative unit tests.

tebeka (approved):

daabr (approved):
I'm approving this PR (with a couple of small comments), and I agree that it's better than #587, but I still think that using `VarP` and a custom type instead of a simple `StringVarP` is not so clearly an improvement, just to handle the rare case of CLI parameters that use env vars which are empty.

cmd/ak/cmd/triggers/create.go
ig-ra: daniel, this is vcode default formatting. I did nothing there. 
I removed empty line between string and map variables and vscode formatted it like this.
I'm sure that your vscode will do the same, so it's not my style preferences but default go formatter ones

cmd/ak/cmd/triggers/create.go
ig-ra: I think you missed the original logic. Which is remain the same, I only used simpler functions defined in common package

PR #599, opened 2024-08-25
feat(metrics): api requests counters + duration

ig-ra:


internal/backend/httpsvc/intercept.go
```diff
@@ -51,7 +54,54 @@ func (r *responseInterceptor) Flush() {
...
+		return errors.New("invalid service path")
+	}
+	service = dotParts[1]
+
+	cntName := fmt.Sprintf("api.%s.%s", service, api)
+	histName := fmt.Sprintf("api.%s.%s.duration", service, api)
+
+	counter, ok := metrics.counters[cntName]
+	if !ok {
+		if counter, err = t.NewCounter(cntName, fmt.Sprintf("HTTP request counter (%s)", cntName)); err != nil {
```
efiShtain:
```suggestion
		if counter, err = t.NewCounter(cntName, fmt.Sprintf("GRPC request counter (%s)", cntName)); err != nil {
```

internal/backend/httpsvc/intercept.go
```diff
@@ -51,7 +54,54 @@ func (r *responseInterceptor) Flush() {
...
+	counters  map[string]metric.Int64Counter
+	durations map[string]metric.Int64Histogram
+}{make(map[string]metric.Int64Counter), make(map[string]metric.Int64Histogram)}
+
+func updateMetric(ctx context.Context, t *telemetry.Telemetry, path string, statusCode int, duration time.Duration) (err error) {
+	if !strings.HasPrefix(path, "/autokitteh.") {
+		return nil // only internal service APIs
+	}
+
+	slashParts := strings.Split(path, "/")
```
efiShtain: could you add a comment with the "algorithm" here ?
not sure I get why you split and check 3 parts and then lower and search for . and do so on...

internal/backend/httpsvc/intercept.go
```diff
@@ -51,7 +54,54 @@ func (r *responseInterceptor) Flush() {
 
 type RequestLogExtractor func(*http.Request) []zap.Field
 
-func intercept(z *zap.Logger, cfg *LoggerConfig, extractors []RequestLogExtractor, next http.Handler) (http.HandlerFunc, error) {
+var metrics = struct {
```
efiShtain: did you try the other option of having just one counter like ```api_request_total_count``` with path/status dimensions ? 
and maybe another one ```api_request_duration``` with path/status dimension as well ?

internal/backend/httpsvc/intercept.go
ig-ra: added comments to the code

internal/backend/httpsvc/intercept.go
ig-ra: done

internal/backend/httpsvc/intercept.go
ig-ra: let's talk

internal/backend/httpsvc/intercept.go
itayd: why not just setting the path as a label?

internal/backend/httpsvc/intercept.go
```diff
@@ -51,7 +55,58 @@ func (r *responseInterceptor) Flush() {
...
+	counters  map[string]metric.Int64Counter
+	durations map[string]metric.Int64Histogram
+}{make(map[string]metric.Int64Counter), make(map[string]metric.Int64Histogram)}
+
+func updateMetric(ctx context.Context, t *telemetry.Telemetry, path string, statusCode int, duration time.Duration) (err error) {
+	// path will be like "/autokitteh.projects.v1.ProjectsService/Create"
+	// 1. check this is an internal API path, e.g. starts with "/autokitteh."
+	// 2. extract service (`projects`) and API name (`create`)
+
+	if !strings.HasPrefix(path, "/autokitteh.") {
```
itayd: why not all?

internal/backend/httpsvc/intercept.go
```diff
@@ -51,7 +55,58 @@ func (r *responseInterceptor) Flush() {
...
+
+	cntName := fmt.Sprintf("api.%s.%s", service, api)
+	histName := fmt.Sprintf("api.%s.%s.duration", service, api)
+
+	counter, ok := metrics.counters[cntName]
+	if !ok {
+		if counter, err = t.NewCounter(cntName, fmt.Sprintf("GRPC request counter (%s)", cntName)); err != nil {
+			return err
+		}
+		metrics.counters[cntName] = counter
```
itayd: you need a lock around this and below

internal/backend/httpsvc/intercept.go
```diff
@@ -73,8 +128,10 @@ func intercept(z *zap.Logger, cfg *LoggerConfig, extractors []RequestLogExtracto
 
 		next.ServeHTTP(rwi, r)
 
-		duration := time.Since(startTime)
-		w.Header().Add("X-AutoKitteh-Duration", duration.Truncate(time.Microsecond).String())
+		duration := time.Since(startTime).Truncate(time.Microsecond)
```
itayd: why truncate?

internal/backend/httpsvc/intercept.go
ig-ra: we've discussed and decided to keep separated metrics. 
resolving?

internal/backend/httpsvc/intercept.go
ig-ra: rewritten using sync.Map

internal/backend/httpsvc/intercept.go
ig-ra: we were reporting it in logs with millisecond resolution (e.g. truncating microseconds). Metrics are also on millisecond resolution

itayd: doesn't matter - i see no use for truncate here.

internal/backend/httpsvc/intercept.go
itayd: ok

internal/backend/httpsvc/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	return &m, nil
+}
+
+func updateMetric(ctx context.Context, t *telemetry.Telemetry, path string, statusCode int, duration time.Duration) (err error) {
+	// path will be like "/autokitteh.projects.v1.ProjectsService/Create"
+	// 1. check this is an internal API path, e.g. starts with "/autokitteh."
+	// 2. extract service (`projects`) and API name (`create`)
+
+	if !strings.HasPrefix(path, "/autokitteh.") {
+		return nil // only internal service APIs
```
itayd, line 51: why

internal/backend/httpsvc/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	dotParts := strings.Split(slashParts[1], ".")
+	if len(dotParts) < 2 {
+		return errors.New("invalid service path")
+	}
+	service = dotParts[1]
+
+	serviceAPI := fmt.Sprintf("%s.%s", service, api) // e.g. "projects.create"
+	m, err := getServiceAPIMetrics(serviceAPI, t)
+	if err != nil {
+		return err
```
itayd: looks like you are trying to make the metric name looks nicer here, but i don't think it's worth the bother.
just convert it to something legible by otel (i guess maybe replaceing special chars with some other chars) and be done with it.

internal/backend/httpsvc/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	counter, err := t.NewCounter(cntName, fmt.Sprintf("GRPC request counter (%s)", cntName))
+	if err != nil {
+		return nil, err
+	}
+	histogram, err := t.NewHistogram(histName, fmt.Sprintf("GRPC request duration (%s)", histName))
+	if err != nil {
+		return nil, err
+	}
+	m := apiMetrics{counter, histogram}
+	metrics.LoadOrStore(serviceAPI, &m)
```
itayd: why not just `Store`?

itayd: ah, you don't want to overwrite in case something else called get in the meanwhile. I would comment about it.

internal/backend/httpsvc/metrics.go
```diff
@@ -0,0 +1,78 @@
...
+	counter  metric.Int64Counter
+	duration metric.Int64Histogram
+}
+
+var metrics sync.Map
+
+// metrics    map[string]apiMetrics
+// metricsMux sync.Mutex
+
+func getServiceAPIMetrics(serviceAPI string, t *telemetry.Telemetry) (*apiMetrics, error) {
```
itayd: i would rename `get` to `acquire` since this has a side effect.

internal/backend/httpsvc/intercept.go
ig-ra: done

internal/backend/httpsvc/metrics.go
ig-ra: this is the standard format according to the otel metric convension https://opentelemetry.io/docs/specs/semconv/general/metrics/.
it will be `ak.projects.create` for example. I'm not trying to anything nicer. Just `service name` + dot + `api name`

internal/backend/httpsvc/metrics.go
ig-ra: we talked abut this with u and Efi in the very beginning of this PR review. The idea was that this PR adds metrics for OUR APIs. If we want to have metrics on other paths we should discuss it and handle in other PR

internal/backend/httpsvc/metrics.go
ig-ra: Otherwise I need to protect this in mutex as well, yes.. and we decided to use sync.Map only because of that. Otherwise (in pure read access, after the counter was added to the map) we don't need to do anything, since OTEL code is handling sync on counter update

ig-ra: added a comment

internal/backend/httpsvc/metrics.go
ig-ra: done

itayd (approved):

PR #600, opened 2024-08-25
fix(db): kill integration in vars

ig-ra:


internal/backend/db/dbgorm/vars.go
```diff
@@ -150,12 +150,11 @@ func (db *gormdb) SetVars(ctx context.Context, vars []sdktypes.Var) error {
 		}
 
 		vr := scheme.Var{
-			ScopeID:       v.ScopeID().AsID().UUIDValue(), // scopeID is varID in DB
-			Name:          v.Name().String(),
-			Value:         v.Value(),
-			IsSecret:      v.IsSecret(),
-			IsOptional:    v.IsOptional(),
-			IntegrationID: iid.UUIDValue(),
```
itayd, line 158: you kill the set but did not remove from the scheme itself. you sure we're not using it?

itayd (approved):

PR #604, opened 2024-08-26
bug(python): Use call context when printing

tebeka:
Fixes ENG-1467


itayd:
don't think it's going to work:
- temporal wants to use `workflow.Go` to attach context. Here it's a bit of a problem, so we should discuss.
- we don't know if a print comes from inside something wrapped as an activity or from the workflow. an activity might print and we need to know it - otherwise there may be inadvertent nondeterminism.

daabr (commented):
@tebeka is this PR still needed, or is it already in the new runtime?

PR #607, opened 2024-08-27
fix: do not overwrite vars when applying a manifest

itayd:
currently when a manifest is applied, all vars are reset. after this fix, if a value is mentioned in manifest it will be reset, otherwise used from existing value.

same goes for cli - additionally set var has an optional value.

### DIFF
--- a/cmd/ak/cmd/vars/set.go
+++ b/cmd/ak/cmd/vars/set.go
@@ -16,7 +16,7 @@ var secret, optional bool
 var setCmd = common.StandardCommand(&cobra.Command{
 	Use:     "set <key> [<value>] [--secret] [--optional] <--env=.. | --connection=....> [--project=...]",
 	Short:   "Set variable",
-	Long:    "Set a variable. If <value> is not specified it will be read from standard input.",
+	Long:    `Set a variable. If <value> is "-", it is read from stdin.`,
 	Aliases: []string{"s"},
 	Args:    cobra.RangeArgs(1, 2),
 
@@ -32,16 +32,35 @@ var setCmd = common.StandardCommand(&cobra.Command{
 		}
 
 		var value string
-		if len(args) == 2 {
-			value = args[1]
-		} else {
-			const maxVarSize = 1 << 20 // 1MB
-			r := io.LimitReader(os.Stdin, maxVarSize)
-			data, err := io.ReadAll(r)
-			if err != nil {
-				return err
+
+		isValueSet := len(args) > 1
+
+		if isValueSet {
+			if args[1] != "-" {
+				value = args[1]
+			} else {
+				const maxVarSize = 1 << 20 // 1MB
+				r := io.LimitReader(os.Stdin, maxVarSize)
+				data, err := io.ReadAll(r)
+				if err != nil {
+					return err
+				}
+				value = string(data)
 			}
-			value = string(data)
+		} else {
+			// get current var in order not to overwrite it with empty.
+
+			ctx, cancel := common.LimitedContext()
+			defer cancel()
+
+			vs, err := vars().Get(ctx, id, n)
+			if err != nil {
+				return fmt.Errorf("get variable: %w", err)
+			}
+
+			if len(vs) > 0 {
+				value = vs[0].Value()
+			}
 		}
 
 		ev := sdktypes.NewVar(n).SetValue(value).SetSecret(secret).SetOptional(optional).WithScopeID(id)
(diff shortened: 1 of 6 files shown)


cmd/ak/cmd/vars/set.go
```diff
@@ -32,16 +32,35 @@ var setCmd = common.StandardCommand(&cobra.Command{
 		}
 
 		var value string
-		if len(args) == 2 {
-			value = args[1]
+
+		isValueSet := len(args) > 1
```
ig-ra, line 36: isValueProvided/Supplied?
because now it's not clear whether we are talking about old value from DB or value supplied by the user

cmd/ak/cmd/vars/set.go
```diff
@@ -32,16 +32,35 @@ var setCmd = common.StandardCommand(&cobra.Command{
...
+			} else {
+				const maxVarSize = 1 << 20 // 1MB
+				r := io.LimitReader(os.Stdin, maxVarSize)
+				data, err := io.ReadAll(r)
+				if err != nil {
+					return err
+				}
+				value = string(data)
+			}
 		} else {
```
ig-ra, line 50: Itay, for me it feels not the right solution to fetch the variable from the DB in order not to reset it if it wasn't provided. It's a lot of logic/checks which could and IMO should be avoided
- let's first discuss whether we allow empty values (I'm still in favor of not allowing them)
- if we are allowing empty values, then we should treat them as normal values, thus allowing user to override any value with empty one. For this user should always(!) provide two(!) args, if the second arg is explicitly "", then value will be overrided with the empty one

internal/manifest/manifest.go
```diff
@@ -53,10 +53,10 @@ func (c Connection) GetKey() string { return c.ProjectKey + "/" + c.Name }
 type Var struct {
 	ParentKey string `yaml:"-" json:"-"` // associated with env or connection.
 
-	Name     string `yaml:"name" json:"name" jsonschema:"required"`
-	Value    string `yaml:"value,omitempty" json:"value,omitempty"`
-	Secret   bool   `yaml:"secret,omitempty" json:"secret,omitempty"`
-	Optional bool   `yaml:"optional,omitempty" json:"optional,omitempty"`
+	Name     string  `yaml:"name" json:"name" jsonschema:"required"`
+	Value    *string `yaml:"value,omitempty" json:"value,omitempty"`
```
ig-ra, line 57: I think empty value should be explicitly passed.
Right now in DB the value type is string. So internally we don't support empty vs. unset, Therefore I think we shouldn't support it here as well

cmd/ak/cmd/vars/set.go
daabr: Or just don't use a var. `len(args) > 1` is trivial and obvious, the var isn't really adding self-documentation.

cmd/ak/cmd/vars/set.go
```diff
@@ -32,16 +32,35 @@ var setCmd = common.StandardCommand(&cobra.Command{
...
-		if len(args) == 2 {
-			value = args[1]
+
+		isValueSet := len(args) > 1
+
+		if isValueSet {
+			if args[1] != "-" {
+				value = args[1]
+			} else {
+				const maxVarSize = 1 << 20 // 1MB
```
daabr, line 42: 1. Pull this outside the function?
2. Do we really want to allow 1 MiB vars? Maybe 1 KiB? I'd say even that would probably merit using a file in the project rather than a var, and FYI Windows env vars are limited to a few KiBs too

cmd/ak/cmd/vars/set.go
```diff
@@ -32,16 +32,35 @@ var setCmd = common.StandardCommand(&cobra.Command{
...
+				if err != nil {
+					return err
+				}
+				value = string(data)
+			}
 		} else {
-			const maxVarSize = 1 << 20 // 1MB
-			r := io.LimitReader(os.Stdin, maxVarSize)
-			data, err := io.ReadAll(r)
+			// get current var in order not to overwrite it with empty.
```
daabr, line 51: Why? It's perfectly reasonable to want to clear the content of a var without deleting it.

tests/system/testdata/manifest/re-apply.txtar
```diff
@@ -1,5 +0,0 @@
-# TODO: The first apply defines a new project.
-
-# TODO: The second apply (of the same file) updates nothing.
-
-# TODO: The third apply (of a different file) updates everything.
```
daabr, line 5: Please move these TODOs to a ticket

PR #608, opened 2024-08-27
fix(integrations): use authenticated HTTP mux when saving connection vars

daabr:
Refs: ENG-1493

### DIFF
--- a/integrations/slack/server.go
+++ b/integrations/slack/server.go
@@ -24,31 +24,37 @@ const (
 	oauthPath = "/slack/oauth"
 )
 
-func Start(l *zap.Logger, mux *http.ServeMux, vs sdkservices.Vars, d sdkservices.Dispatcher) {
-	// Connection UI + save handlers.
+// Start initializes all the HTTP handlers of the Slack integration.
+// This includes connection UIs, initialization webhooks, and event webhooks.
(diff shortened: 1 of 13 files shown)


integrations/slack/server.go
```diff
@@ -24,31 +24,37 @@ const (
...
+	// Init webhooks save connection vars (via "c.Finalize" calls), so they need
+	// to have an authenticated user context, so the DB layer won't reject them.
+	// For this purpose, init webhooks are managed by the "auth" mux, which passes
+	// through AutoKitteh's auth middleware to extract the user ID from a cookie.
+	auth.Handle("GET "+oauthPath, NewHandler(l))
 
-	wsh := websockets.NewHandler(l, vs, d, desc)
-	mux.HandleFunc("POST "+formPath, wsh.HandleForm)
+	wsh := websockets.NewHandler(l, v, d, desc)
+	auth.HandleFunc("POST "+formPath, wsh.HandleForm)
```
ig-ra, line 41: also auth?

ig-ra, line 41: in other places it's called savePath, which is clearer

ig-ra (approved):

integrations/slack/server.go
daabr: Done in a stacked PR: https://github.com/autokitteh/autokitteh/pull/609

PR #610, opened 2024-08-28
kill http integration. new internal webhook integration.

itayd:
- kill http integration
- create an internal webhooks integration
- outgoing http requests are done in the session level (used by starlark)
- connections calls webhooks method on webhook connection creation, so it will automatically add a webhook slug to all new webhooks connections. this will be shown in both connection status and links.

this is phase 1. next phase is making a new webhooks trigger automatically create an http connection if not available yet.

need to fix system tests, but i want to verify first that the approach is ok.


sdk/sdkservices/connections.go
```diff
@@ -20,4 +22,6 @@ type Connections interface {
 	Get(ctx context.Context, id sdktypes.ConnectionID) (sdktypes.Connection, error)
 	Test(ctx context.Context, id sdktypes.ConnectionID) (sdktypes.Status, error)
 	RefreshStatus(ctx context.Context, id sdktypes.ConnectionID) (sdktypes.Status, error)
+
+	Intercept(ConnectionsInterceptorFunc)
```
itayd: move out of the sdk

internal/backend/integrations/webhooks/svc.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	if len(vs) > 0 {
+		return vs[0].Value(), nil
+	}
+
+	return "", nil
+}
+
+func (s *svc) interceptConnection(ctx context.Context, curr, next sdktypes.Connection) (sdktypes.Connection, error) {
```
itayd: document what this does

internal/backend/integrations/webhooks/svc.go
```diff
@@ -0,0 +1,249 @@
...
+
+	if !curr.IsValid() {
+		// create connection.
+
+		slug := shortuuid.New()
+		link := webhookPath(slug)
+
+		sl.With("slug", slug).Infof("connection %v: created webhook %q", next.ID(), link)
+
+		return next.WithLink(webhookURLLinkName, link), nil
```
itayd: output this on manifest apply

PR #616, opened 2024-08-29
feat(python): New Python runtime that will work with remote

tebeka:
This is a rewrite of the Python runtime to work with gRPC, this allows remote runtimes to work with the same Python runtime.

Putting it here for review, I'll shape up the code later.


proto/Makefile
```diff
@@ -39,3 +39,30 @@ py-dist: py-deps
...
+		remote/remote.proto
+
+.PHONY: remote-go
+remote-go:
+	mkdir -p ../runtimes/pythonrt/pb
+	protoc \
+		-I remote \
+		--go_out=../runtimes/pythonrt/pb \
+		--go_opt=paths=source_relative \
+		--go-grpc_out=../runtimes/pythonrt/pb \
```
efiShtain: I think you should generate it next to the other protos in the proto directory

the remotert would ues the same protos as well

efiShtain:
you have some references to your local env specific "autokitteh-miki.ngrok.dev"

can you change it to a more generic thing?

tebeka:
> you have some references to your local env specific "autokitteh-miki.ngrok.dev"
> 
> can you change it to a more generic thing?

This is just for my local testing. Changed.

runtimes/pythonrt/runner/Dockerfile
efiShtain, line 1: do we need a docker file here ?

runtimes/pythonrt/runner/remote_pb2.py
efiShtain, line 1: what's remote_pb2 vs remote_pb ?

runtimes/pythonrt/runner/Dockerfile
tebeka: Was for testing, deleted.

proto/Makefile
tebeka: Done.

runtimes/pythonrt/runner/remote_pb2.py
tebeka: There's only `remote_pb2*` files, that's what Python's gRPC generates.

runtimes/pythonrt/svc.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+func freePort() (int, error) {
+	conn, err := net.Listen("tcp", ":0")
+	if err != nil {
+		return 0, err
+	}
+
+	conn.Close()
+	return conn.Addr().(*net.TCPAddr).Port, nil
```
efiShtain, line 288: if you close the connection, how can you be sure the port is still free ?

runtimes/pythonrt/svc.go
```diff
@@ -0,0 +1,360 @@
...
+}
+
+// Start starts the server on a free port in a new goroutine.
+// It returns the port the server listens on.
+func (s *remoteSvc) Start() error {
+	port, err := freePort()
```
efiShtain, line 294: not sure you need this freePort call

just use 
```go
lis, err := net.Listen("tcp", addr)
if err != nil {
 	return err
 }
s.lis = lis
s.port = lis.Addr().(*net.TCPAddr).Port
```

runtimes/pythonrt/svc.go
```diff
@@ -0,0 +1,360 @@
...
+
+	if syscallFn.IsValid() {
+		svc.syscallFn = syscallFn
+	}
+
+	return &svc
+}
+
+func (s *remoteSvc) Health(context.Context, *pb.HealthRequest) (*pb.HealthResponse, error) {
+	return &pb.HealthResponse{}, nil
```
efiShtain, line 65: I'm thinking about checking with temporal here if the workflow is still active

wdyt ?
let's say someone canceled the workflow, or there was a worker restart somehow, not sure

runtimes/pythonrt/svc.go
efiShtain: do we really need to start a new server for each python process ?

we can add this capability to ak itself, and use the runner ID to get back to the correct go routine of the workflow itself

should be less resources, less error prone (I think)

tebeka:
Closing this since merged from @efiShtain 's branch.

PR #618, opened 2024-08-30
feat(integrations): add event handling for Discord integration

pashafateev:
Currently supports the following events:
- `MessageCreate`
- `MessageDelete`
- `MessageUpdate`

**Refs:**
- eng-1226


integrations/discord/save.go
```diff
@@ -44,7 +46,31 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+func infoWithToken(botToken string) (*discordgo.User, error) {
+	dg, err := discordgo.New("Bot " + botToken)
+	if err != nil {
+		return nil, err
+	}
+
+	bot, err := dg.User("@me")
+	if err != nil {
+		return nil, err
```
itayd:
```suggestion
		return nil, fmt.Errorf("user(me): %w", err)
```

integrations/discord/websockets/socketmode.go
```diff
@@ -0,0 +1,122 @@
...
+		dispatcher:    d,
+		integration:   i,
+		integrationID: i.ID(),
+	}
+}
+
+var (
+	webSocketClients = make(map[string]*discordgo.Session)
+
+	mu = &sync.Mutex{}
```
itayd:
```suggestion
	mu sync.RWMutex
```

integrations/discord/websockets/socketmode.go
```diff
@@ -0,0 +1,122 @@
...
+)
+
+func (h handler) OpenSocketModeConnection(botID, botToken string) {
```
itayd: why do we need to cache connections?

integrations/discord/websockets/socketmode.go
```diff
@@ -0,0 +1,122 @@
...
+		h.logger.Error("Error creating Discord session", zap.Error(err))
+		return
+	}
+
+	addHandlers(dg, h)
+
+	// Open a WebSocket connection to Discord
+	err = dg.Open()
+
+	if err == nil {
```
itayd:
```suggestion
	if err := dg.Open(); err != nil {
```

integrations/discord/websockets/message.go
```diff
@@ -0,0 +1,35 @@
...
+	}
+	cids, err := h.vars.FindConnectionIDs(context.Background(), h.integrationID, vars.BotTokenName, "")
+	if err != nil {
+		h.logger.Error("Failed to find connection IDs", zap.Error(err))
+		return
+	}
+	h.dispatchAsyncEventsToConnections(cids, akEvent)
+}
+
+func (h *handler) HandleMessageCreate(s *discordgo.Session, m *discordgo.MessageCreate) {
```
itayd: these should not be exported

integrations/discord/save.go
```diff
@@ -44,7 +46,31 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 	c.Finalize(sdktypes.NewVars().
-		Set(botToken, r.Form.Get("botToken"), true).
-		Set(authType, integrations.Init, false))
+		Set(vars.BotID, bot.ID, false).
+		Set(vars.BotTokenName, bt, true).
+		Set(vars.AuthType, integrations.Init, false))
+}
+
+func infoWithToken(botToken string) (*discordgo.User, error) {
+	dg, err := discordgo.New("Bot " + botToken)
```
daabr, line 58: Question: if the token is wrong, which call would return an error - `New` or `User`?

integrations/discord/save.go
```diff
@@ -44,7 +46,31 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	bt := r.Form.Get("botToken")
+
+	// get the bot's ID
+	bot, err := infoWithToken(bt)
+	if err != nil {
+		l.Warn("Failed to create new bot", zap.Error(err))
```
daabr: What does it mean to "create" a bot here? Wasn't it already created manually before? In other words, what we're creating here is a runtime client that uses the bot's definitions, so it's more accurate to sat that the problem is that using the token failed, not that the bot wasn't created.

integrations/discord/internal/vars/vars.go
```diff
@@ -0,0 +1,9 @@
+package vars
+
+import "go.autokitteh.dev/autokitteh/sdk/sdktypes"
+
+var (
+	AuthType     = sdktypes.NewSymbol("authType")
+	BotTokenName = sdktypes.NewSymbol("BotToken")
```
daabr: Why not `BotToken`? given that it's in a separate package it doesn't interfere with local variables called `botToken`

integrations/discord/server.go
```diff
@@ -28,5 +33,25 @@ func Start(l *zap.Logger, muxes *muxes.Muxes) {
...
+		data, err := v.Get(context.Background(), sdktypes.NewVarScopeID(cid))
+		if err != nil {
+			l.Error("Missing data for Discord Socket Mode", zap.Error(err))
+			continue
+		}
+
+		botToken := data.GetValue(vars.BotTokenName)
+		botID := data.GetValue(vars.BotID)
+
+		wsh.OpenSocketModeConnection(botID, botToken)
```
daabr: Nit:
```suggestion
		wsh.OpenSocketModeConnection(data.GetValue(vars.BotID), data.GetValue(vars.BotTokenName))
```

integrations/discord/websockets/message.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"context"
+
+	"github.com/bwmarrin/discordgo"
+	"go.uber.org/zap"
+
+	"go.autokitteh.dev/autokitteh/integrations/discord/internal/vars"
+)
+
+func (h *handler) handleMessage(_ *discordgo.Session, m any, eventType string) {
```
daabr: 1. No need to pass the session
2. `m` is short which is good, but when its type doesn't explain what it is, it's better to use a slightly longer name
```suggestion
func (h *handler) handleMessage(msg any, eventType string) {
```

integrations/discord/websockets/message.go
```diff
@@ -0,0 +1,35 @@
...
+	akEvent, err := transformEvent(h.logger, m, eventType)
```
daabr: Why not add `transformEvent` to `handler`, instead of passing `h.logger`?

integrations/discord/save.go
```diff
@@ -44,7 +46,31 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
 	c.Finalize(sdktypes.NewVars().
-		Set(botToken, r.Form.Get("botToken"), true).
-		Set(authType, integrations.Init, false))
+		Set(vars.BotID, bot.ID, false).
+		Set(vars.BotTokenName, bt, true).
+		Set(vars.AuthType, integrations.Init, false))
+}
+
+func infoWithToken(botToken string) (*discordgo.User, error) {
```
daabr, line 57: The function name `infoWithToken` doesn't imply that it's actually initializing a goroutine with a WebSocket.

integrations/discord/save.go
```diff
@@ -44,7 +46,31 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	bt := r.Form.Get("botToken")
+
+	// get the bot's ID
```
daabr: See the comment about the name of the function `infoWithToken`. Likewise, this comment isn't really explaining a lot, it only defines a visual code block (which is good) but also hides the side effect of the function call.

integrations/discord/save.go
```diff
@@ -44,7 +46,31 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	bot, err := infoWithToken(bt)
```
daabr: Question: how does this bot object (and its websocket) survive after this function ends?

integrations/discord/websockets/socketmode.go
daabr: I'm guessing to keep their websockets alive. My question (maybe that's what Itay is really asking): is that really necessary? In other words, if we just create the objects without saving them, are the websockets garbage-collected?

integrations/discord/websockets/socketmode.go
```diff
@@ -0,0 +1,122 @@
...
+
+	dg, err := discordgo.New("Bot " + botToken)
+	if err != nil {
+		h.logger.Error("Error creating Discord session", zap.Error(err))
+		return
+	}
+
+	addHandlers(dg, h)
+
+	// Open a WebSocket connection to Discord
```
daabr:
```suggestion
	// Open a WebSocket connection to Discord.
```

integrations/discord/websockets/socketmode.go
```diff
@@ -0,0 +1,122 @@
...
+	err = dg.Open()
+
+	if err == nil {
+		webSocketClients[botID] = dg
+		return // Normal process termination.
```
daabr: Please clarify more.

integrations/discord/websockets/socketmode.go
```diff
@@ -0,0 +1,122 @@
...
+			zap.Error(err),
+		)
+		return sdktypes.InvalidEvent, err
+	}
+
+	return akEvent, nil
+}
+
+func addHandlers(dg *discordgo.Session, h handler) {
+	// messages
```
daabr: For now (where there are only 3 event types, and all are messages, this comment is redundant

integrations/discord/internal/vars/vars.go
pashafateev: done

integrations/discord/save.go
pashafateev: See comment for the name of the function but I don't think there's any side effect. I removed the comment.

integrations/discord/save.go
pashafateev: ~~I'm not sure how to check that, but I don't think it does.~~

~~We likely don't need it to persist, since we only require the ID, which is static. The ID was added to mirror the approach used in Slack's WebSocket implementation.~~ If it turns out that maintaining a map of WebSocket connections isn't necessary, then we can eliminate the ID as well.

------------
**Update** after another investigating for another comment: I don't think there's a websocket involved here; i.e. there is nothing to keep alive.

integrations/discord/save.go
pashafateev: Done

integrations/discord/save.go
pashafateev: I don't think WebSocket is being used here.
[User()](https://github.com/bwmarrin/discordgo/blob/da9e191069d09e1b467145f5758d9b11cb9cca0d/restapi.go#L326) calls [this function](https://github.com/bwmarrin/discordgo/blob/da9e191069d09e1b467145f5758d9b11cb9cca0d/restapi.go#L171)
It's all HTTP
[Open()](https://github.com/bwmarrin/discordgo/blob/da9e191069d09e1b467145f5758d9b11cb9cca0d/restapi.go#L326) is what starts the websocket connection

integrations/discord/save.go
pashafateev: `User`

integrations/discord/save.go
pashafateev: Done

integrations/discord/websockets/message.go
pashafateev: Done

integrations/discord/websockets/message.go
pashafateev: Done

integrations/discord/websockets/socketmode.go
pashafateev: I checked. I don't think we do. I removed caching.

integrations/discord/websockets/socketmode.go
pashafateev: Done

integrations/discord/websockets/socketmode.go
pashafateev: Done

integrations/discord/websockets/socketmode.go
pashafateev: Done

integrations/discord/websockets/save.go
```diff
@@ -1,12 +1,15 @@
-package discord
+package websockets
```
daabr, line 1: Nit: given that Discord supports only WS events, unlike Slack that has 2 modes, I think it makes sense to move the files in this package to to `discord` and remove the `websockets` package

integrations/discord/websockets/websocket.go
```diff
@@ -0,0 +1,103 @@
...
+		)
+		if err != nil {
+			l.Error("Event dispatch failed", zap.Error(err))
+			return
+		}
+		l.Debug("Event dispatched")
+	}
+}
+
+// Transform the received Discord event into an AutoKitteh event.
```
daabr, line 65: https://go.dev/doc/comment#func

```suggestion
// transformEvent transforms the received Discord event into an AutoKitteh event.
```

integrations/discord/websockets/websocket.go
```diff
@@ -0,0 +1,103 @@
...
+func (h handler) transformEvent(discordEvent any, eventType string) (sdktypes.Event, error) {
+	l := h.logger.With(
+		zap.String("eventType", eventType),
+		zap.Any("event", discordEvent),
```
daabr, line 69: Arguably, this is a bit much to log - but we can always improve this later based on actual experience.

integrations/discord/websockets/websocket.go
```diff
@@ -0,0 +1,103 @@
...
+	integration   sdktypes.Integration
+	integrationID sdktypes.IntegrationID
+}
+
+func NewHandler(l *zap.Logger, v sdkservices.Vars, d sdkservices.Dispatcher, i sdktypes.Integration) handler {
+	return handler{
+		logger:        l,
+		vars:          v,
+		dispatcher:    d,
+		integration:   i,
```
daabr, line 28: See my other comment about moving code from `websockets` to `discord` - which would make the `integration` field unnecessary

integrations/discord/websockets/save.go
pashafateev: Done

integrations/discord/websockets/websocket.go
pashafateev: Done

integrations/discord/websockets/websocket.go
pashafateev: Done

PR #619, opened 2024-09-01
feat(ENG-1316): webhooks redesign

itayd:
see attached issue for further information.

- trigger has a source type now - connection or webhook. in the future: schedule.
- trigger also stores the webhook slug, which is generated when it is created.
- triggers are now valid destinations for events.
- next_event/subscribe can now work on triggers, not only connection names.
- kill http integration. replaced by session level http functions for egress and a webhooks service for ingress.
- webhook service sends events through dispatcher with a trigger destination id.
- scheduler almost entirely rewritten - it becomes very simple with new design. 
- killed `internal/backend/workflows` which contained common code that is no longer common - everything goes through dispatcher.


migrations/postgres/20240901074155_trigger-dests.sql
```diff
@@ -0,0 +1,41 @@
+-- +goose Up
+-- modify "triggers" table
+ALTER TABLE "triggers" ALTER COLUMN "connection_id" DROP NOT NULL, ADD COLUMN "source_type" text NULL, ADD COLUMN "webhook_slug" text NULL;
+-- create index "idx_triggers_source_type" to table: "triggers"
+CREATE INDEX "idx_triggers_source_type" ON "triggers" ("source_type");
+-- create index "idx_triggers_webhook_slug" to table: "triggers"
+CREATE UNIQUE INDEX "idx_triggers_webhook_slug" ON "triggers" ("webhook_slug");
+-- modify "events" table
+ALTER TABLE "events" ADD COLUMN "destination_id" uuid NOT NULL, ADD COLUMN "trigger_id" uuid NULL, ADD
```
efiShtain: this won't pass on a non empty database
you create a column with a not null constraint but all current rows will have a null value

migrations/postgres/20240901074155_trigger-dests.sql
```diff
@@ -0,0 +1,41 @@
...
+ CONSTRAINT "fk_events_trigger" FOREIGN KEY ("trigger_id") REFERENCES "triggers" ("trigger_id") ON UPDATE NO ACTION ON DELETE NO ACTION;
+-- create index "idx_events_destination_id" to table: "events"
+CREATE INDEX "idx_events_destination_id" ON "events" ("destination_id");
+-- create index "idx_events_trigger_id" to table: "events"
+CREATE INDEX "idx_events_trigger_id" ON "events" ("trigger_id");
+-- drop index "idx_connection_id_event_type" from table: "signals"
+DROP INDEX "idx_connection_id_event_type";
+-- modify "signals" table
+ALTER TABLE "signals" ALTER COLUMN "signal_id" TYPE uuid, ALTER COLUMN "connection_id" DROP NOT NULL, ADD COLUMN "destination_id" uuid NOT NULL, ADD COLUMN "trigger_id" uuid NULL, ADD
```
efiShtain: this won't pass on a non empty database
you create a column with a not null constraint but all current rows will have a null value

internal/backend/db/db.go
```diff
@@ -94,9 +95,10 @@ type DB interface {
 	// -----------------------------------------------------------------------
 	CreateTrigger(context.Context, sdktypes.Trigger) error
 	UpdateTrigger(context.Context, sdktypes.Trigger) error
-	GetTrigger(context.Context, sdktypes.TriggerID) (sdktypes.Trigger, error)
+	GetTriggerByID(context.Context, sdktypes.TriggerID) (_ sdktypes.Trigger, _ error)
```
efiShtain: why?
signature is well formed getTrigger(id)

internal/backend/db/dbgorm/events.go
```diff
@@ -76,15 +76,17 @@ func (db *gormdb) SaveEvent(ctx context.Context, event sdktypes.Event) error {
 		return err
 	}
 
-	connectionID := event.ConnectionID() // could be invalid/zero
+	connectionID := event.DestinationID().ToConnectionID()
```
efiShtain, line 79: can this fail?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -162,8 +163,10 @@ type Secret struct {
 
 type Event struct {
 	EventID       sdktypes.UUID  `gorm:"uniqueIndex;type:uuid;not null"`
+	DestinationID sdktypes.UUID  `gorm:"index;type:uuid;not null"`
```
efiShtain, line 166: I'm not sure I like that we have destinationID connectionID and triggerID

nothing currently enforce an invalid state of having both trigger and connection

we can add a constraint on postgres to validate it
```sql
ALTER TABLE events ADD CONSTRAINT event_connection_or_trigger CHECK (
(connection_id IS NULL AND trigger_id IS NOT NULL) 
OR
(connection_id IS NOT NULL AND trigger_id IS NULL)
)
```
another alternvative
```sql
ALTER TABLE events ADD CONSTRAINT event_connection_or_trigger CHECK ( NUM_NONNULLS(trigger_id, connection_id) = 1)
```

and anyhow, each event has a trigger directly or through the connection

why do we need the destinationID at all ?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -478,14 +502,17 @@ func ParseDeploymentWithSessionStats(d DeploymentWithStats) (sdktypes.Deployment
...
+	ConnectionID  *sdktypes.UUID `gorm:"type:uuid"`
+	TriggerID     *sdktypes.UUID `gorm:"type:uuid"`
+	CreatedAt     time.Time
+	WorkflowID    string
+	Filter        string
 
 	// enforce foreign key
 	Connection *Connection
+	Trigger    *Trigger
```
efiShtain, line 509: same questions for the events table

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -375,25 +372,30 @@ func (w *sessionWorkflow) waitOnFirstSignal(wctx workflow.Context, signals []str
...
-	iid := sdktypes.NewIDFromUUID[sdktypes.IntegrationID](signal.Connection.IntegrationID)
-	cid := sdktypes.NewIDFromUUID[sdktypes.ConnectionID](&signal.Connection.ConnectionID)
+	var did sdktypes.EventDestinationID
+	if signal.ConnectionID != nil {
+		did = sdktypes.NewEventDestinationID(sdktypes.NewIDFromUUID[sdktypes.ConnectionID](signal.ConnectionID))
+	} else if signal.TriggerID != nil {
+		did = sdktypes.NewEventDestinationID(sdktypes.NewIDFromUUID[sdktypes.TriggerID](signal.TriggerID))
+	} else {
+		return nil, fmt.Errorf("invalid signal %v: no connection or trigger ids, got %v", signalID, signal.DestinationID)
+	}
```
efiShtain, line 390: can't we just list by connection_id or trigger_id and get the first ? instead of using a destination_id ?

internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -167,21 +168,38 @@ func (w *sessionWorkflow) start(ctx context.Context, args []sdktypes.Value, kwar
...
-	signalID, err := w.createEventSubscription(ctx, connectionName, filter)
+	_, connection := kittehs.FindFirst(w.data.Connections, func(c sdktypes.Connection) bool {
+		return c.Name().String() == name
+	})
+
+	_, trigger := kittehs.FindFirst(w.data.Triggers, func(t sdktypes.Trigger) bool {
+		return t.Name().String() == name
+	})
+
+	if connection.IsValid() && trigger.IsValid() {
```
efiShtain, line 185: you miss the case where both are invalid

internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -238,7 +261,7 @@ func (w *sessionWorkflow) nextEvent(ctx context.Context, args []sdktypes.Value,
 			return sdktypes.InvalidValue, err
 		}
 
-		if signalID == "" {
+		if signalID == (uuid.UUID{}) {
```
efiShtain:
```suggestion
		if signalID == uuid.Nil {
```

internal/backend/db/db.go
itayd: the `_` is idiotic, remnant from a change i was considering.

it's to differentiate from `GetTriggerBySlug`.

internal/backend/db/dbgorm/events.go
itayd: no. it'll return an invalid value if it's not a connection.

internal/backend/db/dbgorm/scheme/scheme.go
itayd: the idea is that find works always on destinationid. i put the other ones to enforce foreign keys, not any other reason.

internal/backend/sessions/sessionworkflows/syscalls.go
itayd: fixed

internal/backend/sessions/sessionworkflows/syscalls.go
itayd: fixed

internal/backend/sessions/sessionworkflows/workflow.go
itayd: that'll will make twice the calls. this is just translating from a stored data to the actual type.

cmd/ak/cmd/triggers/create.go
```diff
@@ -6,20 +6,21 @@ import (
...
-	data                                         map[string]string
+	webhook                                      bool
 )
 
 var createCmd = common.StandardCommand(&cobra.Command{
-	Use: `create -n name --call file:func [-p project] [-e env] -c connection [-E event] [-f filter] [--data key=value]...
-             create -n name --call file:func [-p project] [-e env] -s "schedule"`,
+	Use: `create -n name --call file:func [-p project] [-e env] -c connection [-E event] [-f filter]
+             create -n name --call file:func [-p project] [-e env] -s "schedule"
+  			 create -n name --call file:func [-p project] [-e env] --webhook
```
daabr, line 22: There's nothing inherently wrong here, except that having 1 command that can create 3 different things depending on mutually-exclusive flags is not easy to comprehend unless you're already an expert in using our system.

In other words, our users will 100% struggle before they're able to use triggers effectively, or more likely give up before that.

At this points, it's better to create separate commands for these modes - it won't solve the underlying problem, but at least it would hide it better.

cmd/ak/cmd/triggers/create.go
```diff
@@ -52,41 +53,33 @@ var createCmd = common.StandardCommand(&cobra.Command{
 			eid = e.ID()
 		}
 
-		// Mode 1: connection is required, event and/or filter
-		// and/or data are required, and schedule is not allowed.
```
daabr, line 56: I agree that the code you wrote instead of this is more generic, so it's better, but it also hides the specific needs of each mode: it's not clear what are the required and optional flags for each mode. WDYT?

cmd/ak/cmd/triggers/create.go
```diff
@@ -52,41 +53,33 @@ var createCmd = common.StandardCommand(&cobra.Command{
...
+			}
+
+			t = t.WithConnectionID(cid)
+		} else if schedule != "" {
+			t = t.WithSchedule(schedule)
+		} else if webhook {
+			t = t.WithWebhook()
+		} else {
+			return fmt.Errorf("missing connection, schedule or webhook")
+		}
```
daabr: These flags are mutually exclusive. Use separate `if` blocks instead of a single block with `else` - the code will look cleaner.

cmd/ak/cmd/triggers/create.go
```diff
@@ -110,15 +103,15 @@ func init() {
 	createCmd.MarkFlagsOneRequired("project", "env")
 
 	createCmd.Flags().VarP(common.NewNonEmptyString("", &connection), "connection", "c", "connection name or ID")
+	createCmd.Flags().BoolVarP(&webhook, "webhook", "w", false, "trigger is a webhook")
```
daabr:
```suggestion
	createCmd.Flags().BoolVarP(&webhook, "webhook", "w", false, "trigger uses a webhook")
```

examples/testprograms/http/autokitteh.yaml
```diff
@@ -2,32 +2,10 @@ version: v1
...
-      filter: "data.method == 'POST'"
-      connection: myhttp
-      entrypoint: http.star:on_http_post
-    - name: test
-      connection: myhttp
-      call: http.star:on_http_test
-      data:
-        path: "/test/{a}/{b...}"
+    - name: http_req
+      webhook: {}
```
daabr: What does `{}` mean?

daabr (commented):
A few initial comments. Nothing big, but right now I reviewed only 70/111 files.

An important point for me: "show, don't tell". Is there a working Python project I can look at / run to understand the UX aspects of this PR?

examples/testprograms/http/autokitteh.yaml
itayd: i will add type

cmd/ak/cmd/triggers/create.go
itayd: we'll consider it in the future. for now i think it suffices.

cmd/ak/cmd/triggers/create.go
itayd: can you suggest a comment?

cmd/ak/cmd/triggers/create.go
itayd: i think i disagree with you here. i think this is more self documenting.

cmd/ak/cmd/triggers/create.go
itayd: done

examples/testprograms/http/autokitteh.yaml
itayd: done

cmd/ak/cmd/triggers/create.go
itayd: i changed my mind. fixed.

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: @itayd, could u plz elaborate on what is destination_id?

internal/backend/triggers/triggers.go
```diff
@@ -3,106 +3,114 @@ package triggers
...
-		err = m.scheduler.Create(ctx, triggerID.String(), schedule, triggerID)
+	if next.Equal(curr) {
+		// no update
+		return nil
 	}
 
-	if err != nil {
-		return err
+	if curr.SourceType() != next.SourceType() {
+		return sdkerrors.NewInvalidArgumentError("cannot update source type")
```
ig-ra, line 91: just FYI - before your change user was able to update trigger type, which is IMO much more convenient when redefining trigger in the manifest. Otherwise user will be forced to rename it.
I would suggest somehow to support it - either here in update, or in maifest by deleting curr/adding new next if source type is changing.

itayd:
@efiShtain this is now ready for your SQL migration magic

proto/autokitteh/triggers/v1/trigger.proto
```diff
@@ -3,16 +3,26 @@ syntax = "proto3";
...
-  program.v1.CodeLocation code_location = 6 [(buf.validate.field).required = true];
+  program.v1.CodeLocation code_location = 6;
   string filter = 7;
-  map<string, values.v1.Value> data = 8;
+
+  string connection_id = 50; // if source_type == CONNECTION.
+  string schedule = 51; // if source_type == SCHEDULE.
+
+  // read only.
+  string webhook_slug = 100; // if source_type == WEBHOOK, after creation.
```
efiShtain, line 27: this can probably just be source since all of them are just string
otherwise it probably should be oneOf
wdyt ?

efiShtain: and maybe it should be part of the source type message since you defined one anyway

internal/resolver/resolver.go
```diff
@@ -160,6 +188,31 @@ func (r Resolver) connectionByFullName(ctx context.Context, projNameOrID, connNa
...
+		return sdktypes.InvalidTrigger, sdktypes.InvalidTriggerID, fmt.Errorf("list triggers: %w", err)
+	}
+
+	for _, c := range cs {
+		if c.Name().String() == triggerName {
+			return c, c.ID(), nil
+		}
+	}
+
+	return sdktypes.InvalidTrigger, sdktypes.InvalidTriggerID, NotFoundError{Type: "trigger", Name: fullName}
```
efiShtain, line 213: all of this should be something like
```go
t, err := db.getTriggerByProjectAndName(ctx, projectName, triggerName)
if err != nil {return nil, err}
return t, nil
```

no need to test for project validity or list potentially infinite many triggers just to find one specific

internal/backend/triggers/triggers.go
```diff
@@ -3,106 +3,114 @@ package triggers
...
 		return sdktypes.InvalidTriggerID, err
 	}
 
 	return trigger.ID(), nil
 }
 
-func (m *triggers) Update(ctx context.Context, trigger sdktypes.Trigger) error {
-	triggerID := trigger.ID()
-	prevTrigger, err := m.db.GetTrigger(ctx, triggerID)
+func (m *triggers) Update(ctx context.Context, next sdktypes.Trigger) error {
```
efiShtain: what is next ?

internal/backend/triggers/triggers.go
```diff
@@ -3,106 +3,114 @@ package triggers
...
+	next = next.WithWebhookSlug("") // ignore webhook slug.
+
+	if next.Equal(sdktypes.InvalidTrigger) {
```
efiShtain: not sure i understand, don't we have an isValid convention ?

internal/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,120 @@
...
+}
+
+func New(l *zap.Logger, tc temporalclient.Client) *Scheduler {
+	return &Scheduler{sl: l.Sugar(), temporal: tc}
+}
+
+func (sch *Scheduler) Start(ctx context.Context, dispatcher sdkservices.Dispatcher) error {
+	sch.dispatcher = dispatcher
+
+	w := worker.New(sch.temporal.Temporal(), taskQueueName, worker.Options{Identity: workerID})
```
efiShtain, line 36: didn't we say something about one worker that listen on all queues ?

internal/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,120 @@
...
+		return fmt.Errorf("schedule wf: worker start: %w", err)
+	}
+
+	return nil
+}
+
+func (sch *Scheduler) Create(ctx context.Context, tid sdktypes.TriggerID, schedule string) error {
+	l := sch.sl.With("trigger_id", tid.String())
+
+	_, err := sch.temporal.Temporal().ScheduleClient().Create(
```
efiShtain, line 49: do we have issue with users in this case ? 
same as for delete

internal/backend/scheduler/scheduler.go
```diff
@@ -0,0 +1,120 @@
...
+		ctx,
+		client.ScheduleOptions{
+			ID: tid.String(),
```
efiShtain, line 52: duplicates are handled by temporal ?
in case you register it twice

efiShtain (commented):
some questions about users / temporal in schedule

all in all, this is a huge PR, even though we checked in on going, still hard to go through
next time it should be split to stacked PRs

proto/autokitteh/triggers/v1/trigger.proto
daabr: `SourceType` is just an enum, not a message, so the value can't be added there. As for `oneof` - I personally prefer to avoid it, it complicates generated code and limits future extensibility. But in that case it does make sense to add validation checks:
1. If source_type = CONNECTION then `connection_id` isn't empty, but `schedule` and `webhook_slug` are
2. IF source_type = WEBHOOK, ...
3. IF source_type = SCHEDULE, ...

manifest.schema.yaml
```diff
@@ -98,16 +81,26 @@
         "filter": {
           "type": "string"
         },
+        "type": {
+          "type": "string",
+          "enum": [
+            "schedule",
+            "webhook",
+            "connection"
```
daabr, line 89: Question: what's the default if I don't specify anything? Do I need to specify the type for connections? If I specify a schedule string without the schedule type is that an error?

manifest.schema.yaml
```diff
@@ -98,16 +81,26 @@
...
           "type": "string"
         },
-        "data": {
+        "webhook": {
+          "properties": {},
+          "additionalProperties": false,
           "type": "object"
         },
-        "call": {
+        "connection": {
```
daabr, line 100: Nit: please move `event_type` and `filter` between `type` and `call` - I think the order doesn't matter in a schema, but it's still a good idea to match the logical flow in the same way users will specify it.

docs/autokitteh.yaml
```diff
@@ -52,22 +46,33 @@ project:
...
-        # Each parameter is enclosed in {}. Last parameter
-        # can be suffixed with "..."", which will cause it to
-        # capture the rest of the path.
-        # Last parameter can also be "{$}", which specifies
-        # it expects an exact match.
-        path: "/test/{a}/{b...}"
+      # This is applicable for all trigger types.
+      call: main.star:on_http_get
+      # This indicates that the trigger is a webhook trigger.
+      webhook: {}
```
daabr, line 64: Not `type: webhook`?

cmd/ak/cmd/manifest/deploy.go
```diff
@@ -146,26 +147,38 @@ func PrintSuggestions(cmd *cobra.Command, effects manifest.Effects) {
...
+				fmt.Fprintf(cmd.ErrOrStderr(), "error getting trigger %v: %v\n", tid, err)
+				continue
+			}
 
-		c, err := common.Client().Connections().Get(ctx, cid)
-		if err != nil {
-			fmt.Fprintf(cmd.ErrOrStderr(), "error getting connection %v: %v\n", cid, err)
-			continue
+			if slug := t.WebhookSlug(); slug != "" {
+				fmt.Fprintf(cmd.ErrOrStderr(), "[!!!!] Trigger created. Webhook path is \"%s%s\"\n", webhookssvc.WebhooksPathPrefix, slug)
```
daabr: I prefer this suggestion over adding a missing "." at the end of the line, which is inconsistent with the rest of the output of this command.

```suggestion
				fmt.Fprintf(cmd.ErrOrStderr(), "[!!!!] Trigger created, webhook path is \"%s%s\"\n", webhookssvc.WebhooksPathPrefix, slug)
```

internal/backend/dispatcher/dispatcher_test.go
```diff
@@ -1,179 +0,0 @@
-package dispatcher
```
daabr, line 1: I won't block this PR on lack of unit tests, but this is not cool. Be warned that this is probably the last time that missing/deleted unit tests are accepted.

cmd/ak/cmd/triggers/create.go
```diff
@@ -52,41 +53,32 @@ var createCmd = common.StandardCommand(&cobra.Command{
...
+		if connection != "" {
+			_, cid, err := r.ConnectionNameOrID(ctx, connection, project)
+			if err != nil {
+				return common.ToExitCodeErrorNotNilErr(err, "connection")
+			}
+
+			t = t.WithConnectionID(cid)
+		} else if schedule != "" {
+			t = t.WithSchedule(schedule)
+		} else if webhook {
```
daabr:
```suggestion
		} else if schedule != "" {
			// TODO(ENG-1004): Verify validity of schedule expression.
			t = t.WithSchedule(schedule)
		} else if webhook {
```

internal/backend/projects/projects.go
```diff
@@ -43,13 +41,15 @@ func (ps *Projects) Create(ctx context.Context, project sdktypes.Project) (sdkty
 		return sdktypes.InvalidProjectID, err
 	}
 
-	// ensure there is a default cron connection (for all projects)
-	if cronConnection, err := ps.DB.GetConnection(ctx, sdktypes.BuiltinSchedulerConnectionID); errors.Is(err, sdkerrors.ErrNotFound) {
-		cronConnection = cronConnection.WithID(sdktypes.BuiltinSchedulerConnectionID).WithName(sdktypes.NewSymbol(fixtures.SchedulerConnectionName))
-		if err = ps.DB.CreateConnection(ctx, cronConnection); err != nil && !errors.Is(err, sdkerrors.ErrAlreadyExists) { // just sanity
-			return sdktypes.InvalidProjectID, err
+	/*
```
daabr: Why not delete this block?

cmd/ak/cmd/triggers/create.go
daabr: I don't see it this change here?

internal/backend/scheduler/scheduler.go
daabr: I don't know what you're referring to, but a Temporal worker that listen to multiple Temporal queues is usually an antipattern.

internal/manifest/exec.go
```diff
@@ -203,12 +203,12 @@ func executeAction(ctx context.Context, action actions.Action, execContext *exec
 	case actions.UpdateTriggerAction:
 		trigger := action.Trigger
 
-		if action.ConnectionKey != "" { // convert scheduler -> normal trigger, or just update connection
-			cid, err := execContext.resolveConnectionID(ctx, action.ConnectionKey)
+		if key := action.ConnectionKey; key != nil { // convert scheduler -> normal trigger, or just update connection
```
daabr:
```suggestion
		// convert scheduler -> normal trigger, or just update connection
		if key := action.ConnectionKey; key != nil {
```

tests/system/testdata/deploy/project_deploy.txtar
```diff
@@ -32,7 +32,7 @@ return code == 0
 output contains 'build_id: bld_00000000000000000000000003'
 output contains 'deployment_id: dep_00000000000000000000000004'
 
-# Deploy project from a directory tree with multiple files.
+# Deploy project from a directory tree wit multiple files.
```
daabr: typo

tests/system/testdata/events/list.txtar
```diff
@@ -1,38 +1,38 @@
-# Negative tests: empty events list, with/out --fail flag.
-ak event list --integration http
+ak deploy --manifest project.yaml
```
daabr, line 8: Please back the negative tests: empty events list, with/out --fail flag (using the test integration maybe?)

tests/system/testdata/manifest/re-apply.txtar
```diff
@@ -1,5 +0,0 @@
-# TODO: The first apply defines a new project.
```
daabr, line 1: This TODO file doesn't do anything, but it's a useful reminder, and more visible than a ticket in in a graveyard backlog.

proto/autokitteh/triggers/v1/trigger.proto
itayd: i'd rather have these separated. if we add anything in the future that requires a different type. also, webhook_slug is generated, schedule is not.

internal/backend/triggers/triggers.go
itayd: the new trigger data?

internal/backend/triggers/triggers.go
itayd: i have no idea why i did it that way. fixing.

internal/backend/scheduler/scheduler.go
itayd: maybe in the future. not sure if it's a good idea.

internal/backend/triggers/triggers.go
```diff
@@ -3,106 +3,114 @@ package triggers
...
-		if schedule, _ := trigger.Data()[fixtures.ScheduleExpression].ToString(); schedule != "" {
-			if err := m.scheduler.Create(ctx, trigger.ID().String(), schedule, trigger.ID()); err != nil {
-				return sdktypes.InvalidTriggerID, err
-			}
+	sl := m.sl.With("trigger_id", trigger.ID())
+
+	switch trigger.SourceType() {
+	case sdktypes.TriggerSourceTypeWebhook:
+		trigger = webhookssvc.InitTrigger(trigger)
+		sl.With("slug", trigger.WebhookSlug()).Infof("creating webhook trigger with slug %q", trigger.WebhookSlug())
```
daabr: I think it's better to use a standard message without `%q`, and add the value as a log property. We won't need the value in the log message for any debugging.

And in that case, we need a regular logger, not a sugared logger (same comment for schedule and connection below).

internal/backend/scheduler/scheduler.go
itayd: shouldn't be registered twice. every trigger id is unique.

internal/backend/scheduler/scheduler.go
itayd: and update! yeah, an attacker might do this and the error would come only after the schedule is created. looking into it.

internal/backend/triggers/triggers.go
daabr: The fact that this is not clear means you need to document it, or find a better self-documenting name.

docs/autokitteh.yaml
itayd: this works as well.

internal/backend/dispatcher/dispatcher_test.go
itayd: these are all commented out anyway. i know we need to do these.

proto/autokitteh/triggers/v1/trigger.proto
daabr: @itayd I agree with you, but see my note about validation checks.

cmd/ak/cmd/manifest/deploy.go
daabr: Also need to add a trailing "/" after the slug to avoid Go's automatic redirect for ".../foo" to ".../foo/" - because curl doesn't follow redirects automatically.

internal/backend/triggers/triggers.go
itayd: heh, i changed it because i thought it is more clear. oh well, fixed.

internal/backend/scheduler/scheduler.go
itayd: changed the order of temporal invocation. will happen after db, as the db guards that for now. not that i am happy about it.

manifest.schema.yaml
itayd: it will error out.

manifest.schema.yaml
itayd: disagree - in the code it makes sense to put the type next to the specs.

cmd/ak/cmd/manifest/deploy.go
itayd: done

cmd/ak/cmd/triggers/create.go
itayd: done

internal/backend/projects/projects.go
itayd: oopsie, removed.

internal/manifest/exec.go
itayd: done

tests/system/testdata/deploy/project_deploy.txtar
itayd: fixed

tests/system/testdata/manifest/re-apply.txtar
itayd: ¯\\\_(ツ)\_/¯, done

tests/system/testdata/events/list.txtar
itayd: done

internal/backend/scheduler/scheduler.go
efiShtain: @daabr I didn't see any reference to it an anti pattern
why do you think so ?

plus, 
recommendations I saw is using just one task queue for everything and register everything on the same queue
unless needed to split something between binaries

sdk/sdktypes/connection.go
```diff
@@ -28,11 +28,7 @@ func (ConnectionTraits) Validate(m *ConnectionPB) error {
 }
 
 func (ConnectionTraits) StrictValidate(m *ConnectionPB) error {
-	var errs []error = []error{mandatory("name", m.Name)}
-	if m.ConnectionId != BuiltinSchedulerConnectionID.String() {
-		errs = append(errs, mandatory("project_id", m.ProjectId), mandatory("integration_id", m.IntegrationId))
-	}
-	return errors.Join(errs...)
+	return errors.Join(mandatory("name", m.Name))
```
ig-ra: for non-scheduler connection project_id and integration_id were mandatory. Why did u remove them?

internal/backend/scheduler/scheduler.go
daabr: @efiShtain exactly. **Roughly** you can say that there should be a 1:1 relation between workers and TQs. Usually a Temporal worker is a separate binary, not bundled with other functionalities, and not mixed with other workers in the same binary. And that's why having a single worker responding to multiple TQs is either too many unnecessary TQs, or messy and risky coupling between workers.

manifest.schema.yaml
daabr: is there a test for these cases?

docs/autokitteh.yaml
daabr: Then it's **too** flexible. I'm intimately familiar with AK, and I'm confused by different-but-equivalent configuration options. Imagine how a new user would feel.

itayd: for now we'll keep it - in the future we might use a dedicated webhook object where we'll want to reference from this.

daabr (approved):

internal/backend/triggers/triggers.go
itayd: i don't think changing trigger type is something we need to support. worst case a trigger can be deleted and added.

proto/autokitteh/triggers/v1/trigger.proto
itayd: done

sdk/sdktypes/connection.go
itayd: connection_id cannot be mandatory as a connection is given to the service with no id.
integration_id need to be mandatory though - will re-add it.

itayd: ah project_id, right. yes - readding both.

PR #624, opened 2024-09-03
fix: allow initial run in session to take a while

itayd:
need to finess this, but this is so @efiShtain can try the approach out.
note that calling activities during initial run might not work all the time, depends on timing with the `workflow.Sleep`. We'll work it out if needed.

### DIFF
--- a/internal/backend/sessions/sessionworkflows/workflow.go
+++ b/internal/backend/sessions/sessionworkflows/workflow.go
@@ -524,20 +524,53 @@ func (w *sessionWorkflow) run(wctx workflow.Context) (prints []string, err error
 
 	entryPoint := w.data.Session.EntryPoint()
 
-	run, err := sdkruntimes.Run(
-		ctx,
-		sdkruntimes.RunParams{
-			Runtimes:             w.ws.svcs.Runtimes,
-			BuildFile:            w.data.BuildFile,
-			Globals:              w.globals,
-			RunID:                runID,
-			FallthroughCallbacks: cbs,
-			EntryPointPath:       entryPoint.Path(),
-		},
-	)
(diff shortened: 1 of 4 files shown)


internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -524,20 +524,53 @@ func (w *sessionWorkflow) run(wctx workflow.Context) (prints []string, err error
...
-		return prints, err
+	type runOrErr struct {
+		run sdkservices.Run
+		err error
+	}
+
+	runDone := make(chan runOrErr)
+
+	go func() {
+		time.Sleep(20 * time.Second)
```
efiShtain: please remove this 20 seconds sleep before merging

efiShtain (approved):
Just don't forget to remove the 20 seconds sleep

itayd:
> Just don't forget to remove the 20 seconds sleep

done

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -524,20 +531,49 @@ func (w *sessionWorkflow) run(wctx workflow.Context) (prints []string, err error
...
-	)
-	if err != nil {
-		// TODO(ENG-130): discriminate between infra and real errors.
-		return prints, err
+	type runOrErr struct {
+		run sdkservices.Run
+		err error
+	}
+
+	runDone := make(chan runOrErr)
```
tebeka, line 539: Should make this buffered to avoid goroutine leak

itayd: https://github.com/autokitteh/autokitteh/pull/629

PR #633, opened 2024-09-09
chore: upgrade to Go 1.23 + upgrade dep versions

daabr:
1. Verified again that previous concerns re. HTTP webhooks are no longer observed.

2. Also removed the `shortuuid` dependency - we're already using too many UUID libraries. Use `typeid` instead (but make sure AK generates the base UUID - for testing purposes).

3. While doing (2):
   - Creating TypeIDs can return errors due to invalid input, but if that's not a risk (which is true in our code) then it's OK to use `typeid.Must()` (instead of `kittehs.Must1()`) - no functional impact, but it's more "idiomatic"
   - Fixed a few (new but unrelated) lint warnings

Current binary size is 162 MiB - we need to get rid of unnecessary deps, but that's a task for another time.


internal/backend/webhookssvc/svc.go
```diff
@@ -39,7 +39,7 @@ func (s *Service) Start(muxes *muxes.Muxes) {
 }
 
 func InitTrigger(trigger sdktypes.Trigger) sdktypes.Trigger {
-	return trigger.WithWebhookSlug(shortuuid.DefaultEncoder.Encode(sdktypes.NewUUID()))
+	return trigger.WithWebhookSlug(typeid.Must(typeid.WithPrefix("")).String())
```
itayd: nope - use sdktypes.NewUUID here because it supports the idgen capability (preknown ids)

ig-ra (approved):

internal/backend/webhookssvc/svc.go
daabr: done, ptal

itayd (approved):

PR #634, opened 2024-09-09
fix(db): cant delete used trigger

ig-ra:
soft-delete triggers

### DIFF
--- a/internal/backend/db/dbgorm/scheme/scheme.go
+++ b/internal/backend/db/dbgorm/scheme/scheme.go
@@ -259,28 +259,30 @@ func ParseEnv(e Env) (sdktypes.Env, error) {
 }
 
 type Trigger struct {
-	TriggerID sdktypes.UUID `gorm:"primaryKey;type:uuid;not null"`
-
+	TriggerID    sdktypes.UUID  `gorm:"primaryKey;type:uuid;not null"`
 	ProjectID    sdktypes.UUID  `gorm:"index;type:uuid;not null"`
-	SourceType   string         `gorm:"index"`
 	ConnectionID *sdktypes.UUID `gorm:"index;type:uuid"`
 	EnvID        sdktypes.UUID  `gorm:"index;type:uuid;not null"`
-	Name         string
+
+	SourceType   string `gorm:"index"`
 	EventType    string
 	Filter       string
 	CodeLocation string
 
-	// enforce foreign keys
-	Project    *Project
-	Env        *Env
-	Connection *Connection
-	Ownership  *Ownership `gorm:"polymorphic:Entity;"`
-
+	Name string
 	// Makes sure name is unique - this is the env_id with name.
-	UniqueName string `gorm:"uniqueIndex;not null"`
+	UniqueName string `gorm:"uniqueIndex;not null"` // env_id + name
 
 	WebhookSlug string `gorm:"index"`
 	Schedule    string
+
+	DeletedAt gorm.DeletedAt `gorm:"index"`
+
+	// enforce foreign keys
+	Project    *Project
+	Connection *Connection
+	Env        *Env
+	Ownership  *Ownership `gorm:"polymorphic:Entity;"`
 }
 
 func ParseTrigger(e Trigger) (sdktypes.Trigger, error) {
--- a/internal/backend/db/dbgorm/triggers.go
+++ b/internal/backend/db/dbgorm/triggers.go
@@ -31,9 +31,9 @@ func (gdb *gormdb) deleteTrigger(ctx context.Context, triggerID sdktypes.UUID) e
 		}
 
 		// enforce foreign keys constrains while soft-deleting
-		var count int64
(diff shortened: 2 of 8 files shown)


internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -259,28 +259,30 @@ func ParseEnv(e Env) (sdktypes.Env, error) {
...
-	// enforce foreign keys
-	Project    *Project
-	Env        *Env
-	Connection *Connection
-	Ownership  *Ownership `gorm:"polymorphic:Entity;"`
-
+	Name string
 	// Makes sure name is unique - this is the env_id with name.
-	UniqueName string `gorm:"uniqueIndex;not null"`
+	UniqueName string `gorm:"uniqueIndex;not null"` // env_id + name
```
efiShtain, line 274: was this here before?
we have envID and name, what's the purpose of this unique name ? why not index both of them as unique ?

internal/backend/db/dbgorm/triggers.go
```diff
@@ -29,6 +29,14 @@ func (gdb *gormdb) deleteTrigger(ctx context.Context, triggerID sdktypes.UUID) e
 		if err := tx.isCtxUserEntity(tx.ctx, triggerID); err != nil {
 			return err
 		}
+
+		// enforce foreign keys constrains while soft-deleting
+		var count int64
+		tx.db.Model(&scheme.Event{}).Where("deleted_at is NULL and trigger_id = ?", triggerID).Count(&count)
```
efiShtain: not sure I get what you're doing here
looking for non deleted events ? 
is this a mix and match of soft delete and hard delete ?

efiShtain: I also think you should add limit 1 to this count since you only care if something exists
otherwise you're going to count all relevant events for not real reason

internal/backend/db/dbgorm/scheme/scheme.go
ig-ra: I added only deleted_at, all other were just slightly rearranged.

internal/backend/db/dbgorm/triggers.go
ig-ra: 1. once you soft deleting you cannot enforce foreign keys in db, only in code. We are doing the same in env. Just to prevent deleting trigger before events reference it were deleted
2. regarding count and limit you are totally right

internal/backend/db/dbgorm/scheme/scheme.go
itayd: @efiShtain that's a thing i did (foolishly?) long ago. I'm ok with killing it if you want to.

internal/backend/db/dbgorm/triggers.go
```diff
@@ -29,6 +29,14 @@ func (gdb *gormdb) deleteTrigger(ctx context.Context, triggerID sdktypes.UUID) e
...
+		if count > 0 {
+			return fmt.Errorf("FOREIGN KEY: %w", gorm.ErrForeignKeyViolated)
```
itayd, line 37: do we translate somewhere this error to higher level error? using the `translateError` function.

internal/backend/db/dbgorm/scheme/scheme.go
daabr: Adding my 2 cents: I didn't have an opinion about trigger names when @itayd added them, because they weren't seen anywhere. But now they **are** visible (and important) when creating webhook triggers, so I prefer to keep them as-is. Auto-generating a meaningful name based on a trigger's details would be complicated and sensitive (for example, you can have 2 triggers with the same settings except a different complicated `filter` field).

Edit: I was referring only to the trigger name. If `uniqueName` is not affecting `ak deploy` output then I don't care about it.

internal/backend/db/dbgorm/triggers.go
efiShtain: ok, I understand now the delete part, so my first question is irrelevant

count should be fixed to something else

internal/backend/db/dbgorm/scheme/scheme.go
efiShtain: if it was here before, irrelevant to this PR
but I prefer to clean it up in a different ticket

https://linear.app/autokitteh/issue/ENG-1534/trigger-uniquename-cleanup-in-db

resolving this for now

internal/backend/db/dbgorm/triggers.go
ig-ra: I replaced count by select(1) limit(1)

internal/backend/db/dbgorm/triggers.go
ig-ra: yes, this is done always in API level functions. see DeleteTrigger.
functions working with schema return gorm errors.
API functions return SDK errors by translating gorm ones

efiShtain (approved):
LGTM

PR #637, opened 2024-09-10
refacor: remove embedded ngrok

daabr:
Using this configuration option is undocumented and not beneficial compared to using ngrok in the usual way. Removing this dependency shaves off 1% of AK binary size - not huge, but it's a low hanging fruit.


efiShtain (approved):
Amazing :)

tebeka:
The big win is less code, not binary size.

PR #640, opened 2024-09-10
fix: move config from backend to internal + support list env variables

efiShtain:
- Move config from backend to internal, this is something we wanted to do before, this is a preparation for the CLI to use the same config capabilities
- Support passing env variable that parsed as list of strings
```bash
MYVAR: a,b,c 
```

```go
type Cfg struct {
  ListData []string `koanf:"myvar"`
}
```

### DIFF
--- a/backend/svc/configset.go
+++ b/backend/svc/configset.go
@@ -2,7 +2,7 @@ package svc
 
 import (
 	"go.autokitteh.dev/autokitteh/internal/backend/configset"
-	"go.autokitteh.dev/autokitteh/internal/backend/svc"
+	"go.autokitteh.dev/autokitteh/internal/config"
 )
 
 const (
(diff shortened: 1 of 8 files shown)


tebeka (approved):
Approving but would love to see a test or two

ig-ra (changes requested):
there is already similar try in config dir :)
could u please move it to internal config as well?

ig-ra (approved):

tebeka:
> * Support passing env variable that parsed as list of strings
> 
> ```shell
> MYVAR: a,b,c 
> ```

Use JSON notation for list, don't invent another serialization format.

tebeka (approved):

efiShtain:
Closing for now, 
we might re do it later

PR #643, opened 2024-09-10
feat(config/db): gorm pool connections

ig-ra:


internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -76,6 +76,14 @@ func (db *gormdb) Connect(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("opendb: %w", err)
 	}
+	sqlDB, _ := client.DB()
+
+	if db.cfg.MaxOpenConns > 0 {
+		sqlDB.SetMaxOpenConns(10)
```
efiShtain: Why do you ignore the provided value? 

Set 10 as default and use the provided value if provided

Make sure it is non negative

ig-ra: oops.. leftovers from testing. fixed

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -76,6 +76,14 @@ func (db *gormdb) Connect(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("opendb: %w", err)
 	}
+	sqlDB, _ := client.DB()
+
+	if db.cfg.MaxOpenConns > 0 {
+		sqlDB.SetMaxOpenConns(db.cfg.MaxOpenConns)
+	}
+	if db.cfg.MaxIdleConns > 0 {
```
efiShtain, line 84: is there an issue if maxIdle > then maxOpen ?

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -76,6 +76,14 @@ func (db *gormdb) Connect(ctx context.Context) error {
 	if err != nil {
 		return fmt.Errorf("opendb: %w", err)
 	}
+	sqlDB, _ := client.DB()
```
efiShtain: please handle the possible error

ig-ra: of there is a client with initialized DB (which we opened 3 lines earlier) and we checked for the error 1 line earlier, then we could get DB from it without checking the error. wdyt?

efiShtain: if client.DB can return an error, I think we should check it
if there was no possible error, it would have just return the reference I assume

and it's a crucial part of the system, so 3 more lines of error checking is fine

efiShtain (approved):

PR #646, opened 2024-09-11
fix: get json inputs for session start

RonenMars:


internal/backend/sessionsgrpcsvc/svc.go
```diff
@@ -42,20 +40,11 @@ func (s *server) Start(ctx context.Context, req *connect.Request[sessionsv1.Star
 	}
 
 	for k, v := range msg.JsonInputs {
-		d := json.NewDecoder(strings.NewReader(v))
```
itayd, line 45: you have to decode first, otherwise it'll just convert the json into the string. this is not the problem.

PR #651, opened 2024-09-11
chore: default log level to info, as it should be

itayd:


daabr (commented):
but why? If you don't call `Zap.Level.SetLevel` at all, the default is 0, which means `zapcore.InfoLevel`.

itayd:
> but why? If you don't call `Zap.Level.SetLevel` at all, the default is 0, which means `zapcore.InfoLevel`.

if log level is not explicitly specified, it will be set to info.

daabr (approved):

PR #654, opened 2024-09-11
feat(ENG-1533): allow to ignore prints in session log records get

itayd:


proto/autokitteh/sessions/v1/svc.proto
```diff
@@ -85,6 +85,9 @@ message GetLogRequest {
   // false: all values returned are properly boxed.
   bool json_values = 2;
 
+  // If set, prints will not be returned.
```
daabr: Add in the comment an explanation of what this is useful for, maybe even an example.

daabr (approved):

PR #656, opened 2024-09-12
refactor: db signals api

itayd:
return higher level struct instead of the actual db scheme

### DIFF
--- a/internal/backend/db/db.go
+++ b/internal/backend/db/db.go
@@ -7,7 +7,6 @@ import (
 	"go.uber.org/zap"
 	"gorm.io/gorm"
 
-	"go.autokitteh.dev/autokitteh/internal/backend/db/dbgorm/scheme"
 	"go.autokitteh.dev/autokitteh/internal/backend/health/healthreporter"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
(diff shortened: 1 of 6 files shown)


internal/backend/db/db.go
```diff
@@ -155,3 +154,10 @@ type DB interface {
 	// -----------------------------------------------------------------------
 	GetOwnership(ctx context.Context, entityID sdktypes.UUID) (string, error)
 }
+
+type Signal struct {
```
efiShtain: I think this is the wrong place to put this model definition

please move to internal/backend/models package

later on, we should have all internal models there instead of using sdktypes all over the place

WDYT?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -495,6 +496,25 @@ type Signal struct {
 	Trigger    *Trigger
 }
 
+func ParseSignal(r *Signal) (*db.Signal, error) {
```
efiShtain: just a reference to my previous note about models directory

this line makes no sense, 
you get a Signal (which is a db entity) and return db.Signal - which is business model entity

internal/backend/db/dbgorm/signals.go
```diff
@@ -5,41 +5,43 @@ import (
...
+		Filter:        signal.Filter,
 	}
 
 	return translateError(db.saveSignal(ctx, &s))
 }
 
-func (db *gormdb) ListWaitingSignals(ctx context.Context, dstID sdktypes.EventDestinationID) ([]scheme.Signal, error) {
-	var signals []scheme.Signal
+func (db *gormdb) ListWaitingSignals(ctx context.Context, dstID sdktypes.EventDestinationID) ([]*db.Signal, error) {
+	var rs []*scheme.Signal
```
efiShtain, line 32: what is rs ?
why not signals ?

efiShtain (changes requested):

internal/backend/db/db.go
ig-ra: @efiShtain - it's a good idea to use internal models instead of sdktypes. +1

itayd: not sure this is the wrong place - it's not really a "model" just a struct returned. the goal is just to abstract away the scheme.

internal/backend/db/dbgorm/signals.go
itayd: records

efiShtain (approved):
move to types package

itayd:
> move to types package

done

PR #658, opened 2024-09-12
refactor: make sure all proto reflectors are registered

itayd:
this automatically make sure all the service names are present in the proto lib. then uses it in autokitteh svc.


efiShtain:
first, I'm ok with this change

but,
let me offer another option

this snippet ```generate the exact same file``` you provided automatically

we can add it and run it with go generate or part of make file for protos

pros:
  no need to maintain the file and we know if something change by git diff 

cons:
  it's a utility script we might need to fix for special cases (oauth required a special handling)



```go
package main

import (
	"fmt"
	"os"
	"path/filepath"
	"strings"

	"github.com/iancoleman/strcase"
)

func main() {
	matches, err := filepath.Glob("gen/go/autokitteh/**/v1/*v1connect")
	if err != nil {
		panic(err)
	}

	svcs := make([]string, len(matches))
	imports := make([]string, len(matches))

	for i, m := range matches {
		imports[i] = fmt.Sprintf("\"go.autokitteh.dev/autokitteh/proto/%s\"", m)

		parts := strings.Split(m, "/")
		packageName := parts[len(parts)-1]
		serviceName := parts[len(parts)-3]
		if serviceName == "oauth" {
			serviceName = "OAuth"
		} else {
			serviceName = strcase.ToCamel(serviceName)
		}
		svcs[i] = fmt.Sprintf("%s.%sServiceName", packageName, serviceName)
	}

	f, err := os.OpenFile("svcnames.go", os.O_CREATE|os.O_WRONLY, 0o600)
	if err != nil {
		panic(err)
	}

	fmt.Fprintln(f, "package proto")
	fmt.Println(f, "")
	fmt.Fprintln(f, "import (")
	for _, imp := range imports {
		fmt.Fprintln(f, fmt.Sprintf("\t%s", imp))
	}
	fmt.Fprintln(f, ")")

	fmt.Fprintln(f, "")

	fmt.Fprintln(f, "var ServiceNames = []string{")
	for _, svc := range svcs {
		fmt.Fprintln(f, fmt.Sprintf("\t%s,", svc))
	}
	fmt.Fprintln(f, "}")
}


```

ig-ra:
@efiShtain 
I think the amount of services added/removed is neglected. So IMO we totally could go with hardcoded services file without implementing any automation for it. Just my opinion of coz.. It's just one more thing/utility to support..

tebeka:
What's the motivation for this PR?

itayd:
@efiShtain yeah i think what you say makes sense - but i'd rather do it in a next step. i just went with what i did for validations.

https://linear.app/autokitteh/issue/ENG-1549/

itayd:
> What's the motivation for this PR?

we don't need to manually (and probably forget) adding the names to httpsvc.

efiShtain (approved):

PR #661, opened 2024-09-13
fix: dispatcher now correctly using activities

itayd:
this makes the dispatcher correctly using regular activities, and not local activities. this makes it more robust.


internal/backend/dispatcher/dispatcher.go
```diff
@@ -34,35 +34,58 @@ type Dispatcher struct {
...
+		),
+		workflowName,
+		eventsWorkflowInput{Event: event, Options: opts},
+	)
+	if err != nil {
+		return sdktypes.InvalidEventID, fmt.Errorf("failed starting workflow: %w", err)
 	}
 
-	return eventID, nil
+	sl.With("run", r).Infof("started dispatcher workflow %v for %v", r, eid)
```
daabr, line 79: I strongly disagree that IDs need to be in a log message - they need to be log attributes. The fact that you add them in both places is also a monitoring anti-pattern.

The whole point of structured logging is to use common messages without details, and keep the details as attributes. This allows you to group logs, build smarter analytics and monitoring (dashboard/alerts) based on messages, as well as search easily for all logs related to a specific ID with log filtering.

Furthermore, you keep using `sl.With(...).Log(...)`. This is not how sugared loggers usually accept attributes. Sugared loggers are already expensive compared regular ones, and you're creating a new sugared logger for each log message.

All of these things together are signs of bad logging.

internal/backend/dispatcher/dispatcher.go
```diff
@@ -34,35 +34,58 @@ type Dispatcher struct {
 	sdkservices.Envs
 }
 
-func New(d Dispatcher) *Dispatcher {
-	return &d
+type Dispatcher struct {
+	sl   *zap.SugaredLogger
```
daabr, line 38: Pass a logger, not a sugared logger. This is much more performant, and you can always convert a regular logger to a sugared one when needed.

internal/backend/dispatcher/workflow.go
```diff
@@ -0,0 +1,202 @@
...
+	var sds []sessionData
+
+	if err := workflow.ExecuteActivity(wctx, getEventSessionDataActivityName, event, input.Options).Get(wctx, &sds); err != nil {
+		sl.With("err", err).Errorf("could not get session data: %v", err)
+		return nil, err
+	}
+
+	var started []sdktypes.SessionID
+
+	sl.Infof("found %d dispatch destinations for %v", len(sds), eid)
```
daabr, line 96: Bad log message: if it didn't contain %d and %v we could have created a dashboard for it.

internal/backend/dispatcher/workflow.go
```diff
@@ -0,0 +1,202 @@
...
+
+	sl := d.sl.With("event_id", eid)
+	sl.Debug("started events workflow for %v", eid)
+
+	wctx = temporalclient.WithActivityOptions(wctx, taskQueueName, d.cfg.Activity)
+
+	var sds []sessionData
+
+	if err := workflow.ExecuteActivity(wctx, getEventSessionDataActivityName, event, input.Options).Get(wctx, &sds); err != nil {
+		sl.With("err", err).Errorf("could not get session data: %v", err)
```
daabr, line 90: Bad log message: if it didn't contain %v we could have created an alert for it.

internal/backend/dispatcher/activities.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+
+	var ts []sdktypes.Trigger
+
+	if cid := dstid.ToConnectionID(); cid.IsValid() {
+		if ts, err = d.svcs.Triggers.List(ctx, sdkservices.ListTriggersFilter{ConnectionID: cid}); err != nil {
+			return nil, fmt.Errorf("list triggers: %w", err)
+		}
+
+		sl.Infof("found %d triggers for connection %v", len(ts), cid)
```
daabr, line 107: bad log message: without %d and %v in it we could create a custom metric for it that tracks avg/med/max no. of trigger per connection across our data (with optional filtering per attribute)

daabr (changes requested):
The code itself is good, but I can't approve the logging in this PR. It is inefficient and prevents log-based analytics. The comments I left are just examples. The same issues are in all files.

daabr (approved):
Preemptive approval - agreed to at least store in the struct logger instead of sugared logger.

itayd:
> Preemptive approval - agreed to at least store in the struct logger instead of sugared logger.

done. ftr - we agreed to try out both types of logging styles for now.

PR #660, opened 2024-09-12
feat: extend new pythonrt with remote

efiShtain:
This is a refactor on the new remoteRT
it doesn't change the logic, only split the files in a different way 
so it can be extended to be used with remote runners in a simpler way

PR #663, opened 2024-09-13
[DO NOT MERGE] feat: allow to supply configs to runtimes

itayd:
this is not the most pretty, but it does allow to pass runtime specific configuration.


efiShtain:
please don't merge this one

have a look at my PR https://github.com/autokitteh/autokitteh/pull/660 which uses config for runtime as well (in a different way)

since the pythonrt is in high priority vs starlark, I prefer if you would postpone this PR

itayd:
> please don't merge this one
> 
> have a look at my PR #660 which uses config for runtime as well (in a different way)
> 
> since the pythonrt is in high priority vs starlark, I prefer if you would postpone this PR

we should discuss which approach is best, for now will not merge

itayd:
Obsolete, replaced by #723.

PR #664, opened 2024-09-13
test+fix(ENG-695): test different kind of session errors

itayd:


internal/backend/sessions/sessionworkflows/config.go
```diff
@@ -18,4 +18,6 @@ type Config struct {
 	OSModule bool `koanf:"os_module"`
 
 	SlowOperationTimeout time.Duration `koanf:"slow_operation_timeout"`
+
+	Test bool `koanf:"test"`
```
daabr, line 23: Please add a comment that explains the purpose/function of this flag.

internal/backend/sessions/sessioncalls/call.go
```diff
@@ -66,12 +67,24 @@ func (cs *calls) invoke(ctx context.Context, callv sdktypes.Value, args []sdktyp
...
+		}()
+
+		return caller.Call(ctx, callv, args, kwargs)
+	}()
 	if err != nil {
 		if errors.Is(err, workflow.ErrCanceled) && errors.Is(ctx.Err(), context.DeadlineExceeded) {
 			err = context.DeadlineExceeded
 		}
 
+		if errors.Is(err, test.ErrTestHard) {
```
daabr: Adding test-only code isn't strictly wrong, sometimes it's a valid choice, and I agree with it in this case.

But in general it is considered an anti-pattern because you're adding dead code to production, and risking having tests that test themselves and not the product.

So a short comment about the purpose of this logical branch, and maybe a reference to the test(s) that use it may be good ideas.

internal/backend/sessions/sessioncalls/call.go
```diff
@@ -100,22 +113,12 @@ func (cs *calls) executeCall(ctx context.Context, call sdktypes.SessionCallSpec,
 
 	sl = sl.With(zap.Any("f", callv))
 
-	sl.Debug("calling")
+	sl.Debug("invoking")
```
daabr: This is a clear sign that you shouldn't have passed a sugared logger in the first place.

sdk/sdktypes/value_scalar.go
```diff
@@ -151,10 +151,23 @@ func (BooleanValue) isConcreteValue() {}
 
 func (s BooleanValue) Value() bool { return s.read().V }
 
-func NewBooleanValue(v bool) Value {
+func mkbool(v bool) Value {
```
daabr: `mkbool` is not a self-documenting function name. It's confusing even though it's private.

```suggestion
func boolValue(v bool) Value {
```

sdk/sdktypes/value_scalar.go
```diff
@@ -151,10 +151,23 @@ func (BooleanValue) isConcreteValue() {}
...
 	return forceFromProto[Value](&ValuePB{Boolean: &BooleanValuePB{V: v}})
 }
 
+var (
+	True  = mkbool(true)
+	False = mkbool(false)
```
daabr: Don't let readers get confused by `true` vs. `True`.
```suggestion
	TrueValue  = mkbool(true)
	FalseValue = mkbool(false)
```

sdk/sdktypes/value_scalar.go
```diff
@@ -151,10 +151,23 @@ func (BooleanValue) isConcreteValue() {}
...
+)
+
+func NewBooleanValue(v bool) Value {
```
daabr, line 163: Question: what are you gaining by the change in this file?

daabr (commented):
`connections/vars.txtar` and `sessions/stop.txtar` are still failing

daabr (approved):
preemptive approval

internal/backend/sessions/sessionworkflows/config.go
itayd: done

internal/backend/sessions/sessioncalls/call.go
itayd: i use the sugared logger  in this module everywhere else, not worth desugaring it just for this. adapted to sugared logger elsewhere.

sdk/sdktypes/value_scalar.go
itayd: done

sdk/sdktypes/value_scalar.go
itayd: done, begrudgingly

PR #665, opened 2024-09-13
feat(integrations): add connection tests

pashafateev:


integrations/atlassian/confluence/client.go
```diff
@@ -75,3 +76,33 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
 	})
 }
+
+// connTest is an optional connection test provided by the integration
+// to AutoKitteh. It is used to verify that the connection is working
+// as expected. The possible results are "OK" and "error".
+func connTest(i *integration) sdkintegrations.OptFn {
+	return sdkintegrations.WithConnectionTest(func(ctx context.Context, cid sdktypes.ConnectionID) (sdktypes.Status, error) {
+		if !cid.IsValid() {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "init required"), nil
```
daabr: Please capitalize status messages in this PR.
I'll create a quick PR to do it in all existing `connStatus` functions.

integrations/aws/client.go
```diff
@@ -128,3 +130,33 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+		}
+
+		s3Client := s3.NewFromConfig(cfg)
+
+		_, err = s3Client.ListBuckets(ctx, &s3.ListBucketsInput{})
+		if err != nil {
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, "failed to connect to AWS instance"), nil
+		}
+
+		return sdktypes.NewStatus(sdktypes.StatusCodeOK, "connection test was succesfull"), nil
```
daabr: Remove the success message in all other integrations too.

```suggestion
		return sdktypes.NewStatus(sdktypes.StatusCodeOK, ""), nil
```

integrations/atlassian/confluence/client.go
```diff
@@ -75,3 +76,33 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+		if err != nil {
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, "invalid Atlassian base URL"), nil
+		}
+
+		token := vs.GetValueByString("oauth_AccessToken")
+
+		res, err := accessibleResources(nil, baseURL, token)
+		print(res)
+		if err != nil {
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, "failed to connect to Confluence instance"), nil
```
daabr: Unlike `connStatus`, I think it's more useful in `connTest` to return `err.Message()` rather than an opaque message. Same in all other integrations.

integrations/aws/client.go
```diff
@@ -128,3 +130,33 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+					vs.GetValueByString("SecretKey"),
+					vs.GetValueByString("Token"))),
+		)
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		s3Client := s3.NewFromConfig(cfg)
+
+		_, err = s3Client.ListBuckets(ctx, &s3.ListBucketsInput{})
```
daabr: AWS is a complicated case. User permissions are granular - some people may have S3 access, others may not. Is there something more basic, like getting the user's own details or the user's IAM groups, etc? It won't ensure that the user can you each AWS API, but it will demonstrate that the auth details are valid.

integrations/atlassian/confluence/client.go
pashafateev: Done

integrations/aws/client.go
pashafateev: Done

integrations/atlassian/confluence/client.go
pashafateev: Done

integrations/aws/client.go
pashafateev: Done

integrations/github/client.go
```diff
@@ -3,13 +3,20 @@ package github
 import (
 	"context"
 
+	// "fmt"
```
daabr: Remove unnecessary imports

integrations/twilio/client.go
```diff
@@ -81,3 +85,60 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+		accountSID := decodedVars.AccountSID
+		authSID := accountSID
+		authToken := ""
+
+		if at.Value() == integrations.APIKey {
+			authSID = decodedVars.Username
+		}
+		authToken = decodedVars.Password
+
+		url := fmt.Sprintf("https://api.twilio.com/2010-04-01/Accounts/%s.json", accountSID)
```
daabr: We already have twilio-go as a dependency, so it makes more sense to use its `client.Api.FetchAccount()` instead of this manually-constructed REST API call. You may not even have to care about the auth type, maybe.

integrations/atlassian/confluence/client.go
```diff
@@ -70,3 +71,32 @@ func connStatus(i *integration) sdkintegrations.OptFn {
 		}
 	})
 }
+
+// connTest is an optional connection test provided by the integration
+// to AutoKitteh. It is used to verify that the connection is working
+// as expected. The possible results are "OK" and "error".
```
daabr, line 77: What about the warning status for init required?

daabr: btw, I see the sense in a warning for `connStatus` (because it may be called before a connection is ready to be used), but arguably `connTest` (which is called only on-demand) should consider "Init required" as a failure of the test? otoh, maybe it's good to keep symmetry with `connStatus` - wdyt?

integrations/aws/client.go
```diff
@@ -128,3 +131,36 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+					vs.GetValueByString("SecretKey"),
+					vs.GetValueByString("Token"))),
+		)
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		stsClient := sts.NewFromConfig(cfg)
+
+		_, err = stsClient.GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
```
daabr:
```suggestion
		_, err = sts.NewFromConfig(cfg).GetCallerIdentity(ctx, &sts.GetCallerIdentityInput{})
```

integrations/chatgpt/client.go
```diff
@@ -77,3 +80,24 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+	return sdkintegrations.WithConnectionTest(func(ctx context.Context, cid sdktypes.ConnectionID) (sdktypes.Status, error) {
+		vs, err := i.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		apiKey := vs.Get(apiKeyVar).Value()
+		client := openai.NewClient(apiKey)
+		_, err = client.ListModels(ctx)
+		if err != nil {
```
daabr:
```suggestion
		if _, err = client.ListModels(ctx); err != nil {
```

integrations/discord/client.go
```diff
@@ -68,3 +71,28 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+		}
+
+		token := vs.Get(vars.BotToken).Value()
+		dg, err := discordgo.New("Bot " + token)
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		_, err = dg.User("@me")
+		if err != nil {
```
daabr:
```suggestion
		if _, err = dg.User("@me"); err != nil {
```

integrations/github/client.go
```diff
@@ -78,6 +78,35 @@ func connStatus(i *integration) sdkintegrations.OptFn {
 	})
 }
 
+type ctxKey struct{}
```
daabr: Lint warning: unused type

integrations/google/connections/connections.go
```diff
@@ -0,0 +1,114 @@
...
+			client, err = getOAuthClient(vs.GetValueByString("oauth_AccessToken"))
+		case "jsonKey":
+			client, err = getServiceAccountClient(vs.GetValue(vars.JSON))
+		default:
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, "Unsupported auth type"), nil
+		}
+
+		if err != nil {
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
+		}
```
daabr: Subjective nit: simple var declarations and error checks should be attached to the code that relates to uses these vars / sets these errors

```suggestion
		var client *http.Client
		switch at.Value() {
		case "oauth":
			client, err = getOAuthClient(vs.GetValueByString("oauth_AccessToken"))
		case "jsonKey":
			client, err = getServiceAccountClient(vs.GetValue(vars.JSON))
		default:
			return sdktypes.NewStatus(sdktypes.StatusCodeError, "Unsupported auth type"), nil
		}
		if err != nil {
			return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
		}
```

integrations/google/connections/connections.go
```diff
@@ -0,0 +1,114 @@
...
+
+		// Make a simple API call to verify credentials
```
daabr:
```suggestion
		// Make a simple API call to verify credentials.
```

integrations/google/connections/connections.go
```diff
@@ -0,0 +1,114 @@
...
+		resp, err := client.Get("https://www.googleapis.com/oauth2/v2/userinfo?alt=json")
```
daabr, line 80: 1. Why not use `userInfoService` instead of `Get`? (i.e. no need to fetch a specified-URL)
2. Just making sure - this works with JSON too?

integrations/google/connections/connections.go
```diff
@@ -0,0 +1,114 @@
...
+		if resp.StatusCode != http.StatusOK {
+			body, _ := io.ReadAll(resp.Body)
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, string(body)), nil
+		}
+
+		return sdktypes.NewStatus(sdktypes.StatusCodeOK, ""), nil
+	})
+}
+
+func getOAuthClient(acsTkn string) (*http.Client, error) {
```
daabr: Go does prefer short variable names, but don't sacrifice readability.

Avoid abbreviations when they don't **really** help - either use acronyms if their meaning is clear (e.g. based on the var type - see suggestion below), or human-readable names ("accessToken" or just "token").

integrations/google/connections/connections.go
```diff
@@ -0,0 +1,114 @@
...
+	token := &oauth2.Token{
+		AccessToken: acsTkn,
+		TokenType:   "Bearer",
+	}
+
+	tokenSource := oauth2.StaticTokenSource(token)
```
daabr:
```suggestion
	t := &oauth2.Token{AccessToken: acsTkn, TokenType: "Bearer"}
	tokenSource := oauth2.StaticTokenSource(t)
```

integrations/google/gemini/client.go
```diff
@@ -31,6 +37,58 @@ func New(cvars sdkservices.Vars) sdkservices.Integration {
...
+		apiKey := vs.Get(apiKeyVar)
+		if !apiKey.IsValid() || apiKey.Value() == "" {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
+		}
+		url := "https://generativelanguage.googleapis.com/v1beta/models/gemini-1.5-flash:generateContent?key=" + apiKey.Value()
+		requestBody := `{
+				"contents": [{
+					"parts": [{"text": "Hello, Gemini!"}]
+				}]
+			}`
```
daabr: I know I said not to add new Go libraries when HTTP is enough, but in this specific case (the API is still technically in beta) if might be safer to rely on Google's Go library instead of this hardcoded URL.

Also, I think a "cheaper"/quicker check might be this: https://pkg.go.dev/github.com/google/generative-ai-go/genai#Client.ListModels (like in ChatGPT)

integrations/atlassian/confluence/client.go
pashafateev: I switched to error but with the same message "Init required". I think it being considered a failure makes the most sense.

integrations/github/client.go
pashafateev: Done

integrations/github/client.go
pashafateev: Done.

integrations/google/connections/connections.go
pashafateev: 1. Done
2. Yes

integrations/google/connections/connections.go
pashafateev: Done

integrations/google/gemini/client.go
pashafateev: I think using the library simplifies the code significantly, but just as a side note:
> NOTE: This client uses the v1beta version of the API.

https://pkg.go.dev/github.com/google/generative-ai-go/genai#section-documentation

integrations/twilio/client.go
pashafateev: Done

daabr (approved):
Please fix the unused type and imports in `integrations/github/client.go` before merging.

PR #668, opened 2024-09-14
refactor kittehs errors

daabr:
The core of this PR is removing `internal/kittehs/errors.go`.

Other fixes that I did on the way:

1. `fmt.Errorf` without args → `errors.New` in various places
2. `cmd/ak/common/svc.go`: better docstrings
3. `cmd/ak/common/flags.go`: minor refactoring + better docstrings
4. Disable `MANIFEST` in `.gitignore` - so git won't complain about `.../manifest/...` files 

FYI, other `kittehs` files where docs and tests are more important will be refactored too in the future.

Refs: ENG-881


.gitignore
```diff
@@ -109,4 +109,4 @@ share/python-wheels/
 *.egg-info/
 .installed.cfg
 *.egg
-MANIFEST
```
itayd, line 112: talk to @tebeka before removing this

tebeka: IIRC this is automatically generated by the build process.

tebeka (approved):
OK with it in general. Not sure I like `ErrorWithPrefix` since it voilates [LOB](https://htmx.org/essays/locality-of-behaviour/)

ig-ra (approved):

daabr:
> OK with it in general. Not sure I like `ErrorWithPrefix` since it voilates [LOB](https://htmx.org/essays/locality-of-behaviour/)

Agreed. I'll try to reverse course and get rid of the `ErrorWith` functions, or at least not have them in `kittehs`.

PR #671, opened 2024-09-15
fix(python): singleshot event does not have a dict data.

tebeka:
Refs: ENG-1543


itayd (approved):

runtimes/pythonrt/ak_runner/__main__.py
```diff
@@ -117,12 +117,14 @@ def run(args):
...
     # TODO (ENG-624) change this once we support callbacks to autokitteh
-    body = event.get("data", {}).get("body")
-    if isinstance(body, str):
-        try:
-            event["data"]["body"] = b64decode(body)
-        except ValueError:
-            pass
+    data = event.get("data")
+    if isinstance(data, dict):
+        body = data.get("body")
```
daabr, line 123: 1. See #674 for updates re. `body`
2. So if `data` is a string - what happens then? don't you want to try to convert it into a dict?
3. Did @itayd confirm that singleshot events are indeed expected to have string data instead of a JSON string?

daabr (approved):

PR #674, opened 2024-09-15
refactor(webhooks): update fields in webhook events

daabr:
- Add `raw_url` alongside `url`
- Replace `body` functions with data (`bytes`, `form`, `json` - `None` if missing)
- Remove always-empty fields from `url`: leave only `path`, `query`, `fragment` (i.e. remove `scheme`, `opaque`, `host`)

Refs: ENG-1517, ENG-1518

### DIFF
--- a/internal/backend/webhookssvc/svc.go
+++ b/internal/backend/webhookssvc/svc.go
@@ -24,13 +24,13 @@ import (
 const WebhooksPathPrefix = "/webhooks/"
 
 type Service struct {
-	sl         *zap.SugaredLogger
+	logger     *zap.Logger
 	dispatcher sdkservices.Dispatcher
 	db         db.DB
 }
 
 func New(l *zap.Logger, db db.DB, dispatcher sdkservices.Dispatcher) *Service {
-	return &Service{sl: l.Sugar(), db: db, dispatcher: dispatcher}
+	return &Service{logger: l, db: db, dispatcher: dispatcher}
 }
 
 func (s *Service) Start(muxes *muxes.Muxes) {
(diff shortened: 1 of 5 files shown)


internal/backend/webhookssvc/svc.go
```diff
@@ -24,13 +24,13 @@ import (
 const WebhooksPathPrefix = "/webhooks/"
 
 type Service struct {
-	sl         *zap.SugaredLogger
+	logger     *zap.Logger
```
itayd, line 27: i use `l`, but no problem

internal/backend/webhookssvc/svc.go
```diff
@@ -100,77 +105,73 @@ func requestToData(r *http.Request) (map[string]sdktypes.Value, error) {
...
+	return sdktypes.NewDictValueFromStringMap(
 		map[string]sdktypes.Value{
-			"text":  kittehs.Must1(sdktypes.NewConstFunctionValue("text", sdktypes.NewStringValue(string(body)))),
-			"bytes": kittehs.Must1(sdktypes.NewConstFunctionValue("bytes", sdktypes.NewBytesValue(body))),
-			"json":  jsonBody,
+			"fragment": sdktypes.NewStringValue(u.Fragment),
+			"path":     sdktypes.NewStringValue(u.Path),
+			"query": sdktypes.NewDictValueFromStringMap(
+				kittehs.TransformMapValues(u.Query(), func(vs []string) sdktypes.Value {
+					return sdktypes.NewStringValue(strings.Join(vs, ", "))
```
itayd, line 169:
```suggestion
					return sdktypes.NewStringValue(strings.Join(vs, ","))
```

itayd (approved):

itayd:
i guess kittehub need to be updated as well?

internal/backend/webhookssvc/svc.go
daabr: afacis, canonical implementations and examples of HTTP don't distinguish between:
```
Header: foo, bar
```
And:
```
Header: foo
Header: bar
```
Example: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

PR #682, opened 2024-09-22
feat(db): use gormzap logger

ig-ra:
right our db layer is always using default gorm logger (overriding any other logger), which prevents us to see db logs in datadog
- use zap gorm logger
- switch to default gorm one if debug is requested


efiShtain (approved):
Please open a ticket to fix the debug flag
it should work in production or any other case as well

ig-ra:
> Please open a ticket to fix the debug flag it should work in production or any other case as well

done. ENG-1590

PR #684, opened 2024-09-23
fix(workflow): stop automaticaly delete drained deployments

ig-ra:
This PR removes code that tries to delete draining deployment on last session cleanup. 

- There is a raise condition in deactivate_drain_deployment which is running on every session cleanup. In cleanup it checks whether this is the last session for deploymentID and whether the deployment is draining. The problem/raise condition that deployment could be already deleted by explicit call to project/deployment delete, thus it won't be found and `deactivate_drain_deployment` activity will be stuck forever (activities are not limited in time/attempts)

- Moreover running this code in session cleanup will lock sessions table due to count(*) - when we are trying to check whether this is the last session before deleting deployment.

### DIFF
--- a/internal/backend/sessions/sessionworkflows/activities.go
+++ b/internal/backend/sessions/sessionworkflows/activities.go
@@ -11,7 +11,6 @@ import (
 	"go.temporal.io/sdk/workflow"
 	"go.uber.org/zap"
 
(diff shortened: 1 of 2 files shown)


efiShtain (approved):

efiShtain (approved):
LGTM

PR #687, opened 2024-09-24
feat: custom value

itayd:
custom value allows executors (eg. python) to pass opaque data to be handed back to it later. a custom value also contains the higher level representation that can be used for cross-executor and ui data presentation.


proto/autokitteh/values/v1/values.proto
```diff
@@ -96,6 +96,17 @@ message Function {
...
 
+message Custom {
+  // Executor used to generate this data.
+  string executor_id = 1;
+
+  // Data generated and comprehensible only to the generating executor.
+  bytes data = 2;
+
+  // Value representation of this data for display and cross-executor compatibility.
+  Value value = 3;
```
tebeka, line 107: This should be a string representation of the value. 
We don't have a way to capture all possible user types with our `sdkvalues` type system.

sdk/sdktypes/value.go
```diff
@@ -30,6 +30,15 @@ func validateValue(m *ValuePB) error {
 		objectField[DictValue]("dict", m.Dict),
 		objectField[StructValue]("struct", m.Struct),
 		objectField[ModuleValue]("module", m.Module),
+		objectField[FunctionValue]("function", m.Function),
```
tebeka, line 33: How is this related to the PR? I've expected only `CustomValue` addition here.
This suggests that probably having each type implement a `Validator` interface will be better and more explicit than using protovalidate the reflection magic.

sdk/sdktypes/value_custom.go
```diff
@@ -0,0 +1,65 @@
...
+	)
+}
+
+func (customValueTraits) StrictValidate(m *CustomValuePB) error {
+	return errors.Join(
+		mandatory("executor_id", m.ExecutorId),
+	)
+}
+
+var _ objectTraits[*CustomValuePB] = customValueTraits{}
```
tebeka, line 27: I general I don't like this style. When testing the compiler will fail if the interface is not implemented.

tebeka (changes requested):
See my comment about using a repr string and not a `Value`.

tebeka (approved):

PR #689, opened 2024-09-24
feat(metrics): session invocation delay

ig-ra:
session invocation delay = time between event and session start

### DIFF
--- a/internal/backend/dispatcher/dispatcher.go
+++ b/internal/backend/dispatcher/dispatcher.go
@@ -3,6 +3,7 @@ package dispatcher
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"go.temporal.io/sdk/workflow"
 	"go.uber.org/fx"
@@ -50,6 +51,7 @@ func (d *Dispatcher) Dispatch(ctx context.Context, event sdktypes.Event, opts *s
 	ctx = akCtx.WithRequestOrginator(ctx, akCtx.Dispatcher)
 	ctx = akCtx.WithOwnershipOf(ctx, d.svcs.DB.GetOwnership, event.DestinationID().UUIDValue())
 
+	event = event.WithCreatedAt(time.Now())
 	eid, err := d.svcs.Events.Save(ctx, event)
 	if err != nil {
 		return sdktypes.InvalidEventID, fmt.Errorf("save event: %w", err)
--- a/internal/backend/dispatcher/dispatcher.go
+++ b/internal/backend/dispatcher/dispatcher.go
@@ -56,7 +56,6 @@ func (d *Dispatcher) Dispatch(ctx context.Context, event sdktypes.Event, opts *s
 	if err != nil {
 		return sdktypes.InvalidEventID, fmt.Errorf("save event: %w", err)
 	}
-
 	event = event.WithID(eid)
 
 	sl := d.sl.With("event_id", eid)
--- a/internal/backend/events/events.go
+++ b/internal/backend/events/events.go
@@ -29,7 +29,10 @@ func (e *events) List(ctx context.Context, filter sdkservices.ListEventsFilter) 
 }
 
 func (e *events) Save(ctx context.Context, event sdktypes.Event) (sdktypes.EventID, error) {
-	event = event.WithNewID().WithCreatedAt(time.Now())
+	event = event.WithNewID()
+	if event.CreatedAt() == time.Unix(0, 0).UTC() {
+		event = event.WithCreatedAt(time.Now())
+	}
 
 	if err := e.db.SaveEvent(ctx, event); err != nil {
 		return sdktypes.InvalidEventID, err
--- a/internal/backend/sessions/sessionworkflows/workflows.go
+++ b/internal/backend/sessions/sessionworkflows/workflows.go
@@ -181,7 +181,6 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 		metric.WithAttributes(attribute.Bool("replay", isReplaying)),
 	)
 
-	invocationDelay := time.Since(eventTime)
 	startTime := time.Now() // we want actual start time for metrics.
 	prints, err := runWorkflow(wctx, l, ws, params.Data)
 	duration := time.Since(startTime)
@@ -198,7 +197,7 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 	l = l.With(zap.Duration("duration", duration))
 
 	if eventTime != (time.Time{}) {
-		invocationDelay = time.Since(eventTime)
+		invocationDelay := time.Since(eventTime)
 		l = l.With(zap.Duration("invocation_delay", invocationDelay))
 
 		if !isReplaying {
--- a/internal/backend/sessions/sessionworkflows/workflows.go
+++ b/internal/backend/sessions/sessionworkflows/workflows.go
@@ -126,7 +126,21 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
 
 func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkflowParams) error {
 	wi := workflow.GetInfo(wctx)
-	sid := params.Data.Session.ID()
+	session := params.Data.Session
+	sid := session.ID()
+
+	// get session time (it was created not long after event dispatch) and then update it with actual event time if found
+	eventTime := time.Time{}
+	if eventWrapped, ok := session.Inputs()["event"]; ok {
+		if eventUnwrapped, err := sdktypes.UnwrapValue(eventWrapped); err == nil {
+			if eventMap, ok := eventUnwrapped.(map[string]any); ok {
+				if createdAt, ok := eventMap["created_at"].(time.Time); ok {
+					eventTime = createdAt
+				}
+			}
+		}
+	}
+
 	isReplaying := workflow.IsReplaying(wctx)
 
 	l := ws.l.With(
@@ -167,6 +181,7 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 		metric.WithAttributes(attribute.Bool("replay", isReplaying)),
 	)
 
+	invocationDelay := time.Since(eventTime)
 	startTime := time.Now() // we want actual start time for metrics.
 	prints, err := runWorkflow(wctx, l, ws, params.Data)
 	duration := time.Since(startTime)
@@ -177,16 +192,19 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 	dwctx, done := workflow.NewDisconnectedContext(wctx)
 	defer done()
 
-	sessionDurationHistogram.Record(
-		metricsCtx,
-		duration.Milliseconds(),
-		metric.WithAttributes(
-			attribute.Bool("replay", isReplaying),
-			attribute.Bool("success", err == nil),
-		),
-	)
+	sessionDurationHistogram.Record(metricsCtx, duration.Milliseconds(),
+		metric.WithAttributes(attribute.Bool("replay", isReplaying), attribute.Bool("success", err == nil)))
 
 	l = l.With(zap.Duration("duration", duration))
+
+	if eventTime != (time.Time{}) {
+		invocationDelay = time.Since(eventTime)
+		l = l.With(zap.Duration("invocation_delay", invocationDelay))
+
+		if !isReplaying {
+			sessionInvocationDelayHistogram.Record(metricsCtx, invocationDelay.Milliseconds())
+		}
+	}
 
 	if err != nil {
 		l := l.With(zap.Error(err))
--- a/internal/backend/sessions/sessionworkflows/workflows.go
+++ b/internal/backend/sessions/sessionworkflows/workflows.go
@@ -128,19 +128,6 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 	wi := workflow.GetInfo(wctx)
 	session := params.Data.Session
 	sid := session.ID()
-
-	// get session time (it was created not long after event dispatch) and then update it with actual event time if found
-	eventTime := time.Time{}
-	if eventWrapped, ok := session.Inputs()["event"]; ok {
-		if eventUnwrapped, err := sdktypes.UnwrapValue(eventWrapped); err == nil {
-			if eventMap, ok := eventUnwrapped.(map[string]any); ok {
-				if createdAt, ok := eventMap["created_at"].(time.Time); ok {
-					eventTime = createdAt
-				}
-			}
-		}
-	}
-
 	isReplaying := workflow.IsReplaying(wctx)
 
 	l := ws.l.With(
@@ -150,6 +137,25 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 		zap.String("run_id", wi.WorkflowExecution.RunID),
 		zap.Int32("attempt", wi.Attempt),
 	)
+
+	eventTime := struct {
+		createdAt time.Time
+		valid     bool
+	}{
+		createdAt: time.Time{},
+		valid:     false,
+	}
+
+	if eventWrapped, ok := session.Inputs()["event"]; ok {
+		if eventUnwrapped, err := sdktypes.UnwrapValue(eventWrapped); err == nil {
+			if eventMap, ok := eventUnwrapped.(map[string]any); ok {
+				if createdAt, ok := eventMap["created_at"].(time.Time); ok {
+					eventTime.createdAt = createdAt
+					eventTime.valid = true
+				}
+			}
+		}
+	}
 
 	wctx = temporalclient.WithActivityOptions(wctx, taskQueueName, ws.cfg.Activity)
 
@@ -196,8 +202,8 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 
 	l = l.With(zap.Duration("duration", duration))
 
-	if eventTime != (time.Time{}) {
-		invocationDelay := time.Since(eventTime)
+	if eventTime.valid {
+		invocationDelay := time.Since(eventTime.createdAt)
 		l = l.With(zap.Duration("invocation_delay", invocationDelay))
 
 		if !isReplaying {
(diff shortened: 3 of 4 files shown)


internal/backend/db/dbgorm/sessions.go
```diff
@@ -353,6 +353,7 @@ func (gdb *gormdb) getSessionCallAttemptResult(ctx context.Context, sessionID sd
 }
 
 func (db *gormdb) CreateSession(ctx context.Context, session sdktypes.Session) error {
+	// REVIEW: pass pointer and update createdAT and updatedAt?
```
efiShtain: ?

internal/backend/dispatcher/dispatcher.go
```diff
@@ -50,6 +51,7 @@ func (d *Dispatcher) Dispatch(ctx context.Context, event sdktypes.Event, opts *s
 	ctx = akCtx.WithRequestOrginator(ctx, akCtx.Dispatcher)
 	ctx = akCtx.WithOwnershipOf(ctx, d.svcs.DB.GetOwnership, event.DestinationID().UUIDValue())
 
+	event = event.WithCreatedAt(time.Now()) // ensure that we have a valid creation time
```
efiShtain: comment is kind of redundant

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -87,6 +87,7 @@ func (ws *workflows) StartWorkers(ctx context.Context) error {
 }
 
 func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session) error {
+	fmt.Println("  !!! session: ", session.ID(), session.CreatedAt())
```
efiShtain: no fmt please, if you need log it with debug
if it was for development only, please remove

internal/backend/sessions/sessions.go
```diff
@@ -118,6 +119,8 @@ func (s *sessions) Start(ctx context.Context, session sdktypes.Session) (sdktype
 	ctx = akCtx.WithRequestOrginator(ctx, akCtx.SessionWorkflow)
 	ctx = akCtx.WithOwnershipOf(ctx, s.svcs.DB.GetOwnership, session.EnvID().UUIDValue())
 
+	// REVIEW: pass session by pointer to DB.CreateSession and update createdAT?
+	session = session.WithCreatedAndUpdatedAt(time.Now())
```
efiShtain: db layer already adds the creation time to the session object - this causes different times between what you create here and what's stored in db

need to find a way to reuse the same time or add startExecutionTime to session or something like that

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -126,7 +127,21 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
 
 func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkflowParams) error {
 	wi := workflow.GetInfo(wctx)
-	sid := params.Data.Session.ID()
+	s := params.Data.Session
+	sid := s.ID()
+
+	// get session time (it was created not long after event dispatch) and then update it with actual event time if found
```
efiShtain: you can't really assume that,
assume we get the event but workers are busy, you'll get a delay which might be long from event to event workflow
then you still need to call execute workflow for the session workflow, can have another delay
only then you get here

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -126,7 +127,21 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
 
 func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkflowParams) error {
 	wi := workflow.GetInfo(wctx)
-	sid := params.Data.Session.ID()
+	s := params.Data.Session
```
efiShtain: please use session instead of s

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -126,7 +127,21 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
...
+	sid := s.ID()
+
+	// get session time (it was created not long after event dispatch) and then update it with actual event time if found
+	eventTime := s.CreatedAt()
```
efiShtain: this is a weird assignment, and probably not accurate 
see my comment above

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -126,7 +127,21 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
...
+	if eWr, ok := s.Inputs()["event"]; ok {
```
efiShtain: what is eWr ?
please use meaningful names

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -126,7 +127,21 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
...
+		if eUnWr, err := sdktypes.UnwrapValue(eWr); err == nil {
```
efiShtain: what is eUnWr ?

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -126,7 +127,21 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
...
+			if eMap, ok := eUnWr.(map[string]any); ok {
+				if createdAt, ok := eMap["created_at"].(time.Time); ok {
+					eventTime = createdAt
```
efiShtain: why do you need this time if you check the session time ?

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -177,16 +193,12 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
-		metricsCtx,
-		duration.Milliseconds(),
-		metric.WithAttributes(
-			attribute.Bool("replay", isReplaying),
-			attribute.Bool("success", err == nil),
-		),
-	)
+	sessionDurationHistogram.Record(metricsCtx, duration.Milliseconds(),
+		metric.WithAttributes(attribute.Bool("replay", isReplaying), attribute.Bool("success", err == nil)))
+	sessionInvocationDelayHistogram.Record(metricsCtx, duration.Milliseconds(),
```
efiShtain: you shouldn't log this metric if it's replay
it is only for the first invocation

internal/backend/db/dbgorm/sessions.go
ig-ra: we update session.createdAt only in service. but since its passed by value you cannot benefit from it.
So I'm thinking maybe we could pass by ref here and update time

internal/backend/dispatcher/dispatcher.go
ig-ra: removed

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: already removed

ig-ra: already removed. debugging, yes

internal/backend/sessions/sessions.go
ig-ra: that's exactly what I was marking by REVIEW. I think we could pass event and session by ref. Then in upper layer (e.g. dispatcher) they will have updated time as well

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: assume what?

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: session time is the first close value to the event time. I want just to be sure that we have some valid value that represent something interesting. So session was created pretty soon after saving an event. Therefore I'm using it as a first close thing. In the lines below, I'm updating it by the actual event time if present

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: explain?

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: done

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: done

internal/backend/sessions/sessionworkflows/workflows.go
efiShtain: can't assume the session time is not that different from event time

internal/backend/sessions/sessionworkflows/workflows.go
efiShtain: see my comments above
you can't assume timing on session creation and event creation time - they might differ by a lot 

depends on what we want to measure, we need to use its time only

internal/backend/sessions/sessionworkflows/workflows.go
efiShtain: either user session or event time depends on the requirement

internal/backend/db/dbgorm/sessions.go
efiShtain: let's raise with everyone
not sure what's the correct answer here

ig-ra: rewritten differently. we cannot pass by reference and change sdktype anyway

internal/backend/sessions/sessions.go
ig-ra: since I'm not relying on session time anymore - I removed all this

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: we talked. I won't use session time as a sane default

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: we talked. removed session time. only event time will be used

internal/backend/events/events.go
```diff
@@ -29,7 +29,10 @@ func (e *events) List(ctx context.Context, filter sdkservices.ListEventsFilter)
 }
 
 func (e *events) Save(ctx context.Context, event sdktypes.Event) (sdktypes.EventID, error) {
-	event = event.WithNewID().WithCreatedAt(time.Now())
+	event = event.WithNewID()
+	if event.CreatedAt() == time.Unix(0, 0).UTC() {
```
efiShtain, line 33: I think this should assume this his set from outside and use what ever was provided

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -177,17 +191,20 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
-		),
-	)
+	sessionDurationHistogram.Record(metricsCtx, duration.Milliseconds(),
+		metric.WithAttributes(attribute.Bool("replay", isReplaying), attribute.Bool("success", err == nil)))
 
 	l = l.With(zap.Duration("duration", duration))
 
+	if eventTime != (time.Time{}) {
+		invocationDelay := time.Since(eventTime)
+		l = l.With(zap.Duration("invocation_delay", invocationDelay))
```
efiShtain, line 207: why do we need to add it to every log message ?
or even log it at all?
maybe add some configurable threshold to log warn high latency if above threshold

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -177,17 +191,20 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
-			attribute.Bool("replay", isReplaying),
-			attribute.Bool("success", err == nil),
-		),
-	)
+	sessionDurationHistogram.Record(metricsCtx, duration.Milliseconds(),
+		metric.WithAttributes(attribute.Bool("replay", isReplaying), attribute.Bool("success", err == nil)))
 
 	l = l.With(zap.Duration("duration", duration))
 
+	if eventTime != (time.Time{}) {
```
efiShtain: what does this check ?
event time should always be set

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: there will be only 1 log message with duration (see 197) and invocation delay

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: this is the check we discussed yesterday. I may fail to find event time in inputs.. so if I failed to check then no metric will be added as you requested.

internal/backend/sessions/sessionworkflows/workflows.go
efiShtain: but you replace l with a new l with the addition of invocation_delay

internal/backend/events/events.go
ig-ra: I afraid that this will break the existing assumption that save event will set `created_at`, as it was before. That's why I kept setting created_at here if unset, just for sanity.

Moreover it's also misaligned with other sdktypes/db entities, for example session which is setting created_at in db layer. 
I suggest to open a ticket and decide what will  be the best option to handle it everywhere. It actually could be a part of your suggestion to add more `created_at` fields to other db entities. WDYT?

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: I'm adding invocation_delay to the single log message that would be printed, yes.
previously this message contained duraiton only. Now it will contain invocation delay as well.
This is exactly what I wanted to achieve - 1 message (done/failed/etc) with duration and invocation_delay fields.
You want it to be done differently?

internal/backend/events/events.go
daabr: Setting it here would create inconsistent and unreliable time measurements. Better not to set it here at all - then at least we can see if there's a problem that needs fixing (i.e. having events with an epoch timestamp).

As for an assumption that this function sets timestamps - I don't see that mentioned anywhere, so:
1. Return an error if the event doesn't have a real timestamp
2. Document this function and clarify in that comment that the function requires the caller to set the timestamp in advance

Alternatively, you can remove `WithCreatedAt` and add it as a parameter in `Save`.

Either way I agree that we need to standardize how we handle timestamps across all SDK types - please do open a ticket for that.

FYI, my personal preference is that it should be automatic without the caller having any say in the matter.

API consistency usually enforces measurement consistency, but to be clear measurement consistency in metrics is even more important than API consistency.

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -126,7 +126,21 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
 
 func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkflowParams) error {
 	wi := workflow.GetInfo(wctx)
-	sid := params.Data.Session.ID()
+	session := params.Data.Session
+	sid := session.ID()
+
+	// get session time (it was created not long after event dispatch) and then update it with actual event time if found
```
daabr: Please add punctuation and/or improve the wording, I'm not sure what this comment is saying.

Also, you have to handle and report errors in the block below, you can't assume that the data is correct or that the code will work as expected.

internal/backend/sessions/sessionworkflows/workflows.go
daabr: Checking the result of a code block 60 lines above here is not acceptable. No one should be expected to understand that a value check in line 200 depends on code in line 140.

If you **have** to, you can use a self-documenting bool (e.g. `createTimeFound`) - it's not ideal either, but at least it explains the logical flow.

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -126,7 +126,21 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
...
+				if createdAt, ok := eventMap["created_at"].(time.Time); ok {
+					eventTime = createdAt
+				}
+			}
+		}
+	}
+
 	isReplaying := workflow.IsReplaying(wctx)
 
 	l := ws.l.With(
```
daabr, line 133: It's clearer to set logger attributes as early as possible, i.e. call `l := ws.l.With()` at the beginning of the function without `replay`, and call `l = l.With()` here with only `eventTime` and `replay`.

internal/backend/sessions/sessionworkflows/workflows.go
daabr: I agree with @ig-ra on this. Adding (relevant) attributes to a log message doesn't reduce it's readability.

efiShtain: I didn't say it reduces it's readability 
I meant I just added more attributes to all logs later instead of one specific logs

but I'm ok with it, resolving

internal/backend/events/events.go
ig-ra: Daniel, 
- in the current code the time actually is set *implicitly* *here*, in `Save()`. So for our purposes it was both consistent and reliable I think. I just added a check that event time could be already present in event passed to Save and if not set,  then event's time is updated from epoch to current time.
-  I don't want to change API (e.g. passing time to Save), not in this PR at least. We could discuss this at sync and see what's the best/consistent/convenient option to set event time. This PR tries to introduce a metric on top of existing API.
- I already tried several approaches and unfortunately none of them was convenient :( 
   -  Event is sdktype so every change to it creates new object, thus we cannot pass the reference to Event and update time on db layer before saving
   - we could change API such way that Save till return Event and not EventID, but it requires many changes in code. So I would like to keep this specific PR smaller and deal only with metrics and in other PR deal with where event time is set. WDYT?

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: Converted eventTime to the struct with bool `valid` field which is used to check later on line 200, as u suggested

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: 1. removed the comment - it was remaining of the old flow
2. there is no handling error in the block below. We are trying to extract event time and if we fail, we should continue with the session flow. The only thing that should happen is that metric won't be reported (per efi's request. I would rather to report this metric with session time instead if there is no event time - this what comment was saying)
3. converted event time to the struct with `valid` bool field. I think this is what you requested in other comment

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: done. moved setting logger attrs to the beginning of the functions

daabr (approved):

PR #691, opened 2024-09-25
feat(ENG-1555)!: encrypting/compressing data converter

itayd:
use a data converter to compress and encrypt data given to temporal. 

note that this requires a new key to be supplied in prod, 64 char length hex.

### DIFF
--- a/internal/backend/temporalclient/config.go
+++ b/internal/backend/temporalclient/config.go
@@ -38,6 +38,8 @@ type Config struct {
 	DevServer testsuite.DevServerOptions `koanf:"dev_server"`
 	TLS       tlsConfig                  `koanf:"tls"`
 
+	DataConverter DataConverterConfig `koanf:"data_converter"`
+
 	EnableHelperRedirect bool `koanf:"enable_helper_redirect"`
 }
 
@@ -51,6 +53,12 @@ var (
 	Configs = configset.Set[Config]{
 		Default: &Config{
 			Monitor: defaultMonitorConfig,
+			DataConverter: DataConverterConfig{
+				Compress: true,
+				Encryption: DataConverterEncryptionConfig{
+					Encrypt: true,
+				},
+			},
 		},
 		Dev: &Config{
 			Monitor:               defaultMonitorConfig,
--- a/internal/backend/temporalclient/dataconverter.go
+++ b/internal/backend/temporalclient/dataconverter.go
@@ -6,25 +6,19 @@ import (
 	"encoding/hex"
 	"errors"
 	"fmt"
-	"os"
 	"strings"
 
 	"go.temporal.io/sdk/converter"
-)
-
-const (
-	configKeyEnvVarPrefix = "AK_DATACONV_ENCRYPTION_KEY_"
 )
 
 type DataConverterEncryptionConfig struct {
 	// If `Encrypt` is true, `KeyNames` must have at least one key name specified.
 	Encrypt bool `koanf:"encrypt"`
 
-	// Comma-separated list of key names.
-	// First key in the active encryption key. All the rest are for decryption only.
-	// For each name, the encryption key is fetched from the environment, named as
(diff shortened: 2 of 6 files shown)


internal/backend/temporalclient/config.go
```diff
@@ -38,6 +38,8 @@ type Config struct {
 	DevServer testsuite.DevServerOptions `koanf:"dev_server"`
 	TLS       tlsConfig                  `koanf:"tls"`
 
+	DataConverter DataConverterConfig `koanf:"data_converter"`
```
efiShtain, line 41: do we want to add enableDataConverter to make stuff easier for open source ?
we don't really care about it in this case

internal/backend/temporalclient/dataconverter.go
```diff
@@ -0,0 +1,90 @@
...
+
+	if cfg.Compress {
+		codecs = append(codecs, converter.NewZlibCodec(converter.ZlibCodecOptions{AlwaysEncode: true}))
+	}
+
+	return converter.NewCodecDataConverter(parent, codecs...), nil
+}
+
+func newCipher(name string) (cipher.AEAD, error) {
+	key, ok := os.LookupEnv(configKeyEnvVarPrefix + strings.ToUpper(name))
```
efiShtain: don't use os lookup, use standard way of passing configuration through koanf please

internal/backend/temporalclient/config.go
itayd: not sure the relation to oss? if encryption and compression is not enabled it's just a null converter

internal/backend/temporalclient/dataconverter.go
itayd: problem is that we can't get dynamic names using koanf. see comments in config.

internal/backend/temporalclient/dataconverter.go
```diff
@@ -0,0 +1,90 @@
...
+)
+
+type DataConverterEncryptionConfig struct {
+	// If `Encrypt` is true, `KeyNames` must have at least one key name specified.
+	Encrypt bool `koanf:"encrypt"`
+
+	// Comma-separated list of key names.
+	// First key in the active encryption key. All the rest are for decryption only.
+	// For each name, the encryption key is fetched from the environment, named as
+	// "AK_DATACONV_ENCRYPTION_KEY_<name>". Each key is 64 characters long, hex-encoded.
```
daabr: I don't see any checks that the user has indeed defined all the right keys (what *are* they right keys anyway?), and with valid values as described in this comment. This PR needs to check that and guide the user in a clear, user-friendly way in case of errors.

itayd: code changed, now keys are in a single env var.

internal/backend/temporalclient/config.go
efiShtain: oh, so it will work with default configuration without any values ?

internal/backend/temporalclient/config.go
```diff
@@ -51,6 +53,12 @@ var (
 	Configs = configset.Set[Config]{
 		Default: &Config{
 			Monitor: defaultMonitorConfig,
+			DataConverter: DataConverterConfig{
+				Compress: true,
+				Encryption: DataConverterEncryptionConfig{
+					Encrypt: true,
```
efiShtain, line 59: if you want to set it to true by default, you need to add keys here by default
they can be open source keys

otherwise, make it disabled by default

encrypt:true without keys - the server won't start

efiShtain (approved):
Fine by me

just take a decision on default value
either encrypt is on by default and a set of default keys should be provided
or it's off by default

internal/backend/temporalclient/config.go
itayd: disagree here - i think this will require the user to specify keys when used in production. if you don't set these - this will fail and let the user know that keys are missing.

internal/backend/temporalclient/config.go
itayd: see response below.

PR #692, opened 2024-09-25
feat(ENG-1320): embed web platform in backend

itayd:
- served on port 9982
- enabled only for dev by default


efiShtain:
I think we should at least have the UI version displayed somewhere, just to make sure we understand it

fix the static analysis 

other than this I'm fine with it for now

efiShtain (approved):
see my comments

daabr (changes requested):
Let's discuss this in detail offline (once I feel better, hopefully later today)

itayd:
logged the version

PR #693, opened 2024-09-25
fix(db): trace all DB queries only when db.debug = true

daabr:
It's impossible for me to use debug level with *all* DB queries being logged. But I don't know if this is the right way to disable it.

### DIFF
--- a/internal/backend/db/dbgorm/dbgorm.go
+++ b/internal/backend/db/dbgorm/dbgorm.go
@@ -69,7 +69,7 @@ func (db *gormdb) GormDB() *gorm.DB { return db.db }
 
 func (db *gormdb) Connect(ctx context.Context) error {
 	var l *zap.Logger
-	if db.cfg.Debug	{
+	if db.cfg.Debug {
 		l = db.z.Named("gorm")
 	}
 	client, err := gormkitteh.OpenZ(l, db.cfg, func(cfg *gorm.Config) {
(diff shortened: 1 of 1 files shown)


internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -68,7 +68,11 @@ func New(z *zap.Logger, cfg *Config) (db.DB, error) {
 func (db *gormdb) GormDB() *gorm.DB { return db.db }
 
 func (db *gormdb) Connect(ctx context.Context) error {
-	client, err := gormkitteh.OpenZ(db.z.Named("gorm"), db.cfg, func(cfg *gorm.Config) {
+	var l *zap.Logger
```
itayd, line 71: have you seen https://github.com/autokitteh/autokitteh/blob/460a87efe740a55098028240723f0ce982af8591/internal/backend/gormkitteh/z.go#L27?

daabr: ENG-1590 looks like a more comprehensive approach to logging, but IIUC it'll require extensive changes throughout AK - that's why it's a TODO.

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -68,7 +68,11 @@ func New(z *zap.Logger, cfg *Config) (db.DB, error) {
...
+	if db.cfg.Debug {
+		l = db.z.Named("gorm")
```
itayd, line 73: if no logger on non-debug, even errors would not be reported

PR #694, opened 2024-09-25
feat(integrations): add basic asana integration with pat

pashafateev:


web/static/asana/connect/index.html
```diff
@@ -0,0 +1,73 @@
...
+      </li>
+    </ul>
+
+    <!-- TODO(ENG-799): remove ID once its unused -->
+    <form id="save-form" method="post" action="/asana/save">
+      <div class="form-group">
+        <label for="patSecret">PAT (Personal Access Token)</label>
+        <input
+          type="text"
+          name="patSecret"
```
daabr: Why not just "pat"?

integrations/asana/client.go
```diff
@@ -0,0 +1,116 @@
...
+		}
+
+		at := vs.Get(authType)
+		if !at.IsValid() || at.Value() == "" {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
+		}
+
+		if at.Value() == integrations.Init {
+			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "Initialized"), nil
+		}
```
daabr: It's true that we currently support only PATs, but it looks like they support OAuth apps as well, so it's logical to model this like GitHub.

integrations/asana/client.go
```diff
@@ -0,0 +1,116 @@
...
+
+		req, err := http.NewRequest("GET", "https://app.asana.com/api/1.0/users/me", nil)
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		ps := vs.Get(patSecret).Value()
+		req.Header.Add("Authorization", "Bearer "+ps)
+
+		client := &http.Client{}
```
daabr: If you just use `DefaultClient` (a.k.a. `Client{}`), you might as well just use `http.Get()` without a client at all.

integrations/asana/client.go
```diff
@@ -0,0 +1,116 @@
...
+	"go.autokitteh.dev/autokitteh/sdk/sdkintegrations"
+	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type integration struct{ vars sdkservices.Vars }
+
+var (
+	patSecret     = sdktypes.NewSymbol("pat_secret")
```
daabr: just `pat` - cleaner

integrations/asana/save.go
```diff
@@ -0,0 +1,82 @@
...
+		l.Warn("Failed to parse incoming HTTP request", zap.Error(err))
+		c.AbortBadRequest("form parsing error")
+		return
+	}
+
+	// Test the PAT's usability
+	req, err := http.NewRequest("GET", "https://app.asana.com/api/1.0/users/me", nil)
+	if err != nil {
+		l.Warn("Failed to create HTTP request", zap.Error(err))
+		c.AbortBadRequest("request creation error")
```
daabr: Please distinguish between bad request errors, and internal server errors (where the error probably isn't caused by the details of the request)

integrations/asana/client.go
pashafateev: Done

integrations/asana/client.go
pashafateev: Done

integrations/asana/save.go
pashafateev: Done

web/static/asana/connect/index.html
pashafateev: Done

integrations/asana/save.go
```diff
@@ -0,0 +1,90 @@
...
+		return
+	}
+
+	req.Header.Add("Authorization", "Bearer "+ps)
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
+	if err != nil {
+		l.Error("Failed to execute HTTP request", zap.Error(err))
+		c.AbortServerError("execution error")
```
daabr: This includes auth errors, which are more client errors than server errors

daabr (approved):

integrations/asana/save.go
pashafateev: Done

PR #699, opened 2024-09-26
fix: increase session tests timeout 5 -> 10m

itayd:


daabr (approved):
Increasing timeouts is the easiest but least effective way to debug or solve flakiness, but it'll probably help in the short term 🤷

itayd:
might be just added build time/test time, this will help figure it out

PR #705, opened 2024-09-28
chore!: remove all integrations methods

itayd:
remove all integration methods that do not serve non-connection-testing purpose.

they were used only with starlark.


itayd:
lets discuss this in sync.

sdk/sdkmodule/module.go
```diff
@@ -24,6 +24,8 @@ type module struct {
 	opts moduleOpts
 }
 
+var Empty = New()
```
daabr, line 27: This is meaningless syntactic sugar that wastes human lookup time ("what is `Empty`?") instead of just using the obvious call `New()`).

On top of that, why do we even need to specify an empty module when calling `sdkintegrations.NewIntegration`?

daabr (commented):
Hurray cleaning up dead code!!

But:
1. Are you aware that you completely deleted the gRPC and Redis integrations? Is that intentional?
2. A little more to delete - fix lint warnings in AWS and Gmail
3. Also: these unused-but-not-new functions according to GitHub: `getAWSConfig` function in `aws/config.go`, and `valueWrapper` var in `github/values.go`
4. Remove `store` from system tests

itayd:
> 1. Are you aware that you completely deleted the gRPC and Redis integrations? Is that intentional?

Yes - used only in starlark.

> 2. A little more to delete - fix lint warnings in AWS and Gmail

done

> 3. Also: these unused-but-not-new functions according to GitHub: `getAWSConfig` function in `aws/config.go`, and `valueWrapper` var in `github/values.go`

done

> 4. Remove `store` from system tests

done

daabr (approved):

PR #706, opened 2024-09-30
fix(runner): do not resolve runner

ig-ra:
resolving runner (even localhost) could take time and cause temporal to report deadlock.


efiShtain (approved):
Assuming all tests pass I'm fine with it

PR #708, opened 2024-09-30
feat: webhook to respond with event id

itayd:


daabr (changes requested):
You shouldn't return a response body, much less with a JSON content type, unless the request explicitly contains an [Accept](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept) HTTP header.

Instead, it's much easier to add a custom HTTP header to the response, like `AutoKitteh-Event-ID` (and see [here](https://stackoverflow.com/questions/3561381/custom-http-headers-naming-conventions) about why the header *shouldn't* have an `X-` prefix).

itayd:
hmmm. will change to send data only if it accepts json.
will also add the header.

itayd:
Ended up just setting the header on response.

internal/backend/webhookssvc/svc.go
```diff
@@ -87,12 +87,14 @@ func (s *Service) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-	if _, err := s.dispatcher.Dispatch(ctx, event, nil); err != nil {
+	eid, err := s.dispatcher.Dispatch(ctx, event, nil)
+	if err != nil {
 		sl.Errorw("dispatch failed", "event", event, "err", err)
 		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
 		return
 	}
 
+	w.Header().Set("Autokitteh-Event-ID", eid.String())
```
daabr: There's no rule that only the first letter in header words can be capitalized, e.g. https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers#websockets

```suggestion
	w.Header().Set("AutoKitteh-Event-ID", eid.String())
```

daabr (approved):

PR #716, opened 2024-10-01
chore: remove mistakenly re-added requirements.txt

pashafateev:


daabr (commented):
You also need to add asana in the new https://github.com/autokitteh/autokitteh/blob/main/runtimes/pythonrt/runner/pyproject.toml

daabr (approved):

PR #720, opened 2024-10-03
DRAFT: refactor: use explicit callbacks in python instead of call by value to syscalls

itayd:
- add syscalls to callbacks instead of inferring them from ak module in globals
- currently it still routes to the module, but at the session level. this is to keep the trace in autokitteh showing them.

easiest to review this if you begin with `sdk/sdkservices/runtimes.go`.

i bet there is more code to remove due to this, but just wanted to show the approach for now.

this is something I've been talking with @tebeka for a while. not sure if worth it. lmk.

another route could be just putting the function values themselves in the callbacks or someplace else so you wouldn't need to figure them out.

### DIFF
--- a/internal/backend/sessions/sessionworkflows/syscalls.go
+++ b/internal/backend/sessions/sessionworkflows/syscalls.go
@@ -12,6 +12,7 @@ import (
 	"go.autokitteh.dev/autokitteh/internal/backend/sessions/sessioncontext"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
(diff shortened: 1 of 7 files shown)


internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -23,32 +24,76 @@ const (
 	unsubscribeOp = "unsubscribe"
 )
 
-func (w *sessionWorkflow) syscall(ctx context.Context, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
-	if len(args) == 0 {
-		return sdktypes.InvalidValue, fmt.Errorf("expecting syscall operation name as first argument")
-	}
+func (w *sessionWorkflow) syscalls() *sdkservices.RunSyscalls {
+	call := func(ctx context.Context, name string, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
```
tebeka, line 28: Why is this function defined at runtime? What prevents it from being a method?

internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -23,32 +24,76 @@ const (
...
-		return w.start(ctx, args, kwargs)
-	case subscribeOp:
-		return w.subscribe(ctx, args, kwargs)
-	case nextEventOp:
-		return w.nextEvent(ctx, args, kwargs)
-	case unsubscribeOp:
-		return w.unsubscribe(ctx, args, kwargs)
-	default:
-		return sdktypes.InvalidValue, fmt.Errorf("unknown op %q", op)
+	return &sdkservices.RunSyscalls{
```
tebeka, line 39: Why is `RunSyscalls` a struct and not an interface? I don't understand the purpose of `RunSyscalls` and how it helps the code.

runtimes/pythonrt/pythonrt.go
```diff
@@ -35,8 +35,7 @@ var (
 )
 
 type callbackMessage struct {
-	args           []sdktypes.Value
-	kwargs         map[string]sdktypes.Value
+	f              func(context.Context, sdkservices.RunSyscalls) (sdktypes.Value, error)
```
tebeka, line 38: Find a better name than `f`?

runtimes/pythonrt/pythonrt.go
```diff
@@ -461,7 +460,7 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
 			fn, _ := sdktypes.NewFunctionValue(py.xid, fnName, r.Data, nil, pyModuleFunc)
 			py.call(ctx, fn)
 		case cb := <-py.channels.callback:
-			val, err := py.cbs.Call(ctx, py.runID, py.syscallFn, cb.args, cb.kwargs)
+			val, err := cb.f(ctx, *py.cbs.Syscalls)
```
tebeka, line 463: Why do we pass by value and use a pointer field?

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -157,82 +161,71 @@ func (s *workerGRPCHandler) Activity(ctx context.Context, req *pb.ActivityReques
 
 // ak functions
 
-func makeCallbackMessage(args []sdktypes.Value, kwargs map[string]sdktypes.Value) *callbackMessage {
+func rpcSyscall[Resp proto.Message](
```
tebeka, line 164: How is generics helping us here?

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -157,82 +161,71 @@ func (s *workerGRPCHandler) Activity(ctx context.Context, req *pb.ActivityReques
...
+	rid string,
+	callf func(context.Context, sdkservices.RunSyscalls) (sdktypes.Value, error),
+	errf func(error) Resp,
+	vf func(sdktypes.Value) Resp,
+) (Resp, error) {
+	w.mu.Lock()
+	runner, ok := w.runnerIDsToRuntime[rid]
```
tebeka, line 171: If `w.runnerIdsToRuntime` is nil, this code can panic and then we don't unlock the mutex.

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -157,82 +161,71 @@ func (s *workerGRPCHandler) Activity(ctx context.Context, req *pb.ActivityReques
...
-	runner.channels.callback <- msg
-
-	select {
-	case err := <-msg.errorChannel:
-		err = status.Errorf(codes.Internal, "sleep(%f) -> %s", secs, err)
-		return &pb.SleepResponse{Error: err.Error()}, nil
-	case <-msg.successChannel:
-		return &pb.SleepResponse{}, nil
-	}
+	return rpcSyscall(
```
tebeka, line 199: This code is shorter but less explicit than the original version. I prefer more verbose and explicit.

sdk/sdkservices/runtimes.go
```diff
@@ -48,6 +51,20 @@ type (
...
 	RunPrintFunc = func(ctx context.Context, rid sdktypes.RunID, text string)
 	NewRunIDFunc = func() sdktypes.RunID
+
+	RunSleepFunc       = func(ctx context.Context, t time.Duration) error
+	RunStartFunc       = func(ctx context.Context, entrypoint sdktypes.CodeLocation, inputs map[string]sdktypes.Value, memo map[string]string) (string, error)
+	RunSubscribeFunc   = func(ctx context.Context, name, filter string) (uuid.UUID, error)
+	RunUnsubscribeFunc = func(ctx context.Context, id uuid.UUID) error
+	RunNextEventFunc   = func(ctx context.Context, subscriptions []uuid.UUID, t time.Duration) (sdktypes.Value, error)
+
+	RunSyscalls struct {
```
tebeka, line 61: Why not use an interface. IMO even a concrete struct that holds the required state will be better/

sdk/sdkservices/runtimes.go
```diff
@@ -95,3 +114,44 @@ func (rc *RunCallbacks) SafePrint(ctx context.Context, rid sdktypes.RunID, text
...
+
+func (rc *RunCallbacks) SafeSyscalls() *RunSyscalls {
+	sc := rc.Syscalls
+
+	if rc == nil {
+		sc = &RunSyscalls{}
+	}
+
+	if sc.Sleep == nil {
+		sc.Sleep = func(ctx context.Context, t time.Duration) error {
```
tebeka, line 126: Why do we provide these defaults/stubs? How does it help us?

tebeka (commented):
Let's talk on this one.

PR #724, opened 2024-10-04
refactor: config registration

itayd:
- config printing and registration so it wont be necessary to actually initialize the service to populate all configurations for printing. 
- add `Validate()` method to all config to validate without actually starting stuff up.

the underlying motivation here is that the next step I'd like also to document each configuration param in code (tbd how, either field tag or a custom method?).


internal/backend/auth/authloginhttpsvc/config.go
```diff
@@ -28,20 +50,44 @@ type descopeConfig struct {
...
+func (c descopeConfig) Validate() error {
+	if !c.Enabled {
+		return nil
+	}
+
+	if c.ProjectID == "" {
+		return errors.New("project_id is required")
+	}
+
+	if c.ManagementKey == "" {
```
daabr: AFAIK management key isn't required. For example, I have a Descope config in my local server that doesn't have it.

internal/backend/auth/authloginhttpsvc/config.go
```diff
@@ -28,20 +50,44 @@ type descopeConfig struct {
...
 type Config struct {
 	GoogleOAuth oauth2Config  `koanf:"google_oauth"`
 	GithubOAuth oauth2Config  `konf:"github_oauth"`
 	Descope     descopeConfig `koanf:"descope"`
 
 	// Allowed login patterns, separated by commas.
 	// Pattern format is either of:
-	// - "*"       - matches any login
+	// - "" or "*" - matches any login
 	// - "*@host"  - matches any login from host
```
daabr:
```suggestion
	// - "*@domain"  - matches any login, but only from domain
```

daabr (commented):
Didn't finish my review, but here are a couple of comments

backend/svc/config.go
```diff
@@ -0,0 +1,17 @@
+package svc
+
+import (
+	"go.autokitteh.dev/autokitteh/internal/backend/config"
+)
+
+type Config = config.Config
```
tebeka, line 7: I don't like the liberal use we do with type aliases, they are intended for code refactoring.

internal/backend/config/set.go
```diff
@@ -0,0 +1,77 @@
+package config
+
+import (
+	"fmt"
+
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+)
+
+type ComponentConfig interface {
```
tebeka, line 9: Rename to `Validated` or `Validator`

internal/backend/config/set.go
```diff
@@ -0,0 +1,77 @@
...
+
+type emptyComponentConfig struct{}
+
+func (emptyComponentConfig) Validate() error { return nil }
+
+var EmptyComponentConfig emptyComponentConfig
+
+type baseSet interface {
+	hasDefault() bool
+	default_() ComponentConfig
```
tebeka: Not happy with `default_` and in general lower case, unexported, functions in an interface.

internal/backend/config/set.go
```diff
@@ -0,0 +1,77 @@
...
+
+	hasDev() bool
+	dev() ComponentConfig
+
+	hasTest() bool
+	test() ComponentConfig
+}
+
+type Set[T ComponentConfig] struct {
```
tebeka, line 23: IMO we can get rid of `Set` and have only one configuration. The settings of `dev`, `test` and `default` should be set from the outside.

internal/backend/config/set.go
```diff
@@ -0,0 +1,77 @@
...
+func (s Set[T]) dev() ComponentConfig      { return *s.Dev }
+func (s Set[T]) test() ComponentConfig     { return *s.Test }
+
+func (s Set[T]) hasDefault() bool { return s.Default != nil }
+func (s Set[T]) hasDev() bool     { return s.Dev != nil }
+func (s Set[T]) hasTest() bool    { return s.Test != nil }
+
+var EmptySet = Set[emptyComponentConfig]{Default: &EmptyComponentConfig}
+
+func (set *Set[T]) Choose(mode Mode) (chosen T, err error) {
```
tebeka, line 31: Why do we need this and not access `s.Default` etc?

internal/backend/secrets/config.go
```diff
@@ -17,11 +20,17 @@ type Config struct {
...
 }
 
-var (
-	Configs = configset.Set[Config]{
-		Default: &Config{},
-		Dev: &Config{
-			Provider: SecretProviderDatabase,
-		},
+func (c Config) Validate() error {
+	if c.Provider != "" && !kittehs.ContainedIn(providers...)(c.Provider) {
```
tebeka: Use built-in `slices.Contains`

internal/backend/store/store.go
```diff
@@ -23,7 +24,16 @@ type Config struct {
 	ServerURL string `koanf:"server_url"`
 }
 
-var Configs = configset.Set[Config]{
+func (c Config) Validate() error {
+	if c.ServerURL == "" || c.ServerURL == miniredisURL {
+		return nil
+	}
+
+	_, err := url.Parse(c.ServerURL)
```
tebeka, line 32: Try to connect as well?

tebeka (approved):
Approving for the added validation. Still think code is too complex.

internal/backend/auth/authloginhttpsvc/config.go
itayd: removed mgmt key all togethor.

internal/backend/auth/authloginhttpsvc/config.go
itayd: done

backend/svc/config.go
itayd: this is only for exposing internal stuff through this specific package. I don't think I am using aliases anywhere else tbh

internal/backend/config/set.go
itayd: the intention is to add more functionality in the future, so `Validator` is too narrow

internal/backend/config/set.go
itayd: removed all functions, now only constraining known sets.

internal/backend/config/set.go
itayd: don't think so - we want to be able in the future, or even very soon, to list all values and not just keys per configuration mode.

internal/backend/config/set.go
itayd: it's choosing the right one depending on mode - not sure i understand the q.

internal/backend/secrets/config.go
itayd: oooohhhh nice

internal/backend/store/store.go
itayd: nah, that actually has a side effect.

PR #728, opened 2024-10-05
fix(ENG-1648): properly unwrap into jsonable presentation

itayd:


itayd:
btw, sorry for the formatting changes. seems like my editor do these automatically. the actual fix is on line 274.

efiShtain (approved):

PR #732, opened 2024-10-07
fix: pythonrt now builds on windows

itayd:


runtimes/pythonrt/local_runner_windows.go
```diff
@@ -0,0 +1,7 @@
+//go:build windows
+
+package pythonrt
+
+import "os/exec"
+
+func setCmdSysAttrPGID(cmd *exec.Cmd) {}
```
tebeka, line 7: Does this mean that on windows if `ak` fails the Python process will continue to run? @efiShtain ?

runtimes/pythonrt/pythonrt_test.go
```diff
@@ -16,13 +16,14 @@ import (
...
-	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
-	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 	"go.uber.org/zap"
 	"golang.org/x/net/http2"
 	"golang.org/x/net/http2/h2c"
 
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+
```
tebeka, line 26: Where are all this formatting changes come from?

tebeka (commented):
I'd love to see @efiShtain answering my comment.
Also: Is there a product requirement to run on windows?

runtimes/pythonrt/local_runner_windows.go
daabr: +1. Maybe it's better to have a Windows version of `Start` that ensures the runner is shut down without PGID?

runtimes/pythonrt/pythonrt_test.go
daabr: AFAIK the Go formatter can handle correctly both 2-block and 3-block import sections, it just doesn't know that *our* default is 3 blocks. All the rest below is automatic for gofumpt.

PR #733, opened 2024-10-07
bug(python): Drain prints before ending flow.

tebeka:
Fixes: ENG-1660

### DIFF
--- a/runtimes/pythonrt/pythonrt.go
+++ b/runtimes/pythonrt/pythonrt.go
@@ -489,6 +489,7 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
 	}
 }
 
+// drainPrints drains the print channel at the end of a run.
 func (py *pySvc) drainPrints(ctx context.Context) {
 	for {
 		select {
--- a/runtimes/pythonrt/pythonrt.go
+++ b/runtimes/pythonrt/pythonrt.go
@@ -468,6 +468,10 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
 				cb.successChannel <- val
 			}
 		case v := <-py.channels.done:
+			pCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
+			defer cancel()
+			py.drainPrints(pCtx)
+
 			done = v
 			if done.Error != "" {
 				perr := sdktypes.NewProgramError(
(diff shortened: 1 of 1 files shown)


tebeka:
Tested manually, ran @itayd 's workflow before fix and saw missing every 3-4 runs. Now see all prints every time.

runtimes/pythonrt/pythonrt.go
```diff
@@ -485,6 +489,17 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
 	}
 }
 
+func (py *pySvc) drainPrints(ctx context.Context) {
```
daabr, line 493: Please add a docstring explaining this function's purpose: it's easy to understand what it does, but not why.

daabr (approved):

runtimes/pythonrt/pythonrt.go
itayd: this is go, not py :) so you mean comment.

runtimes/pythonrt/pythonrt.go
```diff
@@ -468,6 +468,10 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
 				cb.successChannel <- val
 			}
 		case v := <-py.channels.done:
+			pCtx, cancel := context.WithTimeout(ctx, 3*time.Second)
```
efiShtain: maybe we should check if the process is still up instead of counting on time ?
3 seconds would probably work most of the times, but we saw many problems when system is under load
so I guess it might not really work sometimes

runtimes/pythonrt/pythonrt.go
tebeka: Done

runtimes/pythonrt/pythonrt.go
tebeka: @efiShtain : I thought of something like:
```go
	ctx, cancel := context.WithCancel(ctx)
	defer cancel()
	go func() {
		var req pb.HealthRequest
		for range time.Tick(100 * time.Millisecond) {
			if _, err := py.runner.Health(ctx, &req); err != nil {
				cancel()
			}
		}
	}()
```

What do you think?

tebeka: Leaving at 1 sec wait. We'll see...

PR #734, opened 2024-10-07
feat(projects): Export project

tebeka:
Fixes: ENG-746


cmd/ak/cmd/export/export.go
```diff
@@ -0,0 +1,143 @@
...
+	triggers, err := r.Client.Triggers().List(ctx, tf)
+	if err != nil {
+		return err
+	}
+
+	// We print to buffer until the end so in case of errors we won't print anything but the error
+	var buf bytes.Buffer
+	fmt.Fprintln(&buf, "version: v1")
+	fmt.Fprintln(&buf)
+	fmt.Fprintln(&buf, "project:")
```
itayd: isn't a better way to use the `manifest` package and just serialize it?

tebeka: Yes it is, forgot about that - will change.

proto/autokitteh/projects/v1/svc.proto
```diff
@@ -104,6 +104,15 @@ message DownloadResourcesResponse {
   map<string, bytes> resources = 2;
 }
 
+message ExportRequest {
+  string project_id = 1 [(buf.validate.field).string.min_len = 1];
+}
+
+message ExportResponse {
+  // Zip of the mainfest + files
```
itayd, line 112: i would just return the files unziped in a map with the manifest, but i that's fine as well.

internal/backend/projects/projects.go
```diff
@@ -114,3 +118,153 @@ func (ps *Projects) SetResources(ctx context.Context, projectID sdktypes.Project
 func (ps *Projects) DownloadResources(ctx context.Context, projectID sdktypes.ProjectID) (map[string][]byte, error) {
 	return ps.DB.GetProjectResources(ctx, projectID)
 }
+
+func (ps *Projects) Export(ctx context.Context, projectID sdktypes.ProjectID) ([]byte, error) {
+	const manifestFileName = "autokitteh.yaml"
+	hasManifest := false
```
itayd: i think we should always export it, because stuff might have changed. the files will contain the original manifest, but we should also return the updated somehow as well.

cmd/ak/cmd/export/export.go
tebeka: Done

internal/backend/projects/projects.go
tebeka: It's confusing since there are two `autokitteh.yaml` files in the zip.

tebeka: Renamed the original and added the current.

efiShtain (approved):

PR #736, opened 2024-10-07
fix(pythonrt): propagate killed local python state

ig-ra:
python runner could be killed by the OS, e.g. EOM. We should handle this case and restart the session

### DIFF
--- a/internal/backend/sessions/sessionworkflows/workflows.go
+++ b/internal/backend/sessions/sessionworkflows/workflows.go
@@ -234,6 +234,9 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 				err = nil
 			} else {
 				l.Sugar().Errorf("session workflow error: %v", err)
+				if errors.As(err, &sdkerrors.RetryableErrorType) {
+					panic("panic session to retry")
+				}
 			}
 		}
 	} else {
--- a/internal/backend/sessions/sessionworkflows/workflows.go
+++ b/internal/backend/sessions/sessionworkflows/workflows.go
@@ -235,7 +235,7 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 				l.Sugar().Errorf("session workflow error: %v", err)
 				if errors.As(err, &sdkerrors.RetryableErrorType) {
 					sessionsRetryErrorsCounter.Add(metricsCtx, 1)
-					l.DPanic("panic session to retry")
+					l.Panic("panic session to retry")
 				} else {
 					sessionsErroredCounter.Add(metricsCtx, 1)
 				}
--- a/internal/backend/sessions/sessionworkflows/workflows.go
+++ b/internal/backend/sessions/sessionworkflows/workflows.go
@@ -235,7 +235,7 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 			} else {
 				l.Sugar().Errorf("session workflow error: %v", err)
 				if errors.As(err, &sdkerrors.RetryableErrorType) {
-					panic("panic session to retry")
+					l.DPanic("panic session to retry")
 				}
 			}
 		}
--- a/internal/backend/sessions/sessionworkflows/workflows.go
+++ b/internal/backend/sessions/sessionworkflows/workflows.go
@@ -236,9 +236,8 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 				if errors.As(err, &sdkerrors.RetryableErrorType) {
 					sessionsRetryErrorsCounter.Add(metricsCtx, 1)
 					l.Panic("panic session to retry")
-				} else {
-					sessionsErroredCounter.Add(metricsCtx, 1)
 				}
+				sessionsErroredCounter.Add(metricsCtx, 1)
 			}
 		}
 	} else {
--- a/internal/backend/sessions/sessionworkflows/workflows.go
+++ b/internal/backend/sessions/sessionworkflows/workflows.go
@@ -221,7 +221,6 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 
 			ws.stopped(dwctx, sid)
 		} else {
-			sessionsErroredCounter.Add(metricsCtx, 1)
 
 			ws.errored(dwctx, sid, err, prints)
 
@@ -235,7 +234,10 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 			} else {
 				l.Sugar().Errorf("session workflow error: %v", err)
 				if errors.As(err, &sdkerrors.RetryableErrorType) {
+					sessionsRetryErrorsCounter.Add(metricsCtx, 1)
 					l.DPanic("panic session to retry")
+				} else {
+					sessionsErroredCounter.Add(metricsCtx, 1)
 				}
 			}
 		}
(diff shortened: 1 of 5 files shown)


runtimes/pythonrt/local_runner.go
```diff
@@ -145,6 +144,32 @@ func (r *LocalPython) Start(pyExe string, tarData []byte, env map[string]string,
 }
 
 func (r *LocalPython) Health() error {
+	if _, err := os.FindProcess(r.proc.Pid); err != nil {
```
tebeka: IMO do this after `r.proc.Signal(syscall.Signal(0))` and only if there's an error.

runtimes/pythonrt/local_runner.go
```diff
@@ -145,6 +144,32 @@ func (r *LocalPython) Start(pyExe string, tarData []byte, env map[string]string,
...
+		return fmt.Errorf("runner heath: no runner proc found. %w ", err)
+	}
+
+	var status syscall.WaitStatus
+	if pid, err := syscall.Wait4(r.proc.Pid, &status, syscall.WNOHANG, nil); err == nil && pid == r.proc.Pid {
```
tebeka: Too much happening on this line. Maybe
```go
pid, err := syscall.Wait4(r.proc.Pid, &status, syscall.WNOHANG, nil)
if err != nil {
    return err
}
if pid != r.proc.Pid { // When does this happen?
    return nil // Or an error?
}
if !status.Signaled() {
    return nil
}

switch sig := status.Signal(); sig {
// ...
}
```

runtimes/pythonrt/local_runner.go
```diff
@@ -145,6 +144,32 @@ func (r *LocalPython) Start(pyExe string, tarData []byte, env map[string]string,
...
+		if status.Signaled() {
+			sig := status.Signal()
+			switch sig {
+			case syscall.SIGKILL, syscall.SIGTERM, syscall.SIGABRT, syscall.SIGSEGV, syscall.SIGBUS:
+				err = fmt.Errorf("runner health: proc terminated by signal %v", sig)
```
tebeka: Add pid?

runtimes/pythonrt/local_runner.go
```diff
@@ -145,6 +144,32 @@ func (r *LocalPython) Start(pyExe string, tarData []byte, env map[string]string,
...
+			default:
+				// maybe process communicates with itself? do nothing meanwhile
+				return nil
+			}
+			if err == nil {
+				err = fmt.Errorf("runner health: unexpected proc state")
+			}
+
+			r.Close() // FIXME: not ideal :( should be called by the caller?
```
tebeka: I don't like this. Health should not cleanup. We can wait & hang in a goroutine and then do cleanup.

runtimes/pythonrt/pythonrt.go
```diff
@@ -438,7 +438,10 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
 				return
 			case <-time.After(10 * time.Second):
 				if err := runnerManager.RunnerHealth(ctx, py.runnerID); err != nil {
-					runnerHealthChan <- err
+					py.log.Warn("health loop check err", zap.Error(err))
+
+					// FIXME: need a better solution. Now just stuck here and let temporal panic on deadlock and retry
+					// runnerHealthChan <- err
```
tebeka: Remove commented out code.

tebeka (commented):
I'll let @efiShtain decide.
Is there a issue requesting this?

runtimes/pythonrt/local_runner.go
ig-ra: done

runtimes/pythonrt/local_runner.go
ig-ra: done

runtimes/pythonrt/local_runner.go
ig-ra: meanwhile pid is not relevant IMO. We are not trackign it neither on start, nor on stop.
we could do it tho, but adding it only here feels redundant for me meanwhile

runtimes/pythonrt/local_runner.go
ig-ra: you are totally right. I marked it with FIXME to discuss with you/Efi and get feedbacks how to implement it better. Because `Close()` meanwhile isn't invoked on process stop/ ctx invalidation.

runtimes/pythonrt/pythonrt.go
ig-ra: done

runtimes/pythonrt/local_runner.go
ig-ra: removed meanwhile. Will handle this in different PR - see ENG-1675

runtimes/pythonrt/pythonrt.go
```diff
@@ -440,10 +441,23 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
...
+				healthReq := pb.HealthRequest{}
+
+				resp, err := py.runner.Health(ctx, &healthReq)
+				if err != nil { // no network/lost packet.load? for sanity check the state locally via IPC/signals
+					err = runnerManager.RunnerHealth(ctx, py.runnerID)
+				} else if resp.Error != "" {
+					err = fmt.Errorf("grpc: %s", resp.Error)
+				}
+				if err != nil {
+					py.log.Warn("runner health failed", zap.Error(err))
```
itayd: why not error?

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -234,6 +234,9 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 				err = nil
 			} else {
 				l.Sugar().Errorf("session workflow error: %v", err)
+				if errors.As(err, &sdkerrors.RetryableErrorType) {
+					panic("panic session to retry")
```
itayd: should be `l.Panic(...)`

itayd: i also wouldn't count it as a session error for the statistics. i would make it have another stat.

runtimes/pythonrt/pythonrt.go
ig-ra: done

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: 1. Done
2. Done

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -234,6 +233,12 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 				err = nil
 			} else {
 				l.Sugar().Errorf("session workflow error: %v", err)
+				if errors.As(err, &sdkerrors.RetryableErrorType) {
+					sessionsRetryErrorsCounter.Add(metricsCtx, 1)
+					l.DPanic("panic session to retry")
```
itayd: l.Panic

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -234,6 +233,12 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
+				} else {
```
itayd: no need for else

itayd (approved):

internal/backend/sessions/sessionworkflows/workflows.go
ig-ra: sure

tebeka (approved):

PR #739, opened 2024-10-08
fix(ENG-1661): allow to delete stopped sessions

itayd:
changed from checking explicit states to just checking if state is final

### DIFF
--- a/internal/backend/sessions/sessions.go
+++ b/internal/backend/sessions/sessions.go
@@ -89,7 +89,7 @@ func (s *sessions) Delete(ctx context.Context, sessionID sdktypes.SessionID) err
 
 	// delete only finalized sessions.
 	if state := session.State(); !state.IsFinal() {
-		return fmt.Errorf("%w: cannot delete session, not in final state: %s, session_id: %v", sdkerrors.ErrFailedPrecondition, state, sessionID)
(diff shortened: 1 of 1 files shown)


internal/backend/sessions/sessions.go
```diff
@@ -87,10 +87,9 @@ func (s *sessions) Delete(ctx context.Context, sessionID sdktypes.SessionID) err
 		return err
 	}
 
-	// delete only failed or finished sessions
-	state := session.State()
-	if state != sdktypes.SessionStateTypeCompleted && state != sdktypes.SessionStateTypeError {
-		return fmt.Errorf("%w: cannot delete session, invalid state: %s, session_id: %s", sdkerrors.ErrFailedPrecondition, session.State(), sessionID.String())
+	// delete only finalized sessions.
+	if state := session.State(); !state.IsFinal() {
+		return fmt.Errorf("%w: cannot delete session, not in final state: %s, session_id: %s", sdkerrors.ErrFailedPrecondition, session.State(), sessionID.String())
```
daabr: "Final state" is something that users won't understand.

```suggestion
		return fmt.Errorf("%w: cannot delete session while in progress, state: %s, session_id: %s", sdkerrors.ErrFailedPrecondition, session.State(), sessionID.String())
```

daabr (approved):

PR #740, opened 2024-10-08
fix(slack): parse channel_archive events correctly

daabr:
`is_moved` should be a boolean, but it's sent as a number (0) in `channel_archive` events, thereby breaking Go's JSON unmarshaling.

There's no documentation to say if it's a bug, or if it's specific just to this event, so I chose to continue with the expected behavior rather than the actual one.

Refs: ENG-980

### DIFF
--- a/integrations/slack/events/channel.go
+++ b/integrations/slack/events/channel.go
@@ -1,9 +1,9 @@
 package events
 
 import (
-	"bytes"
 	"encoding/json"
 	"net/http"
+	"strconv"
 
 	"go.uber.org/zap"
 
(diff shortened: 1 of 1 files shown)


integrations/slack/events/channel.go
```diff
@@ -80,6 +81,9 @@ func ChannelGroupHandler(l *zap.Logger, w http.ResponseWriter, body []byte, cb *
 		}
 	}
 
+	// Workaround for ENG-980: "is_moved: 0" should be "is_moved: false".
+	body = bytes.ReplaceAll(body, []byte(`"is_moved":0`), []byte(`"is_moved":false`))
```
itayd: wow this is ugly. i think that instead you can write a custom UnmarshalJSON function for the struct that knows how to do this.

tebeka: Why are we changing the event that comes from slack?

daabr: That's the point of ENG-980: `is_moved` **should** be a boolean, but it's sent as a number, thereby breaking Go's JSON unmarshaling. There's no documentation to say if it's a bug, or if it's specific just to this event, so I chose to continue with the expected behavior rather than the actual one.

daabr: I can do it, but than involved a custom bool-or-int type with a custom unmarshal function - is that really better than a single line of code?

daabr: done

daabr: see previous comment - change approach to have more permissive JSON parsing

itayd (approved):

PR #742, opened 2024-10-08
feat: add session id to all session logs (+python runner)

efiShtain:
- Use zapio writer so python runner logs add session_id to every log
- pass session_id to runtimes


cmd/ak/cmd/runtimes/run.go
```diff
@@ -135,7 +135,7 @@ func run(ctx context.Context, b *sdkbuildfile.BuildFile, path string) (map[strin
 		defer cancel()
 	}
 
-	run, err := runtimes().Run(ctx, sdktypes.NewRunID(), path, b, nil, cbs)
+	run, err := runtimes().Run(ctx, sdktypes.NewRunID(), sdktypes.InvalidSessionID, path, b, nil, cbs)
```
tebeka, line 138: Will this generate a new valid session ID? And will the runtime pass it to Python?

runtimes/pythonrt/local_runner.go
```diff
@@ -126,9 +136,11 @@ func (r *LocalPython) Start(pyExe string, tarData []byte, env map[string]string,
 		"--code-dir", r.userDir,
 	)
 	cmd.Env = overrideEnv(env, r.runnerDir, r.userDir)
+
 	if r.logRunnerCode {
-		cmd.Stdout = os.Stdout
-		cmd.Stderr = os.Stderr
+		r.runnerLogger = &zapio.Writer{Log: r.log, Level: zap.InfoLevel}
```
tebeka: In the `remote` project we have a `Log` RPC endpoint. The idea was to add a log handler to the Python logger that will send RPC calls with logs. This way you'll get log levels as well. These logs are meant for the runner only and not the user code. But currently they go to stderr/stdout and will be captured by your logger.

IMO have 2 loggers, one for stdout and one for stderr, each with a `stream` field and also the runtime.

tebeka (changes requested):

cmd/ak/cmd/runtimes/run.go
efiShtain: No real sessionID for cli call for run. It's like a manual run. So we don't really care about session (there is no session)

tebeka: OK

runtimes/pythonrt/runner/log.py
```diff
@@ -1,12 +1,27 @@
...
-    datefmt="%Y-%m-%dT%H:%M:%S",
-)
+class JsonFormatter(logging.Formatter):
+    """Formatter to dump error message into JSON"""
 
-_log = logging.getLogger("runner")
+    def format(self, record: LogRecord) -> str:
+        record_dict = {
+            "level": record.levelname,
+            "date": self.formatTime(record),
```
tebeka, line 12: Not needed, we'll have it from zap

runtimes/pythonrt/local_runner.go
```diff
@@ -126,9 +136,11 @@ func (r *LocalPython) Start(pyExe string, tarData []byte, env map[string]string,
...
+		cmd.Stdout = r.runnerLogger
```
tebeka: Different logger for each with a field.

runtimes/pythonrt/runner/log.py
```diff
@@ -1,12 +1,27 @@
...
+            "message": record.getMessage(),
+            "module": record.module,
```
tebeka: I prefer the file name, not the module.

tebeka (approved):

PR #744, opened 2024-10-08
chore(python): Shutdown server cleanly

tebeka:
@efiShtain Do you have a scenario to test where the server got stuck before you added `os._exit`?


runtimes/pythonrt/runner/main.py
```diff
@@ -20,9 +20,10 @@
 from grpc_reflection.v1alpha import reflection
 from syscalls import SysCalls
 import threading
-import os
 import time
 
+SERVER_GRACE = 3  # seconds
```
daabr: Optional nit: maybe rename or add a comment that this is a server stop/shutdown grace period? "Server grace" is a little vague.

runtimes/pythonrt/runner/main.py
```diff
@@ -125,7 +127,7 @@ def ShouldKeepRunning(self, intial_delay=10, period=10):
             time.sleep(period)
 
         log.error("could not verify if should keep running, killing self")
-        os._exit(1)
+        self.server.stop(SERVER_GRACE)
```
daabr: Is it intentional that the proc's return value is no longer 1? Also, I'm assuming there's a reason you didn't use `sys.exit()`?

(Same in line 307)

daabr (approved):
Approved, but please reply to the comment on `self.server.stop(SERVER_GRACE)`

runtimes/pythonrt/runner/main.py
tebeka: There's no need, server stop will cause the ` server.wait_for_termination()` to exit which will exit the process.

PR #748, opened 2024-10-09
feat!(ENG-1662): kill drain grpc. deactivates drains first.

itayd:
- kill drain endpoint. deactivate will always drain first.
- auto deactivate deployment once all sessions are done.


daabr (approved):
I'm uncomfortable with the non-atomic nature of the `DeploymentHasActiveSessions` function. We can have sessions start between line 225 and line 229, and switch to "running" state between lines 212 and 220, i.e. while this function is running.

But I don't have a way to solve this, because we can't globally lock sessions every time this function is running. WDYT?

itayd:
> I'm uncomfortable with the non-atomic nature of the `DeploymentHasActiveSessions` function. We can have sessions start between line 225 and line 229, and switch to "running" state between lines 212 and 220, i.e. while this function is running.
> 
> But I don't have a way to solve this, because we can't globally lock sessions every time this function is running. WDYT?

This is always happening always in a transaction.

internal/backend/db/dbgorm/deployments.go
```diff
@@ -201,3 +202,29 @@ func (db *gormdb) ListDeployments(ctx context.Context, filter sdkservices.ListDe
 		return kittehs.TransformError(ds, scheme.ParseDeployment)
 	}
 }
+
+func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.DeploymentID) (bool, error) {
+	r, err := db.ListSessions(ctx, sdkservices.ListSessionsFilter{
+		DeploymentID: id,
+		StateType:    sdktypes.SessionStateTypeRunning,
```
ig-ra: not for this PR of coz, but maybe ListSessions could receive several states somehow? e.g as a list `Running, Created`? or bit mask: `Running | Created`?

itayd: maybe in the future.

PR #753, opened 2024-10-09
fix(google): accept calendar/form IDs from web UI

daabr:
When initializing Google connections, the backend assumed that calendar and form IDs are specified only in POST requests, but the web UI sends GET requests.

All other aspects of parsing `/google/save` requests already accept GET as well as POST.

Refs: UI-700


RonenMars (approved):
LGTM, just take into account that you had Session Tests failed, better check it.

PR #754, opened 2024-10-10
feat(discord): add support for message reaction, presence and thread events

pashafateev:
**Notes:** 
- I decided to rename `authorID` because the field is specific to messages, and the term "author" doesn’t seem appropriate for all event types.
- There is no way to tell who initiated a `ThreadUpdate` event without checking the audit logs which is too involved for right now.

**Refs:** eng-1574


integrations/discord/events.go
```diff
@@ -13,15 +13,26 @@ import (
...
+		initiatorID = "" // Deleted messages don't have an author
+	case *discordgo.MessageReactionAdd:
+		initiatorID = e.UserID
+	case *discordgo.MessageReactionRemove:
+		initiatorID = e.UserID
+	case *discordgo.PresenceUpdate:
+		initiatorID = e.User.ID
+	case *discordgo.ThreadCreate:
+		initiatorID = e.OwnerID
+	case *discordgo.ThreadUpdate:
```
daabr, line 32: Clarification question: what does the thread update even mean? Is it only the creator of the thread who can update it?

integrations/discord/websocket.go
```diff
@@ -117,4 +120,10 @@ func (h handler) addHandlers(dg *discordgo.Session) {
 	dg.AddHandler(h.handleMessageCreate)
 	dg.AddHandler(h.handleMessageDelete)
 	dg.AddHandler(h.handleMessageUpdate)
+	dg.AddHandler(h.handleMessageReactionAdd)
+	dg.AddHandler(h.handleMessageReactionRemove)
+	dg.AddHandler(h.handlePresenceUpdate)
+	dg.AddHandler(h.handleThreadCreate)
+	dg.AddHandler(h.handleThreadUpdate)
+	dg.AddHandler(h.handleThreadDelete)
```
daabr: Please sort these lines alphabetically, i.e. `ThreadDelete` before `ThreadUpdate`

integrations/discord/events.go
pashafateev: Yeah, I had to look it up too because it's not obvious. `ThreadUpdate` event occurs when thread properties change. Here is an example of thread properties:
image
Discord does not provide in detail documentation and I had to find this out by asking chatGPT and then testing. chatGPT says that thread creators and privileged users can make changes to threads.

integrations/discord/websocket.go
pashafateev: Done

integrations/discord/events.go
daabr: In the case of update, I'm assuming the "owner" ID belongs to the original creator. Is there an ID of who initiated the current change, which may be different than the ID of the thread owner?

pashafateev: There is not... Confirmed on the `discordgo` discord channel. 
chatGPT says to check the audit logs which sounds like a pain... Guy on `discordgo` confirmed this.

daabr (approved):

PR #761, opened 2024-10-12
feat(ENG-1637): start new sessions from a session

itayd:
this adds the `start` syscall to python


efiShtain:
do we really need this capability? 
does the UI knows how to handle multi session stuff ?
can't we just trigger a different webhook and let it run as any other workflow ?
do we have any relation between this sessions ? how do we handle replay ? how do we handle error in the child sessions ?

efiShtain (approved):
fix formatting before merge

runtimes/pythonrt/runner/syscalls.py
```diff
@@ -36,6 +37,33 @@ def call(self, fn, args, kw):
...
+            json_data = json.dumps(data)
+        except (ValueError, TypeError) as err:
+            raise ValueError(f"start: invalid data: {err}")
+
+        try:
+            json_memo = json.dumps(memo)
+        except (ValueError, TypeError) as err:
+            raise ValueError(f"start: invalid memo: {err}")
+
+        req = pb.StartSessionRequest(runner_id=self.runner_id, location=loc, data=bytearray(json_data), memo=bytearray(json_memo))
```
tebeka:
```suggestion
        req = pb.StartSessionRequest(runner_id=self.runner_id, location=loc, data=json_data.encode(), memo=json_memo.encode())
```

tebeka (approved):

runtimes/pythonrt/runner/syscalls.py
itayd: done

PR #765, opened 2024-10-15
feat(ENG-1609): fetch web platform instead of having it in repo

itayd:
- fetch according to `web/webplatform/VERSION`.
- display version in up.
- release and docker builds are handled as well.


internal/backend/svc/banner.go
```diff
@@ -27,7 +27,7 @@ var banner string
 
 var bannerTemplate = template.Must(template.New("banner").Parse(banner))
 
-func printBanner(cfg *bannerConfig, opts RunOptions, addr, wpAddr, temporalFrontendAddr, temporalUIAddr string) {
+func printBanner(cfg *bannerConfig, opts RunOptions, addr, wpAddr, wpVersion, temporalFrontendAddr, temporalUIAddr string) {
```
daabr, line 30: You've officially crossed the threshold of "too many parameters, better use a struct".

Also, "addr" and "wp*" are unclear param names.

Also, please document this function, and mention the boolean config key to control showing this banner (`banner.show`).

I'm ok with doing this in a separate PR, but if so please do it soon-ish.

internal/backend/svc/banner.txt
```diff
@@ -1,8 +1,8 @@
 
-        _..--~~~~--.._             autokitteh   {{ .Version }}
-     _--              --_          PID:         {{ .PID }}
-    '                    `         gRPC/HTTP:   {{ .Addr }}
-   '                      `        Web:         {{ .WebAddr }}
+        _..--~~~~--.._             autokitteh    {{ .Version }}
+     _--              --_          PID:          {{ .PID }}
+    '                    `         gRPC/HTTP:    {{ .Addr }}
+   '                      `        Web{{ .WebVer }}  {{ .WebAddr }}
```
daabr, line 5: It's very easy to make this look messed up when you inject a variable-length string in the left side, especially with no space after "Web", and no ":" before the address. Maybe `Web: address (version)` instead? Or separate "Web Address" and "Web Version"?

web/webplatform/VERSION
```diff
@@ -0,0 +1 @@
+2.29.1
```
daabr, line 1: Question: any plan to make this dynamic?

web/webplatform/nothing.zip
```diff
@@ -0,0 +1 @@
+This is not really a zip file, just a placeholder to prevent go embed from failing if no distribution was downloaded.
```
daabr, line 1: But shouldn't it fail in this case?

web/webplatform/fetch.sh
```diff
@@ -0,0 +1,34 @@
...
+    exit 0
+fi
+
+echo "not downloaded yet."
+
+PREV="$(ls -1 autokitteh-web-v*.zip > /dev/null 2>&1 || true)"
+
+echo "fetching ${FILENAME}..."
+
+curl -sL "https://github.com/autokitteh/web-platform/releases/download/v${VERSION}/${FILENAME}" > "${FILENAME}"
```
daabr, line 23: Can you also read and check the file's hash?

web/webplatform/VERSION
itayd: nope. we need to be explicit.

web/webplatform/nothing.zip
itayd: nope, sometimes web platform is not required.

web/webplatform/VERSION
daabr: ok, then please add a document in the catcafe to clarify who does what, how, and when / how often.

web/webplatform/fetch.sh
itayd: doesn't look like we publish the hash on the release, we'll have to do that. @RonenMars ?

RonenMars: It's possible, and I'll implement it on my side.

Do we want the hash to be part of the file name or a separate text file?
I guess a separate text file would be a better option, right?

itayd: Separate text file. see https://github.com/temporalio/temporal/releases/tag/v1.25.1 for example, file `checksums.txt` which is the standard i think

daabr (approved):

internal/backend/svc/banner.go
itayd: in a futute pr.

internal/backend/svc/banner.txt
itayd: in a futute pr.

PR #771, opened 2024-10-16
feat: introduce aux port, place banner on api port

itayd:
- introduces aux port where dashboard is served (default 9983).
- integrations connections forms are served on main port.
- dashboard connection initialization methods (init, postInit) moved to `connectionsinitsvc` which does not belong anymore to dashboard.
- main port is considered API port, '/' displays autokitteh banner same as in console (might want to hide it on prod?).

NOTE: `AK_HTTP__SERVICE_URL` now need to be properly set. For dev default it is `localhost:9980`. For prod default is `api.autokitteh.cloud`.

image

in prod, a minimal banner is shown including only version, mode and service url.


internal/backend/httpsvc/config.go
```diff
@@ -34,49 +34,60 @@ type CORSConfig struct {
 
 type Config struct {
 	// local server address, set to run server on different port
-	Addr string `koanf:"addr"`
+	Addr    string `koanf:"addr"`
+	AuxAddr string `koanf:"aux_addr"`
```
daabr, line 38: Undocumented, and the reader as no chance to understand what this is/does on their own

internal/backend/httpsvc/config.go
```diff
@@ -34,49 +34,60 @@ type CORSConfig struct {
...
-		H2C:                  httpH2CConfig{Enable: true},
-		EnableGRPCReflection: true,
-		CORS: CORSConfig{
-			AllowedOrigins:   []string{"*"},
-			AllowedMethods:   []string{"GET", "POST", "PUT", "DELETE", "OPTIONS", "PATCH", "HEAD"},
-			AllowedHeaders:   []string{"*"},
-			AllowCredentials: true,
+var defaultConfg = Config{
+	Addr:                 "0.0.0.0:" + sdkclient.DefaultPort,
+	AuxAddr:              "0.0.0.0:9983",
```
daabr, line 61: 1. We use 9982 after 9980 because 9981 is reserved. 9983 is free, but why not continue with the same skip and use 9984?

2. Either way, ports should be configurable as integers, not full-address strings.

3. Why does 9980 have an `sdkclient` const and this one doesn't?

internal/backend/httpsvc/http.go
```diff
@@ -86,30 +96,89 @@ func New(lc fx.Lifecycle, z *zap.Logger, cfg *Config, reflectors []string, extra
...
+
+			return nil
+		},
+		OnStop: func(ctx context.Context) error {
+			l.Named("main_server").Info("shutting down")
+			return mainServer.Shutdown(ctx)
+		},
+	})
+
+	interceptedAuxMux := http.NewServeMux()
```
daabr, line 126: Missing documentation for the new code below

internal/backend/muxes/muxes.go
```diff
@@ -2,4 +2,16 @@ package muxes
 
 import "net/http"
 
-type Muxes struct{ Auth, NoAuth *http.ServeMux }
+type (
```
daabr, line 5: In general, “main” and “aux” are meaningless jargon, I don’t think anyone will see “aux” and think “ah! I know what this means/does!” or understand the logic of what is/isn’t “main”.

From an architecture perspective, we started with a single mux, then we duplicated it, then we merged the pair, and now we’re creating a tree of muxes - this is too complex to maintain by median developers, so that’s not healthy or safe.

internal/backend/oauth/webhook.go
```diff
@@ -25,8 +25,12 @@ type svc struct {
...
 func InitWebhook(l *zap.Logger, muxes *muxes.Muxes, c sdkservices.Services) {
 	s := &svc{logger: l, svcs: c}
-	muxes.Auth.HandleFunc("GET /oauth/start/{intg}", s.start)
-	muxes.NoAuth.HandleFunc("GET /oauth/redirect/{intg}", s.exchange)
+
+	muxes.Main.Auth.HandleFunc("GET /oauth/start/{intg}", s.start)
+	muxes.Main.NoAuth.HandleFunc("GET /oauth/redirect/{intg}", s.exchange)
+
+	muxes.Aux.Auth.HandleFunc("GET /oauth/start/{intg}", s.start)
+	muxes.Aux.NoAuth.HandleFunc("GET /oauth/redirect/{intg}", s.exchange)
```
daabr, line 33: See my comment above about the excessive complexity of mutex trees. This specifically is an example of a foot gun: which is the right mutex to use, given that both support the same URLs?

Please document what it means/does that both ports support identical URLs.

internal/backend/svc/banner.go
```diff
@@ -3,82 +3,139 @@ package svc
...
+	"go.autokitteh.dev/autokitteh/internal/backend/temporalclient"
+	"go.autokitteh.dev/autokitteh/internal/backend/webplatform"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/internal/version"
 )
 
 type bannerConfig struct {
-	Show bool `koanf:"show"`
+	Show    bool `koanf:"show"`
+	Minimal bool `koanf:"minimal"`
```
daabr, line 24: First document this config section in "Server Config" in Docusaurus, as requested in the past. Then we could discuss adding more config values.

internal/backend/svc/banner.txt
```diff
@@ -1,13 +1,13 @@
...
-    '                    `         gRPC/HTTP:    {{ .Addr }}
-   '                      `        Web{{ .WebVer }}  {{ .WebAddr }}
-  '                        `       {{ .Mode }}
- |                    |\   |       
- |                    | \  |       {{ .Temporal }}
- |          .,    _,-~   \ |       {{ .TemporalUI }}
-  \         \ `.-'        \/ 
+        _..--~~~~--.._             autokitteh     {{ .Version }}
+     _--              --_          PID:           {{ .PID }}
+    '                    `         gRPC/HTTP API: {{ .Addr }}
```
daabr, line 4: Web version: commented before but not fixed yet: you're making assumptions about the web ver string that are neither obvious nor stable.

internal/backend/svc/banner.txt
```diff
@@ -1,13 +1,13 @@
...
+   '                      `        Web{{ .WebVer }}   {{ .WebAddr }}
+  '                        `       Aux:           {{ .AuxAddr }}
```
daabr, line 6: 99% of users: "WTF is Aux? Is that a new protocol I never heard of? Oh, it's a web UI/API which isn't *the* web UI, which uses HTTP but isn't on *the* HTTP port?"

internal/backend/svc/svc.go
```diff
@@ -303,34 +326,35 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
...
-			fx.Invoke(func(cfg *bannerConfig, lc fx.Lifecycle, z *zap.Logger, httpsvc httpsvc.Svc, tclient temporalclient.Client, wp *webplatform.Svc) {
-				HookSimpleOnStart(lc, func() {
-					temporalFrontendAddr, temporalUIAddr := tclient.TemporalAddr()
-					printBanner(cfg, opts, httpsvc.Addr(), wp.Addr(), wp.Version(), temporalFrontendAddr, temporalUIAddr)
+			fx.Invoke(func(lc fx.Lifecycle, b banner, muxes *muxes.Muxes) {
+				if b.Cfg.Show {
+					HookSimpleOnStart(lc, func() { b.Print(os.Stderr, opts, false) })
+				}
+
+				muxes.Main.NoAuth.HandleFunc("/{$}", func(w http.ResponseWriter, r *http.Request) {
```
daabr, line 341: Please document what this handler is/does

daabr (commented):
In general, I think this PR is more brittle and complex than the current state. So (1) it’s a step back, not forward, (2) it’s too risky to merge right now.

Regardless of timing - it feels like over-engineering when what you really want is a single flag to disable the dashboard (by default in production?).

With relation to timing - based on past experience, it might take all of us weeks to find, debug, and fix issues resulting from this PR, and we can’t afford this especially now.

Is there a compelling reason to focus on this right now? Can you share more info on what and how much testing did you do on this PR?

internal/backend/connectionsinitsvc/svc.go
```diff
@@ -0,0 +1,157 @@
...
+		return
+	}
+
+	var u string
+	switch origin {
+	case "vscode":
+		u = "vscode://autokitteh.autokitteh?cid=%s"
+	default:
+		// Another redirect just to get rid of the secrets in the URL.
+		u = "/connections/%s/success"
```
daabr, line 106: Isn't this one of the result pages that you deleted? I'm not sure.

internal/backend/httpsvc/config.go
```diff
@@ -34,49 +34,60 @@ type CORSConfig struct {
...
+	ServiceURL:           sdkclient.DefaultCloudURL,
```
daabr, line 62: `api.autokitteh.cloud` belongs to our cloud, and only our cloud. It is inappropriate to use it as a default value in an open-source project.

daabr:
> * main port is considered API port, '/' displays autokitteh banner same as in console (might want to hide it on prod?).

The banner contains very helpful hints for attackers when they fingerprint the server - in other words, serving it over HTTP is a security risk, even in non-prod, it should never be a default

itayd:
> In general, I think this PR is more brittle and complex than the current state. So (1) it’s a step back, not forward, (2) it’s too risky to merge right now.

I don't think it's too brittle, but omv.
 
> Regardless of timing - it feels like over-engineering when what you really want is a single flag to disable the dashboard (by default in production?).

I don't want to disable it - I want to still be able to use it via VPN. also to add more stuff to the aux port such as Temporal codec et al.

> 
> With relation to timing - based on past experience, it might take all of us weeks to find, debug, and fix issues resulting from this PR, and we can’t afford this especially now.
> 
> Is there a compelling reason to focus on this right now? Can you share more info on what and how much testing did you do on this PR?

This is all for discussing in sync.

efiShtain:
I tend to agree with Daniel here, the aux/main names are meaningless
we can go with public / internal / api prefixes if you want to make more sense

also, we agreed not to invest in this internal dashboard and we keep investing in it

I prefer one of these options:
- split the dashboard web server to be served as a totally different process that only serves the dashboard, It can contact ak using the API. This will also get us in the way of splitting it from the code

- Disable it in prod totally, run it locally to work with the same DB and temporal over VPN in case you need it

- Try some cidr block logic to allow access to specific endpoints only from the VPC internal subnets. It's a code change but very limited and might affect only this dashboard

- Route the load balancer with high priority rule for /dashboard so it will point to error page or something - no code changes at all

PR #774, opened 2024-10-17
feat(python): generate JWTs without exposing the private key

daabr:
Refs: ENG-1391


runtimes/pythonrt/runner/syscalls.py
```diff
@@ -26,9 +26,9 @@ def __init__(self, runner_id, worker):
         self.ak_funcs = {
             "next_event": self.ak_next_event,
             "sleep": self.ak_sleep,
+            "start": self.ak_start,
```
itayd, line 29: daniel

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -373,6 +375,37 @@ func (s *workerGRPCHandler) Unsubscribe(ctx context.Context, req *pb.Unsubscribe
 	}
 }
 
+func (s *workerGRPCHandler) EncodeJWT(ctx context.Context, req *pb.EncodeJWTRequest) (*pb.EncodeJWTResponse, error) {
+	// GitHub's JWTs must be signed using the RS256 algorithm:
+	// https://docs.github.com/en/apps/creating-github-apps/authenticating-with-a-github-app/generating-a-json-web-token-jwt-for-a-github-app
+	if req.Algorithm != jwt.SigningMethodRS256.Name {
```
itayd, line 381: why do you even pass it in in the first place oh well

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -373,6 +375,37 @@ func (s *workerGRPCHandler) Unsubscribe(ctx context.Context, req *pb.Unsubscribe
...
+	}
+
+	claims := jwt.MapClaims{}
+	for key, value := range req.Payload {
+		claims[key] = value
+	}
+
+	token := jwt.NewWithClaims(jwt.SigningMethodRS256, claims)
+
+	pem, ok := os.LookupEnv("GITHUB_PRIVATE_KEY")
```
itayd, line 392: are you sure this is available at the runner?

itayd (approved):
some questions, but nothing blocking

itayd (approved):

runtimes/pythonrt/worker_grpc_handler.go
daabr: in case the value is incorrect - we should get a clear error message.
Why check here and not before? To keep the passthrough code as generic as possible, in case this criterion changes in the future.

runtimes/pythonrt/worker_grpc_handler.go
daabr: As long as the runner is a part of the AK server - yes. This is a server-wide setting, not connection/runner-specific.

itayd: fyi @efiShtain

PR #775, opened 2024-10-17
fix(python): deal with errors gracefully

itayd:
- currently when there is an error inside an activity, python attempts to make a print. when this happens there is a deadlock since `initialCall` blocks the select. the fix is to accumulate the prints and logs and emit them all together when this is done.
- traces are not emitted in a single print, which should make @daabr a happier person.
- errors look much cleaner now, no non-user related gibrish.

Refs: ENG-1700

### DIFF
--- a/runtimes/pythonrt/worker_grpc_handler.go
+++ b/runtimes/pythonrt/worker_grpc_handler.go
@@ -104,6 +104,9 @@ func (s *workerGRPCHandler) Log(ctx context.Context, req *pb.LogRequest) (*pb.Lo
 	m := &logMessage{level: req.Level, message: req.Message}
 
 	if !runner.firstCall {
+		// If we're not in the first call, the prints channel is not listened on since
+		// the select loop is stuck on the initialCall function.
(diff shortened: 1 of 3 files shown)


daabr (approved):

daabr (changes requested):
I tested the branch manually with a workflow that crashes: but no errors are printed, and the session state is "COMPLETED". In other words, this PR makes errors disappear which is worse, not better.

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -104,6 +104,9 @@ func (s *workerGRPCHandler) Log(ctx context.Context, req *pb.LogRequest) (*pb.Lo
 	m := &logMessage{level: req.Level, message: req.Message}
 
 	if !runner.firstCall {
+		// If we're not in the first call, the prints channel is not listened on since
+		// the select loop is stuck on the initialCall function.
+		// Cccumulate the prints which will be reported when the call is done.
```
daabr: Typo

daabr (approved):

PR #777, opened 2024-10-17
fix(google): reduce sensitivity of requested Drive scopes

daabr:
Refs: ENG-1701


internal/backend/oauth/oauth.go
```diff
@@ -182,7 +183,7 @@ func New(l *zap.Logger) sdkservices.OAuth {
 					forms.FormsResponsesReadonlyScope,
 					sheets.SpreadsheetsScope,
 					// Restricted.
-					drive.DriveScope,
+					// drive.DriveScope, // See ENG-1701
```
itayd, line 186: not sure i understand the scopes, can you elaborate?

itayd (approved):

PR #781, opened 2024-10-20
feat: install pysdk directly from code instead of pypi

efiShtain:


runtimes/pythonrt/local_runner.go
```diff
@@ -365,16 +368,30 @@ func createVEnv(pyExe string, venvPath string) error {
...
+		return err
+	}
+
+	if err = install(venvPy, path.Join(tmpDir, "/py-sdk"), []string{"-m", "pip", "install", "."}); err != nil {
+		return fmt.Errorf("install autokitteh py sdk %w", err)
+	}
+	return nil
+}
+
+func install(pyPath, cwd string, args []string) error {
```
tebeka, line 387: Rename to `runInDir` or something like that?

tebeka (approved):
One very minor comment

runtimes/pythonrt/local_runner.go
efiShtain: I think working directory is standard term in this command line execution stuff

PR #784, opened 2024-10-21
feat: acceptance test for basic usability of integrations

daabr:
This can be used for acceptance tests for new environments and auth types, verification of connection UIs and flows, misc Python bug reproductions, etc.

It is meant to be simpler and quicker than running all the samples one after the other - a single deployment, trivial API calls without runtime parameters, and no extra logic in triggers. In other words, is checks that the rest of the projects are ready to be tested.

It's not meant to offer the same functionality coverage or educational value as other projects in this repo.

FYI, some functionalities are currently broken in the cloud platform, but are working in on-prem servers (e.g. #774 and #779 not deployed in prod, and ENG-1680).

### DIFF
--- a/tests/acceptance/README.md
+++ b/tests/acceptance/README.md
@@ -1,0 +1,42 @@
+# Acceptance Test Project
+
+This AutoKitteh project is used as a testing tool for:
+
+- Acceptance of new cloud environments
+- Connection auth types, UIs, and flows
+- Python runtime issues involving connections
+
+It is meant to be simpler and quicker than running all the projects in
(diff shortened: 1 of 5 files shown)


itayd (approved):
I brief README would be nice

itayd:
also maybe a better path is `tests/projects/auth`?

daabr:
> I brief README would be nice

Done.

> also maybe a better path is `tests/projects/auth`?

I'm keeping the current path for now. For me `tests/projects` means "tests of projects", not "tests with projects". But this is easy to change later.

PR #786, opened 2024-10-22
refactor: IsRetryableError

itayd:
simple fix to not reuse a global var. in prep for https://linear.app/autokitteh/issue/ENG-1721/review-activities-and-use-temporalnewnonretryableapplicationerror


itayd:
efi approved

PR #787, opened 2024-10-22
testing: print tests timing info in ci

itayd:
will give us more info to debug sporadic cancellations of actions


efiShtain:
you can just show the timestamps of everything, no need to add code

Screenshot 2024-10-22 at 11 01 53

PR #788, opened 2024-10-22
refactor: pull connection init logic out of dashboard

itayd:
we need it even when we deprecate/hide the dashboard.


efiShtain (approved):
not sure why these are not standard GRPC endpoints with connect making them http endpoints as well

but it doesn't block merging it

approved

efiShtain:
Does this require UI sync ? if so, we need to make sure this is a breaking change version so we won't deploy it without the updated UI

internal/backend/connectionsinitsvc/svc.go
```diff
@@ -0,0 +1,157 @@
...
+	if err := json.NewEncoder(w).Encode(st); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	http.Redirect(w, r, fmt.Sprintf("/connections/%s", id), http.StatusFound)
+}
+
+func (s Svc) success(w http.ResponseWriter, _ *http.Request) {
+	fmt.Fprint(w, "success - you can now close this tab.")
```
daabr, line 156: In case we can't close the tab, this message is... underwhelming.

It's preferable to return an HTML page like before, not just a text string. And then it would be up to Ronen to define the look & feel of this page.

internal/backend/connectionsinitsvc/svc.go
```diff
@@ -0,0 +1,157 @@
...
+		return v.WithScopeID(sdktypes.NewVarScopeID(cid))
+	})
+
+	if err := s.svcs.Vars.Set(r.Context(), data...); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
+		return
+	}
+
+	var u string
+	switch origin {
```
daabr, line 101: 1. Switch with one option + default --> if/else
2. Please explain: how are CLI invocations of "ak connection init" handled now? (They used to have a different URL target that used the dashboard, what's the new behavior?)

itayd:
> not sure why these are not standard GRPC endpoints with connect making them http endpoints as well

need sometimes oauth redirect and such, cannot be grpc

itayd:
> Does this require UI sync ? if so, we need to make sure this is a breaking change version so we won't deploy it without the updated UI

nope, all routes are the same

internal/backend/connectionsinitsvc/svc.go
itayd: in case of redir from vscode, we do a redir back there.

other than that i don't mind looks rn.

internal/backend/connectionsinitsvc/svc.go
itayd: > 2. Please explain: how are CLI invocations of "ak connection init" handled now? (They used to have a different URL target that used the dashboard, what's the new behavior?)

the URL did not change

itayd: > 1. Switch with one option + default --> if/else

i think a switch is a good fit here - we might add options, also this is very readable.

daabr: This PR changes the SaaS init flow because the case for it - that was the point of my question. Since it's already deployed I'll open a bug on this.

PR #790, opened 2024-10-22
fix: timeout is ignored on next_event in pythonrt

efiShtain:


runtimes/pythonrt/runner/syscalls.py
```diff
@@ -100,10 +100,17 @@ def ak_subscribe(self, args, kw):
         return resp.signal_id
 
     def ak_next_event(self, args, kw):
-        (id,) = extract_args(["subscription_id"], args, kw)
+        log.info("ak_next_event: %r %r", args, kw)
+        (
+            id,
+            timeout,
+        ) = extract_args(["subscription_id", "timeout?"], args, kw)
```
daabr, line 107:
```suggestion
        id, timeout = extract_args(["subscription_id", "timeout?"], args, kw)
```

daabr, line 107: In `py-sdk/events.py`, the type of the ID is `str` - which your code supports - but I'm sure Miki wrote somewhere that it can also be a variadic sequence of ID strings (e.g. `next_event("conn1", "conn2", ...)` in order to get the first event from any number of unrelated connections - and IIUC this is not supported in this PR? I'm not even sure `extract_args` supports it before your PR... This is something to check with Miki unrelated to this PR

See https://github.com/autokitteh/docs.autokitteh.com/pull/124 (not sure why it wasn't auto-deployed when the PR was merged)

runtimes/pythonrt/runner/syscalls.py
```diff
@@ -100,10 +100,17 @@ def ak_subscribe(self, args, kw):
...
         if not id:
             raise ValueError("empty subscription_id")
+
         req = pb.NextEventRequest(runner_id=self.runner_id, signal_ids=[id])
+        if timeout:
+            req.timeout_ms = int(timeout * 1000)
```
daabr, line 113: In `py-sdk/events.py`, the type of the timeout is `timedelta`, not `int`, which actually makes sense.

BTW, unrelated to your PR, do you think we should make `timeout` a kwarg-only parameter? (i.e. `def next_event(id, *, timeout)`)

runtimes/pythonrt/runner/syscalls.py
efiShtain: formatting is automatic, not my decision

efiShtain: please open a ticket as this is a different issue
this PR only solves the timeout not being passed

runtimes/pythonrt/runner/syscalls.py
efiShtain: @tebeka what's your take on it ?
I'm ok with both options

daabr (approved):
The only thing to fix is the expected type of `timeout` (`timedelta` instead of `int`)

runtimes/pythonrt/runner/syscalls.py
daabr: I see in https://github.com/autokitteh/autokitteh/blob/main/runtimes/pythonrt/py-sdk/autokitteh/events.py#L21 that Miki already added `*` before `timeout` - so the BTW question is already answered

efiShtain:
Merging after talking with @daabr 
the time delta / int issue will be solved by @daabr  / @tebeka

PR #791, opened 2024-10-23
refactor: remove unused function from kittehs

daabr:


sdk/sdkclients/sdkclients.go
```diff
@@ -77,3 +78,18 @@ func (c *client) Sessions() sdkservices.Sessions         { return c.sessions() }
 func (c *client) Store() sdkservices.Store               { return c.store() }
 func (c *client) Triggers() sdkservices.Triggers         { return c.triggers() }
 func (c *client) Vars() sdkservices.Vars                 { return c.vars() }
+
+// lazyCache wraps a function and a single input. The first call to the wrapper
+// calls the wrapped function. Subsequent calls return the first result.
+func lazyCache[T, P any](f func(P) T, p P) func() T {
```
itayd: why not keep it in kittehs?

daabr: because it's used only once throughout all of AK - so it's not really a utility function. Anyway, the rename from "lazy1" & doc header is more important to me than the function's location.

itayd: ok, i think it's useful enough to be in kittehs. can you do the doc there?

daabr: done

itayd (approved):

PR #792, opened 2024-10-23
fix: don't allow Atlassian OAuth connections to run API calls

daabr:
This is until we synchronize new refresh tokens across sessions via connection vars, so OAuth tokens can be refreshed more than once (because Atlassian OAuth refreshes change the refresh token too).

Refs: ENG-1655


itayd:
we can't do it right now since nsure is using it. discuss with joey first.

itayd (approved):

PR #794, opened 2024-10-23
fix(ENG-1715): remove schedule when its trigger is kaput

itayd:


internal/backend/scheduler/scheduler.go
```diff
@@ -109,11 +113,29 @@ func (sch *Scheduler) Update(ctx context.Context, tid sdktypes.TriggerID, schedu
 func (sch *Scheduler) workflow(wctx workflow.Context, tid sdktypes.TriggerID) error {
 	sl := sch.sl.With("trigger_id", tid)
 
-	event := sdktypes.NewEvent(tid).WithType("tick")
-
 	ctx := temporalclient.NewWorkflowContextAsGOContext(wctx)
 
-	eid, err := sch.dispatcher.Dispatch(ctx, event, nil)
+	// It's ok to run all these straight in the workflow - they're all should be
```
efiShtain: disagree
under load it can take time
and it will cause dead lock fails

should be in an activity, same as the delete schedule

internal/backend/scheduler/scheduler.go
```diff
@@ -109,11 +113,29 @@ func (sch *Scheduler) Update(ctx context.Context, tid sdktypes.TriggerID, schedu
...
+		sl.Warnf("trigger %v not found, removing schedule", tid)
+
+		if err := sch.Delete(ctx, tid); err != nil {
+			return fmt.Errorf("delete schedule: %w", err)
+		}
+
+		return nil
+	}
+
+	eid, err := sch.dispatcher.Dispatch(ctx, sdktypes.NewEvent(tid).WithType("tick"), nil)
```
efiShtain, line 150: this is slow, how does it even work in the workflow itself ?

itayd: it's not slow - it's starting a new workflow.

internal/backend/scheduler/scheduler.go
itayd: if it's stuck, there'll be panic and the workflow will retry. but i get it - will turn these into activities.

itayd: fixed

efiShtain (approved):
Looks ok

But that's now another worker we have, meaning it's harder to maintain how many concurrent workflows / activities we have in parallel

itayd:
> But that's now another worker we have, meaning it's harder to maintain how many concurrent workflows / activities we have in parallel

We already had a worker there. It's just using the workflow to start an activity as well.

PR #795, opened 2024-10-23
fix: next_event py signature using int instead of timedelta

itayd:
same as sleep

### DIFF
--- a/runtimes/pythonrt/py-sdk/autokitteh/events.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/events.py
@@ -18,7 +18,7 @@ def unsubscribe(subscription_id: str) -> None:
     pass
 
 
-def next_event(subscription_id: str, *, timeout: timedelta = None) -> AttrDict:
+def next_event(subscription_id: str, *, timeout: int = None) -> AttrDict:
     """Get the next event from the subscription."""
     # Dummy implementation for local development.
     return AttrDict()


runtimes/pythonrt/py-sdk/autokitteh/events.py
```diff
@@ -18,7 +18,7 @@ def unsubscribe(subscription_id: str) -> None:
     pass
 
 
-def next_event(subscription_id: str, *, timeout: timedelta = None) -> AttrDict:
```
daabr, line 21: All of our docs and samples refer to using a `timedelta` and not an `int`. The main reason is that `int` doesn't have a clear meaning - is it seconds? milliseconds? I was going to fix it in the other direction after #790 anyway.

tebeka (changes requested):
I think timeout should be timedelta. See https://github.com/autokitteh/autokitteh/pull/797

runtimes/pythonrt/py-sdk/autokitteh/events.py
daabr: See https://github.com/autokitteh/autokitteh/pull/797

itayd: @tebeka @daabr i get it, but isn't it a bit tedious to use? sleep is using int. up to you.

PR #796, opened 2024-10-23
fix: race condition in python rt print

itayd:
simplified mechanism, no need for additional accumulator for print and logs. just using a larger buffered channel without checking the done channel.

couldn't test it yet, ran into some trouble. @efiShtain if you can test it out that'll be great. if not I'll do it tomorrow.

### DIFF
--- a/runtimes/pythonrt/pythonrt.go
+++ b/runtimes/pythonrt/pythonrt.go
@@ -74,8 +74,6 @@ type pySvc struct {
 	runnerID string
 
 	firstCall bool // first call is the trigger, other calls are activities
-	accPrints []*logMessage
-	accLogs   []*logMessage
 
 	channels comChannels
 
@@ -156,8 +154,8 @@ func newSvc(cfg *Config, l *zap.Logger) (sdkservices.Runtime, error) {
 			done:     make(chan *pb.DoneRequest, 1),
 			err:      make(chan string, 1),
 			request:  make(chan *pb.ActivityRequest, 1),
-			print:    make(chan *logMessage, 1),
-			log:      make(chan *logMessage, 1),
+			print:    make(chan *logMessage, 1024),
+			log:      make(chan *logMessage, 1024),
 			callback: make(chan *callbackMessage, 1),
 		},
 	}
@@ -469,10 +467,6 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
 			if healthErr != nil {
 				return sdktypes.InvalidValue, sdkerrors.NewRetryableError("runner health: %w", healthErr)
 			}
-		case r := <-py.channels.log:
-			py.accLogs = append(py.accLogs, r)
-		case r := <-py.channels.print:
-			py.accPrints = append(py.accPrints, r)
 		case r := <-py.channels.request:
 			var (
 				fnName = "pyFunc"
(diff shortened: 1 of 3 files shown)


runtimes/pythonrt/pythonrt.go
```diff
@@ -522,21 +516,19 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
...
 func (py *pySvc) emitPrintsAndLogs(ctx context.Context) {
-	for ; len(py.accPrints) > 0; py.accPrints = py.accPrints[1:] {
-		p := py.accPrints[0]
-		py.cbs.Print(ctx, py.runID, p.message)
+	for r := range py.channels.print {
+		py.cbs.Print(ctx, py.runID, r.message)
 
-		if p.doneChannel != nil {
-			close(p.doneChannel)
+		if r.doneChannel != nil {
```
efiShtain, line 522: why do you close the channel inside the loop?

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -91,6 +91,33 @@ func (s *workerGRPCHandler) IsActiveRunner(ctx context.Context, req *pb.IsActive
 	return &pb.IsActiveRunnerResponse{}, nil
 }
 
+func pushMsg(initialCall bool, ch chan *logMessage, m *logMessage) error {
+	if initialCall {
```
efiShtain, line 95: initial call is implementation details of the pythonrt run 
please keep it there and not in this handler side

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -91,6 +91,33 @@ func (s *workerGRPCHandler) IsActiveRunner(ctx context.Context, req *pb.IsActive
...
+		// If we're not in the first call, the prints channel is not listened on since
+		// the select loop is stuck on the initialCall function.
+		// Accumulate the prints which will be reported when the call is done.
+
+		select {
```
efiShtain, line 100: so technically no prints / logs until the session is done ?

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -91,6 +91,33 @@ func (s *workerGRPCHandler) IsActiveRunner(ctx context.Context, req *pb.IsActive
...
+		case ch <- m:
+			return nil
+		case <-time.After(time.Second):
+			// just to check print is not full.
```
efiShtain, line 104: and if it's full ? you kill the workflow ?

runtimes/pythonrt/pythonrt.go
```diff
@@ -156,8 +154,8 @@ func newSvc(cfg *Config, l *zap.Logger) (sdkservices.Runtime, error) {
 			done:     make(chan *pb.DoneRequest, 1),
 			err:      make(chan string, 1),
 			request:  make(chan *pb.ActivityRequest, 1),
-			print:    make(chan *logMessage, 1),
-			log:      make(chan *logMessage, 1),
+			print:    make(chan *logMessage, 1024),
```
tebeka, line 157: Why 1024? (vs any other number)

tebeka, line 157: I don't see any code sending to the `print` channel. @efiShtain ?

runtimes/pythonrt/pythonrt.go
```diff
@@ -522,21 +516,19 @@ func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktyp
 }
 
 func (py *pySvc) emitPrintsAndLogs(ctx context.Context) {
-	for ; len(py.accPrints) > 0; py.accPrints = py.accPrints[1:] {
-		p := py.accPrints[0]
-		py.cbs.Print(ctx, py.runID, p.message)
+	for r := range py.channels.print {
```
tebeka, line 519: BUG: This will block until someone closes the channel, and I don't see any code closing this channel.

runtimes/pythonrt/pythonrt.go
tebeka: This is per item in the print channel to notify the print was handled. 
But, closing a closed or nil channel will panic, there's a race condition here if some other goroutine closes this. Not sure it'll happen though since here it's the only code that need to do that.

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -91,6 +91,33 @@ func (s *workerGRPCHandler) IsActiveRunner(ctx context.Context, req *pb.IsActive
...
+	}
+
+	m.doneChannel = make(chan struct{})
+
+	ch <- m
+
+	select {
+	case <-m.doneChannel:
+		return nil
+	case <-time.After(runnerChResponseTimeout):
```
tebeka, line 116: Related to @efiShtain's comment above. Not sure the right thing is to err if we can't add a print log.

tebeka (changes requested):
In general, the amount of effort we spend on prints shows we have a serious design flaw here. It all started from the restriction that a print will be from the same goroutine that started the `Call`. If this restriction is removed a lot of hairy code can be deleted.

itayd:
> In general, the amount of effort we spend on prints shows we have a serious design flaw here. It all started from the restriction that a print will be from the same goroutine that started the `Call`. If this restriction is removed a lot of hairy code can be deleted.

The root issue here is that it must be deterministic, not just from the same go routine.

itayd:
replaced by https://github.com/autokitteh/autokitteh/pull/799
@tebeka @efiShtain ptal

PR #799, opened 2024-10-24
fix: race condition in python rt print

itayd:
much simpler approach

replaces https://github.com/autokitteh/autokitteh/pull/796


efiShtain:
isn't this the original solution that we had before ?

itayd:
> isn't this the original solution that we had before ?

you didn't have the go statement that listens for the channels in the Execute method when it's not an initialCall.

efiShtain (approved):

PR #800, opened 2024-10-24
feat: compare checksums when downloading web-platform; bump ver

itayd:


web/webplatform/VERSION
```diff
@@ -1 +1 @@
-2.29.3
+2.37.2 dfc6f412248826db2a070e9c5bbece778a69185224636deadc084d19b5fbc7ab
```
daabr: The point of a checksum is that no one is allowed to specify the expected value other than the trusted source:

You download the expected checksum, and compare it to the actual checksum of the versioned tar you have/downloaded - and if they match you know the tar file wasn't tampered.

If you don't download the checksum, whoever tampers with the local tar file can also tamper with the version file.

web/webplatform/fetch.sh
```diff
@@ -4,28 +4,45 @@ set -euo pipefail
...
 
-echo "want version ${VERSION}"
+echo "want version ${VERSION}."
+
+checksum() {
+    echo "verifying checksum..."
+
+    sha=$(shasum -a 256 "$1" | cut -d ' ' -f 1)
+
+    if [[ "${sha}" != "${SHA}" ]]; then
```
daabr: This is hard to understand, please use clearer and unique var names like "expected" and "actual" instead of "SHA" and "sha".

And please use a single convention for var names: either "FOO" or "foo".

web/webplatform/VERSION
itayd: i'm not sure we're on the same page. a trusted source updates the VERSION file, that way you cannot hijack the link or whatever. This change also checks against the current file even if not download happened btw.

web/webplatform/fetch.sh
itayd: fixed

itayd: oh, re convention - "sha" is a "local" var, others are inputs/globals/whatever

web/webplatform/VERSION
itayd: talked to @tebeka , opined that this is the way for extra paranoids, so we're good.

efiShtain:
this way is more secure than fetching the checksum from GitHub
so it's fine by me
we can reiterate later if we want

efiShtain (approved):

PR #803, opened 2024-10-25
refactor: wait logic in pythonrt

itayd:
while messing around with it to debug i refactored the wait logic. it can now be responsive to cancels and have configurable waiting time. 

not a must.

### DIFF
--- a/runtimes/pythonrt/config.go
+++ b/runtimes/pythonrt/config.go
@@ -1,6 +1,10 @@
 package pythonrt
 
-import "go.autokitteh.dev/autokitteh/internal/backend/configset"
+import (
+	"time"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/configset"
+)
 
 type Config struct {
 	RemoteRunnerEndpoints []string `koanf:"remote_runner_endpoints"`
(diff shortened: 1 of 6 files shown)


runtimes/pythonrt/config.go
```diff
@@ -13,17 +17,22 @@ type Config struct {
...
 
 var Configs = configset.Set[Config]{
 	Default: &Config{
-		RunnerType: "local",
+		RunnerType:    "local",
+		StartDuration: 20 * time.Second,
 	},
 	Test: &Config{
 		LazyLoadLocalVEnv: true,
+		StartDuration:     20 * time.Second,
```
efiShtain, line 31: if you set the default to 20, no need to set the others as well

runtimes/pythonrt/config.go
```diff
@@ -13,17 +17,22 @@ type Config struct {
 	LogBuildCode      bool `koanf:"log_build_code"`
 
 	RunnerType string `koanf:"runner_type"`
+
+	StartDuration time.Duration `koanf:"start_duration"`
```
efiShtain, line 21: something like MAX_RUNNER_START_TIMOUET suits this use case more than start duration

runtimes/pythonrt/runner_manager.go
```diff
@@ -48,37 +49,56 @@ type Healther interface {
...
 	}
 
-	return fmt.Errorf("%s not ready after %v", name, timeout)
+	tmo := time.After(timeout)
+
+	const retryInterval = 50 * time.Millisecond
+
+	startTime := time.Now()
+
+	for {
```
efiShtain, line 63: I think you can refactor this in a more readable way as a loop with max retries like 5 or 10 or whatever configurable times instead of multiple timeouts

runtimes/pythonrt/runner_manager.go
```diff
@@ -48,37 +49,56 @@ type Healther interface {
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-tmo:
+			return errors.New("timeout waiting for " + name + " server")
```
efiShtain, line 68: name is always runner, it makes the error not very informative in case we have many runners going on

PR #804, opened 2024-10-25
fix: properly classify activity errors

itayd:
this enables as to discern between retryable and not retryable activity errors. it is probable that not all our methods return proper errors, so we'll have to treat these when we bump into them.

this also makes the error display in temporal more understandable.

### DIFF
--- a/sdk/sdkerrors/errors.go
+++ b/sdk/sdkerrors/errors.go
@@ -10,15 +10,15 @@ import (
 
 var (
 	ErrNotImplemented     = errors.New("not implemented")
-	ErrRPC                = errors.New("rpc")
 	ErrAlreadyExists      = errors.New("already exists")
 	ErrNotFound           = errors.New("not found")
 	ErrConflict           = errors.New("conflict")
 	ErrUnauthorized       = errors.New("unauthorized")
 	ErrUnauthenticated    = errors.New("unauthenticated")
-	ErrUnknown            = errors.New("unknown")
+	ErrUnknown            = NewRetryableErrorf("unknown")
 	ErrFailedPrecondition = errors.New("failed precondition")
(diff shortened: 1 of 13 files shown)


sdk/sdkerrors/errors.go
```diff
@@ -97,7 +93,36 @@ func AsConnectError(err error) error {
...
 	case errors.Is(err, ErrFailedPrecondition):
 		return connect.NewError(connect.CodeFailedPrecondition, err)
+	case errors.Is(err, ErrProgram):
+		fallthrough
 	default:
 		return connect.NewError(connect.CodeUnknown, err)
 	}
 }
+
+func ErrorType(err error) string {
```
tebeka, line 119: I prefer that each error (e.g. `ErrNotFound`) will implement a method (say `Type() string`). Then you can do something like:

```go
type Typer interface {
	Type() string
}

func ErrorType(e error) string {
	if t, ok := e.(Typer); ok {
		return t.Type()
	}

	return "unknown error"
}
```

sdk/sdktypes/program_error.go
```diff
@@ -95,6 +96,11 @@ func WrapError(err error) ProgramError {
 
 type programError ProgramError
 
+var _ error = programError{}
```
tebeka, line 99: Delete

tebeka (changes requested):

sdk/sdktypes/program_error.go
itayd: porque?

sdk/sdkerrors/errors.go
itayd: cool, fixed. ptal

tebeka (approved):

PR #805, opened 2024-10-26
feat(python): next_event support both timedelta and int

itayd:
i think will be less confusing.

### DIFF
--- a/runtimes/pythonrt/py-sdk/autokitteh/events.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/events.py
@@ -18,7 +18,9 @@ def unsubscribe(subscription_id: str) -> None:
     pass
 
 
-def next_event(subscription_id: str | list, *, timeout: timedelta = None) -> AttrDict:
+def next_event(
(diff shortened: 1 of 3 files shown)


efiShtain (approved):
fix the else / raise issue

itayd:
> fix the else / raise issue

done

runtimes/pythonrt/py-sdk/autokitteh/events.py
```diff
@@ -18,7 +18,9 @@ def unsubscribe(subscription_id: str) -> None:
     pass
 
 
-def next_event(subscription_id: str | list, *, timeout: timedelta = None) -> AttrDict:
+def next_event(
+    subscription_id: str | list[str], *, timeout: timedelta | int = None
```
tebeka, line 22: Why int and not float? I can't do a 500ms timeout.

itayd: right! will fix

itayd: see https://github.com/autokitteh/autokitteh/pull/815

PR #811, opened 2024-10-29
feat(ENG-1736): limit amount and size of vars that can be saved

itayd:


internal/backend/vars/vars.go
```diff
@@ -46,6 +63,25 @@ func (v *Vars) Set(ctx context.Context, vs ...sdktypes.Var) error {
...
+		if v.cfg.MaxValueSize != 0 && len(va.Value()) > v.cfg.MaxValueSize {
+			return fmt.Errorf("%w: value size %d exceeds max value size %d", sdkerrors.ErrLimitExceeded, len(va.Value()), v.cfg.MaxValueSize)
+		}
+
+		scids[va.ScopeID()] = true
+	}
+
+	if maxN := v.cfg.MaxNumVarsPerScope; maxN != 0 && len(vs) > maxN {
+		for scid := range scids {
+			n, err := v.db.CountVars(ctx, scid)
```
efiShtain: does this do N db queries to where N == Number of vars you try to set ?
if so, please refactor to query once in a transaction,
otherwise, I'm probably missing something
and anyhow you check the size and save without transaction, so it is not atomic

itayd: consider maybe change the vars api to work in a single scope per call

itayd: now with transaction

efiShtain (approved):

PR #812, opened 2024-10-29
DRAFT: feat: user audit log

itayd:
suggestion for a user audit log. the intention is that it can be later displayed in UI and will also be served for telemetry.


internal/backend/db/dbgorm/users_audit_logs.go
```diff
@@ -0,0 +1,48 @@
...
+	r := scheme.UserAuditLog{
+		UserID:      uuid,
+		Description: description,
+		Data:        jsonData,
+		Timestamp:   time.Now().UTC(),
+		Error:       errStr,
+	}
+
+	if err := db.db.WithContext(ctx).Create(&r).Error; err != nil {
+		l.Error("failed to create user audit log", zap.Error(err))
```
itayd, line 51: add more info for compliance

internal/backend/db/dbgorm/users_audit_logs.go
```diff
@@ -0,0 +1,48 @@
...
+
+func (db *gormdb) addUserAuditLog(ctx context.Context, description string, data any, err error) {
+	uidStr := authcontext.GetAuthnUserID(ctx)
+
+	l := db.z.With(zap.String("uid", uidStr), zap.String("description", description), zap.Error(err))
+
+	jsonData, err := json.Marshal(data)
+	if err != nil {
+		l.DPanic("failed to marshal data", zap.Error(err))
+		jsonData = []byte("null")
```
tebeka, line 23:
```suggestion
		jsonData, _ = json.Marshal(map[string]string{"error": err.Error()})
```

tebeka (commented):
IMO we need to first write a spec on what we want to have in the audit logs and what do we log there.

itayd:
> IMO we need to first write a spec on what we want to have in the audit logs and what do we log there.

did an initial once over with @efiShtain , this is just to get us started and to have something. i think it's a good start.

PR #816, opened 2024-10-30
feat(conn init): make origin optional

itayd:
makes it easier to start stuff from the dashboard and debug.

### DIFF
--- a/internal/backend/oauth/oauth.go
+++ b/internal/backend/oauth/oauth.go
@@ -413,10 +413,6 @@ func (o *oauth) StartFlow(ctx context.Context, intg string, cid sdktypes.Connect
 		return "", errors.New("invalid connection ID")
 	}
 
-	if origin == "" {
-		return "", errors.New("missing origin")
-	}
-
 	// Identify the relevant connection when we get an OAuth response.
 	state := strings.Replace(cid.String(), "con_", "", 1) + "_" + origin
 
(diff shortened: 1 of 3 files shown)


internal/backend/oauth/oauth.go
```diff
@@ -413,10 +413,6 @@ func (o *oauth) StartFlow(ctx context.Context, intg string, cid sdktypes.Connect
 		return "", errors.New("invalid connection ID")
 	}
 
-	if origin == "" {
```
daabr: To answer @efiShtain - there's no security aspect

internal/backend/oauth/webhook.go
```diff
@@ -85,7 +85,7 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Ensure the state parameter is well-formed.
-	sub := regexp.MustCompile(`^(.+)_([a-z]+)$`).FindStringSubmatch(state)
```
daabr, line 88: The change from `+` to `*` is incorrect because the `state` param is supposed to be an AK ID.

daabr (changes requested):
The `origin` param is supposed to be a required parameter - the specified value affects the way we end connection flows depending on whether they were initiated by the web UI / CLI / dashboard / etc. So currently there is no "default origin", and I don't think it's appropriate to have one, but it there was it wouldn be the web UI and not the dashboard.

internal/backend/oauth/oauth.go
efiShtain, line 416: do we use it for any security aspect ?

PR #818, opened 2024-10-30
bug(python): Fix "start" in another module.

tebeka:
Fixes: ENG-1749

Note: This currently doesn't solve the bug, see issue for more details.


tebeka:
Currently there are sever logs that are helping me debug. Not sure if to keep them or remove.

runtimes/pythonrt/py-sdk/autokitteh/events.py
```diff
@@ -9,7 +9,7 @@
 def subscribe(connection_name: str, filter: str) -> str:
     """Subscribe to events on connection. Optional filter is a CEL expression."""
     # Dummy implementation for local development.
-    return uuid4().hex
+    return f"sig_{uuid4().hex}"
```
itayd, line 12: should be typeid

runtimes/pythonrt/py-sdk/autokitteh/events.py
```diff
@@ -30,7 +30,7 @@ def next_event(
     return AttrDict()
 
 
-def start(loc: str, data: dict[str, any], memo: dict[str, str] = None) -> str:
+def start(loc: str, data: dict = None, memo: dict = None) -> str:
     """Start a new session."""
     # Dummy implementation for local development.
-    pass
+    return f"ses_{uuid4().hex}"
```
itayd, line 36: ditto

runtimes/pythonrt/py-sdk/autokitteh/events.py
tebeka: These are dummy functions that are replaced by the runner to call `ak`. I don't think adding `typeid-python` as dependency to SDK just to be compatible is worth it. Same for below.

itayd:
@tebeka what's the status here?

tebeka:
>  what's the status here?
@itayd Ready, need approval

itayd:
Does it actually solve the issue?

tebeka:
> Does it actually solve the issue?

@itayd I see calls to `Start`, but no call from `ak` to start a new session. So it solves start in another module, but not start.
IMO let's merge this and then work on the other issue.

runtimes/pythonrt/runner/call.py
```diff
@@ -1,23 +1,13 @@
...
-from time import sleep
 
-import autokitteh
 from autokitteh import decorators
 
 import log
 from deterministic import is_deterministic
 
-# Functions that are called back to ak
-AK_FUNCS = {
```
itayd, line 13: where did this go?

itayd: ah i see in 200

itayd:
i'm really confused about this pr. let's discuss when you can.

itayd (approved):

PR #820, opened 2024-10-31
fix(ENG-1755): update trigger now updating

itayd:


internal/backend/db/dbgorm/triggers.go
```diff
@@ -125,29 +124,17 @@ func (db *gormdb) DeleteTrigger(ctx context.Context, id sdktypes.TriggerID) erro
...
+	r.Schedule = trigger.Schedule()
 
-	// Note: building trigger record involves non-transactionl fetching connectionID and projectID from the DB
-	t, err := db.triggerToRecord(ctx, trigger)
-	if err != nil {
-		return err
-	}
-	// update and set null fields as well, e.g. nullify connection, data etc..
-	updateData := map[string]any{
-		"ConnectionID": t.ConnectionID,
```
efiShtain, line 144: don't we need connectionID update ? you didn't move it in the refactor

itayd: not sure i understand, but updating connection id does not make sense.

internal/backend/db/dbgorm/triggers.go
```diff
@@ -125,29 +124,17 @@ func (db *gormdb) DeleteTrigger(ctx context.Context, id sdktypes.TriggerID) erro
...
-		return err
+	r, err := db.getTriggerByID(ctx, trigger.ID().UUIDValue())
+	if r == nil || err != nil {
+		return translateError(err)
 	}
 
-	if envID := trigger.EnvID(); envID.IsValid() && curr.EnvID() != envID {
-		return sdkerrors.ErrConflict
-	}
+	r.CodeLocation = trigger.CodeLocation().CanonicalString()
```
daabr, line 136: IIUC, you removed the sanity check for the trigger's envID in favor of ignoring it in the update? It took me a while to carefully trace the code to make sure it's not dangerous. This needs a very verbose comment to explain what you're doing.

Consider the root cause of this bug: someone changed the trigger, but didn't know/remember they also need to update this function (because it's hidden) - this is the real problem we need to fix, beyond this specific bug about not updating a specific field during update.

If someone tries to update an ignored field, this function won't complain in any way - this is another maintainability issue. It's not worse than before, but it's not better either.

internal/backend/db/dbgorm/triggers.go
```diff
@@ -125,29 +124,17 @@ func (db *gormdb) DeleteTrigger(ctx context.Context, id sdktypes.TriggerID) erro
...
+	r.EventType = trigger.EventType()
+	r.Filter = trigger.Filter()
+	r.Schedule = trigger.Schedule()
```
daabr, line 139: Your forgot `Name`, `UniqueName`, and `WebhookSlug`

daabr (commented):
See comments - 1 specific bug, 1 maintainability issue, and a general lack of documentation and unit testing to show that the modified functions work as expected.

internal/backend/db/dbgorm/triggers.go
itayd: in the future we'll use FieldMask to signify what fields is being updated. for now i'll document this.

internal/backend/db/dbgorm/triggers.go
itayd: WebhookSlug cannot be updated.
I'll fix Name (and hence UniqueName).

daabr (approved):
Discussed offline, approving in advance

internal/backend/db/dbgorm/triggers.go
itayd: commented

internal/backend/db/dbgorm/triggers.go
itayd: done

PR #822, opened 2024-11-01
fix: list envs without a project filter

daabr:
If the project ID filter in env list requests is nil - AK should return all (accessible) envs, not an empty list.

Refs: ENG-1762


daabr:
> you basically allow any user fetch all envs with this change

The use-case I'm trying to fix is running "ak env list" - which currently returns nothing.

You're saying there's no automatic limitation in env queries to see only envs that I own? If so, that's an urgent thing to fix (and then we can merge this PR).

efiShtain:
you basically allow any user fetch all envs with this change

efiShtain:
ok I missed this line
```
q := gdb.withUserEnvs(ctx)
```

so you're good to go

efiShtain (approved):

PR #825, opened 2024-11-02
feat(ENG-1303)!: user objects

itayd:
- refactor auth to use actual user objects with generated ids.
- when configured, users are generated when first logged in.
- authcontext always stores entire user, not just id.
- jwt tokens store only user id. these are verified as existing when used.

currently users is an internal auth service. when we have proper rbac we should consider making it external to enable orgs, etc.

This is based on #832 .


internal/backend/usersgrpcsvc/nope.go
```diff
@@ -0,0 +1,5 @@
+package usersgrpcsrc
```
itayd: remove

tests/system/client.go
```diff
@@ -26,12 +26,14 @@ var (
 	sessionStateFinal = regexp.MustCompile(`state:SESSION_STATE_TYPE_(COMPLETED|ERROR|STOPPED)`)
 	sessionStateAll   = regexp.MustCompile(`state:SESSION_STATE_TYPE_`)
 	token             = "INVALID_TOKEN"
+
+	testUserEmail = "zumi@localhost"
```
efiShtain, line 30: this is an invalid email, we shuold test for valid emails

sdk/sdktypes/user.go
```diff
@@ -10,48 +10,39 @@ import (
...
 
-	return
-}
+func (u User) ID() UserID     { return kittehs.Must1(ParseUserID(u.read().UserId)) }
+func (u User) Email() string  { return u.read().Email }
+func (u User) Disabled() bool { return u.read().Disabled }
+
+var DefaultUser = kittehs.Must1(UserFromProto(&UserPB{
+	UserId: kittehs.Must1(ParseUserID("usr_3vser000000000000000000001")).String(),
+	Email:  kittehs.GetenvOr("DEFAULT_USER_EMAIL", "autokitteh@localhost"),
```
efiShtain: not a valid email

sdk/sdktypes/id.go
```diff
@@ -137,6 +137,8 @@ func newNamedIDString(name, kind string) string {
 	// a hint to the original name is encoded in the ID.
 	idName := strings.Map(func(r rune) rune {
 		switch r {
+		case '@':
```
efiShtain: what's the meaning of @ ?
could you add comments next to each case indicating it's purpose ?

internal/backend/svc/svc.go
```diff
@@ -126,15 +127,24 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
...
+			authhttpmiddleware.Configs,
 			fx.Provide(authhttpmiddleware.New),
-			fx.Provide(authhttpmiddleware.AuthorizationHeaderExtractor)),
+			fx.Provide(authhttpmiddleware.AuthorizationHeaderExtractor),
+		),
+		Component(
+			"authusers",
+			authusers.Configs,
+			fx.Provide(authusers.New),
+			fx.Invoke(func(lc fx.Lifecycle, users authusers.Users) { HookOnStart(lc, users.Setup) }),
```
efiShtain: why do you need to hook on start here ?

internal/backend/db/dbgorm/scheme/scheme.go
```diff
@@ -535,13 +547,7 @@ type Ownership struct {
...
-	Key       string        `gorm:"primaryKey;not null"`
-	Value     []byte
-
-	UpdatedAt time.Time
-
-	// foreign key
-	Project *Project
+type User struct {
+	UserID sdktypes.UUID `gorm:"primaryKey;type:uuid;not null"`
+	Email  string        `gorm:"uniqueIndex;not null"`
```
efiShtain: we also get the display name I think, might be useful
not a must

internal/backend/auth/authusers/users.go
```diff
@@ -0,0 +1,91 @@
...
+	"go.uber.org/zap"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/configset"
+	"go.autokitteh.dev/autokitteh/internal/backend/db"
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type Config struct {
+	TestUsersEmails string `koanf:"test_users_emails"`
```
efiShtain: why do we need this configuration as an actual configuration?
what do we try to achieve ? could you elaborate ?

if it's a default user, it should be named defulat user
for testing purposes we shuold use valid API of the system instead of adding stuff for the sole purpose of testing

internal/backend/auth/authusers/users.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+
+	if l == nil {
+		l = zap.NewNop()
+	}
+
+	return &users{db: db, cfg: cfg, l: l}
+}
+
+func (us *users) Setup(ctx context.Context) error {
```
efiShtain: if we need to seed stuff, better to have a seed migrations instead of running this code on every server restart
I also think if you need such stuff it should run in dev mode or test mode only

internal/backend/db/dbgorm/vars.go
```diff
@@ -21,10 +22,8 @@ func (gdb *gormdb) withUserVars(ctx context.Context) *gorm.DB {
 func (gdb *gormdb) setVar(ctx context.Context, vr *scheme.Var) error {
 	vr.VarID = vr.ScopeID // just ensure
 
-	uid, err := userIDFromContext(ctx)
-	if err != nil {
-		return err
-	}
+	uid := authcontext.GetAuthnUser(ctx).ID().String()
```
efiShtain, line 25: can this panic somehow on nil reference ?

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -52,12 +53,24 @@ func newTokensMiddleware(next http.Handler, deps Deps) http.HandlerFunc {
...
 					}
-					ctx = authcontext.SetAuthnUser(ctx, user)
+
+					u, err := deps.Users.GetByID(r.Context(), uid)
+					if err != nil {
+						http.Error(w, "unknown user", http.StatusUnauthorized)
+						return
+					}
+
+					if u.Disabled() {
```
efiShtain, line 69: what is disabled ? how do we control it? when do we disable users ?

internal/backend/auth/authjwttokens/jwt.go
```diff
@@ -53,12 +53,7 @@ func New(cfg *Config) (authtokens.Tokens, error) {
 	return &tokens{signKey: key}, nil
 }
 
-func (js *tokens) Create(user sdktypes.User) (string, error) {
-	uj, err := user.MarshalJSON()
-	if err != nil {
-		return "", fmt.Errorf("marshal JSON: %w", err)
-	}
-
+func (js *tokens) Create(uid sdktypes.UserID) (string, error) {
```
efiShtain: i think we shuold keep the email and userID in jwt, and we soon would need organizationID and maybe tier, 

I know you prefer minimal JWT, I do as well
but these are 4 elements that would make life much easier (maybe except for email which is not mandatory)

tier and organization_id would allow us to implement rate limit without any db access at the load balancer level

internal/backend/auth/authloginhttpsvc/svc.go
```diff
@@ -24,28 +27,27 @@ type Deps struct {
...
-	}
-
+	svc := &svc{Deps: deps}
 	return svc.registerRoutes(deps.Muxes)
 }
 
+type loginData struct {
+	Email        string
+	ProviderName string
+	ProviderData any
```
efiShtain: maybe map of string:string is better ?

efiShtain (commented):
left some comments, the gist looks ok

internal/backend/db/dbgorm/scheme/scheme.go
itayd: makes sense

sdk/sdktypes/id.go
itayd: it's an invalid typeid character. should probably change the logic here to be more clear.

itayd: changed

internal/backend/db/dbgorm/scheme/scheme.go
itayd: done

internal/backend/auth/authusers/users.go
itayd: for testing. problem is that currently we don't have an externally accessible API to add users. will do so when we have some minimal RBAC. i think we should leave this until then. lmk if it doesn't make sense to you.

internal/backend/db/dbgorm/vars.go
itayd: no, GetAuthnUser returns InvalidUser (or default user), which its ID is empty, and thus the string is empty.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: currently no api to do so, so maybe i should remove. the idea is that we can disable users because maybe abuse, non payment, etc.

internal/backend/auth/authjwttokens/jwt.go
itayd: but we need to get the user anyway from the database on each login. in that case we would have the info there?

internal/backend/auth/authloginhttpsvc/svc.go
itayd: we don't really do anything with this rn, so i'd just leave it as is, or maybe remove it?

internal/backend/svc/svc.go
itayd: for the test emails seeding.

tests/system/client.go
itayd: we should discuss. not sure how much validation we can do here.

internal/backend/auth/authusers/users.go
efiShtain: in testing you have full access to db, 
I prefer having a testing migration that runs in tests mode instead of adding test code to production 
what do you think ?

internal/backend/auth/authjwttokens/jwt.go
efiShtain: on login sure, but after login we don't need to fetch the user if we can, on any API call

internal/backend/auth/authloginhttpsvc/svc.go
efiShtain: if you can remove it even better

internal/backend/auth/authhttpmiddleware/authmiddleware.go
efiShtain: I think it's a premature capability, 
we don't have any actual reason to disable users
we should have it as part of a whole users system

internal/backend/auth/authjwttokens/jwt.go
efiShtain: pass the user to remind

itayd: now i put the entire user in the token, which is essential email and title.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: diagree. i want the ability to disable users as a defensive mechanism.

internal/backend/auth/authloginhttpsvc/svc.go
itayd: removed

itayd:
fixed seeding

efiShtain (approved):

PR #829, opened 2024-11-05
feat: autodetect required tools and versions during build

itayd:


Makefile
```diff
@@ -62,11 +62,11 @@ ak: bin/ak
 bin: bin/ak
 
 .PHONY: bin/ak
-bin/ak:
+bin/ak: require/go
```
daabr, line 65: Nit: I don't think you need .PHONY for bin/ak because it's an actual file, but I don't really care.

Makefile
```diff
@@ -136,35 +134,55 @@ test-sessions:
...
-generate-migrations:
+generate-migrations: require/atlas
 	@read -p "Enter migration name: " migration_name; \
 	atlas migrate diff $$migration_name --env sqlite; \
 	atlas migrate diff $$migration_name --env postgres
 
 # Requires nodejs installed
 .PHONY: tailwindcss
-tailwindcss:
+tailwindcss: require/npx
```
daabr: 1. With this PR making requirements explicit, you can remove the comment about the target.

2. I think we can get rid of the entire `tailwindcss` target - we don't need a good-looking integrations page in AK anymore. But that's an FYI beyond the scope of this PR.

Makefile
```diff
@@ -136,35 +134,55 @@ test-sessions:
...
 	npx --yes tailwindcss build -o web/static/tailwind.css
 
 .PHONY: webplatform
-webplatform:
+webplatform: require/go
 	make -C ./web/webplatform
+
+.PHONY: require/go
```
daabr, line 165: I think it actually makes sense NOT to use `.PHONY` here - the targets point to actual files, and if you run them once you don't need to run them again.

README.md
```diff
@@ -108,10 +108,9 @@ $ ak version
 ## Build Requirements (_Full_ Builds Only)
 
 - buf
-- docker
 - go >= 1.23
+- shellcheck
```
daabr, line 112: It makes more sense to keep `go` and `golangci-lint` together, and put `shellcheck` below them.

README.md
```diff
@@ -120,6 +119,7 @@ $ ak version
 - jq (for advanced output formatting)
 - atlasgo (for new DB migrations)
 - nodejs >= 20 (only if updating the UI)
+- docker
```
daabr, line 122: Question: even though it's not official, don't we actually expect on-prem instances of AK to use Docker for runtimes? Aren't there features that depend on that instead of local execution of Python?

go.mod
```diff
@@ -1,8 +1,6 @@
 module go.autokitteh.dev/autokitteh
 
-go 1.23
-
-toolchain go1.23.1
```
daabr, line 5: The `toolchain` line was added automaticaly IIRC. Are you **sure** it's a good idea to remove it? (I don't know, so I'm asking)

scripts/require/atlas.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -euo pipefail
+
+cd "$(dirname "$0")"
+
+# shellcheck disable=SC1091
```
daabr, line 7: https://www.shellcheck.net/wiki/SC1091 recommends a different way to work around this issue.

Makefile
itayd: it is a file, but since we can't really figure out the dependencies in the Makefile, we have to do it from scratch every time.

Makefile
itayd: we can't leave .PHONY out, see explanation in previous comment.

README.md
itayd: not always i think. for now we should allow docker-less.

go.mod
itayd: go mod tidy removed it, not me. so i guess that's fine.

scripts/require/atlas.sh
itayd: we already do these and they don't work, no idea why.

Makefile
itayd: removed target.

README.md
itayd: not sure about this, lets discuss if you're up to it.

README.md
efiShtain: we require docker for running shell check anyway

efiShtain:
will these changes work on windows machines ?

README.md
tebeka: > we require docker for running shell check anyway

Not anymore, see the fix in the Makefile.

scripts/require/common.sh
```diff
@@ -0,0 +1,25 @@
...
+}
+
+require() {
+    BIN="$1"
+    REQ_VER="$2"
+    VER_CMD="$3"
+    VER_PATTERN="$4"
+
+    if ! which "${BIN}" >& /dev/null; then
+        fail "missing ${BIN} binary, please install golang ${REQ_VER} or later."
```
tebeka, line 15:
```suggestion
        fail "missing ${BIN} binary, please install ${REQ_VER} or later."
```

tebeka (commented):
This is nice, but I'm not sure how much it improves developer experience vs the added complexity.

Also, I prefer to assume `make` is running in a valid environment and maybe have a `validate-env.sh` file that will do all these checks.

Also, also, I think it'll be better to have a single `check-version` script that will have what common has + a list of dependencies and you'll run it with `check-version buf`.

PR #832, opened 2024-11-06
feat(ENG-1783)!:  kill environments

itayd:
This is a breaking change that kills the environment objects. now everything belongs to projects.

TODO:
- [x] fix tests
- [x] do proper migration


go.sum
daabr, line 1: New entries in `go.sum` without `go.mod`? Does this file belong in this PR / is the other file missing by accident?

cmd/ak/cmd/deploy.go
```diff
@@ -59,24 +59,17 @@ var deployCmd = common.StandardCommand(&cobra.Command{
...
 		}
 
-		bid, err := common.BuildProject(project, dirPaths, filePaths)
+		bid, pid, err := common.BuildProject(project, dirPaths, filePaths)
 		if err != nil {
 			return err
 		}
 		logFunc(cmd, "exec")(fmt.Sprintf("create_build: created %q", bid))
 
-		// Step 3: parse the optional environment argument.
```
daabr, line 68: Remove the `env` var at the top of this file, and the `env` flag at the bottom.

cmd/ak/cmd/deploy.go
```diff
@@ -59,24 +59,17 @@ var deployCmd = common.StandardCommand(&cobra.Command{
...
-			return err
-		}
-
-		// Step 4: deploy the build
+		// Step 3: deploy the build
 		// (see also the "deployment" and "project" parent commands).
 		deployment, err := sdktypes.DeploymentFromProto(&sdktypes.DeploymentPB{
-			EnvId:   eid.String(),
-			BuildId: bid.String(),
+			ProjectId: pid.String(),
```
daabr, line 81: Question: is this really needed? Build objects already have a project ID field.

cmd/ak/cmd/deployments/create.go
```diff
@@ -14,7 +14,7 @@ import (
 var activate bool
 
 var createCmd = common.StandardCommand(&cobra.Command{
-	Use:     "create <--build-id=...> <--env=...> [--activate]",
+	Use:     "create <--build-id=...> <--project=...> [--activate]",
```
daabr, line 17: Same question as in `cmd/deploy.go`: if build objects already have a project ID field, why does the user need to specify it manually here?

proto/autokitteh/store/v1/svc.proto
```diff
@@ -11,9 +11,8 @@ service StoreService {
 }
 
 message GetRequest {
-  string env_id = 1;
-  string project_id = 2;
-  repeated string keys = 3 [(buf.validate.field).repeated.items.string.min_len = 1];
+  string project_id = 1;
```
daabr, line 14: I'm fine with this, but to be clear you're committing multiple violations here at a point in time that AK is no longer in its "early days":
- Delete a field instead of deprecating it
- Change a field ID

daabr (commented):
I really liked the concept of envs, but this is the right thing to do for now, so yay code cleanup! Left a few small comments - will approve once you reply to / address them, and fix all the tests and lint warnings.

go.sum
itayd: /shrug
i just ran go tidy again

cmd/ak/cmd/deploy.go
itayd: done

proto/autokitteh/store/v1/svc.proto
itayd: nobody is using the old proto yet, so I'm fine doing these kind of stuff while we can get away with it.

cmd/ak/cmd/deploy.go
itayd: i think having project id in deployment is important, having a project_id in build is only for auth. I'm ok with leaving this as is for now. we might revisit in the future.

cmd/ak/cmd/deployments/create.go
itayd: see above

efiShtain:
Just noting this requires a lot of UI updates I guess, need to make sure UI has time to adhere to this change before we merge

cmd/ak/common/build.go
```diff
@@ -43,51 +43,51 @@ func Build(rts sdkservices.Runtimes, srcFS fs.FS, paths []string, syms []sdktype
 	return b, nil
 }
 
-func BuildProject(project string, dirPaths, filePaths []string) (sdktypes.BuildID, error) {
+func BuildProject(project string, dirPaths, filePaths []string) (sdktypes.BuildID, sdktypes.ProjectID, error) {
```
efiShtain: why do you need to return the extra argument here ?

cmd/ak/cmd/projects/deploy.go
```diff
@@ -22,28 +21,19 @@ var deployCmd = common.StandardCommand(&cobra.Command{
...
-		if err = common.ToExitCodeWithSkipNotFoundFlag(cmd, err, "environment"); err != nil {
-			return err
-		}
-
-		// Step 3: deploy the build (see the "deployment" parent command).
+		// Step 2: deploy the build (see the "deployment" parent command).
 		deployment, err := sdktypes.DeploymentFromProto(&sdktypes.DeploymentPB{
-			EnvId:   eid.String(),
-			BuildId: bid.String(),
+			ProjectId: pid.String(),
```
efiShtain, line 50: the command's Use description says deploye <project name or ID>
why do we get the pid from the buildProject result instead of user input?
if this is the only reason to change buildProject signature to return projectId you can revert the change of buildProject

cmd/ak/cmd/triggers/create.go
```diff
@@ -38,28 +38,21 @@ var createCmd = common.StandardCommand(&cobra.Command{
 		}
 
 		// Project and/or environment are required.
```
efiShtain:
```suggestion
		// Project is required.
```

cmd/ak/cmd/deploy.go
```diff
@@ -59,24 +59,17 @@ var deployCmd = common.StandardCommand(&cobra.Command{
 			dirPaths = append(dirPaths, filepath.Dir(manifestPath))
 		}
 
-		bid, err := common.BuildProject(project, dirPaths, filePaths)
+		bid, pid, err := common.BuildProject(project, dirPaths, filePaths)
```
efiShtain: again projectId is provided from user, no need to get it as result from buildProject

cmd/ak/cmd/deploy.go
efiShtain: i also prefer leaving the projectID in deployment, it has much more use on the API and UI side
we can remove the projectID from build, since build belongs to one deployment only, we don't share builds

internal/backend/db/dbgorm/projects.go
```diff
@@ -48,6 +48,31 @@ func (gdb *gormdb) deleteProject(ctx context.Context, projectID sdktypes.UUID) e
 	})
 }
 
+func (gdb *gormdb) deleteProjectVars(ctx context.Context, id sdktypes.UUID) error {
+	// NOTE: should be transactional
+	db := gdb.db.WithContext(ctx)
+
+	// enforce foreign keys constrains while soft-deleting - should be no active deployments
+	var count int64
+	db.Model(&scheme.Deployment{}).Where("deleted_at is NULL and project_id = ?", id).Count(&count)
```
efiShtain, line 56: does this list only active deployments ? or any deployment ?

internal/backend/db/dbgorm/projects.go
```diff
@@ -48,6 +48,31 @@ func (gdb *gormdb) deleteProject(ctx context.Context, projectID sdktypes.UUID) e
 	})
 }
 
+func (gdb *gormdb) deleteProjectVars(ctx context.Context, id sdktypes.UUID) error {
+	// NOTE: should be transactional
```
efiShtain, line 52: so why isn't it a transaction ?

internal/backend/db/dbgorm/projects.go
```diff
@@ -197,18 +218,13 @@ type DeploymentState struct {
 func (db *gormdb) getProjectDeployments(ctx context.Context, pid sdktypes.UUID) ([]DeploymentState, error) {
 	var pds []DeploymentState
 	res := db.db.WithContext(ctx).Model(&scheme.Deployment{}).
-		Joins("join Envs on Envs.env_id = Deployments.env_id").
-		Where("Envs.project_id = ?", pid).
+		Joins("join Projects on Projects.project_id = Deployments.project_id").
```
efiShtain: deployment now has project id set, no need to join projects 
I'm also not sure whay the signature of this function is getProjectDeployments while is returns a list of states

this can be simplified to 
```sql 
select deployment_id, state from deployments where project_id = pid
```

no join and no need to distinct

the function should also be renamed to getProjectDeploymentsStates or it should get a fields list to return like
```go
func (db *gormdb) getProjectDeployments(ctx context.Contest, pid sdktypes.ProjectId, ...fields string)
```

internal/backend/db/dbgorm/triggers.go
```diff
@@ -75,24 +75,18 @@ func (gdb *gormdb) listTriggers(ctx context.Context, filter sdkservices.ListTrig
...
-	envID := trigger.EnvID()
-	env, err := db.GetEnvByID(ctx, envID)
-	if err != nil {
-		return nil, fmt.Errorf("get trigger env: %w", err)
-	}
-	projID := env.ProjectID()
+	pid := trigger.ProjectID()
 
-	uniqueName := triggerUniqueName(envID.String(), trigger.Name())
+	uniqueName := triggerUniqueName(pid.String(), trigger.Name())
```
efiShtain, line 85: this is not backward compatible if i understand correctly,
do we hav a migratino plan ?

internal/backend/sessions/sessiondata/data.go
```diff
@@ -58,32 +56,22 @@ func downloadBuild(ctx context.Context, buildID sdktypes.BuildID, builds sdkserv
 func Get(ctx context.Context, svcs *sessionsvcs.Svcs, session sdktypes.Session) (*Data, error) {
```
efiShtain, line 56: why get session data download the build data ?
it's not related to this PR but i just saw it here

internal/backend/sessions/sessiondata/data.go
```diff
@@ -58,32 +56,22 @@ func downloadBuild(ctx context.Context, buildID sdktypes.BuildID, builds sdkserv
...
-		if data.Env, err = retrieve(ctx, envID, svcs.Envs.GetByID); err != nil {
-			return nil, err
-		}
-
-		if data.ProjectID = data.Env.ProjectID(); !data.ProjectID.IsValid() {
-			return nil, fmt.Errorf("sessions can only run on projects")
-		}
-
-		cfilter := sdkservices.ListConnectionsFilter{ProjectID: data.ProjectID}
+	if pid := session.ProjectID(); pid.IsValid() {
```
efiShtain, line 63: i think env wasn't required before
but can we have a session without a valid projectID ?
maybe it should return an error if the projectID is invalid on the session ?

internal/backend/sessions/sessionworkflows/modules/store/store.go
```diff
@@ -17,12 +17,12 @@ const moduleName = "store"
...
 var ExecutorID = sdktypes.NewExecutorID(fixtures.NewBuiltinIntegrationID("store"))
 
-func New(envID sdktypes.EnvID, projectID sdktypes.ProjectID, client *redis.Client) sdkexecutor.Executor {
+func New(pid sdktypes.ProjectID, client *redis.Client) sdkexecutor.Executor {
 	mod := redisint.NewInternalModule(
 		moduleName,
 		ExecutorID,
 		client,
-		func(s string) string { return store.Prefix(projectID, envID) + s },
+		func(s string) string { return store.Prefix(pid) + s },
```
efiShtain, line 25: again breaking change,
it's not only API breaking change, its data breaking change that won't be so easy to migrate

internal/backend/store/store.go
```diff
@@ -65,18 +65,18 @@ func New(z *zap.Logger, cfg *Config) (sdkservices.Store, *redis.Client, error) {
...
-func (s *store) List(ctx context.Context, envID sdktypes.EnvID, projectID sdktypes.ProjectID) ([]string, error) {
+func (s *store) List(ctx context.Context, projectID sdktypes.ProjectID) ([]string, error) {
 	if s.client == nil {
 		// return nil in order not to fail any query that doesn't check if redis enabled or not.
 		return nil, nil
 	}
 
 	var (
-		prefix = Prefix(projectID, envID)
+		prefix = Prefix(projectID)
```
efiShtain, line 79: do we still need the prefix function ?

internal/manifest/plan.go
```diff
@@ -102,12 +97,12 @@ func planProject(ctx context.Context, mproj *Project, client sdkservices.Service
 	}
 
 	// TODO: Remove all non-default environments.
```
efiShtain: remove comment

internal/manifest/plan.go
```diff
@@ -102,12 +97,12 @@ func planProject(ctx context.Context, mproj *Project, client sdkservices.Service
...
-	envActions, err := planDefaultEnv(ctx, mproj.Vars, client, mproj.Name, pid, optfns...)
+	pvarsActions, err := planProjectVars(ctx, mproj.Vars, client, mproj.Name, pid, optfns...)
```
efiShtain, line 99: is this something new?
i don't see this code before
why do we need it? am i missing something ?

migrations/postgres/20241106064610_killenv.sql
```diff
@@ -0,0 +1,40 @@
+-- +goose Up
```
efiShtain: did you check this migrations on a full db with actual data ?
does it require down time ?
how do we make sure running code doesn't try to create envs while you run migrations and old code still running?

Maybe we should plan this change in multiple deployments

proto/autokitteh/deployments/v1/svc.proto
```diff
@@ -44,7 +44,7 @@ message TestRequest {
 message TestResponse {}
 
 message ListRequest {
-  string env_id = 1;
```
efiShtain, line 47: all breaking API changes should be cheked with UI before merging

consider having a backward compatible change, 
mark the fields as deprecated
and if someone provided env_id still, assume it is a project_id and continue with it in the backend

once ui is ready and updated, remove the env_id

proto/autokitteh/dispatcher/v1/svc.proto
```diff
@@ -26,8 +26,8 @@ message DispatchResponse {
 
 message RedispatchRequest {
   string event_id = 1 [(buf.validate.field).string.min_len = 1];
-  string env_id = 3;
-  string deployment_id = 2;
+  string project = 2; // either id or name.
```
efiShtain, line 29: why do we need the project_id for re / dispatching event ? we needed the env cause it might have different arguments, but now deployment should be enough, no need for project

sdk/sdkservices/dispatcher.go
```diff
@@ -7,8 +7,8 @@ import (
 )
 
 type DispatchOptions struct {
-	// If set, dispatch only to deployments in this environment. Can be either an env ID or full path (project/env).
-	Env string
+	// If set, dispatch only to deployments in this project. Can be either n project ID or name.
+	Project string
```
efiShtain, line 11: no need for project

efiShtain (commented):
requires some change
a really risky change
requires a dedicated TESTED plan for migrating live environments, backups before executing the plan, might require down time

cmd/ak/common/build.go
itayd: you need the project id further on, this is just kind of a shortcut.

internal/backend/db/dbgorm/projects.go
itayd: looks like all deployments

itayd: any non-deleted one that is

internal/backend/db/dbgorm/triggers.go
itayd: i'm going to re-add envID back, so we'll be able to preserve it and run some migration to move from env to proj. after migration we can remove the envid.

internal/backend/sessions/sessiondata/data.go
itayd: we need to give the runtime the build?

internal/backend/sessions/sessiondata/data.go
itayd: so for session tests there is no project id. theoretically one might launch sessions without a project.

internal/backend/sessions/sessionworkflows/modules/store/store.go
itayd: nobody is using this for now, so it's ok. i might even totally remove it.

internal/manifest/plan.go
itayd: instead of setting up a default env, it's just setting all the vars. before vars belong to env, now they belong to project.

migrations/postgres/20241106064610_killenv.sql
itayd: oh we're going to work on this in a separate change. we'll discuss.

proto/autokitteh/deployments/v1/svc.proto
itayd: we'll discuss how exactly to do this - i think we should just merge for now, but deploy in tandem backend + UI. so there wont be any crosstalk between versions. we have low enough volume rn that we can pull it off.

proto/autokitteh/dispatcher/v1/svc.proto
itayd: it's essentially an convinience that says "just throw it to the active deployment of that project".

sdk/sdkservices/dispatcher.go
itayd: see above

cmd/ak/cmd/deploy.go
itayd: i'm going to leave it in as optional in build as well for now for record keeping. might remove later.

cmd/ak/cmd/projects/deploy.go
itayd: not sure i follow you here, but project in build is optional.

cmd/ak/cmd/deploy.go
itayd: ok, moved some things around.

internal/backend/db/dbgorm/projects.go
itayd: it is called in a transaction changed the comment.

internal/backend/db/dbgorm/projects.go
itayd: good find. done.

internal/backend/store/store.go
itayd: i think it's better this way since it's more standard. nobody is currently using it anyway currently and we're going to redo this.

internal/manifest/plan.go
itayd: done

cmd/ak/cmd/projects/deploy.go
efiShtain: I say it can't be optional, has to be mandatory
you can't build anything, you have to build a project

cmd/ak/common/build.go
efiShtain: I don't like that BuildProject get some random project name and returns projectID
it should get a project object or a projectID and build that

proto/autokitteh/dispatcher/v1/svc.proto
efiShtain: so why do we need the deployment_id if the only option is the active deployment any way ?

internal/backend/sessions/sessiondata/data.go
efiShtain: it just make the system more complicated for stuff we don't really need, nor there are requirements for it
I think error it better here

internal/backend/sessions/sessiondata/data.go
efiShtain: ok, 
thought this is an external API

internal/backend/db/dbgorm/projects.go
efiShtain:
```
// enforce foreign keys constrains while soft-deleting - should be no active deployments
```

the comment state it should check active deployments 

so which is wrong? code or comment ? 
I guess the comment since the context is delete env

efiShtain (approved):
some stuff, not a blocker for such a big change

cmd/ak/common/build.go
itayd: that's what i did now...

cmd/ak/cmd/projects/deploy.go
itayd: the api lets you build arbitrary stuff. we'll limit it with rbac later.

internal/backend/sessions/sessiondata/data.go
itayd: i think it is useful, not just for testing. will limit it with rbac later.

proto/autokitteh/dispatcher/v1/svc.proto
itayd: you can have "testing" deployments. these are deployments that receive events that are specifically sent to them.

PR #839, opened 2024-11-11
bug(python): User functions should not run as activities.

tebeka:
Fixes: ENG-1777

This changes the code to use a custom importer that will patch all files in user code directory.


efiShtain (approved):
Please test it in a container before merging

PR #843, opened 2024-11-13
fix: isinstance called with multiple args

efiShtain:


runtimes/pythonrt/runner/syscalls.py
```diff
@@ -113,7 +113,7 @@ def ak_next_event(self, args, kw):
 
         req = pb.NextEventRequest(runner_id=self.runner_id, signal_ids=ids)
         if timeout:
-            if isinstance(timeout, int, float):
+            if isinstance(timeout, int) or isinstance(timeout, float):
```
daabr: https://stackoverflow.com/questions/33311258/python-check-if-variable-isinstance-of-any-type-in-list

```suggestion
            if isinstance(timeout, int | float):
```

daabr (approved):

PR #844, opened 2024-11-13
fix(tests): print in acceptance test project

daabr:
Shouldn't clash with built-in function `print`

### DIFF
--- a/tests/acceptance/pretty_print.py
+++ b/tests/acceptance/pretty_print.py
@@ -1,5 +1,5 @@
 import json
 
 
-def pretty_json(label, data):
+def print_json(label, data):
     print(label + ":\n" + json.dumps(data, indent=2))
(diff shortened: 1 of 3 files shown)


tests/acceptance/pretty_print.py
```diff
@@ -1,5 +1,5 @@
 import json
 
 
-def pretty_json(label, data):
+def print_json(label, data):
```
efiShtain, line 4:
```suggestion
def json(label, data):
```

efiShtain (approved):
suggested changing pretty_print.print_json to pretty_print.json

tests/acceptance/pretty_print.py
daabr: Sorry, I set this PR to auto-merge on approve.

But anyway, the suggestion wouldn't work because the `pretty_print` module already imports the Python module `json`, so we can't have a function called `json` too.

efiShtain: I think you can import as another name, and exposing a nice API is better then technical issue in your packgae
but it's water under the bridge

PR #849, opened 2024-11-17
feat(lint): Create Linter

tebeka:
Start of ENG-1029

Initial linter code. Here for comments.


cmd/ak/cmd/projects/lint.go
```diff
@@ -0,0 +1,83 @@
...
+		return err
+	}
+
+	resources := map[string][]byte{
+		"autokitteh.yaml": data,
+	}
+
+	ctx, cancel := common.LimitedContext()
+	defer cancel()
+	vs, err := projects().Lint(ctx, sdktypes.InvalidProjectID, resources)
```
efiShtain: why pass an invalid project ID ? 
can this lint run local in cli ? 
I prefer if we have lint only function without the projectID If it is not required

internally on set resources we can call lint anyway

cmd/ak/cmd/projects/lint.go
```diff
@@ -0,0 +1,83 @@
...
+	if err != nil {
+		return err
+	}
+
+	w := cmd.OutOrStdout()
+
+	ok := true
+	for _, v := range vs {
+		level := levelName(v.Level)
+		fmt.Fprintf(w, "%s:%d - %s - %s\n", v.FileName, v.Line, level, v.Message)
```
efiShtain: I think we use abstracted writers so output format is controlled (text / json / etc...)

cmd/ak/cmd/projects/lint.go
```diff
@@ -0,0 +1,83 @@
...
+
+		if v.Level == projectsv1.CheckViolation_LEVEL_ERROR {
```
efiShtain: we tend to wrap this in sdktypes

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,108 @@
...
+
+import (
+	"fmt"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/projectsgrpcsvc"
+	"go.autokitteh.dev/autokitteh/internal/manifest"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type Checker func(projectID sdktypes.ProjectID, manifest *manifest.Manifest, resources map[string][]byte) []*sdktypes.CheckViolation
```
efiShtain, line 23: I think you can wrap the inputs in a type
it will be easier to extend later and it makes all the checker signatures simple, instead of stuff like

```go
func checkNoTriggers(_ sdktypes.ProjectID, m *manifest.Manifest, _ map[string][]byte) []*sdktypes.CheckViolation
```

or 
```go
func checkEmptyVars(_ sdktypes.ProjectID, m *manifest.Manifest, _ map[string][]byte) []*sdktypes.CheckViolation
```

you can do 
```go
type ProjectDescriptor struct {
 pid sdktypes.ProjectID
 resources map[string][]byte
 manifest manifest.Manifest
}

type Checker func(desc ProjectDescriptor) []*sdktypes.CheckViolation

func Validate(projectID sdktypes.ProjectID, manifest *manifest.Manifest, resources map[string][]byte) []*sdktypes.CheckViolation {
 	var vs []*sdktypes.CheckViolation
        desc := ProjectDescriptor{
              pid: projectID,
              manifest: manifest,
              resources: resources
        }
 	for _, checker := range lintCheckers {
 		vs = append(vs, checker(desc)...)
 	}

 	return vs
 }
```

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,108 @@
...
+	var vs []*sdktypes.CheckViolation
+	for _, checker := range lintCheckers {
+		vs = append(vs, checker(projectID, manifest, resources)...)
+	}
+
+	return vs
+}
+
+func checkNoTriggers(_ sdktypes.ProjectID, m *manifest.Manifest, _ map[string][]byte) []*sdktypes.CheckViolation {
+	if len(m.Project.Triggers) == 0 {
```
efiShtain, line 73: if manifest is nil here it's gonna panic
better check that it is not nil or add a checkManifestProvidedChecker before

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,108 @@
...
+		}
+	}
+
+	return nil
+}
+
+func checkEmptyVars(_ sdktypes.ProjectID, m *manifest.Manifest, _ map[string][]byte) []*sdktypes.CheckViolation {
+	var vs []*sdktypes.CheckViolation
+	for _, v := range m.Project.Vars {
+		if v.Value == "" {
```
efiShtain, line 92: are you sure we don't allow empty variables ?
sometimes I use empty vars as booleans, just to indicate if it exists or not

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,108 @@
...
+			vs = append(vs, &sdktypes.CheckViolation{
+				FileName: manifestFile,
```
efiShtain: technically all the violations are in the manifest file, why do we need this FileName ? 
future user code errors ?

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return vs
+}
+
+func checkNoTriggers(_ sdktypes.ProjectID, m *manifest.Manifest, _ map[string][]byte) []*sdktypes.CheckViolation {
+	if len(m.Project.Triggers) == 0 {
+		return []*sdktypes.CheckViolation{
+			{
+				FileName: manifestFile,
+				Level:    sdktypes.ViolationError,
```
efiShtain: we support projects without triggers, so it is at most a warning not an error

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,108 @@
...
+
+	return vs
+}
+
+const (
+	mb = 1 << 20
+)
+
+func checkSize(_ sdktypes.ProjectID, _ *manifest.Manifest, resources map[string][]byte) []*sdktypes.CheckViolation {
+	maxProjectSize := projectsgrpcsvc.Configs.Default.MaxUploadSize
```
efiShtain, line 127: you need to get the correct set of configs using DI like the other services

make validator a dependency of projects or a separate service and inject it to projects (even better)
so you can access the configs

might be possible to reuse it directly from CLI for faster validation

efiShtain: also, this check is at the wrong place, we don't want to allow people upload 1GB of data just to fail on listing, it can crash the service
size should be checked on client side for fast fail and on setResources endpoint before the full size arrive at the backend

internal/backend/projects/projects.go
```diff
@@ -266,3 +266,22 @@ func findConnection(id sdktypes.ConnectionID, conns []sdktypes.Connection) (sdkt
...
+			return nil, err
+		}
+	}
+
+	manifest, err := manifest.Read(data, "autokitteh.yaml")
+	if err != nil {
+		return nil, err
+	}
+
+	violations := Validate(projectID, manifest, resources)
```
efiShtain: see my comments on config above

internal/backend/projectsgrpcsvc/svc.go
```diff
@@ -282,3 +282,16 @@ func (s *Server) Export(ctx context.Context, req *connect.Request[projectsv1.Exp
 
 	return connect.NewResponse(&resp), nil
 }
+
+func (s *Server) Lint(ctx context.Context, req *connect.Request[projectsv1.LintRequest]) (*connect.Response[projectsv1.LintResponse], error) {
+	// TODO: Need to work with our without project
+	vs, err := s.projects.Lint(ctx, sdktypes.InvalidProjectID, req.Msg.Resources)
```
efiShtain: why invalid project id instead of passing what you get from lint request message ?

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,108 @@
...
+
+	total := 0
+	for _, data := range resources {
```
efiShtain, line 130: technically user can upload millions of zero size data and still cause issues
maybe we should also limit number of files ?

cmd/ak/cmd/projects/lint.go
tebeka: Added TODO for getting project ID from user or manifest.

cmd/ak/cmd/projects/lint.go
tebeka: Added TODO to support JSON

cmd/ak/cmd/projects/lint.go
tebeka: It was there, just didn't use it  :) Fixed.

internal/backend/projects/lint.go
tebeka: Currently I'm OK with 3 parameters. If it'll grow then we can think about it, don't like to much future proofing.

internal/backend/projects/lint.go
tebeka: If manifest is `nil` we should panic

internal/backend/projects/lint.go
tebeka: That's what the RFC says. I'll check with @haimzlato

internal/backend/projects/lint.go
tebeka: I agree client should check on deploy/build - but not a linter issue.
As to get correct set of configs, it's very backwards to create a service just to get a config and currently there's only one default config.

internal/backend/projectsgrpcsvc/svc.go
tebeka: There's a TODO there about working with or without a project. Need to discuss with @itayd and @haimzlato

internal/backend/projects/lint.go
tebeka: Good idea, added a check to the RFC

internal/backend/projects/lint.go
efiShtain: I care more about the multiple functions that most ignore parameters with the _ 
the future proofing is just a good side effect

internal/backend/projects/lint.go
efiShtain: this is backend code
you can't panic

internal/backend/projects/lint.go
efiShtain: it's 3 simple changes

create a struct for the validator
use fx to instantiate it
di the validator to projects

it also adhere to current convention

internal/backend/projects/lint.go
tebeka: I can panic in a handler, gRPC recovers.

internal/backend/projects/lint.go
tebeka: A struct only hides the fact you don't use all the variables, this way it's more explicit.

efiShtain: not sure I agree, but ok

internal/backend/projects/lint.go
efiShtain: still prefer not to panic and leave it as last resort
if you think of an error, we need to handle it

proto/autokitteh/projects/v1/svc.proto
```diff
@@ -113,6 +113,30 @@ message ExportResponse {
   bytes zip_archive = 1 [(buf.validate.field).bytes.min_len = 10];
 }
 
+message LintRequest {
+  string project_id = 1 [(buf.validate.field).string.min_len = 1];
+  map<string, bytes> resources = 2;
```
itayd, line 112: if this is on the project, why do you need resources and manifest file? do it directly on the saved resources?

proto/autokitteh/projects/v1/svc.proto
```diff
@@ -113,6 +113,30 @@ message ExportResponse {
...
+
+message CheckViolation {
+  enum Level {
+    LEVEL_UNSPECIFIED = 0;
+    LEVEL_WARNING = 1;
+    LEVEL_ERROR = 2;
+  }
+
+  string file_name = 1;
+  uint32 line = 2;
```
itayd: this can be CodeLocation

sdk/sdkclients/sdkprojectsclient/client.go
```diff
@@ -211,3 +211,25 @@ func (c *client) Export(ctx context.Context, pid sdktypes.ProjectID) ([]byte, er
...
+	req := projectsv1.LintRequest{
+		ProjectId:    pid.String(),
+		Resources:    resources,
+		ManifestFile: manifestFile,
+	}
+	resp, err := c.client.Lint(ctx, connect.NewRequest(&req))
+	if err != nil {
+		return nil, rpcerrors.ToSDKError(err)
+	}
+
```
itayd, line 230: you need to add protovalidate validation check here

sdk/sdktypes/project.go
```diff
@@ -47,3 +47,13 @@ func (p Project) WithNewID() Project { return p.WithID(NewProjectID()) }
 func (p Project) WithID(id ProjectID) Project {
 	return Project{p.forceUpdate(func(pb *ProjectPB) { pb.ProjectId = id.String() })}
 }
+
+type (
```
itayd, line 61: I guess you don't want to do this in the style of all other types here. I can live with it for now since it's quite simple.

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,371 @@
...
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+type Checker func(projectID sdktypes.ProjectID, manifest *manifest.Manifest, resources map[string][]byte) []*sdktypes.CheckViolation
+
+var (
+	lintCheckers []Checker
+)
+
+const manifestFile = "autokitteh.yaml"
```
itayd: manifestFilePath?

internal/backend/projects/lint.go
itayd: this needs to be a warning. and with a way to disable it.

internal/backend/projects/lint.go
itayd: it's a warning. but i'd like a way to disable it.

internal/backend/projects/lint.go
itayd: yeah we should not mention it's a manifestFile issue... more as global settings?

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,371 @@
...
+				RuleId:   "E2",
+			},
+		}
+	}
+
+	return nil
+}
+
+func checkConnectionNames(_ sdktypes.ProjectID, m *manifest.Manifest, _ map[string][]byte) []*sdktypes.CheckViolation {
+	names := make(map[string]int) // name -> count
```
itayd, line 152: when checking names, we need to make sure it paths validation through `sdktypes.ParseSymbol`.

itayd: also for triggers, projects, etc...

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,371 @@
...
+	}
+
+	return vs
+}
+
+func checkHandlers(_ sdktypes.ProjectID, m *manifest.Manifest, resources map[string][]byte) []*sdktypes.CheckViolation {
+	var vs []*sdktypes.CheckViolation
+
+	for _, t := range m.Project.Triggers {
+		// FIXME: Is it OK to have a trigger without "Call"? There are in the tests
```
itayd: yes

internal/backend/projects/lint.go
```diff
@@ -0,0 +1,371 @@
...
+		if t.Call == "" {
+			continue
+		}
+
+		fileName, handler, found := strings.Cut(t.Call, ":")
```
itayd: you can just use `sdktypes.ParseCodeLocation`

internal/backend/projects/lint.go
tebeka: IMO passing `nil` manifest is an unrecoverable error for the linter. I'm OK with panicking here. There's no meaning to returning an error here.

proto/autokitteh/projects/v1/svc.proto
tebeka: Since we don't want to save new resources every time we lint. Also initially there's no project.

proto/autokitteh/projects/v1/svc.proto
tebeka: "A little copying is better than a little dependency." :) I don't think the linter should be tied to program.
LMK if it's important to you and I'll change.

sdk/sdkclients/sdkprojectsclient/client.go
tebeka: Where? before calling `c.client.Lint`? Also, do you mean `internal.Validate`?

sdk/sdktypes/project.go
tebeka: How does it differ? It a type alias like the other types.

internal/backend/projects/lint.go
tebeka: Done

internal/backend/projects/lint.go
tebeka: Done

internal/backend/projects/lint.go
tebeka: Made into a warning. Opened ENG-1955 for disabling linter.

internal/backend/projects/lint.go
tebeka: Created ENG-1956 for configuration.

tebeka: Added the check for size in the command line.

proto/autokitteh/projects/v1/svc.proto
tebeka: Done.

sdk/sdktypes/project.go
itayd: tbh i have no idea what i meant here.

proto/autokitteh/projects/v1/svc.proto
itayd: please document that in the proto.

proto/autokitteh/projects/v1/svc.proto
itayd: I think it is - there should be one way to specify location. Please change.

internal/backend/projects/lint.go
efiShtain: I don't mind panicking on the cli side, but I'm not sure it's a valid option in backend

did you verify that panicking here does not break the whole server?

internal/backend/projects/lint.go
efiShtain: @tebeka ?

internal/backend/projectsgrpcsvc/svc.go
efiShtain: @itayd  / @haimzlato  any decision here ?
If it's a discussion, let's open a ticket
if there is a decision let's resolve or fix

internal/backend/projects/lint.go
tebeka: Manually added a `panic` in `Validate` and ran `ak project lint`. The server continue to run.

internal/backend/projectsgrpcsvc/svc.go
tebeka: This is outdated. I'm using project ID from the manifest or from `-p` switch.

proto/autokitteh/projects/v1/svc.proto
tebeka: Done.

internal/backend/projects/lint.go
tebeka: It is a warning. We have a different issue on how to disable linter.

internal/backend/projects/lint.go
tebeka: A linter issue is connect to a file & line number. That's what users excpect. I don't see the harm in using the manifest file.

sdk/sdkclients/sdkprojectsclient/client.go
itayd:
```
	if err := internal.Validate(resp.Msg); err != nil {
		return nil, err
	}
```

same as the function above, on the response from lint.

itayd (approved):
approved with a comment

PR #850, opened 2024-11-17
chore: improve ownership error logs

daabr:


itayd:
i'd rather not merging this. i'm redoing the whole thing and this will interrupt me.

PR #855, opened 2024-11-21
fix: remove old web platform distributions on VERSION update

itayd:
when running `make ak`, properly discard old downloaded versions. currently it doesn't because bash is HARD, which causes the server to possibly use the wrong version.

also:
- alert and disable webplatform when used version does not match VERSION, do not use older versions.
- specifying the sha is no longer mandatory.
- upgrade to latest web-platform.


web/webplatform/fetch.sh
tebeka, line 1: IMO this file is complex enough to merit a rewrite in Go.

web/webplatform/fetch.sh
```diff
@@ -4,13 +4,18 @@ set -euo pipefail
...
 
 checksum() {
+    if [[ -z "${EXPECTED_SHA}" ]]; then
+        echo "no checksum provided"
+        return
+    fi
+
     echo "verifying checksum..."
 
     sha=$(shasum -a 256 "$1" | cut -d ' ' -f 1)
```
tebeka: This works only because it's the first field. Use awk here as well?

daabr (approved):

web/webplatform/webplatform.go
```diff
@@ -27,17 +30,35 @@ func DistFilename() (string, error) {
 		return "", err
 	}
 
-	_, de := kittehs.FindFirst(des, func(de fs.DirEntry) bool { return distRegex.MatchString(de.Name()) })
-	if de == nil {
+	des = kittehs.Filter(des, func(de fs.DirEntry) bool { return distRegex.MatchString(de.Name()) })
```
tebeka: Why do we need to glob? You can know the exact file name from `VERSION`

web/webplatform/webplatform_test.go
```diff
@@ -6,12 +6,13 @@ import (
...
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	"go.uber.org/zap"
 
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 )
 
 func TestLoad(t *testing.T) {
-	fs, v, err := LoadFS()
+	fs, v, err := LoadFS(zap.NewNop())
```
tebeka: This will quell errors, use `zaptest`?

web/webplatform/webplatform_test.go
```diff
@@ -27,3 +28,20 @@ func TestLoad(t *testing.T) {
 	require.NoError(t, err)
 	require.NotNil(t, f)
 }
+
+func TestVersion(t *testing.T) {
+	l, _ := zap.NewDevelopment() // if test fails, will show detailed version error.
```
tebeka: Check for error, otherwise test failures will be hard to understand.

tebeka (approved):
Some minor comment but can merge.

web/webplatform/fetch.sh
itayd: if we take it further somehow, will consider.

web/webplatform/webplatform.go
itayd: oh wow, right. will try.

web/webplatform/fetch.sh
itayd: done

web/webplatform/webplatform_test.go
itayd: done

web/webplatform/webplatform_test.go
itayd: removed test

itayd:
greatly simplified the code since now only using version in VERSION. Thanks @tebeka for the tip. going to push this in, lmk if any issues - I'll fix.

PR #858, opened 2024-11-28
bug(python): Stop run when there's a call error

tebeka:
Fixes: ENG-1696


runtimes/pythonrt/runner/main.py
```diff
@@ -194,14 +194,12 @@ def Start(self, request: pb.runner.StartRequest, context: grpc.ServicerContext):
 
     def Execute(self, request: pb.runner.ExecuteRequest, context: grpc.ServicerContext):
         with self.lock:
-            call = self.activity_calls.pop() if self.activity_calls else None
```
itayd: ah good i wanted to talk to you about that. if you do `pop` it will not work on retries.

itayd (commented):
i see mostly logging changes?

runtimes/pythonrt/runner/main.py
tebeka: Any action item for me?

tebeka:
> i see mostly logging changes?

No. The main change is sending `Done` when there's an error so it'll stop the workflow.

runtimes/pythonrt/runner/main.py
itayd: nope

itayd (approved):

PR #866, opened 2024-12-05
feat(integrations): add custom OAuth for slack

pashafateev:


web/static/slack/connect/index.html
```diff
@@ -45,6 +45,9 @@ <h1>Slack - Create a New Connection</h1>
       <button class="toggle" id="toggletab2" onclick="toggleTab('tab2')">
         OAuth v2
```
daabr:
```suggestion
        Default OAuth v2
```

web/static/slack/connect/index.html
```diff
@@ -125,10 +127,68 @@ <h1>Slack - Create a New Connection</h1>
...
+          <input
+            type="text"
+            name="signing_secret"
+            autocomplete="off"
+            spellcheck="false"
+            required
+          />
+        </div>
+
+        <button type="submit" class="submit-btn">Start Custom OAuth Flow</button>
```
daabr, line 190: I see the details that the user is supposed to provide to us, but what about the webhook slug that we provide to them to accept events? Or is that a separate phase?

sdk/sdkservices/oauth.go
```diff
@@ -16,5 +16,5 @@ type OAuth interface {
 	Register(ctx context.Context, id string, cfg *oauth2.Config, opts map[string]string) error
 	Get(ctx context.Context, id string) (*oauth2.Config, map[string]string, error)
 	StartFlow(ctx context.Context, integration string, cid sdktypes.ConnectionID, origin string) (string, error)
-	Exchange(ctx context.Context, integration, code string) (*oauth2.Token, error)
+	Exchange(ctx context.Context, integration, code string, cid sdktypes.ConnectionID) (*oauth2.Token, error)
```
daabr: A more appropriate location for `cid` - before `code`

integrations/slack/internal/vars/vars.go
```diff
@@ -15,6 +15,11 @@ type Vars struct {
 var (
 	AuthType = sdktypes.NewSymbol("authType")
 
+	// Custom OAuth
+	ClientID         = sdktypes.NewSymbol("clientID")
+	ClientSecret     = sdktypes.NewSymbol("clientSecret")
+	SigningSecret    = sdktypes.NewSymbol("signingSecret")
```
daabr: I know the casing you're using here is consistent with the other vars, but I want to slowly migrate to `snake_case`, for all vars. Please use this in the new vars to avoid future migrations.

integrations/slack/oauth.go
```diff
@@ -11,17 +11,20 @@ import (
...
 // to receive and dispatch asynchronous event notifications.
 type handler struct {
 	logger *zap.Logger
+	oauth  sdkservices.OAuth
+	vars   sdkservices.Vars
 }
 
-func NewHandler(l *zap.Logger) http.Handler {
-	return handler{logger: l}
+func NewHandler(l *zap.Logger, o sdkservices.OAuth, v sdkservices.Vars) handler {
```
daabr: I'm assuming the actual usage of the OAuth service is pending in this PR and not a separate one? Otherwise, please postpone this addition.

integrations/slack/server.go
```diff
@@ -20,14 +20,18 @@ const (
 	// new OAuth-based connections.
```
daabr:
```suggestion
	// new OAuth-based connections using the server's default app.
```

integrations/slack/server.go
```diff
@@ -36,7 +40,9 @@ func Start(l *zap.Logger, muxes *muxes.Muxes, v sdkservices.Vars, d sdkservices.
 	// to have an authenticated user context, so the DB layer won't reject them.
 	// For this purpose, init webhooks are managed by the "auth" mux, which passes
 	// through AutoKitteh's auth middleware to extract the user ID from a cookie.
-	muxes.Auth.Handle("GET "+oauthPath, NewHandler(l))
+	h := NewHandler(l, o, v)
+	muxes.Auth.Handle("GET "+oauthPath, h)
```
daabr:
```suggestion
	muxes.Auth.Handle("GET "+defaultOauthPath, h)
```

integrations/slack/save.go
```diff
@@ -0,0 +1,70 @@
...
+	headerContentType = "Content-Type"
+	contentTypeForm   = "application/x-www-form-urlencoded"
+)
+
+func (h handler) handleSave(w http.ResponseWriter, r *http.Request) {
+	c, l := sdkintegrations.NewConnectionInit(h.logger, w, r, desc)
+
+	// Check "Content-Type" header.
+	contentType := r.Header.Get(headerContentType)
+	if r.Method == http.MethodPost && !strings.HasPrefix(contentType, contentTypeForm) {
```
daabr: Actually, I think the web UI uses GET even for non-OAuth connections. Please check with @RonenMars, and anyway try to be as close as possible to GitHub's `pat.go`, or Google Calendar/Forms which also do a 2-step flow.

daabr (commented):
Left a few tiny comments, but I think it would help to do a walkthrough together to understand it better. Unfortunately, that won't be possible until Tuesday. Anyway, the direction looks good, and I agree with Itay's previous comment.

sdk/sdkservices/oauth.go
pashafateev: Help me understand the decision behind this for the future. Is it because `integration` and `code` are both strings and the shorthand is being used?

web/static/slack/connect/index.html
pashafateev: I don't understand. What webhook slug? Are you referring to what replaced our http connections? That webhook?
i.e.:
```
  triggers:
    - name: cross_repo
      type: webhook
      event_type: get
      call: program.py:on_cross_repo
```

integrations/slack/oauth.go
pashafateev: I'm not sure. I think this is a relic of when I was trying something different. I removed it.

integrations/slack/internal/vars/vars.go
pashafateev: Done

web/static/slack/connect/index.html
daabr: Oh, my mistake - I kept thinking about having a unique OAuth response webhook, but I guess it's redundant in the way you implemented it (which is good).

sdk/sdkservices/oauth.go
daabr: No, it's because there's an implicit hierarchy: integration ID is the most general/basic detail to identify the entity we're working with, connection ID is a narrower focus than the integration type (single connection compared to multiple connections), and code is the most specific and ephemeral aspect of the connection init flow.

integrations/slack/save.go
RonenMars: Actually for non-OAuth connections we use POST, so we're good.

tebeka: IMO the router should accept only POST requests to the handler. The handler don't need to check.
Also, why do we care about the content-type. Let `ParseForm` fail and return an error.

integrations/slack/save.go
```diff
@@ -0,0 +1,70 @@
...
+	// Read and parse POST request body.
+	if err := r.ParseForm(); err != nil {
+		l.Warn("Failed to parse incoming HTTP request", zap.Error(err))
+		c.AbortBadRequest("form parsing error")
+		return
+	}
+
+	clientID := r.FormValue("client_id")
+	clientSecret := r.FormValue("client_secret")
+	signingSecret := r.FormValue("signing_secret")
```
tebeka, line 39: Validate they are not empty, maybe other limitations?

integrations/slack/webhooks/webhook.go
```diff
@@ -108,21 +111,59 @@ func checkRequest(w http.ResponseWriter, r *http.Request, l *zap.Logger, wantCon
...
 		return nil
 	}
-	signingSecret := os.Getenv(signingSecretEnvVar)
-	if !verifySignature(signingSecret, ts, sig, b) {
-		l.Error("Slack signature verification failed")
+
+	appID, teamID, enterpriseID, err := h.extractIDs(body, wantContentType, l)
+	if err != nil {
+		l.Error("Failed to extract IDs",
+			zap.Error(err),
```
tebeka: Why the newline?

internal/backend/oauth/oauth.go
```diff
@@ -395,6 +397,18 @@ func (o *oauth) Register(ctx context.Context, intg string, cfg *oauth2.Config, o
 	return nil
 }
 
+// TODO: is this needed? if yes and only saving scopes, then this can be simplified
+func (o *oauth) saveConnectionConfig(ctx context.Context, cid sdktypes.ConnectionID, scopes []string) error {
+	vs := sdktypes.NewVars().
+		Set(sdktypes.NewSymbol("scopes"), strings.Join(scopes, ","), false)
```
daabr:
```suggestion
	vs := sdktypes.NewVars().Set(sdktypes.NewSymbol("scopes"), strings.Join(scopes, ","), false)
```

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -409,7 +409,8 @@ func (s *workerGRPCHandler) RefreshOAuthToken(ctx context.Context, req *userCode
 	}
 
 	// Get the integration's OAuth configuration.
-	cfg, _, err := oauth.New(runner.log).Get(ctx, req.Integration)
+	// TODO(ENG-1879): pass vars service instead of nil
```
daabr:
```suggestion
	// TODO(INT-46): pass vars service instead of nil
```

internal/backend/oauth/webhook.go
```diff
@@ -138,3 +146,10 @@ func redirect(w http.ResponseWriter, r *http.Request, intg, cid, origin, param,
 	u := fmt.Sprintf("/%s/oauth?cid=con_%s&origin=%s&%s=%s", intg, cid, origin, param, value)
 	http.Redirect(w, r, u, http.StatusFound)
 }
+
+func transformState(state string) string {
+	if state == "" {
+		return state
+	}
+	return "con_" + strings.Split(state, "_")[0]
```
daabr: I thought the `state` string is already not including the origin (`sub[2]` above). Even though this will work either way, are you sure it's needed?

internal/backend/oauth/oauth.go
```diff
@@ -447,3 +475,12 @@ func authCode(opts map[string]string) []oauth2.AuthCodeOption {
...
 	return acos
 }
+
+func (o *oauth) isCustomOAuth(ctx context.Context, cid sdktypes.ConnectionID) bool {
+	vs, err := o.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+	if err != nil {
+		return false
+	}
+
+	return vs.GetValueByString("client_id") != "" && vs.GetValueByString("client_secret") != ""
```
daabr: We may want to save client ID in the future for all OAuth connection, so only the existence of a client **secret** is the real indication that this is is a custom OAuth connection.

Also, please add function header that explains this (i.e. why "no OAuth client secret connection variable" means "custom OAuth connection").

internal/backend/oauth/oauth.go
```diff
@@ -420,13 +444,17 @@ func (o *oauth) StartFlow(ctx context.Context, intg string, cid sdktypes.Connect
 	// Identify the relevant connection when we get an OAuth response.
 	state := strings.Replace(cid.String(), "con_", "", 1) + "_" + origin
 
+	if o.vars == nil || !o.isCustomOAuth(ctx, cid) {
```
daabr: `o.vars == nil` should never happen  after this PR, in which you initialize `o` with `vars`, so it's an unnecessary check.

integrations/slack/webhooks/webhook.go
```diff
@@ -108,21 +111,59 @@ func checkRequest(w http.ResponseWriter, r *http.Request, l *zap.Logger, wantCon
...
+		)
+		http.Error(w, "Bad Request", http.StatusBadRequest)
+		return nil
+	}
+
+	// Get signing secret
```
daabr:
```suggestion
	// Get signing secret.
```

integrations/slack/webhooks/webhook.go
```diff
@@ -108,21 +111,59 @@ func checkRequest(w http.ResponseWriter, r *http.Request, l *zap.Logger, wantCon
...
+	cids, err := h.listConnectionIDs(r.Context(), appID, enterpriseID, teamID)
+	if err != nil {
+		l.Error("Failed to list connection IDs",
+			zap.Error(err),
+		)
+		http.Error(w, "Bad Request", http.StatusBadRequest)
```
daabr: This is actually an Internal Server Error (500) - this error has nothing to do with the incoming request.

integrations/slack/webhooks/webhook.go
```diff
@@ -108,21 +111,59 @@ func checkRequest(w http.ResponseWriter, r *http.Request, l *zap.Logger, wantCon
...
+		return nil
+	}
+	signingSecret := ""
+	for _, cid := range cids {
```
daabr: No need for loop-and-break, you can safely assume that the first item (if there is one) is correct. Even if there are multiple connections for the same app/enterprise/workspace IDs, they necessarily share the same signing secret (if it exists).

However, please document this subtle point here for posterity.

integrations/slack/webhooks/webhook.go
```diff
@@ -108,21 +111,59 @@ func checkRequest(w http.ResponseWriter, r *http.Request, l *zap.Logger, wantCon
...
+		if s := secret.GetValue(vars.SigningSecret); s != "" {
+			signingSecret = s
+			break
+		}
+	}
+	if signingSecret == "" {
+		signingSecret = os.Getenv(signingSecretEnvVar)
+	}
+
+	// Verify signature
```
daabr:
```suggestion
	// Verify signature.
```

integrations/slack/save.go
daabr: I agree with Miki, partially.

The context here is that in the past we supported both GET and POST, but now this `mux` (and `/save`) are already limited to only POST, so the HTTP method check is unnecessary.

But the content type check is still needed - `r.ParseForm` is not limited to `POST` at all, and may have weird results if we send data in a valid-but-unexpected way.

integrations/slack/save.go
daabr: No need. This is an (uncommon) case where client-side checks are sufficient. If a user sends a request manually, and it's malformed, that's their problem because we fail safely anyway.

integrations/slack/save.go
```diff
@@ -0,0 +1,70 @@
...
+		l.Warn("Failed to save client ID and secret", zap.Error(err))
+		c.AbortBadRequest("failed to save client ID and secret")
+		return
+	}
+
+	redirectURL := fmt.Sprintf("/oauth/start/slack?cid=%s&origin=%s", c.ConnectionID, c.Origin)
+	http.Redirect(w, r, redirectURL, http.StatusFound)
+}
+
+func (h handler) saveClientIDAndSecret(ctx context.Context, c sdkintegrations.ConnectionInit, clientID, clientSecret, signingSecret string) error {
```
daabr:
```suggestion
func (h handler) saveClientIDAndSecrets(ctx context.Context, c sdkintegrations.ConnectionInit, clientID, clientSecret, signingSecret string) error {
```

integrations/slack/save.go
```diff
@@ -0,0 +1,70 @@
...
+		l.Warn("Failed to parse incoming HTTP request", zap.Error(err))
+		c.AbortBadRequest("form parsing error")
+		return
+	}
+
+	clientID := r.FormValue("client_id")
+	clientSecret := r.FormValue("client_secret")
+	signingSecret := r.FormValue("signing_secret")
+
+	if err := h.saveClientIDAndSecret(r.Context(), c, clientID, clientSecret, signingSecret); err != nil {
```
daabr:
```suggestion
	if err := h.saveClientIDAndSecrets(r.Context(), c, clientID, clientSecret, signingSecret); err != nil {
```

integrations/slack/save.go
```diff
@@ -0,0 +1,70 @@
...
+		l.Warn("Failed to save client ID and secret", zap.Error(err))
+		c.AbortBadRequest("failed to save client ID and secret")
```
daabr:
```suggestion
		c.AbortBadRequest("failed to save OAuth configuration")
```

integrations/slack/save.go
pashafateev: Done

integrations/slack/webhooks/webhook.go
pashafateev: Done

integrations/slack/webhooks/webhook.go
pashafateev: Done

internal/backend/oauth/oauth.go

internal/backend/oauth/oauth.go

internal/backend/oauth/webhook.go
pashafateev: you're right, it's not needed

pashafateev:
split into #909 and #910

PR #867, opened 2024-12-05
fix(github): OAuth code missing sometimes

daabr:
When GitHub redirects OAuth flows back to us on success, sometimes the `code` parameter is missing.

Instead of exchanging it for an OAuth token in order to get the app ID (and other details), store the app ID as an environment variable, just just the app name. We don't need OAuth tokens anyway, we generate JWTs.

On the way, we also report an accurate error message if the user tries to initialize a GitHub connection with an app in a GitHub org, but isn't an org admin.

Lastly, we also save more useful debug details in the connection vars.

Refs: INT-7

### DIFF
--- a/integrations/github/jwt.go
+++ b/integrations/github/jwt.go
@@ -46,13 +46,13 @@ func (i integration) NewClient(ctx context.Context) (*github.Client, error) {
 	if pat := data.Get(vars.PAT); pat.IsValid() {
 		return github.NewTokenClient(ctx, pat.Value()), nil
 	} else {
-		return i.newClientWithInstallJWT(data)
+		return newClientWithInstallJWT(data)
 	}
 }
 
-// NewClientWithInstallJWT returns a GitHub client that
+// newClientWithInstallJWT returns a GitHub client that
 // uses a newly-generated GitHub app installation JWT.
-func (i integration) newClientWithInstallJWT(data sdktypes.Vars) (*github.Client, error) {
+func newClientWithInstallJWT(data sdktypes.Vars) (*github.Client, error) {
 	// Initialize and return a GitHub client with a JWT.
 	s := data.GetValue(vars.AppID)
 	if s == "" {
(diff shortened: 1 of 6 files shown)


integrations/github/jwt.go
```diff
@@ -46,13 +46,13 @@ func (i integration) NewClient(ctx context.Context) (*github.Client, error) {
...
 	} else {
-		return i.newClientWithInstallJWT(data)
+		return newClientWithInstallJWT(data)
 	}
 }
 
-// NewClientWithInstallJWT returns a GitHub client that
+// newClientWithInstallJWT returns a GitHub client that
 // uses a newly-generated GitHub app installation JWT.
-func (i integration) newClientWithInstallJWT(data sdktypes.Vars) (*github.Client, error) {
```
pashafateev, line 55: what's the motivation behind making it a standalone function?

integrations/github/oauth.go
```diff
@@ -46,100 +48,98 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
...
+		Set(vars.AppID, appID, false).
+		Set(vars.AppName, string(*i.AppSlug), false).
+		Set(vars.InstallID, installID, false).
+		Set(vars.TargetID, strconv.FormatInt(*i.TargetID, 10), false).
+		Set(vars.TargetName, name, false).
+		Set(vars.TargetType, string(*i.TargetType), false).
+		Set(vars.RepoSelection, string(*i.RepositorySelection), false).
+		Set(vars.Permissions, ps, false).
+		Set(vars.Events, events, false).
+		Set(vars.UpdatedAt, i.UpdatedAt.Format(time.RFC3339), false).
```
pashafateev, line 143: These are being saved but currently not being used anywhere?

itayd:
> > Instead of exchanging it for an OAuth token in order to get the app ID (and other details), store the app ID as an **environment variable**, just just the app name. We don't need OAuth tokens anyway, we generate JWTs.
>
> You mean connection variable, right? 

No, OS env var for the server's default app - see https://github.com/autokitteh/docs.autokitteh.com/pull/133/files - but the app ID is also saved as a connection var, which will support custom OAuth

integrations/github/oauth.go
itayd: @daabr likes to keep this stuff around for debugging

internal/backend/oauth/webhook.go
```diff
@@ -103,6 +103,25 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Special case: GitHub gives us what we need for generating JWTs in the future
+	// (i.e. the GitHub app's installation ID) without exchanging the code below.
+	if intg == "github" {
```
itayd, line 108: ugh, i wonder if there is a way to put this code into the github module or generalize this somehow.

integrations/github/oauth.go
daabr: Exactly - specifically in this case it can help identify and explain issues if an org admin (or another connection init) changes the app's repo selection / approved permissions / selected events.

integrations/github/jwt.go
daabr: Even though I don't use this specific function in this PR, it doesn't use any of the integration structs's fields, so there's no point in keeping it attached to the integration.

internal/backend/oauth/webhook.go
daabr: There are 3 options, none of them makes sense, at least now:

1. Move all the token exchange logic to each integration's handler - violates the DRY principle in a big way
2. Keep this function as a shared function, but not a webhook, and call it in all integrations except GitHub - will require reconfiguration of all existing OAuth apps which is very disruptive
3. Expose a token-exchange-customization function in each integration, but implement only in GitHub - feels like overkill **for now** if only GitHub needs it

pashafateev (approved):

itayd:
> > > Instead of exchanging it for an OAuth token in order to get the app ID (and other details), store the app ID as an **environment variable**, just just the app name. We don't need OAuth tokens anyway, we generate JWTs.
> > 
> > 
> > You mean connection variable, right?
> 
> No, OS env var for the server's default app - see https://github.com/autokitteh/docs.autokitteh.com/pull/133/files - but the app ID is also saved as a connection var, which will support custom OAuth

will this work with custom oauth?

daabr:
> will this work with custom oauth?

Yes!

This env var will be used only once when creating a connection with the default OAuth (to save it as a connection var).

In all other times the connection var will determine the app ID - and that does align with custom OAuth (custom OAuth connections will specify the app ID when initializing the connection instead of reading the server's env var).

So in effect, connections won't care if they're using default or custom OAuth.

PR #870, opened 2024-12-08
feat(python): Warn on I/O outside of an activity

tebeka:
Fixes: ENG-1893


itayd:
is this only for io? i don't see anything limits it, i guess this is what audit is for?

tebeka:
> is this only for io? i don't see anything limits it, i guess this is what audit is for?

It's only for using the `open` function. It's not limiting only warning (per @haimzlato's request)

itayd (approved):

PR #873, opened 2024-12-10
feat: add event handler for "app_home_opened" Slack

mario99logic:
Added a new AppHomeOpenHandler to process app_home_opened events from Slack. This handler validates the event payload and ignores self-triggered events.

Refs: INT-93


integrations/slack/events/app_home_opened.go
```diff
@@ -0,0 +1,39 @@
...
+    // Ignore self-triggered events.
+    for _, a := range cb.Authorizations {
+        if a.UserID == cb.Event.User {
+            l.Debug("Ignoring self-triggered event")
+            return nil
+        }
+    }
+
+    j := &appHomeOpenContainer{}
+    if err := json.Unmarshal(body, j); err != nil {
```
tebeka:
```suggestion
    var j appHomeOpenContainer
    if err := json.Unmarshal(body, &j); err != nil {
```

integrations/slack/events/app_home_opened.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"go.uber.org/zap"
+)
+
+
+type AppHomeOpen struct {
```
tebeka: Why do we need this is we return `any` from `AppHomeOpenHandler`?
You can use a `map[string]any` and get rid of the ad-hoc types.

mario99logic: Yes, I can use map[string]any, but I followed the pattern from app_mention, which also uses ad-hoc types even though its handler function returns any. 
what do you recommend?
@tebeka @daabr

integrations/slack/webhooks/bot_events.go
```diff
@@ -21,6 +21,7 @@ var BotEventHandlers = map[string]BotEventHandler{
 	"app_mention": events.AppMentionHandler,
 	// TODO: app_rate_limit
 	// TODO: app_uninstalled
+	"app_home_opened": events.AppHomeOpenHandler,
```
daabr: Nit: let's keep this ordered alphabetically, i.e. put this line above "app_mention"

integrations/slack/events/app_home_opened.go
daabr: Personally, I try to minimize using `var`. It feels less explicit to me, but I admit it's just a personal stylistic choice.

@tebeka your suggestion works too, but can you explain why you think it's better? If it is, we could do a fix across all other events (in a separate PR).

integrations/slack/events/app_home_opened.go
```diff
@@ -0,0 +1,39 @@
+package events
```
daabr: I tried to create a file for each group of events, not for each event. Please move this into `app_mention.go` and rename the file `app.go`

integrations/slack/events/app_home_opened.go
```diff
@@ -0,0 +1,39 @@
...
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"go.uber.org/zap"
+)
+
+
```
daabr:
```suggestion
// https://api.slack.com/events/app_home_opened
// https://api.slack.com/surfaces/app-home
```

integrations/slack/events/app_home_opened.go
daabr: 1. The type's name should be like the event: `AppHomeOpened` and not `AppHomeOpen` (same in other places in this file)
2. We need to return `any` in order to group all event handlers in a map - see `webhooks/bot_events.go`
3. Incomplete struct specification (missing `view`) - see https://api.slack.com/events/app_home_opened
4. I agree it's possible to use `map` in this specific case, let's try that - ignore other comments about the struct

integrations/slack/events/app_home_opened.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func AppHomeOpenHandler(l *zap.Logger, w http.ResponseWriter, body []byte, cb *Callback) any {
+	// Ignore self-triggered events.
+	for _, a := range cb.Authorizations {
+		if a.UserID == cb.Event.User {
+			l.Debug("Ignoring self-triggered event")
+			return nil
+		}
+	}
```
daabr: Unlike other events can can be triggered by live users as well as bots, this event is **always** triggered by a live user (when the user goes to the bot's "home" in Slack). Therefore this check is unnecessary in this specific event handler.

integrations/slack/events/app.go
```diff
@@ -45,3 +45,23 @@ func AppMentionHandler(l *zap.Logger, w http.ResponseWriter, body []byte, cb *Ca
...
+	}
+
+	if event, ok := eventData["event"]; ok {
+		return event
+	}
+
+	l.Warn("Event key not found in the JSON for app_home_opened event",
+		zap.ByteString("json_body", body),
+	)
+	return nil
```
daabr:
```suggestion
	event, ok := eventData["event"]
	if !ok {
		l.Warn("Event field not found in the Slack event app_home_opened", zap.Any("event", eventData))
		return nil
	}

	return event
```

daabr (approved):
Approved with a few tiny cosmetic fixes to do

integrations/slack/events/app.go
mario99logic: The ok after the semicolon is causing an error because !ok is checked immediately afterward. This can be simplified to avoid unnecessary double-checking.

mario99logic: Another issue is zap.ByteString("event", eventData). The eventData should be marshaled, as zap.ByteString expects the raw JSON body. We can solve that by returning the original body parameter

daabr: Oops, fixed both issues in the suggestion. Also remember to rename `event` to `innerEvent` and `eventData` to `event` before merging.

daabr: Oops! Your fixes are ok, but I made a tiny suggestion on top of them.

integrations/slack/events/app.go
```diff
@@ -27,6 +27,26 @@ type appMentionContainer struct {
...
+	if err := json.Unmarshal(body, &event); err != nil {
+		invalidEventError(l, w, body, err)
+		return nil
+	}
+
+	innerEvent, ok := event["event"]
+	if !ok {
+		l.Warn("Event field not found in the Slack event app_home_opened",
+			zap.ByteString("json_body", body))
+		return nil
```
daabr, line 44:
```suggestion
		l.Warn("Event field not found in the Slack event app_home_opened", zap.Any("event", event))
		return nil
```

PR #874, opened 2024-12-10
feat(python): Add stack information to gRPC errors

tebeka:
Fixes ENG-1895


runtimes/pythonrt/runner/main.py
```diff
@@ -110,6 +110,14 @@ def killIfStartWasntCalled(runner):
         os._exit(1)
 
 
+def abort_with_exception(context, status, err):
+    io = StringIO()
+    print(err, file=io)
```
itayd, line 114: this wont be going to the print hijacking you did?

tebeka: It will, so the user will be able to see it.

daabr (approved):
Not sure what @itayd means in his question (i.e. if there's some concern there), but I'm personally waiting for this

itayd:
> Not sure what @itayd means in his question (i.e. if there's some concern there), but I'm personally waiting for this

he did a print method hijacking to capture all logs. i wonder if it'll go there or some kind of unified error.

PR #892, opened 2024-12-16
fix(python): SDK docs build failing

tebeka:


efiShtain:
you should probably add a workflow to verify that the generated code matches the current code and fail otherwise

tebeka:
> you should probably add a workflow to verify that the generated code matches the current code and fail otherwise

Done

daabr (approved):

daabr (commented):
Don't we need to update the SDK version number in `py-sdk/docs/conf.py` too?

tebeka:
> Don't we need to update the SDK version number in `py-sdk/docs/conf.py` too?

Good idea, added code to read it from `pyproject.toml`

PR #896, opened 2024-12-17
fix(ENG-1910): change web-ui url display on start

RonenMars:
"Locally-delivered resources such as those with http://127.0.0.1 URLs, http://localhost and http://*.localhost URLs (e.g. http://dev.whatever.localhost/), and file:// URLs are also considered to have been delivered securely."
(reference: https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts)

Since we display and run it on HTTP://0.0.0.0 it's not considered secured and we are loosing some functionalities.

### DIFF
--- a/internal/backend/webplatform/webplatform.go
+++ b/internal/backend/webplatform/webplatform.go
@@ -65,7 +65,7 @@ func (w *Svc) Start(context.Context) error {
 	fsrv := http.FileServer(http.FS(webfs))
 
 	srv := &http.Server{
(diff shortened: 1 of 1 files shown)


itayd:
not sure this will work in AWS, since localhost is 127.0.0.1, which means it binds to the loopback interface and not all interfaces. @efiShtain ?

RonenMars:
@itayd On AWS - we won't need it on port 0.0.0.0 - it will always have a separate UI, isn't it?

itayd:
looks like this is for web only, so it's ok.

itayd (approved):

PR #898, opened 2024-12-18
fix: retrieve valid auth0 env var value and define its name as a constant

pashafateev:


integrations/auth0/oauth.go
```diff
@@ -48,7 +52,14 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Test the OAuth token's usability and get authoritative installation details.
-	req, err := http.NewRequest("GET", fmt.Sprintf("https://%s/api/v2/roles", os.Getenv(domain.String())), nil)
+	d := os.Getenv(domainEnvVarName)
+	if d == "" {
+		l.Error("Missing Auth0 domain env var")
+		c.AbortBadRequest("missing Auth0 domain env var")
```
daabr: The user who sees this message doesn't need to know how the domain is stored, only we do

```suggestion
		c.AbortBadRequest("unknown Auth0 domain")
```

daabr (approved):

PR #903, opened 2024-12-18
refactor: CI and release workflows

daabr:
Notable changes:
- Don't run most CI checks when merging an approved PR to the main branch
  - In theory it's good, but we use "main" linearly as a repo setting, and never merge parallel project branches or use release-lifecycle branches, so this is just wasting time and resources with no value, and all failures there are false positives 
- Docker images:
  - Moved "publish Docker image" job from Go CI workflow to a new "Main Branch Merge" workflow (instead of `if: github.ref == 'refs/heads/main'` in the Go CI)
  - Removed `with: fetch-depth: 0` - it means *full* history, not just the latest commit (which is the default behavior) - it's needed only in Go Releaser
  - Decreased timeout from 30m to 10m (actual runtimes are always around 5m)
  - Manual Docker image workflow: no need for `env: WORKING_DIRECTORY: .` ever since we committed to single-repo instead of "fake repo" directories
- Go CI:
  - "Build" step: `make build bin` -> `make bin/ak` (no need to build twice, and it takes twice the time because we use `.phony` make targets)
  - Deleted obsolete file `cmd/docgen/main.go` (related to the build step being redundant)
  - Increased timeout of "session tests" job from 10m to 15m (it's bad and needs to be fixed in the code, but for now we have to face it that even when it works as expected it's too close to 10m)
  - Updated `golangci-lint-action` version in CI workflow and Makefile
  - Added `make webplatform` to the "test-docker-build" jobs (previously named "verify-docker-builds"), because it's used in other Docker jobs

Refs: ENG-2014


efiShtain:
I disagree with the idea that we shouldn't run all check on merge to main
wasting CI minutes is not a valid case to remove everything

we should fix the false positive failures

I'm fine with the rest

daabr:
> I disagree with the idea that we shouldn't run all check on merge to main wasting CI minutes is not a valid case to remove everything

The problem isn't the wasted minutes but the lack of value because of how we work (we never merge parallel project branches or use release-lifecycle branches). But I don't mind adding back checks. Which ones do you consider useful after merging to main?

> we should fix the false positive failures

I agree, but all of them are due to timeouts, not actual failures - this requires a deeper review beyond the scope of this PR, not just increasing the timeout.

.github/workflows/go.yml
```diff
@@ -22,16 +17,17 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
```
tebeka, line 23: I prefer to stay in 3.11 for now until the docker image we use support 3.12

tebeka (approved):
I'm OK with the changes, but please close the loop with @efiShtain

PR #913, opened 2024-12-21
fix(ENG-1914): fix exception handling

itayd:
when an exception that its `str(ex)` is empty is propagated from python the runtime does not know that it happened.
this fixes it by always having a non empty error string in the done message.

this pr also adds a capability for session tests to trim non deterministic paths from exception printing so they can be compared against.

### DIFF
--- a/cmd/ak/cmd/sessions/test.go
+++ b/cmd/ak/cmd/sessions/test.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"regexp"
 	"slices"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


cmd/ak/cmd/sessions/test.go
```diff
@@ -20,8 +21,12 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+var trimPythonExceptionPath bool
+
+var trimRe = regexp.MustCompile(`\/.*\/ak-(user|runner)-.*?\/`)
+
 var testCmd = common.StandardCommand(&cobra.Command{
-	Use:   "test <txtar-file> [--build-id=...] [--env=...] [--deployment-id=...] [--entrypoint=...] [--quiet] [--timeout DURATION] [--poll-interval DURATION] [--no-timestamps]",
```
daabr, line 24: this shouldn't be an optional flag, this should always be the behavior

daabr (approved):

cmd/ak/cmd/sessions/test.go
itayd: done

PR #914, opened 2024-12-21
fix: `ak up -m test` get stuck sometimes

itayd:
looks like the session tests get stuck because temporal start dev fails to run from the binary. this attempts to start it multiple times on failure. i tested it lots of times and now it looks to be solved using this very heavy handed fix.

we should consider not starting temporal ourselves, but that might be bothersome to the users in test mode.


daabr (approved):
LGTM, you should be ashamed of this :)

FYI, I filed INT-1924 to fix this better in the future, but not anytime soon.

PR #920, opened 2024-12-24
feat: Add HubSpot connTest

mario99logic:
add connTest function for HubSpot integration 

refs: INT-105

### DIFF
--- a/integrations/hubspot/client.go
+++ b/integrations/hubspot/client.go
@@ -2,6 +2,8 @@ package hubspot
 
 import (
 	"context"
+	"fmt"
+	"net/http"
 
 	"go.autokitteh.dev/autokitteh/integrations"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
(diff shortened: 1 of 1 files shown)


mario99logic:
@daabr Question: what do I return in line 98 in case resp.StatusCode != http.StatusOK

integrations/hubspot/client.go
```diff
@@ -63,9 +65,39 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+		if !cid.IsValid() {
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, "Init required"), nil
+		}
+
+		vs, err := i.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		token := vs.GetValueByString("oauth_AccessToken")
```
daabr: This isn't sufficient because the access token stored when you create the connection is effective only for 30 minutes. You need to read `oauth_RefreshToken` and use https://developers.hubspot.com/docs/reference/api/app-management/oauth#get-information-for-oauth-2.0-refresh-token

daabr: committed a fix

pashafateev (approved):

PR #921, opened 2024-12-24
feat: Add PAT option to Atlassian connTest 

mario99logic:
Add case to check if the PAT connection is valid for jira and confluence connTest functions 

Refs: INT-77

### DIFF
--- a/integrations/atlassian/confluence/client.go
+++ b/integrations/atlassian/confluence/client.go
@@ -2,6 +2,8 @@ package confluence
 
 import (
 	"context"
+	"fmt"
+	"net/http"
 
 	"go.autokitteh.dev/autokitteh/integrations"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
@@ -9,6 +11,7 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+	"go.uber.org/zap"
 )
 
 type integration struct {
@@ -86,17 +89,60 @@ func connTest(i *integration) sdkintegrations.OptFn {
 			return sdktypes.InvalidStatus, err
 		}
 
-		baseURL, err := apiBaseURL()
+		at := vs.Get(authType)
+		if !at.IsValid() || at.Value() == "" {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
+		}
+
+		oauthURL, err := apiBaseURL()
 		if err != nil {
 			return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
 		}
 
-		token := vs.GetValueByString("oauth_AccessToken")
+		switch at.Value() {
+		case integrations.OAuth:
+			token := vs.GetValueByString("oauth_AccessToken")
+			_, err := accessibleOAuth(nil, oauthURL, token)
+			if err != nil {
+				return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
+			}
+		case integrations.APIToken: // Isnt it supposed to be PAT?
+			patURL := vs.Get(baseURL).Value()
+			email := vs.Get(email).Value()
+			token := vs.Get(token).Value()
+			err := accessiblePAT(nil, patURL, email, token)
+			if err != nil {
+				return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
+			}
 
-		_, err = accessibleResources(nil, baseURL, token)
-		if err != nil {
-			return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
 		}
+
 		return sdktypes.NewStatus(sdktypes.StatusCodeOK, ""), nil
 	})
+}
+
+func accessiblePAT(l *zap.Logger, baseURL string, email string, token string) error {
+	u := fmt.Sprintf("%s/rest/api/3/myself", baseURL)
+	req, err := http.NewRequest("GET", u, nil)
+	if err != nil {
+		logWarnIfNotNil(l, "Failed to construct HTTP request for PAT test", zap.Error(err))
+		return err
+	}
+
+	req.SetBasicAuth(email, token)
+	req.Header.Set("Accept", "application/json")
+
+	resp, err := httpClient.Do(req)
+	if err != nil {
+		logWarnIfNotNil(l, "Failed to request accessible resources for PAT token", zap.Error(err))
+		return err
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		logWarnIfNotNil(l, "Unexpected response on accessible resources", zap.Int("status", resp.StatusCode))
+		return fmt.Errorf("accessible resources: unexpected status code %d", resp.StatusCode)
+	}
+
+	return nil
 }
--- a/integrations/atlassian/confluence/oauth.go
+++ b/integrations/atlassian/confluence/oauth.go
@@ -53,7 +53,7 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Test the OAuth token's usability and get authoritative installation details.
-	res, err := accessibleResources(l, url, oauthToken.AccessToken)
+	res, err := accessibleOAuth(l, url, oauthToken.AccessToken)
 	if err != nil {
 		c.AbortBadRequest(err.Error())
 		return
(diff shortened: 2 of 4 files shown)


mario99logic:
This is the initial version of the changes. I have the following questions:
1.Should we have a single connTest function for both Jira and Atlassian, or keep them separate?
2.Why Do we get apiToken instead of PAT (had to make a case for apiToken instead of PAT line 109), is it related to the frontend?

integrations/atlassian/confluence/client.go
```diff
@@ -86,17 +89,60 @@ func connTest(i *integration) sdkintegrations.OptFn {
...
 
-		token := vs.GetValueByString("oauth_AccessToken")
+		switch at.Value() {
+		case integrations.OAuth:
+			token := vs.GetValueByString("oauth_AccessToken")
+			_, err := accessibleOAuth(nil, oauthURL, token)
+			if err != nil {
+				return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
+			}
+		case integrations.APIToken: // Isnt it supposed to be PAT?
```
pashafateev, line 109: https://support.atlassian.com/atlassian-account/docs/manage-api-tokens-for-your-atlassian-account/

For our use cases Atlassian uses api tokens and not PAT. @daabr please correct me if I'm wrong, but I think Atlassian uses PAT for their Enterprise products. Or at least for products that we currently don't support. To be frank, I don't completely understand it myself.

integrations/atlassian/confluence/client.go
```diff
@@ -86,17 +89,60 @@ func connTest(i *integration) sdkintegrations.OptFn {
 			return sdktypes.InvalidStatus, err
 		}
 
-		baseURL, err := apiBaseURL()
+		at := vs.Get(authType)
+		if !at.IsValid() || at.Value() == "" {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
+		}
+
+		oauthURL, err := apiBaseURL()
```
pashafateev, line 97: why did you rename `baseURL` to `oauthURL`?

integrations/atlassian/confluence/client.go
```diff
@@ -86,17 +89,60 @@ func connTest(i *integration) sdkintegrations.OptFn {
...
-		_, err = accessibleResources(nil, baseURL, token)
-		if err != nil {
-			return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
 		}
+
 		return sdktypes.NewStatus(sdktypes.StatusCodeOK, ""), nil
 	})
 }
+
+func accessiblePAT(l *zap.Logger, baseURL string, email string, token string) error {
```
pashafateev, line 124: Are you sure `accessiblePAT` and `accessibleOAuth` need to be two separate functions?
If yes, I personally would rename this function. It doesn't return resources like the function in `oauth.go`. 

I would call it something like `validateAuth` or `verifyAPIToken` or something like that.
What do you think @daabr?

integrations/atlassian/confluence/oauth.go
```diff
@@ -89,11 +89,11 @@ type resource struct {
 	AvatarURL string   `json:"avatarUrl"`
 }
 
-// accessibleResources retrieves the Atlassian Cloud metadata associated with
+// accessibleOAuth retrieves the Atlassian Cloud metadata associated with
 // an OAuth token, which is necessary for API calls and webhook events. Based on:
 // https://developer.atlassian.com/cloud/confluence/oauth-2-3lo-apps/#3--make-calls-to-the-api-using-the-access-token
 // https://developer.atlassian.com/cloud/jira/platform/oauth-2-3lo-apps/#3--make-calls-to-the-api-using-the-access-token
-func accessibleResources(l *zap.Logger, baseURL string, token string) ([]resource, error) {
+func accessibleOAuth(l *zap.Logger, baseURL string, token string) ([]resource, error) {
```
pashafateev, line 96: I would leave the old name because it's returning resources. See comment above.
@daabr I also realized I don't understand the purpose of these resources.

pashafateev (commented):
All my comments for Confluence apply to Jira as well

integrations/atlassian/confluence/client.go
daabr: Correct: Atlassian Cloud uses "API tokens" and Atlassian on-prem (whatever it's called) uses "Personal Access Tokens". Effectively, they're **exactly the same thing**, so because we focus on Atlassian Cloud we use `APIToken`.

integrations/atlassian/confluence/client.go
mario99logic: When adding this line patURL := vs.Get(baseURL).Value(), I got an error because it confused the baseURL from vars.go with baseURL from baseURL, err := apiBaseURL()

integrations/atlassian/confluence/oauth.go
daabr: I agree with @pashafateev - the API endpoint we use is called "accessible-resources", even though it's specific to OAuth, so we should preserve that name.

To answer your question: it lists the domain's details and scopes available for the app, as a proof that it's installed.

integrations/atlassian/confluence/client.go
mario99logic: The main difference between the two functions, aside from the different URLs (which isn’t an issue if we combine them), is the warning comments—one specific to OAuth and the other to PAT. Another difference is that the OAuth function returns resources, though these aren’t used when calling the function. However, if we want to make changes to the OAuth function, it should be handled in a separate ticket in my opinion.

daabr:
> 1. Should we have a single connTest function for both Jira and Atlassian, or keep them separate?

To clarify, you're asking about "Jira" and "Confluence" - both being "Atlassian" products. The answer is I don't know for sure, it depends on how we can test each one's API token / PAT. But given they're very similar, it's logical the same URL endpoint / API call is available for both.

> 2. Why Do we get apiToken instead of PAT (had to make a case for apiToken instead of PAT line 109), is it related to the frontend?

See my reply to Pasha's comment about this. tl;dr - "API key" is for Atlassian Cloud (which is our focus), "PAT" is for Atlassian on-prem (which we support but isn't our focus). Effectively though, they're exactly the same, just different names.

integrations/atlassian/confluence/client.go
```diff
@@ -2,13 +2,16 @@ package confluence
...
+	"fmt"
+	"net/http"
 
 	"go.autokitteh.dev/autokitteh/integrations"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/sdk/sdkintegrations"
 	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+	"go.uber.org/zap"
```
daabr, line 14: This is an incorrect location for this import. Please rebase your PR to main - After #956 golangci-lint should fix this automatically when your format.

integrations/atlassian/confluence/client.go
daabr: @mario99logic @pashafateev important correction: I was mistaken in my previous comment, I see in `save.go` that there IS a difference between (cloud) API tokens and (on-prem) PATs: if the user saves a connection with a secret and an email address it's  a (cloud) API token; if the user saves a connection with a secret but without an email address it's an (on-prem) PAT.

See:
- Confluence: https://github.com/autokitteh/autokitteh/blob/0a71ffed2dcab58ae6eb7246be4707ed35d19a0d/integrations/atlassian/confluence/save.go#L64-L69
- Jira: https://github.com/autokitteh/autokitteh/blob/0a71ffed2dcab58ae6eb7246be4707ed35d19a0d/integrations/atlassian/jira/save.go#L51-L56

So to clarify the task in this PR: you need to implement conn checks for both API keys and PATs, not just PATs. In face, you won't be able to test PATs for now.

integrations/atlassian/confluence/client.go
daabr: @mario99logic yes, but "OAuth path" is an incorrect term here.

My first suggestion is that instead of changing this function, you rename it to `oauthConnTest`, and create new `apiKeyConnTest` (based on your `accessiblePAT` function) and `patConnTest`, and then a new `connTest` function that chooses which one to run. Combining all of them together is too messy.

Unrelated to that, you're correct that the local var `baseURL` is shadowing the package var `baseURL`. My suggestion is that you rename the local var to `bu` or something like that.

integrations/atlassian/confluence/client.go
daabr: Yes, I agree too. This needs to be a separate function from OAuth, and needs to have a unique name because it doesn't get accessible resources, it deals with user identity. Also please add API link(s) as documentation, like in `accessibleResources`.

integrations/atlassian/confluence/client.go
mario99logic: I think you mean connTest here and not connStatus? 
Instead of renaming the function and making a new connTest which chooses which one to run I will keep the current connTest because it does exactly that I will optimize it and make it more clear that it chooses between apiKeyConnTest and oauthConnTest, the new function will be OauthConnTest I will create it to be able to use accessableResources (to keep the name "accessableResources")

PR #925, opened 2024-12-25
fix(python): Handle exceptions in activities.

tebeka:
Fixes: ENG-1913


tests/sessions/python_exception.txtar
```diff
@@ -1,12 +1,6 @@
...
-Traceback (most recent call last):
-  File "runner/main.py", line 323, in on_event
-    result = fn(event)
-             ^^^^^^^^^
-  File "main.py", line 4, in main
-    raise X()
-main.X
+Error: X()
+ [0] /tmp/ak-runner-2212711217/runner/main.py:318.1,_call _call
+ [1] /tmp/ak-user-1034797789/handler.py:6.1,main main
```
daabr: The full path is useless and confusing for users. It's better to always remove the prefix `.../ak-*/`. For example:

```
for file, line, func, text in traceback.extract_stack():
    file = re.sub(r'^.+/ak-(runner|user)-.+?/', "", file)
```

But feel free to do it in a different PR.

runtimes/pythonrt/runner/main.py
```diff
@@ -46,17 +46,15 @@ def parse_entry_point(entry_point):
     return file_name[:-3], func_name
 
 
-def exc_traceback(err):
-    """Format traceback to JSONable list."""
-    te = TracebackException.from_exception(err)
+def pb_traceback(tb):
```
daabr, line 49: Why remove the docstring?

daabr (commented):
I left a few tiny comments, but I tried to review this PR but I can't.

I can't determine if your changes work, if there are any general approach oversights, or minor details that are wrong, or if you're adding any new regressions by mistake.

Combine that with the lack of documentation - in the PR, and the existing code in general.

I need a deeper walkthrough to understand this PR, and also I recommend you spend a few days on documentation afterwards.

runtimes/pythonrt/runner/main.py
tebeka: It's a different function, will add docstring.

tests/sessions/python_exception.txtar
tebeka: I agree, I also want to remove the the lines of the stack trace coming from the runner code.
Created ENG-1932

runtimes/pythonrt/runner/main.py
```diff
@@ -262,22 +247,26 @@ def ActivityReply(
...
             )
 
-        if request.error:
-            call.fut.set_exception(ActivityError(request.error))
-            context.abort(
-                grpc.StatusCode.ABORTED,
-                f"activity error: {request.error}",
-            )
+        if result.error:
+            # TODO: Restore exception traceback
```
itayd: can you explain?

itayd (commented):
i don't see you using ProgramError for relaying uncaught exceptions. Am I right? The worry is that it wont show correctly, or result in a proper error in the session log.

runtimes/pythonrt/runner/main.py
tebeka: When you pickle an exception, the stored data does not have the traceback associated with the exception (you can't pickle a stack trace). I want to add the traceback back to the exception somehow, though not urgent since it's printed out.

tebeka:
> i don't see you using ProgramError for relaying uncaught exceptions. Am I right? The worry is that it wont show correctly, or result in a proper error in the session log.

There's a ProgramError in `Done`. See `pythonrt.go`

itayd:
oh wow all tests pass. is this ready @tebeka ?

runtimes/pythonrt/runner/main.py
```diff
@@ -100,21 +99,43 @@ def fix_http_body(event):
             pass
 
 
-def killIfStartWasntCalled(runner):
```
efiShtain, line 103: why?

runtimes/pythonrt/runner/main.py
```diff
@@ -209,49 +230,43 @@ def Start(self, request: pb.runner.StartRequest, context: grpc.ServicerContext):
...
-        result = err = None
+        result = self._call(call.fn, call.args, call.kw)
         try:
-            result = call.fn(*call.args, **call.kw)
-        except Exception as e:
-            log.error("%s failed: %s", func_name, e)
-            display_err(call.fn, e)
-            err = e
+            data = pickle.dumps(result)
+        except (TypeError, pickle.PickleError) as err:
```
efiShtain, line 241: what about other errors ?

runtimes/pythonrt/runner/main.py
tebeka: It wasn't used anywhere.

runtimes/pythonrt/runner/main.py
tebeka: These are the errors pickle raises. The rule-of-thumb in Python is to catch specific exceptions and not `Exception`. If you catch `Exception` all the time you might mask errors you don't know how to handle.

tebeka:
> oh wow all tests pass. is this ready @tebeka ?

IMO yes.

runtimes/pythonrt/runner/main.py
efiShtain: that's probably a bug
I think it should be used
but we can leave it to a different PR

cmd/ak/cmd/sessions/test.go
```diff
@@ -21,7 +22,31 @@ import (
...
-var trimRe = regexp.MustCompile(`\/.*\/ak-(user|runner)-.*?\/`)
+var (
+	// /tmp/ak-user-2767870919/main.py:6.1,main
+	userRe = regexp.MustCompile(`\/.*\/ak-user-.*?\/`)
+	// runner/main.py:6.1,main, in _call
+	runnerRe = regexp.MustCompile(`.*runner.*/.*\.py.*`)
+
+	// File "/opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result`
+	pyLibRe = regexp.MustCompile(`File ".*/lib/python3\.\d+/(.*\.py)", line (\d+), in (.*)`)
+)
```
efiShtain, line 33: is this linux only? 
or will it work on a Mac as well ?

runtimes/pythonrt/pythonrt.go
```diff
@@ -384,6 +388,11 @@ func (py *pySvc) call(ctx context.Context, val sdktypes.Value, args []sdktypes.V
 
 	if _, err = py.runner.ActivityReply(ctx, &req); err != nil {
 		py.log.Error("activity reply error", zap.Error(err))
+		req := pbUserCode.DoneRequest{
+			RunnerId: py.runnerID,
+			Error:    err.Error(),
+		}
+		py.channels.done <- &req
```
efiShtain, line 395: so you mimic the runner code here assuming if you can't call activity reply then you send s "done request" to the channel ?
it feels a bit weird
the health check should handle this one I think

runtimes/pythonrt/runner/main.py
```diff
@@ -209,49 +230,43 @@ def Start(self, request: pb.runner.StartRequest, context: grpc.ServicerContext):
...
-            tb = exc_traceback(err)
-            resp.traceback.extend(tb)
-
         return resp
 
     def ActivityReply(
         self, request: pb.runner.ActivityReplyRequest, context: grpc.ServicerContext
     ):
-        if request.error and not request.result.custom.data:
+        if not request.result.custom.data:
```
efiShtain, line 261: any chance custom is not define ?

runtimes/pythonrt/runner/main.py
```diff
@@ -209,49 +230,43 @@ def Start(self, request: pb.runner.StartRequest, context: grpc.ServicerContext):
...
-                error=request.error,
+                error="activity reply not a Custom value",
             )
-            resp = self.worker.Done(req)
-            if resp.error:
-                log.error("on_event: done send error: %r", resp.error)
+            try:
+                self.worker.Done(req)
+            except grpc.RpcError as err:
+                log.error("done send error: %r", err)
```
efiShtain, line 269: maybe we should crash the runner here?
if it can't send the done request, it won't be able to run the activityReplyResponse any way I think

cmd/ak/cmd/sessions/test.go
tebeka: Looks like it :)

runtimes/pythonrt/runner/main.py
```diff
@@ -209,49 +230,43 @@ def Start(self, request: pb.runner.StartRequest, context: grpc.ServicerContext):
...
-        except Exception as e:
-            log.error("%s failed: %s", func_name, e)
-            display_err(call.fn, e)
-            err = e
+            data = pickle.dumps(result)
+        except (TypeError, pickle.PickleError) as err:
+            # Print so it'll get to session log
+            print(f"error: cannot pickle result - {err}")
+            print(pickle_help)
+            context.abort(grpc.StatusCode.INTERNAL, f"can't pickle result - {err}")
```
efiShtain, line 245: why context abort instead of telling ak there is an error and call done ?

runtimes/pythonrt/runner/main.py
```diff
@@ -100,21 +99,43 @@ def fix_http_body(event):
...
     io = StringIO()
-    print(err, file=io)
     for line in format_exception(err):
         io.write(line)
-    context.abort(status, io.getvalue())
+    text = io.getvalue()
+    context.abort(status, text)
+
+
+def set_exception_args(err):
```
efiShtain, line 110: so this is enriching an exception ?

runtimes/pythonrt/pythonrt.go
tebeka: I want to terminate the flow, this is the way to do it. The health check won't get the right error to the session log.

runtimes/pythonrt/runner/main.py
tebeka: Nope:

```python
In [1]: import pb
In [2]: req = pb.runner.ActivityReplyRequest()
In [3]: req.result
Out[3]:
In [4]: req.result.custom
Out[4]:
In [5]: req.result.custom.data
Out[5]: b''
```

runtimes/pythonrt/runner/main.py
tebeka: Agree and done.

runtimes/pythonrt/runner/main.py
tebeka: No, it about making the exception unpicklable. See https://stackoverflow.com/questions/41808912/cannot-unpickle-exception-subclass (it's in the function docstring).

runtimes/pythonrt/pythonrt.go
efiShtain: got ya, ok
weird but fine

efiShtain (approved):
that's a high risk change
it's really hard to understand that changes here 
but the main idea makes sense

should be tested very well once we merge it

runtimes/pythonrt/runner/main.py
tebeka: Opened ENG-1963

runtimes/pythonrt/runner/main.py
tebeka: When `Execute` returns and error, the `Call` in pythonrt.go (line 657) will return error and we fail the workflow. This can be an issue with user code that returns non-pickleable value and we print the exception. Do you think we should close the runner as well?

daabr: Your explanation sounds like an excellent comment to add to the code

cmd/ak/cmd/sessions/test.go
```diff
@@ -21,7 +22,31 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-var trimRe = regexp.MustCompile(`\/.*\/ak-(user|runner)-.*?\/`)
+var (
+	// /tmp/ak-user-2767870919/main.py:6.1,main
+	userRe = regexp.MustCompile(`\/.*\/ak-user-.*?\/`)
```
daabr, line 27: This won't work on Windows, and while we don't officially support it right now, we shouldn't add new issues to fix down the road, or at least we should document them with comments, or (even better) TODO comments with a linked issue.

cmd/ak/cmd/sessions/test.go
```diff
@@ -21,7 +22,31 @@ import (
...
+	// runner/main.py:6.1,main, in _call
+	runnerRe = regexp.MustCompile(`.*runner.*/.*\.py.*`)
+
+	// File "/opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result`
+	pyLibRe = regexp.MustCompile(`File ".*/lib/python3\.\d+/(.*\.py)", line (\d+), in (.*)`)
```
daabr, line 32: 1. `, in (.*)` is unnecessary for your purposes
2. Why `/lib/`? This feels overly specific

cmd/ak/cmd/sessions/test.go
```diff
@@ -21,7 +22,31 @@ import (
...
+)
+
+func normalizePath(p string) string {
+	// Remove location specific prefix of Python standard library.
+	line := pyLibRe.ReplaceAllString(p, `py-lib/$1, line XXX, in $3`)
```
daabr, line 37: Non-blocker comments:
1. Looks like your regex is reversed: you want to replace the path, but you're treating everything **else** as a submatch instead
2. I don't **really** mind, but why are you dropping the double-quotes around the file path?
3. I'm assuming you're hiding the line number for test stability. If so, don't specify it at all, since this code affects only session tests
4. "python" is more familiar and potentially correct compared to "py-lib"

daabr (approved):

PR #928, opened 2024-12-26
fix: proper migration from 0.11.x versions to 0.12.x users and orgs

itayd:
When the authz migration is done, we need to backfill proper user and orgs configurations. For concrete example, all users now need to have a default org id which points to a real org, and be a member of this org. Also, all projects need to have an org associated with, which this migration sets as the project's owner personal org.

### DIFF
--- a/internal/backend/db/dbgorm/dbgorm.go
+++ b/internal/backend/db/dbgorm/dbgorm.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"sync"
+	"time"
 
 	"github.com/google/uuid"
 	"github.com/pressly/goose/v3"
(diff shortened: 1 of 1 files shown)


internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -158,7 +167,146 @@ func (db *gormdb) Migrate(ctx context.Context) error {
...
+	}
+
+	if err := db.backfillUsersAndOrgs(ctx); err != nil {
+		return fmt.Errorf("backfill users and orgs: %w", err)
+	}
+
+	return nil
+}
+
+func (db *gormdb) backfillUsersAndOrgs(ctx context.Context) error {
```
daabr, line 186: 1. Document this function *in detail*, and add a TODO with a ticket ID for removing it
2. If this function runs only when AK does a migration, it actually won't run in production unless we run it explicitly, so what are you gaining here?

daabr (commented):
Missing detailed PR description and in-code documentation for this sensitive PR: what is the migration exactly? How and when are you doing it? Why this approach and not a regular migration script? Side effects or lack thereof? Etc.

daabr (approved):

internal/backend/db/dbgorm/dbgorm.go
itayd: 1. done
2. we need to discuss proper procedures re prod here, out of scope for oss.

PR #938, opened 2024-12-26
feat(opa): do not report logs < warn by default

itayd:
- do not report opa logs at level less than WARN by default.
- authz now reports debug log for any decision.
- simplify opa bundle loading configuration, now if path is empty - load default embedded config.

all these will cause during normal operation to not have >= info logs from opa. if there is any error in opa we will see logs. we will still see warnings on unauthorized access attempts.

### DIFF
--- a/internal/backend/auth/authz/policy.go
+++ b/internal/backend/auth/authz/policy.go
@@ -44,8 +44,11 @@ func NewPolicyCheckFunc(l *zap.Logger, db db.DB, decide policy.DecideFunc) Check
 			return fmt.Errorf("authz opa decision: not a boolean")
 		}
 
+		l := l.With(zap.Any("input", input), zap.Any("result", result))
+		l.Debug("authz opa decision")
+
 		if !decision {
-			l.Warn("authz opa decision: denied", zap.Any("input", input), zap.Any("result", result))
+			l.Warn("authz opa decision: denied")
 
 			if cfg.convertForbiddenToNotFound {
 				return sdkerrors.ErrNotFound
--- a/internal/backend/policy/opapolicy/opa.go
+++ b/internal/backend/policy/opapolicy/opa.go
@@ -10,7 +10,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/open-policy-agent/contrib/logging/plugins/ozap"
 	"github.com/open-policy-agent/opa/v1/sdk"
 	sdktest "github.com/open-policy-agent/opa/v1/sdk/test"
 	"go.uber.org/zap"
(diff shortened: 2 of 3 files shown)


internal/backend/auth/authz/policy.go
```diff
@@ -44,8 +44,11 @@ func NewPolicyCheckFunc(l *zap.Logger, db db.DB, decide policy.DecideFunc) Check
 			return fmt.Errorf("authz opa decision: not a boolean")
 		}
 
+		l := l.With(zap.Any("input", input), zap.Any("result", result))
+		l.Debug("authz opa decision")
```
daabr, line 48: There are too many decisions, so there's no need to log *every* one, only denials.

internal/backend/policy/opapolicy/opa.go
```diff
@@ -21,18 +20,17 @@ import (
...
 
-const defaultConfig = "default"
+// Embedded config path in opa_bundles.FS.
+const embeddedDefaultConfigPath = "default"
 
 type Config struct {
-	// If empty, use bundled config with the name `defaultConfig`.
-	// If begings with "!", use bundled config with that name (without the "!" prefix).
-	ConfigPath string `koanf:"config_path"`
+	ConfigPath         string `koanf:"config_path"`       // if empty, use embedded default config.
```
daabr, line 27: Having the policy's path as a configurable parameter seems like a security risk (allows an attacker to modify the server without oversight), and there's no real reason for it anyway.

If you want to modify the policy, just rebuild AK with it.

If you want to allow the user to select between different existing flavors - that's fine and safe, as long as we don't allow arbitrary customizations, only a selection between options.

internal/backend/policy/opapolicy/ozap.go
```diff
@@ -0,0 +1,114 @@
+// This is an adaptation of https://github.com/open-policy-agent/contrib/blob/main/logging/plugins/ozap/zap.go.
```
daabr, line 1: Good comment!

internal/backend/auth/authz/policy.go
itayd: hence why this is debug - i think that's proper.

internal/backend/policy/opapolicy/opa.go
itayd: the config path is not the policy path. the config path is a path to server configuration where it can specify in that configuration where the policy is. see https://www.openpolicyagent.org/docs/latest/configuration/.

if an attacker can modify our deployment env vars, we have much, much bigger issues.

internal/backend/policy/opapolicy/ozap.go
itayd: awwwww

internal/backend/auth/authz/policy.go
daabr: Fine, we can remove it later when we have more confidence.

daabr (approved):
Approved - but see my comment about configurable policy paths, but I'm not sure it's worthwhile solving this in this specific PR.

itayd:
> Approved - but see my comment about configurable policy paths, but I'm not sure it's worthwhile solving this in this specific PR.

i answered it - not sure it needs solving.

PR #940, opened 2024-12-27
fix(db): remove obsolete config param `ownership`

itayd:


itayd:
this is a trivial change, merging. @daabr lmk if any issues.

PR #941, opened 2024-12-27
fix(auth0): use system user context in OAuth exchange

daabr:
Small fix, but the reasoning behind it took a lot of deep thinking and nuance.

### DIFF
--- a/internal/backend/oauth/oauth.go
+++ b/internal/backend/oauth/oauth.go
@@ -19,6 +19,7 @@ import (
 	googleoauth2 "google.golang.org/api/oauth2/v2"
 	"google.golang.org/api/sheets/v4"
 
+	"go.autokitteh.dev/autokitteh/internal/backend/auth/authcontext"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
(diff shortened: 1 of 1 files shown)


internal/backend/oauth/oauth.go
```diff
@@ -507,9 +508,13 @@ func (o *oauth) StartFlow(ctx context.Context, intg string, cid sdktypes.Connect
...
-	// Convert the received temporary authorization code
-	// into a refresh token / user access token.
-	cfg, opts, err := o.getConfigWithConnection(ctx, integration, cid)
+	// Convert the received temporary authorization code into a refresh token / user access token.
+	// ATTENTION: This method may be called by an UNAUTHENTICATED webhook handler, however we CAN
+	// use the system user (SetAuthnSystemUser) securely, because only the 3rd-party OAuth
+	// provider can provide us a valid "code" parameter related to the connection ID. In other
+	// words, if the "code" is fake, or the CID is valid-but-unrelated to the beginning of the
+	// OAuth flow, the OAuth provider will reject the exchange, so data leakage is not possible.
+	cfg, opts, err := o.getConfigWithConnection(authcontext.SetAuthnSystemUser(ctx), integration, cid)
```
pashafateev, line 517:
```suggestion
	// Convert the received temporary authorization code into a refresh token / user access token.
	// ATTENTION: This method may be called by an UNAUTHENTICATED webhook handler. 
	// Using the system user (SetAuthnSystemUser) is secure because only the 
	// 3rd-party OAuth provider can issue a valid "code" parameter for the connection ID. 
	// If the "code" is fake or the CID is unrelated to the OAuth flow, 
	// the provider will reject the exchange, preventing data leakage.
	cfg, opts, err := o.getConfigWithConnection(authcontext.SetAuthnSystemUser(ctx), integration, cid)
```

pashafateev: nit: Edited by chatGPT. I think it's shorter and easier to read, but it's not a big deal.

pashafateev (approved):
approved with nit

PR #947, opened 2024-12-28
feat(orgs,users): cli; updates with fieldmask

itayd:
when updating orgs and users, now it is possible to optionally specify a fieldmask in order to update specific fields only. this is backward compatible - if fieldmask is not specified, previous behavior remains unchanged.

only top level fields are currently permitted.


cmd/ak/cmd/orgs/add_member.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var addMemberCmd = common.StandardCommand(&cobra.Command{
+	Use:     "add-member <org id> <user id>",
```
efiShtain, line 13: I think it's finally time to introduce context to the cli

context should include current org, current project, current token

so most commands won't require the basic args like org id and project id and users would be able to switch context easily

something like

```bash
# Create Context
ak context create --name "my context" --token my_token --server https://api.autokitteh.cloud # use default org , server is cloud by default, can be localhost by default or by setting --local instead of server

ak context create --name "my context" --token my_token --org org_id # use any org id allowed for token

ak context create --name "my context" --token my_token --org org_id --project project_id # use specific project, not sure this is required but can be useful

# List Context
ak context list

# Switch context
ak context use context_name
```

internal/backend/db/dbgorm/dbtime/dbtime.go
```diff
@@ -0,0 +1,13 @@
+package dbtime
```
daabr: You can merge this as-is, but in my opinion it's misleading to call this "DB time" when the purpose is to enable testing with fake time, and the usability of this utility extends beyond the DB.

internal/backend/db/dbgorm/dbtime/dbtime.go
```diff
@@ -0,0 +1,13 @@
...
+
+import "time"
+
+var now = func() time.Time { return time.Now() }
+
+func Now() time.Time { return now() }
+
+// Used for testing.
+func Freeze() {
```
daabr: 1. The comment is useful in an IDE but not in a code review, if the function name also reflected that it's only for testing it would be safer (so reviewers who see this being called know that it must be only in tests)
2. Nit: the comment should be "<Function name> is used for testing"

internal/kittehs/proto.go
```diff
@@ -28,3 +31,44 @@ func MarshalProtoMapJSON[K comparable, V proto.Message](xs map[K]V) ([]byte, err
 
 	return json.Marshal(rs)
 }
+
+// ProtoToMap converts a protobuf message to a map, using the provided field mask.
+// If fm is nil, all fields are included.
+// This function deals only with top level fields, nested fields are not supported.
+func ProtoToMap(pb proto.Message, fm *fieldmaskpb.FieldMask) (map[string]interface{}, error) {
```
daabr, line 39: Maybe I'm missing something, but why aren't you just converting the proto to JSON and then to a map?

internal/kittehs/proto_test.go
```diff
@@ -0,0 +1,32 @@
...
+	pb := &apipb.Method{
+		Name:              "meow",
+		RequestTypeUrl:    "woof",
+		ResponseStreaming: true,
+		Syntax:            1,
+	}
+
+	m, err := ProtoToMap(pb, Must1(fieldmaskpb.New(pb, "name", "response_streaming", "syntax")))
+	if assert.NoError(t, err) {
+		assert.Equal(t, map[string]interface{}{
```
daabr, line 21: Another fundamental case to add here: a message field, i.e. a sub-map.

cmd/ak/cmd/users/getorgs.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var getOrgsCmd = common.StandardCommand(&cobra.Command{
+	Use:   "get-orgs [email or id] [--fail]",
```
daabr:
```suggestion
	Use:   "get-orgs [user email or ID] [--fail]",
```

cmd/ak/cmd/orgs/update.go
```diff
@@ -0,0 +1,48 @@
...
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var updateCmd = common.StandardCommand(&cobra.Command{
+	Use:     "update <org-id> [--display-name display-name]",
+	Short:   "Update an org",
+	Aliases: []string{"u"},
```
daabr: This is the only command on the orgs context that starts with "u", so regular auto-complete should be enough, this alias isn't needed

cmd/ak/cmd/orgs/remove_member.go
```diff
@@ -0,0 +1,40 @@
...
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var removeMemberCmd = common.StandardCommand(&cobra.Command{
+	Use:     "remove-member <org id> <user id>",
```
daabr, line 13: Why not names too?

cmd/ak/cmd/orgs/orgs.go
```diff
@@ -0,0 +1,32 @@
...
+import (
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+)
+
+var orgsCmd = common.StandardCommand(&cobra.Command{
+	Use:   "orgs",
+	Short: "Orgs: create, get, add-member, list-members, get-member, remove-member, update-member",
```
daabr, line 12: Maybe add a sub command for members? Just an idea, not blocking this PR.

cmd/ak/cmd/users/update.go
```diff
@@ -0,0 +1,74 @@
...
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var enabled bool
+
+var updateCmd = common.StandardCommand(&cobra.Command{
+	Use:     "update [email or id] [--email email] [--display-name display-name] [--disabled] [--enabled]",
+	Short:   "Update a user",
+	Aliases: []string{"u"},
```
daabr: Unnecessary alias with auto-complete

cmd/ak/cmd/users/update.go
```diff
@@ -0,0 +1,74 @@
...
+			if _, uid, err = r.User(ctx, args[0]); err != nil {
+				return fmt.Errorf("resolve user: %w", err)
+			}
+		}
+
+		u := sdktypes.NewUser().WithID(uid)
+
+		fm := &sdktypes.FieldMask{}
+
+		if cmd.Flags().Changed("email") {
```
daabr: Isn't the email address immutable?

cmd/ak/cmd/orgs/list_members.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/internal/resolver"
+)
+
+var listMembersCmd = common.StandardCommand(&cobra.Command{
+	Use:     "list-members <org id>",
```
daabr, line 13: Why not "name or ID"?

cmd/ak/cmd/orgs/create.go
```diff
@@ -0,0 +1,38 @@
...
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var displayName string
+
+var createCmd = common.StandardCommand(&cobra.Command{
+	Use:     "create [--display-name display-name]",
```
daabr, line 15: IIUC, this is a required single field - if so, it's better to use an arg

cmd/ak/cmd/orgs/add_member.go
itayd: ack, in a different pr though

internal/backend/db/dbgorm/dbtime/dbtime.go
itayd: will move to kittehs

internal/backend/db/dbgorm/dbtime/dbtime.go
itayd: done

cmd/ak/cmd/orgs/remove_member.go
itayd: orgs and users dont have login names (yet?)

cmd/ak/cmd/orgs/orgs.go
itayd: let's not overdo it. maybe in the future.

cmd/ak/cmd/users/update.go
itayd: right!

cmd/ak/cmd/orgs/list_members.go
itayd: no name

cmd/ak/cmd/orgs/create.go
itayd: it's not a required field. you can create an unnamed org.

cmd/ak/cmd/users/update.go
itayd: ok

cmd/ak/cmd/orgs/update.go
itayd: ok...

cmd/ak/cmd/orgs/create.go
daabr: so an optional arg?

itayd: we might add more stuff in the future. i think this should be a --thingy.

internal/kittehs/proto.go
itayd: i still want to retain the actual values as non-json. converting to json will convert everything to json.

internal/kittehs/proto_test.go
itayd: added a note that only scalars are currently supported.

daabr (approved):

PR #948, opened 2024-12-28
refactorish(authn): default user is just a user, and simpler authentication and thorough testing

itayd:
simplified authentication workflow, now well defined behaviour of priorities between session, tokens and default user, as well as enforced in testing.

this is not a pure refactor - default user is now used only when not authenticated by any other mean. prior - default user was always used and rejected other authn methods.

### DIFF
--- a/internal/backend/auth/authhttpmiddleware/authmiddleware.go
+++ b/internal/backend/auth/authhttpmiddleware/authmiddleware.go
@@ -29,7 +29,7 @@ var Configs = configset.Set[Config]{
 
 // The middleware passes around internally the authenticated user ID,
 // so that we can check only once in the AuthMiddlewareDecorator for
-// the common stuff - if the user exists and if it is not disabled.
+// the common stuff - that the user exists and that it is not disabled.
 // The AuthMiddlewareDecorator will call eventually the authcontext.SetAuthnUser
 // if the user is deemed worthy of authentication.
 type userIDCtxKey string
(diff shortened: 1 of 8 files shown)


internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -25,6 +27,28 @@ var Configs = configset.Set[Config]{
 	Dev:     &Config{UseDefaultUser: true},
 }
 
+// The middlewares passes around internally the authenticated user id,
```
daabr:
```suggestion
// The middleware passes around internally the authenticated user ID,
```

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -25,6 +27,28 @@ var Configs = configset.Set[Config]{
...
+// so that we can check only once in the AuthMiddlewareDecorator for
+// the common stuff - if the user exists and if it is not disabled.
```
daabr:
```suggestion
// the common stuff - that the user exists and that it is not disabled.
```

internal/backend/users/users.go
```diff
@@ -18,7 +18,12 @@ import (
 
 type Config struct {
 	// If set, do not create a personal org for new users. Instead, set this as their default org.
-	DefaultOrgID sdktypes.OrgID `json:"default_org_id"`
+	// This is useful for single tenants setups where all users belong to the same org.
```
daabr:
```suggestion
	// This is useful for single-tenant setups where all users belong to the same org.
```

daabr (approved):
Approved with tiny nits

PR #955, opened 2024-12-31
feat: users can invite other users to orgs

itayd:
- each org has members, each member has a status
- status can be ACTIVE, INVITED or REJECTED.
- invite is sent by adding a member in status INVITED by the inviting user. inviting user must be a member of the org.
- invite is responded to by updating the status to ACTIVE or REJECTED by the invited user.
- add method to translate user emails to user ids without getting the user. this can be done by anyone.


cmd/ak/cmd/orgs/add_member.go
```diff
@@ -7,10 +7,13 @@ import (
...
 	"go.autokitteh.dev/autokitteh/cmd/ak/common"
 	"go.autokitteh.dev/autokitteh/internal/resolver"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+var status string
+
 var addMemberCmd = common.StandardCommand(&cobra.Command{
-	Use:     "add-member <org id> <user id>",
+	Use:     "add-member <org id> <user id> [--status status]",
```
daabr: We don't want to allow users to add someone with the `declined` status, so I suggest you replace this with a boolean flag: `--invite` (true = `invited`, false = `active`) or `--force` (reversed true/false meaning).

BTW, that's the intended default status if not specified? Document it in the CLI.

cmd/ak/cmd/users/get_id.go
```diff
@@ -0,0 +1,30 @@
+package users
+
+import (
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+)
+
+var getIDCmd = common.StandardCommand(&cobra.Command{
```
daabr, line 9: What's the point of this command, compared to a regular "get"?

proto/autokitteh/orgs/v1/org.proto
```diff
@@ -3,9 +3,21 @@ syntax = "proto3";
...
+  string user_id = 1;
+  string org_id = 2;
+  OrgMemberStatus status = 3;
+}
+
+enum OrgMemberStatus {
+  ORG_MEMBER_STATUS_UNSPECIFIED = 0;
+  ORG_MEMBER_STATUS_ACTIVE = 1;
+  ORG_MEMBER_STATUS_INVITED = 2;
+  ORG_MEMBER_STATUS_REJECTED = 3;
```
daabr: Please change to "DECLINED" throughout this PR - it's more appropriate for invitations, as opposed to requests.

cmd/ak/cmd/orgs/add_member.go
itayd: done

cmd/ak/cmd/users/get_id.go
itayd: does not return the entire user, just the user id. don't want people to be able to see user details they're not allowed to see.

proto/autokitteh/orgs/v1/org.proto
itayd: done

cmd/ak/cmd/users/get_id.go
daabr: the point is this: if I'm not allowed to see your details, why should I be allowed to see your user ID?

cmd/ak/cmd/orgs/add_member.go
```diff
@@ -7,10 +7,13 @@ import (
...
 	"go.autokitteh.dev/autokitteh/cmd/ak/common"
 	"go.autokitteh.dev/autokitteh/internal/resolver"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+var active bool
+
 var addMemberCmd = common.StandardCommand(&cobra.Command{
-	Use:     "add-member <org id> <user id>",
+	Use:     "add-member <org id> <user id> [--active]",
```
daabr, line 16: Suggestion: user email (which is also unique and immutable) instead of user ID:
1. The caller doesn't need to know the user ID, AK handles this translation under the hood
2. If the user isn't in the server, we could decide whether to invite them to join AK, or return an error

configs/opa_bundles/default/authz.rego
```diff
@@ -46,13 +46,28 @@ allow if {
...
 }
 
+allow if {
+	authn
+	input.kind == "usr"
+	input.action == "get-org-member-status"
+	member_of_single_assosicated_org_id
+}
+
+# anyone can translate an email to an id.
```
daabr, line 75: I don't agree with this rule, let's discuss it offline.

configs/opa_bundles/default/authz.rego
```diff
@@ -46,13 +46,28 @@ allow if {
...
+}
+
 #
 # Orgs
 #
 
 allow if {
 	authn
 	input.kind == "org"
+	input.action in ["delete", "update", "remove-member"]
```
daabr, line 89: So any member of an org can take over the org by removing all other members, or even delete the org?

It's bad but not a merge blocker because we don't support RBAC yet, and in fact being specific about actions is better than before. I just want to confirm I'm understanding this correctly.

configs/opa_bundles/default/authz.rego
```diff
@@ -62,6 +77,25 @@ allow if {
 	input.action == "create"
 }
 
+# new members must be invited.
+allow if {
+	authn
+	input.kind == "org"
+	input.action == "add-member"
+	is_org_member_of_resource
+	input.data.status == "INVITED"
```
daabr, line 105: Does this mean we can't auto/forcefully add users to an org? Does this conflict with the `--active` flags in the `add-member` CLI command?

internal/backend/db/dbgorm/orgs.go
```diff
@@ -141,35 +175,50 @@ func (gdb *gormdb) RemoveOrgMember(ctx context.Context, oid sdktypes.OrgID, uid
...
 
-func (gdb *gormdb) IsOrgMember(ctx context.Context, oid sdktypes.OrgID, uid sdktypes.UserID) (bool, error) {
+func (gdb *gormdb) GetOrgMemberStatus(ctx context.Context, oid sdktypes.OrgID, uid sdktypes.UserID) (sdktypes.OrgMemberStatus, error) {
 	if oid == authusers.DefaultOrg.ID() {
-		return uid == authusers.DefaultUser.ID(), nil
+		if uid == authusers.DefaultUser.ID() {
+			return sdktypes.OrgMemberStatusActive, nil
+		}
+
+		return sdktypes.OrgMemberStatusUnspecified, nil
```
daabr, line 184: So the default org technically doesn't have members?

internal/backend/orgs/orgs.go
```diff
@@ -99,54 +99,96 @@ func (o *orgs) ListMembers(ctx context.Context, id sdktypes.OrgID) ([]sdktypes.U
...
+		if err != nil {
+			return err
+		}
+
+		if err := authz.CheckContext(
+			ctx,
+			oid,
+			"write:update-member-status",
+			authz.WithAssociationWithID("user", uid),
+			authz.WithData("next_status", status.String()),
```
daabr: "Next" implies a sequence of states, or an order, why not just "new_status"? If you change this, remember to change the Rego policy file too.

internal/resolver/resolver.go
```diff
@@ -366,28 +366,41 @@ func (r Resolver) SessionID(ctx context.Context, id string) (s sdktypes.Session,
...
+
+	if sdktypes.IsUserID(emailOrID) {
+		return sdktypes.StrictParseUserID(emailOrID)
+	}
+
+	uid, err = r.Client.Users().GetID(ctx, emailOrID)
+	return
+}
+
+// User returns a user, based on the given name, email or ID supplied by emailOrID.
```
daabr:
```suggestion
// User returns a user, based on the given email address or user ID.
```

configs/opa_bundles/default/authz.rego
itayd: discussed offline.

internal/backend/orgs/orgs.go
itayd: will change to `new`

internal/backend/db/dbgorm/orgs.go
itayd: i will turn the defaults into ordinary records in a future pr.

daabr (approved):

internal/backend/orgs/orgs.go
itayd: done

PR #957, opened 2025-01-02
feat: quick and dirty remote registry usage

efiShtain:
This is a quick kind of native implementation for building images on the build step instead of the run workflow step

remote registry is optional, with minimal options (address only) and is off by default

### DIFF
--- a/runtimes/pythonrt/build.go
+++ b/runtimes/pythonrt/build.go
@@ -85,7 +85,7 @@ func (py *pySvc) Build(ctx context.Context, fsys fs.FS, path string, values []sd
 
 		hash := md5.Sum(data)
 		version := fmt.Sprintf("u%x", hash)
-		imageRef := fmt.Sprintf("localhost:5001/usercode:%s", version)
+		imageRef := fmt.Sprintf("usercode:%s", version)
 
 		if err := drm.client.BuildImage(ctx, imageRef, codePath); err != nil {
 			return sdktypes.InvalidBuildArtifact, fmt.Errorf("build image: %w", err)
--- a/runtimes/pythonrt/build.go
+++ b/runtimes/pythonrt/build.go
@@ -4,6 +4,7 @@ import (
 	"archive/tar"
 	"bytes"
 	"context"
+	"crypto/md5"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -69,6 +70,28 @@ func (py *pySvc) Build(ctx context.Context, fsys fs.FS, path string, values []sd
 		}
 
 		compiledData[hdr.Name] = nil
+	}
+
+	if configuredRunnerType == runnerTypeDocker {
+		drm, ok := runnerManager.(*dockerRunnerManager)
+		if !ok {
+			return sdktypes.InvalidBuildArtifact, fmt.Errorf("configured runner type is docker but runner manager is not docker")
+		}
+
+		codePath, err := prepareUserCode(data, false)
+		if err != nil {
+			return sdktypes.InvalidBuildArtifact, fmt.Errorf("prepare user code: %w", err)
+		}
+
+		hash := md5.Sum(data)
+		version := fmt.Sprintf("u%x", hash)
+		imageRef := fmt.Sprintf("localhost:5001/usercode:%s", version)
+
+		if err := drm.client.BuildImage(ctx, imageRef, codePath); err != nil {
+			return sdktypes.InvalidBuildArtifact, fmt.Errorf("build image: %w", err)
+		}
+
+		compiledData[imageKey] = []byte(imageRef)
 	}
 
 	var art sdktypes.BuildArtifact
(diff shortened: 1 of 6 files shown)


runtimes/pythonrt/build.go
```diff
@@ -70,6 +71,28 @@ func (py *pySvc) Build(ctx context.Context, fsys fs.FS, path string, values []sd
...
+		}
+
+		codePath, err := prepareUserCode(data, false)
+		if err != nil {
+			return sdktypes.InvalidBuildArtifact, fmt.Errorf("prepare user code: %w", err)
+		}
+
+		hash := md5.Sum(data)
+		version := fmt.Sprintf("u%x", hash)
+		imageRef := fmt.Sprintf("localhost:5001/usercode:%s", version)
```
daabr: What is `localhost:5001`? Shouldn't it be configurable?

daabr (approved):
No objections to merging, but please answer these questions before you do that:
1. Use-case: who will want to use this / when?
2. Use-case: example for a registry that supports usernames and passwords (i.e. why are these parameters desirable)
3. Design/UX beyond the scope of this PR: the build step can't be and was never meant to be instantaneous, but so far it was relatively quick. How will the UX change if the build can take a minute instead of a second?
4. Design/UX beyond the scope of this PR: if the image upload to the registry fails because of some tiny and temporary network issue, it means that the failure happens long after the user initiated the build - how do they know about this? How does this affect the rest of their interaction with AK?

runtimes/pythonrt/build.go
```diff
@@ -70,6 +71,28 @@ func (py *pySvc) Build(ctx context.Context, fsys fs.FS, path string, values []sd
 		compiledData[hdr.Name] = nil
 	}
 
+	if configuredRunnerType == runnerTypeDocker {
+		drm, ok := runnerManager.(*dockerRunnerManager)
+		if !ok {
+			return sdktypes.InvalidBuildArtifact, fmt.Errorf("configured runner type is docker but runner manager is not docker")
```
itayd, line 78: i think this should be a panic, but you wouldn't like it :)

runtimes/pythonrt/build.go
```diff
@@ -70,6 +71,28 @@ func (py *pySvc) Build(ctx context.Context, fsys fs.FS, path string, values []sd
...
+		codePath, err := prepareUserCode(data, false)
+		if err != nil {
+			return sdktypes.InvalidBuildArtifact, fmt.Errorf("prepare user code: %w", err)
+		}
+
+		hash := md5.Sum(data)
+		version := fmt.Sprintf("u%x", hash)
+		imageRef := fmt.Sprintf("localhost:5001/usercode:%s", version)
+
+		if err := drm.client.BuildImage(ctx, imageRef, codePath); err != nil {
```
itayd, line 90: how much time can this take?

efiShtain: same as before, if cache is on the machine from previous builds, almost instantly. if not cache, 1-2 minutes on standard cases. can take more if user want to install more stuff.

runtimes/pythonrt/build.go
efiShtain: it's build, not workflow
if we panic the server will fail
I think we should never panic on known errors, we can handle them gracefully and tell the user what happened

runtimes/pythonrt/build.go
efiShtain: good catch, it is configurable but I missed passing it here

efiShtain:
> No objections to merging, but please answer these questions before you do that:
> 
> 1. Use-case: who will want to use this / when?
> 2. Use-case: example for a registry that supports usernames and passwords (i.e. why are these parameters desirable)
> 3. Design/UX beyond the scope of this PR: the build step can't be and was never meant to be instantaneous, but so far it was relatively quick. How will the UX change if the build can take a minute instead of a second?
> 4. Design/UX beyond the scope of this PR: if the image upload to the registry fails because of some tiny and temporary network issue, it means that the failure happens long after the user initiated the build - how do they know about this? How does this affect the rest of their interaction with AK?

1. it is mandatory to distribute the image between many runners. anyone who wants to use it can use it.
2. not sure if you want an example or ask why is it needed?
3. ux not solved and should be updated to support long running requests

runtimes/pythonrt/build.go
itayd: the idea here is that if this fails we're in a really weird situation - this should never fail. but ok.

runtimes/pythonrt/build.go
itayd: what worries me is that this is done via a synchronous blocking call. when a user press "build" it might be stuck for a while. should we change the build to be async?

itayd: this is obviously a concern for another pr... but we might even fall on http timeouts here.

efiShtain:
waiting for remote builds new architecture cause we can't build without DIND in current solution and we don't want it

PR #958, opened 2025-01-03
feat: allow users to invite other users

itayd:
- add status field to user, which one of the values is INVITED
- allow users to create INVITED users only
- a user can accept the invitation by logging in


proto/autokitteh/users/v1/user.proto
```diff
@@ -8,6 +8,14 @@ message User {
...
-  bool disabled = 4;
+  bool disabled = 4; // obsolete, use status instead.
   string default_org_id = 5; // org to use for projects, if not otherwise specified.
+  UserStatus status = 6;
+}
+
+enum UserStatus {
+  USER_STATUS_UNSPECIFIED = 0;
+  USER_STATUS_ACTIVE = 1;
+  USER_STATUS_INVITED = 2; // not active yet, just invited.
```
daabr, line 19: Why do we need the `INVITED` status for users?
- Do we want to block inviting a single email address more than once?
- Are there functional differences between an invited user and an active user?
- Do we need a user to be in the DB before the sign-in for the first time?
- Do we want to track invited-but-never-logged-in users in the AK DB?

proto/autokitteh/users/v1/user.proto
```diff
@@ -8,6 +8,14 @@ message User {
...
+  USER_STATUS_DISABLED = 3; // disabled, cannot login.
```
daabr, line 20: If we allow disabling users (which is good), we should consider adding a `STATUS_REFERENCE` string field which is usually empty but can be a URL that describes/tracks why that user is disabled (e.g. doc, linear ticket). This is not useful for self-hosted servers, but critical for our cloud ops, and trivial to add here.

configs/opa_bundles/default/authz.rego
```diff
@@ -43,9 +43,28 @@ allow if {
 allow if {
 	authn
 	input.kind == "usr"
+	input.action_type == "read"
 	input.user_id == input.resource_id
 }
 
+# Allow any user to invite other users as long as they
```
daabr, line 50: I agree with this rule, but I think it's too weak: in general we don't want the inviter to set **any** user details (including `user_id`, `status`, `default_org_id`.

Also, a risk to mitigate: we need to ensure that any future changes in the user proto will be reflected here too.

proto/autokitteh/users/v1/user.proto
itayd: when we add a user to org we need a user id, this allows it.

proto/autokitteh/users/v1/user.proto
itayd: in a future pr.

daabr (approved):

PR #960, opened 2025-01-04
feat(deployments): auto deactivate drained deployments

itayd:
- improve deactivation logic by performing it in a single SQL query.
- periodically auto-deactivate drained deployments in the deployments module. use interval and jitter to minimize collisions between instances.

### DIFF
--- a/internal/backend/db/db.go
+++ b/internal/backend/db/db.go
@@ -111,6 +111,10 @@ type DB interface {
 	CreateDeployment(ctx context.Context, deployment sdktypes.Deployment) error
 	DeleteDeployment(ctx context.Context, deploymentID sdktypes.DeploymentID) error
 	DeploymentHasActiveSessions(ctx context.Context, deploymentID sdktypes.DeploymentID) (bool, error)
+	// Returns number of deployments affected.
+	DeactivateAllDrainedDeployments(ctx context.Context) (int, error)
+	// Returns true if the deployment was deactivated.
+	DeactivateDrainedDeployment(ctx context.Context, did sdktypes.DeploymentID) (bool, error)
 
 	// -----------------------------------------------------------------------
 	CreateSession(ctx context.Context, session sdktypes.Session) error
--- a/internal/backend/db/dbgorm/deployments.go
+++ b/internal/backend/db/dbgorm/deployments.go
@@ -106,7 +106,7 @@ func (gdb *gormdb) listDeploymentsCommonQuery(ctx context.Context, filter sdkser
 		q = q.Limit(int(filter.Limit))
 	}
 
-	return q.Order("deployments.created_at desc")
+	return q.Order("deployments.deployment_id desc")
 }
 
 func (db *gormdb) listDeploymentsWithStats(ctx context.Context, filter sdkservices.ListDeploymentsFilter) ([]scheme.DeploymentWithStats, error) {
@@ -217,4 +217,44 @@ func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.D
 	}
 
 	return r.TotalCount > 0, nil
+}
+
+var finalSessionStateTypes = kittehs.Transform(sdktypes.FinalSessionStateTypes, func(s sdktypes.SessionStateType) int { return s.ToInt() })
+
+func (db *gormdb) DeactivateAllDrainedDeployments(ctx context.Context) (int, error) {
+	q := db.db.WithContext(ctx).Exec(`UPDATE deployments
+SET state = ?
+WHERE state = ?
+AND NOT EXISTS (
+    SELECT 1
+    FROM sessions
+    WHERE sessions.deployment_id = deployments.deployment_id
+		AND sessions.current_state_type NOT IN (?)
+);`,
+		sdktypes.DeploymentStateInactive.ToInt(), sdktypes.DeploymentStateDraining.ToInt(), finalSessionStateTypes)
+
+	if err := q.Error; err != nil {
+		return 0, translateError(err)
+	}
+
+	return int(q.RowsAffected), nil
+}
+
+func (db *gormdb) DeactivateDrainedDeployment(ctx context.Context, did sdktypes.DeploymentID) (bool, error) {
+	q := db.db.WithContext(ctx).Exec(`UPDATE deployments
+SET state = ?
+WHERE state = ? AND deployment_id = ?
+AND NOT EXISTS (
+	SELECT 1
+	FROM sessions
+	WHERE sessions.deployment_id = deployments.deployment_id
+		AND sessions.current_state_type NOT IN (?)
+);`,
+		sdktypes.DeploymentStateInactive.ToInt(), sdktypes.DeploymentStateDraining.ToInt(), did.UUIDValue(), finalSessionStateTypes)
+
+	if err := q.Error; err != nil {
+		return false, translateError(err)
+	}
+
+	return int(q.RowsAffected) > 0, nil
 }
--- a/internal/backend/deployments/deployments.go
+++ b/internal/backend/deployments/deployments.go
@@ -3,11 +3,14 @@ package deployments
 import (
 	"context"
 	"fmt"
+	"math/rand/v2"
+	"time"
 
 	"go.uber.org/zap"
 
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authcontext"
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authz"
+	"go.autokitteh.dev/autokitteh/internal/backend/configset"
 	"go.autokitteh.dev/autokitteh/internal/backend/db"
 	"go.autokitteh.dev/autokitteh/internal/backend/telemetry"
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
@@ -15,14 +18,61 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-type deployments struct {
-	z  *zap.Logger
-	db db.DB
+// Draining will happen each interval + rand(jitter).
+// Jitter is to prevent all instances from deactivating at the same time.
+// Note that deactivation will happen if these are not set, as these are more
+// to safeguard from a deployment being stuck in draining state.
+type Config struct {
+	AutoDrainingDeactivationInterval time.Duration `koanf:"draining_deactivation_interval"`
+	AutoDrainingDeactivationJitter   time.Duration `koanf:"draining_deactivation_jitter"`
 }
 
-func New(z *zap.Logger, db db.DB, telemetry *telemetry.Telemetry) sdkservices.Deployments {
+var Configs = configset.Set[Config]{
+	Default: &Config{
+		AutoDrainingDeactivationInterval: 5 * time.Minute,
+		AutoDrainingDeactivationJitter:   1 * time.Minute,
+	},
+}
+
+type deployments struct {
+	l   *zap.Logger
+	db  db.DB
+	cfg *Config
+}
+
+func New(l *zap.Logger, cfg *Config, db db.DB, telemetry *telemetry.Telemetry) sdkservices.Deployments {
 	initMetrics(telemetry)
-	return &deployments{z: z, db: db}
+
+	ds := &deployments{l: l, db: db, cfg: cfg}
+
+	if cfg.AutoDrainingDeactivationInterval > 0 || cfg.AutoDrainingDeactivationJitter > 0 {
+		go ds.Autodrain()
+	}
+
+	return ds
+}
+
+func (d *deployments) Autodrain() {
+	d.l.
+		With(zap.Duration("interval", d.cfg.AutoDrainingDeactivationInterval),
+			zap.Duration("jitter", d.cfg.AutoDrainingDeactivationJitter)).
+		Info("periodically autoinactivating drained deployments")
+
+	for {
+		t := d.cfg.AutoDrainingDeactivationInterval + time.Duration(rand.Float32()*float32(d.cfg.AutoDrainingDeactivationJitter))
+
+		d.l.Debug("waiting before next autodeactivation", zap.Duration("duration", t))
+
+		time.Sleep(t)
+
+		n, err := d.db.DeactivateAllDrainedDeployments(context.Background())
+		if err != nil {
+			d.l.Error("autodeactivation failed", zap.Error(err))
+			continue
+		}
+
+		d.l.Sugar().Infof("autodeactivated drained %d deployments", n)
+	}
 }
 
 func (d *deployments) Activate(ctx context.Context, id sdktypes.DeploymentID) error {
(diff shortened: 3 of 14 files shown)


internal/backend/db/dbgorm/deployments.go
```diff
@@ -218,3 +218,46 @@ func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.D
 
 	return r.TotalCount > 0, nil
 }
+
+func (db *gormdb) DeactiveDrainedDeployments(ctx context.Context, id sdktypes.DeploymentID) (int, error) {
+	q := db.db.WithContext(ctx)
```
daabr: What's the point of initializing this `q` if you never use it? Or is that a bug?

internal/backend/db/dbgorm/deployments.go
```diff
@@ -218,3 +218,46 @@ func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.D
...
+
+	finalStates := []int{
+		int(sdktypes.SessionStateTypeCompleted.ToProto()),
+		int(sdktypes.SessionStateTypeError.ToProto()),
+		int(sdktypes.SessionStateTypeStopped.ToProto()),
+	}
+
+	if id.IsValid() {
```
daabr: It is very bad for code readability and maintainability to cram 2 nested SQL queries into an Go if-else block. The meaning of both Go and SQL is lost on the reader until they spend **minutes** to read it carefully. This is not the standard we should strive for.

internal/backend/db/db.go
```diff
@@ -111,6 +111,8 @@ type DB interface {
 	CreateDeployment(ctx context.Context, deployment sdktypes.Deployment) error
 	DeleteDeployment(ctx context.Context, deploymentID sdktypes.DeploymentID) error
 	DeploymentHasActiveSessions(ctx context.Context, deploymentID sdktypes.DeploymentID) (bool, error)
+	// Returns number of deployments affected.
+	DeactiveDrainedDeployments(ctx context.Context, did sdktypes.DeploymentID) (int, error)
```
daabr:
```suggestion
	DeactivateDrainedDeployments(ctx context.Context, did sdktypes.DeploymentID) (int, error)
```

internal/backend/db/dbgorm/deployments.go
```diff
@@ -218,3 +218,46 @@ func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.D
...
+	WHERE NOT EXISTS (
+		SELECT 1
+		FROM sessions
+		WHERE sessions.deployment_id = ? AND sessions.current_state_type NOT IN (?)
+	)
+) AND state = ?;`, sdktypes.DeploymentStateInactive.ToProto(), id.UUIDValue(), id.UUIDValue(), finalStates, sdktypes.DeploymentStateDraining.ToProto())
+	} else {
+		q = db.db.Exec(`UPDATE deployments
+SET state = ?
+WHERE deployment_id IN (
```
daabr: You're doing an UPDATE query with `WHERE deployment_id IN` and a nested query that extracts IDs from the same table?! Why not a single flat query?

Also note that it's more efficient to run the subquery only on draining deployments, i.e. check `state = DRAINING` before the subquery.

```
UPDATE deployments
SET state = ?
WHERE state = ?
AND NOT EXISTS (
    SELECT 1
    FROM sessions
    WHERE sessions.deployment_id = deployments.deployment_id
      AND sessions.current_state_type NOT IN (?)
)
```

internal/backend/db/dbgorm/deployments.go
```diff
@@ -218,3 +218,46 @@ func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.D
...
+		int(sdktypes.SessionStateTypeCompleted.ToProto()),
+		int(sdktypes.SessionStateTypeError.ToProto()),
+		int(sdktypes.SessionStateTypeStopped.ToProto()),
+	}
+
+	if id.IsValid() {
+		// TODO: This probably can be greatly simplified.
+		q = db.db.Exec(`UPDATE deployments
+SET state = ?
+WHERE deployment_id = ? AND deployment_id IN (
```
daabr: See comment below about simplifying this query too

internal/backend/deployments/deployments.go
```diff
@@ -3,34 +3,83 @@ package deployments
...
 	"go.autokitteh.dev/autokitteh/internal/backend/db"
 	"go.autokitteh.dev/autokitteh/internal/backend/telemetry"
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+type Config struct {
+	AutoDrainingDeactivationInterval time.Duration `koanf:"draining_deactivation_interval"`
+	AutoDrainingDeactivationJitter   time.Duration `koanf:"draining_deactivation_jitter"`
```
daabr, line 27: Please document each of these values:
- Interval: it's clear what it is, but not obvious why it's needed, i.e. explain why AK doesn't **really** auto-drain without a polling loop
- jitter: what does "jitter" even mean in this context, in additions to why it's needed

internal/backend/deployments/deployments.go
```diff
@@ -3,34 +3,83 @@ package deployments
...
+}
+
+var Configs = configset.Set[Config]{
+	Default: &Config{
+		AutoDrainingDeactivationInterval: 5 * time.Minute,
+		AutoDrainingDeactivationJitter:   1 * time.Minute,
+	},
+	Dev: &Config{
+		// No autodeactivating of drain deployments in dev.
```
daabr: Why not? Remember, this is the AutoKitteh **we** use for development, as well as talented external users and OSS developers.

https://youtu.be/3t6L-FlfeaI?t=121

internal/backend/deployments/deployments.go
```diff
@@ -3,34 +3,83 @@ package deployments
...
+	}
+
+	return ds
+}
+
+func (d *deployments) Autodrain() {
+	d.l.
+		With(zap.Duration("interval", d.cfg.AutoDrainingDeactivationInterval),
+			zap.Duration("jitter", d.cfg.AutoDrainingDeactivationJitter)).
+		Info("periodically autoinactivating drained deployments")
```
daabr, line 59: "Autoinactivating" is not a word, and "inactive" is a state not an action. Maybe "auto deactivating"?

internal/backend/db/dbgorm/deployments.go
itayd: it's just adding the WithContext so it'll be take affect on the Execs.

internal/backend/db/db.go
itayd: done

internal/backend/deployments/deployments.go
itayd: done

internal/backend/deployments/deployments.go
itayd: Because it might hide bugs. This is more of a "garbage collector" for orphaned sessions.

internal/backend/deployments/deployments.go
itayd: done

internal/backend/db/dbgorm/deployments.go
itayd: what do you propose to do then?

internal/backend/db/dbgorm/deployments.go
itayd: How would you do it in a flat query?

internal/backend/db/dbgorm/deployments.go
```diff
@@ -218,3 +218,48 @@ func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.D
...
+	finalStates := []int{
+		int(sdktypes.SessionStateTypeCompleted.ToInt()),
+		int(sdktypes.SessionStateTypeError.ToInt()),
+		int(sdktypes.SessionStateTypeStopped.ToInt()),
+	}
+
+	if id.IsValid() {
+		// TODO: This probably can be greatly simplified.
+		q = db.db.Exec(`UPDATE deployments
+SET state = ?
```
efiShtain, line 234: why do you need where deployment_id= and deployment_id in ?
it will always match the one deployment_id=

internal/backend/db/dbgorm/deployments.go
efiShtain: the order of the ands is irrelevant, postgres will optimize it how ever it want

but it might be more readable for us humans to write the state check before the not exists

internal/backend/db/dbgorm/deployments.go
```diff
@@ -218,3 +218,48 @@ func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.D
...
+WHERE deployment_id = ? AND deployment_id IN (
+    SELECT d.deployment_id
+    FROM deployments d
+	WHERE NOT EXISTS (
+		SELECT 1
+		FROM sessions
+		WHERE sessions.deployment_id = ? AND sessions.current_state_type NOT IN (?)
```
efiShtain, line 241: I'm not sure this is atomic
there might be new session between the subquery exeuction and the update statement

internal/backend/db/dbgorm/deployments.go
```diff
@@ -218,3 +218,48 @@ func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.D
...
+SET state = ?
+WHERE deployment_id IN (
+    SELECT d.deployment_id
+    FROM deployments d
+	WHERE NOT EXISTS (
+		SELECT 1
+		FROM sessions
+		WHERE sessions.deployment_id = d.deployment_id AND sessions.current_state_type NOT IN (?)
+	)
+) AND state = ?;`,
```
efiShtain: there is no index on state, so it will perform a sequential scan on the deployments table which is not what we want assuming only a minimal number of deployments will be in draining state

internal/backend/db/dbgorm/deployments.go
efiShtain: I think i agree with @daabr here, assuming i understand correctly

this function should deactivate one deployment by its ID, or if you want, pass a list of deployments

having this internal logic that if the ID is not valid, perform full blown update on DB is a "surprise" for who ever tries to use this function

I suggest to have a dedicated separated function that perform the all table maintanance 
and let this function handle one deployment by id

internal/backend/db/dbgorm/deployments.go
efiShtain: I think you gain more performance (low level optimization) if you use < 3 (state is running or created) instead of not in (1,2)
need to test with big tables and see

internal/backend/db/dbgorm/deployments.go
efiShtain: anyhow, @daabr is simpler and more readable so i'd go with that, maybe changing the not in to lower than and add an index on the deployments state

internal/backend/deployments/deployments.go
daabr: Then maybe set it to 1 hour or 1 day in dev? An interested developer will definitely notice that, but a regular user of a self-hosted server won't feel it for long.

internal/backend/db/dbgorm/deployments.go
daabr: No, it's not :) you're overwriting q below. There's also a static analysis warning here.

internal/backend/db/dbgorm/deployments.go
```diff
@@ -218,3 +218,48 @@ func (db *gormdb) DeploymentHasActiveSessions(ctx context.Context, id sdktypes.D
...
 	return r.TotalCount > 0, nil
 }
+
+func (db *gormdb) DeactivateDrainedDeployments(ctx context.Context, id sdktypes.DeploymentID) (int, error) {
+	q := db.db.WithContext(ctx)
+
+	finalStates := []int{
+		int(sdktypes.SessionStateTypeCompleted.ToInt()),
+		int(sdktypes.SessionStateTypeError.ToInt()),
+		int(sdktypes.SessionStateTypeStopped.ToInt()),
```
daabr: Unnecessary type conversion: `int(...ToInt())`

internal/backend/db/dbgorm/deployments.go
daabr: Agree with @efiShtain - splitting this into separate functions, even if only private helper functions, is much more readable. For example, even though I understood what this code does (eventually) I missed the "surprising" aspect of it that Efi points out.

Other possible tactics: split the SQL queries from the Go code with local or package vars.

internal/backend/deployments/deployments.go
```diff
@@ -3,34 +3,87 @@ package deployments
...
+	"go.autokitteh.dev/autokitteh/internal/backend/configset"
 	"go.autokitteh.dev/autokitteh/internal/backend/db"
 	"go.autokitteh.dev/autokitteh/internal/backend/telemetry"
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+type Config struct {
+	// Draining will happen each interval + rand(jitter).
```
daabr: Note that the location of this comment attaches it to the `AutoDrainingDeactivationInterval` field. You might want to move it above `type Config struct`, or split it into 2 comment (1 for each field).

internal/backend/deployments/deployments.go
```diff
@@ -3,34 +3,87 @@ package deployments
...
+	// Jitter is to prevent all instances from deactivating at the same time.
```
daabr: Jitter is relevant when you have agents that are synchronized (e.g. threads in a single process, or synchronized processes) or when the amount of agents is large enough that concurrency is expected most of the time anyway.

This is not the case here. I'm saying using jitter is bad, I'm saying it's solving a problem you shouldn't have anyway.

internal/backend/db/dbgorm/deployments.go
itayd: oopsy

internal/backend/deployments/deployments.go
itayd: oh well. just made it the same as default.

internal/backend/db/dbgorm/deployments.go
itayd: added index

internal/backend/db/dbgorm/deployments.go
itayd: i don't really like it since it'll force some ordering on the state types, which might have migration issues.

internal/backend/db/dbgorm/deployments.go
itayd: will split.

internal/backend/db/dbgorm/deployments.go
itayd: done

internal/backend/deployments/deployments.go
itayd: tbh i think it is clearer the way it is. i moved it to above the struct anyway.

internal/backend/deployments/deployments.go
itayd: It's just to add another level of randomness in case weird things happen.

daabr (approved):

PR #962, opened 2025-01-05
feat: add slack channel_rename event handler

mario99logic:
ChannelRenameHandler has been added to handle the channel_rename event. Additionally, the same handler has been implemented to manage the group_rename event, as it uses an identical JSON structure.

refs: INT-87

### DIFF
--- a/integrations/slack/events/channel.go
+++ b/integrations/slack/events/channel.go
@@ -11,18 +11,22 @@ import (
 )
 
 // https://api.slack.com/events/channel_created
-type ChannelCreatedEvent struct {
+// https://api.slack.com/events/channel_rename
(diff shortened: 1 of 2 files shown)


daabr (commented):
Good observation that you can use the same code for 2 events!

But it looks like we can go further: there's no need to add new structs & parsing for the 2 rename events, The existing `ChannelCreated` event has the exact same structure.

Please rename `ChannelCreated` to `ChannelCreatedRename` and use that instead of `ChannelRename`.

mario99logic:
@daabr I actually thought of the same idea! but didn't notice the struct of the channelCreated, nice.

daabr (approved):

PR #964, opened 2025-01-07
fix(python): enable project unit tests in Kittehub

daabr:
Return dummy GitHub and Slack clients instead of `ConnectionInitError` when the AutoKitteh Python SDK is running within unit tests (determined by the existence of a special environment variable).

Usage of this fix requires releasing to PyPI after merging.

Refs: INT-159

### DIFF
--- a/runtimes/pythonrt/py-sdk/autokitteh/github.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/github.py
@@ -27,6 +27,9 @@ def github_client(connection: str, **kwargs) -> Github:
     """
     check_connection_name(connection)
 
+    if os.getenv("AUTOKITTEH_UNIT_TEST"):
+        return Github(auth=Auth.Token("dummy_pat"), **kwargs)
+
     # Optional: GitHub Enterprise Server
     base_url = os.getenv("GITHUB_ENTERPRISE_URL")
     if base_url:
--- a/runtimes/pythonrt/py-sdk/autokitteh/slack.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/slack.py
@@ -30,6 +30,9 @@ def slack_client(connection: str, **kwargs) -> WebClient:
     """
     check_connection_name(connection)
 
+    if os.getenv("AUTOKITTEH_UNIT_TEST"):
+        return WebClient("dummy_bot_token", **kwargs)
+
     bot_token = os.getenv(connection + "__oauth_AccessToken")  # OAuth v2
     if not bot_token:
         bot_token = os.getenv(connection + "__BotToken")  # Socket Mode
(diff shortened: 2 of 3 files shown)


mario99logic (approved):

tebeka:
I don't like this approach. In general code shouldn't be aware it's tested.
If you're running unit tests - use `pytest.monkeypatch` and `unittest.mock`

tebeka (changes requested):
See my comments

daabr:
> In general code shouldn't be aware it's tested.

In theory I agree, product logic that behaves differently under test is risky. On the other hand, test code that replaces product logic and tests itself is equally risky and also useless. So in practice there has to be a balance.

> If you're running unit tests - use `pytest.monkeypatch` and `unittest.mock`

I do use mocks, but it's irrelevant for the problem that this PR solves. See https://github.com/autokitteh/kittehub/pull/149 and its test result https://github.com/autokitteh/kittehub/actions/runs/12647739289/job/35240695875?pr=149

itayd:
not clear to me why we need dummy pats and stuff. why not go the normal route?

daabr:
> not clear to me why we need dummy pats and stuff. why not go the normal route?

When running AK project unit tests, i.e. running project code without an AK server, i.e. without integration configs and connection vars in environment variables, the SDK throws `ConnectionInitError`.

The point of these unit tests is not to use clients "for real" - we mock them - but mocking can't overcome the raising of `ConnectionInitError` because client initializations happen when the modules are loaded, before any test begins to run.

tebeka:
> I do use mocks, but it's irrelevant for the problem that this PR solves. See [autokitteh/kittehub#149](https://github.com/autokitteh/kittehub/pull/149) and its test result https://github.com/autokitteh/kittehub/actions/runs/12647739289/job/35240695875?pr=149

I disagree :) You can patch `slack_client` before importing `markdown`

```python
from unittest.mock import MagicMock
import autokitteh.slack
autokitteh.slack.slack_client = MagicMock()

import markdown
```

daabr:
> You can patch `slack_client` before importing `markdown`
> 
> ```python
> from unittest.mock import MagicMock
> import autokitteh.slack
> autokitteh.slack.slack_client = MagicMock()
> 
> import markdown
> ```

~No, I can't. It conflicts with a patch in a different test module:~

```
@mock.patch.object(slack_cmd, "slack", autospec=True)
@mock.patch.object(slack_cmd, "data_helper", autospec=True)
class SlackCmdTest(unittest.TestCase):
```

~Which results in this error:~

```
unittest.mock.InvalidSpecError:
Cannot autospec attr 'slack' from target 'slack_cmd' as it has already been mocked out.
[target=<module 'slack_cmd' from 'kittehub/purrr/slack_cmd.py'>]
```

~Fixing this means removing `autospec=True` from the patch - and that's the point where I consider my approach better. Without `autospec=True` this test is testing itself and not the usage of a mock Slack client.~

Update: actually, I can use `spec=slack_sdk.web.client import WebClient` instead of `autospec=True`. So you're right, and I'm closing this PR. Thanks @tebeka !

PR #965, opened 2025-01-07
switch from soft delete to hard delete

efiShtain:
- remove all soft delete functions
- add relevant references and foreign keys
- use cascade delete deferred so database can handle all deletions correctly without foreign key issue
- added lots of unit tests for delete propagation 
- went table by table, checked which table is referenced by which other table and added a matching test
    - event is pointing to a trigger, so removing the trigger remove the event as well
    - session_call_attempts is referencing a session, so delete session delete all its attempts
    - trigger/signal/events are referencing connections - delete connection delete all triggers/signals/events
    - etc...

did some manual tests as well


internal/backend/db/dbgorm/builds_test.go
```diff
@@ -52,6 +52,42 @@ func TestDeleteBuild(t *testing.T) {
...
+	f, p := preBuildTest(t)
+	b := f.newBuild(p)
+	f.saveBuildsAndAssert(t, b)
+
+	// test deleteProject
+	assert.NoError(t, f.gormdb.deleteProject(f.ctx, p.ProjectID))
+	f.assertBuildDeleted(t, b.BuildID)
+}
+
+func TestDeleteBuildDeleteDeployment(t *testing.T) {
```
itayd, line 65: That's kinda scary - how do we make sure we don't delete an active deployment?

internal/backend/db/dbgorm/builds_test.go
```diff
@@ -52,6 +52,42 @@ func TestDeleteBuild(t *testing.T) {
...
+
+	d := f.newDeployment(b, p)
+	f.createDeploymentsAndAssert(t, d)
+
+	// test deleteBuild
+	assert.NoError(t, f.gormdb.deleteBuild(f.ctx, b.BuildID))
+	f.assertDeploymentsDeleted(t, d)
+}
+
+func TestDeleteBuildDeleteSession(t *testing.T) {
```
itayd, line 78: ditto

internal/backend/db/dbgorm/builds_test.go
tebeka: This is a test file, and we should never run tests in deployments.

efiShtain:
We went with a simplified version, handling only events and sessions for now. 
https://github.com/autokitteh/autokitteh/pull/992

Closing this PR for now, we can reiterate later if we want to switch to fully hard delete

PR #966, opened 2025-01-08
refactor(authn): simplify sub-middlewares and testing

itayd:
instead of passing internal context, just explicitly pass the user id in the sub-middleware level.
this makes the code simpler and greatly simplifies testing.

### DIFF
--- a/internal/backend/auth/authhttpmiddleware/authmiddleware.go
+++ b/internal/backend/auth/authhttpmiddleware/authmiddleware.go
@@ -33,7 +33,7 @@ type AuthMiddlewareDecorator func(http.Handler) http.Handler
 type Deps struct {
 	fx.In
 
-	L        *zap.Logger
+	Logger   *zap.Logger
 	Cfg      *Config
 	Users    sdkservices.Users
 	Sessions authsessions.Store `optional:"true"`
@@ -42,14 +42,15 @@ type Deps struct {
 
 type middlewareError struct {
 	code int
-	msg  string
+	msg  string // must never contain sensitive data.
 }
 
 func (m *middlewareError) apply(w http.ResponseWriter) { http.Error(w, m.msg, m.code) }
 
-func newMiddlewareError(code int, msg string) *middlewareError {
-	return &middlewareError{code: code, msg: msg}
-}
+var (
+	invalidAuthHeaderErr = &middlewareError{http.StatusUnauthorized, "invalid authorization header"}
+	invalidTokenErr      = &middlewareError{http.StatusUnauthorized, "invalid token"}
+)
 
 // middlewareFn is a function that extracts a user ID from a request.
 // It returns the user ID and an error if the request is invalid.
@@ -65,12 +66,12 @@ func newTokensMiddleware(tokens authtokens.Tokens) middlewareFn {
 		kind, payload, _ := strings.Cut(authHdr, " ")
 
 		if kind != "Bearer" {
-			return sdktypes.InvalidUserID, newMiddlewareError(http.StatusUnauthorized, "invalid authorization header")
+			return sdktypes.InvalidUserID, invalidAuthHeaderErr
 		}
 
 		u, err := tokens.Parse(payload)
 		if err != nil {
-			return sdktypes.InvalidUserID, newMiddlewareError(http.StatusUnauthorized, "invalid token")
+			return sdktypes.InvalidUserID, invalidTokenErr
 		}
 
 		return u.ID(), nil
@@ -125,7 +126,7 @@ func New(deps Deps) AuthMiddlewareDecorator {
 				}
 			}
 
-			l := deps.L
+			l := deps.Logger
 
 			if mwErr != nil {
 				l.Info("auth middleware error", zap.Error(errors.New(mwErr.msg)))
--- a/internal/backend/auth/authhttpmiddleware/authmiddleware.go
+++ b/internal/backend/auth/authhttpmiddleware/authmiddleware.go
@@ -1,11 +1,12 @@
 package authhttpmiddleware
 
 import (
-	"context"
+	"errors"
 	"net/http"
 	"strings"
 
 	"go.uber.org/fx"
+	"go.uber.org/zap"
 
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authcontext"
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authsessions"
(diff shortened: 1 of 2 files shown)


internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -27,145 +28,141 @@ var Configs = configset.Set[Config]{
...
-
-	return v.(sdktypes.UserID)
-}
-
 type AuthMiddlewareDecorator func(http.Handler) http.Handler
 
 type Deps struct {
 	fx.In
 
+	L        *zap.Logger
```
daabr:
```suggestion
	Logger   *zap.Logger
```

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -27,145 +28,141 @@ var Configs = configset.Set[Config]{
...
+func newMiddlewareError(code int, msg string) *middlewareError {
+	return &middlewareError{code: code, msg: msg}
 }
 
-func newTokensMiddleware(next http.Handler, tokens authtokens.Tokens) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
-		authHdr := r.Header.Get("Authorization")
+// middlewareFn is a function that extracts a user ID from a request.
+// It returns the user ID and an error if the request is invalid.
+type middlewareFn func(*http.Request) (sdktypes.UserID, *middlewareError)
```
daabr, line 57: This should be an interface, not a functor type.

itayd: strong disagree here - there is no reason to create a struct and define a function on it. @tebeka ?

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -27,145 +28,141 @@ var Configs = configset.Set[Config]{
...
+}
 
-		if getCtxUserID(r.Context()).IsValid() {
-			next = yes
-		}
+func (m *middlewareError) apply(w http.ResponseWriter) { http.Error(w, m.msg, m.code) }
 
-		next.ServeHTTP(w, r)
-	})
+func newMiddlewareError(code int, msg string) *middlewareError {
```
tebeka: Use value semantics?

internal/backend/auth/authhttpmiddleware/authmiddleware.go
tebeka: Agree with @itayd :)

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -27,145 +28,141 @@ var Configs = configset.Set[Config]{
...
+	return &middlewareError{code: code, msg: msg}
 }
 
-func newTokensMiddleware(next http.Handler, tokens authtokens.Tokens) http.HandlerFunc {
-	return func(w http.ResponseWriter, r *http.Request) {
-		authHdr := r.Header.Get("Authorization")
+// middlewareFn is a function that extracts a user ID from a request.
```
tebeka, line 55: The name `middleware` is for wrappers around HTTP handlers. This should be called `extractUID` or something like that.

Also, it's better to pass `context.Context` and not the request.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -27,145 +28,141 @@ var Configs = configset.Set[Config]{
...
-					return
-				}
+			l := deps.L
 
-				next.ServeHTTP(w, r.WithContext(authcontext.SetAuthnUser(ctx, u)))
-			}),
-			/* not authenticated: */ http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			if mwErr != nil {
+				l.Info("auth middleware error", zap.Error(errors.New(mwErr.msg)))
+				mwErr.apply(w)
```
tebeka, line 133: This is a security risk. You don't know what's in `w.msg`.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
```diff
@@ -27,145 +28,141 @@ var Configs = configset.Set[Config]{
...
-		if sessions != nil {
-			f = ifAuthenticated(f, newSessionsMiddleware(f, sessions))
-		}
+			ctx := r.Context()
 
-		// Evaluated first.
-		if tokens != nil {
-			f = ifAuthenticated(f, newTokensMiddleware(f, tokens))
-		}
+			u, err := users.Get(authcontext.SetAuthnSystemUser(ctx), uid, "")
```
tebeka, line 149: Not sure I love setting context for system user here. Need to make sure we remove it (defer?)

tebeka (approved):

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: not following?

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: > Also, it's better to pass context.Context and not the request.

Whaaaaa? I don't follow here why

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: it's setting it specifically for this call.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: i guess you mean just return without the ptr?

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: done

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: this seems to complicate things, i'd rather use it as is.

internal/backend/auth/authhttpmiddleware/authmiddleware.go
itayd: we do know - it's only a static string. will make it more clear.

PR #967, opened 2025-01-08
DRAFT: feat: asynchronous builds

itayd:


proto/autokitteh/builds/v1/build.proto
```diff
@@ -2,10 +2,24 @@ syntax = "proto3";
...
 
+import "autokitteh/program/v1/program.proto";
 import "google/protobuf/timestamp.proto";
 
 message Build {
   string build_id = 1;
   string project_id = 2;
+
+  BuildStatus status = 4;
+  program.v1.Error error = 5; // populated if status == ERROR.
```
efiShtain, line 13: if we are going to do image building, errors might be other than program

timeout
too large
out of storage
maybe other stuff

proto/autokitteh/builds/v1/svc.proto
```diff
@@ -16,6 +16,7 @@ message GetResponse {
 message ListRequest {
   uint32 limit = 1;
   string project_id = 2;
+  BuildStatus status = 3;
```
efiShtain, line 20: do we want to list by status ?

proto/autokitteh/builds/v1/svc.proto
```diff
@@ -24,13 +25,12 @@ message ListResponse {
...
-    expression: "has(this.build) && this.build.build_id == ''"
-  };
-
-  Build build = 1;
+  // If build.build_id is not set, a new build_id will be generated.
+  // If build.build_id is populated, it will be used if exists. The other
+  // content of the build message will be matched to the current one,
+  // and if there is any difference, the request will be rejected.
+  // build.status must not be set.
+  Build build = 1 [(buf.validate.field).required = true];
```
efiShtain: maybe only Create should create a build
and save / update should update by ID without comparing other fields? 
let the user do what he wants

proto/autokitteh/builds/v1/svc.proto
```diff
@@ -61,11 +61,37 @@ message DescribeResponse {
...
+}
+
+message CreateResponse {
+  string build_id = 1 [(buf.validate.field).string.min_len = 1];
+}
+
+message UpdateBuildStatusRequest {
+  string build_id = 1 [(buf.validate.field).string.min_len = 1];
+  BuildStatus status = 2;
```
efiShtain, line 83: what about the error in case of an error?

proto/autokitteh/projects/v1/svc.proto
```diff
@@ -76,10 +76,16 @@ message ListResponse {
 
 message BuildRequest {
   string project_id = 1 [(buf.validate.field).string.min_len = 1];
+
+  // if set to true, just create the build and return.
+  // the build will be processed asynchronously.
+  bool async = 2;
```
efiShtain, line 82: can't we make it always async ?

the user sends a build request
and then he can poll status immediately to check if its ready

I don't see much value in letting the user wait on the wire for a build

efiShtain (commented):
Left some comments

I think we agree on the main changes

but, 
I think we can simplify and refactor a bit

- build service should handle builds (today project build handle builds)
- project service can call the build service if we want to allow it
- everything should be always async, no need to support both options
- not sure what Download build would mean for an image, I think we should remove it
- we need to add type to build - image / code
- not sure what's the different between describe / get

proto/autokitteh/builds/v1/build.proto
itayd: this can be encoded in a program error, just leave the callstack and locations out. this will allow to give also detailed build errors re actual user code issues.

proto/autokitteh/builds/v1/svc.proto
itayd: completely separated flows now.

proto/autokitteh/builds/v1/svc.proto
itayd: modified

proto/autokitteh/projects/v1/svc.proto
itayd: backward compatability - currently the ui is relying on this to be blocking.

itayd:
> * build service should handle builds (today project build handle builds)
> * project service can call the build service if we want to allow it

Agreed, but I think this should be addressed in a different PR.

> * everything should be always async, no need to support both options

See comment below re legacy and compatability.

> * not sure what Download build would mean for an image, I think we should remove it
> * we need to add type to build - image / code

We'll discuss - I think that the fact that its either an image or a compiled code should be encoded in the data.


> * not sure what's the different between describe / get

It's a UI thing, not sure if being used. Will find out.

proto/autokitteh/projects/v1/svc.proto
efiShtain: so let's do v2 ?

itayd: ugh, think it's an overkill for something like this. it weaves nicely into the current api.

proto/autokitteh/builds/v1/build.proto
tebeka: We need a systemic wait of signaling an internal error (such as out of storage) to user errors (can't build due to syntax error).
And discuss if `image too large` is internal or user one.

proto/autokitteh/builds/v1/build.proto
```diff
@@ -2,10 +2,24 @@ syntax = "proto3";
...
+
   google.protobuf.Timestamp created_at = 3;
+  google.protobuf.Timestamp updated_at = 6;
+}
+
+enum BuildStatus {
+  BUILD_STATUS_UNSPECIFIED = 0;
+  BUILD_STATUS_PENDING = 1;
+  BUILD_STATUS_IN_PROGRESS = 2;
+  BUILD_STATUS_READY = 3;
```
tebeka, line 23:
```suggestion
  BUILD_STATUS_SUCESS = 3;
```

proto/autokitteh/projects/v1/svc.proto
tebeka: If were not going to use `async` now, why add it?

proto/autokitteh/builds/v1/build.proto
itayd: i actually like READY better... @efiShtain ? 
also, SUCCESS

proto/autokitteh/projects/v1/svc.proto
itayd: the whole idea is to be ready to use it :)

proto/autokitteh/builds/v1/build.proto
itayd: will discuss, but i think internal errors should not be sent to the users, we need to mitigate them. if we can't - we need to deliver them to the user, no way around it. are you suggesting there will be some sort of code for internal error?

proto/autokitteh/builds/v1/build.proto
efiShtain: The build is ready
the build request was successful

so I think ready is better

proto/autokitteh/builds/v1/build.proto
efiShtain: I think any internal error like out of storage / builders not available / any error the user can't do anything with it, should be just Internal Error, probably with some ID the user can report to us so we can help him if he requests help / support

Error related to user should be returned to him - Image to large / Security scan fail / Not supported libraries / Runner can't start / maybe there are more

PR #969, opened 2025-01-08
fix(jira): reuse webhooks, delete duplicates

daabr:
Fixes:
- When creating a new webhook, don't append `"_jira"` to the JQL query, it's redundant
- When looking for existing webhooks, use `strings.Contains` instead of `strings.HasSuffix`
- When looking for existing webhooks, delete duplicate webhooks if more than one exists
- More logging

Refs: INT-130

### DIFF
--- a/integrations/atlassian/jira/webhooks.go
+++ b/integrations/atlassian/jira/webhooks.go
@@ -78,12 +78,12 @@ type webhookListResponse struct {
 
 // getWebhook checks whether the given Jira domain already has a
 // registered dynamic webhook for this AutoKitteh server. Based on:
-// https://developer.atlassian.com/cloud/jira/platform/webhooks/
+// https://developer.atlassian.com/cloud/jira/platform/webhooks/#using-the-rest-api--fetching-registered-webhooks
+// https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-webhooks/#api-rest-api-3-webhook-get
 // https://developer.atlassian.com/server/jira/platform/webhooks/
-// https://developer.atlassian.com/cloud/jira/platform/rest/v3/api-group-webhooks/
-func getWebhook(l *zap.Logger, base, token string) (int, bool) {
+func getWebhook(l *zap.Logger, baseURL, token string) (int, bool) {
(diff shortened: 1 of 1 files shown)


integrations/atlassian/jira/webhooks.go
```diff
@@ -126,14 +126,64 @@ func getWebhook(l *zap.Logger, base, token string) (int, bool) {
...
-		if strings.HasSuffix(v.JQLFilter, webhookBase) {
-			return v.ID, true
+		if strings.Contains(v.JQLFilter, webhookBase) {
+			if id == 0 {
+				id = v.ID
+			} else {
+				// Already found a webhook for this AutoKitteh server.
+				// Delete duplicates and return the first one's ID.
+				deleteWebhook(l, baseURL, token, v.ID)
+			}
```
pashafateev, line 138: Questions: 
1. How do we know that the `JQLFilter` will contain the `WEBHOOK_ADDRESS`?
2. Why are there duplicates in the first place?

integrations/atlassian/jira/webhooks.go
```diff
@@ -169,7 +219,7 @@ func registerWebhook(l *zap.Logger, base, token string) (int, bool) {
 				// "GET .../webhook" doesn't show webhook URLs in the response,
 				// so we use a trick: we specify the AutoKitteh server address in
 				// the JQL filter, without affecting the actual event filtering.
-				JQLFilter: fmt.Sprintf("project != %s_jira", webhookBase),
```
pashafateev, line 172: why was `_jira` added originally?

pashafateev (approved):

integrations/atlassian/jira/webhooks.go
daabr: 1. See the function `registerWebhook` in `webhooks.go`
2. Because the "_jira" suffix in `webhooks.go` shouldn't have been there

integrations/atlassian/jira/webhooks.go
daabr: Originally I wanted to protect against confusing between Jira and Confluence webhooks, but that was before I realized they're implemented in entirely different ways.

PR #973, opened 2025-01-09
feat(slack): add app_uninstalled event

mario99logic:
Added function `AppUninstalledHandler`.

Refs: INT-140


daabr (commented):
This PR is a good start: there's no need to add a struct for parsing in this case.

However, please also add to this PR the `tokens_revoked` event (https://api.slack.com/events/tokens_revoked). Based on the `app_uninstalled` event's documentation, there's no point in implementing this event without the other one (also with generic JSON parsing without a specific struct).

daabr (approved):

PR #974, opened 2025-01-10
fix(pythonrt): properly deal with functors

itayd:
Also added some more tests for various function and method calls

Ref: ENG-1948


tebeka:
Duplicate (my bad).

PR #975, opened 2025-01-10
feat(authz): roles

itayd:
each org member has roles. if "admin" in roles, it is an org admin and has extra privilidges such as inviting and remove users, updating their roles, etc.


tests/system/testdata/orgs/roles.txtar
```diff
@@ -0,0 +1,79 @@
...
+
+ak orgs list-members $meow_oid
+return code == 0
+output equals_jq .[].org_id $meow_oid
+output equals_jq .[].user_id $GIZMO_UID
+output equals_jq .[].status ORG_MEMBER_STATUS_ACTIVE
+
+
+-- test-config.yaml --
+exclusive: true
```
itayd:
```suggestion
```

daabr (commented):
After ignoring proto-gen files and trivial stuff, there are still 26 large diffs to review, so this going to take a while...

sdk/sdkclients/sdkorgsclient/client.go
```diff
@@ -55,6 +53,21 @@ func (c *client) GetByID(ctx context.Context, oid sdktypes.OrgID) (sdktypes.Org,
 	return sdktypes.OrgFromProto(resp.Msg.Org)
 }
 
+func (c *client) BatchGetByIDs(ctx context.Context, oids []sdktypes.OrgID) ([]sdktypes.Org, error) {
+	resp, err := c.client.BatchGet(ctx, connect.NewRequest(&orgsv1.BatchGetRequest{
```
efiShtain: we have to have a limit here, since it is a public API

proto/autokitteh/orgs/v1/svc.proto
```diff
@@ -22,6 +22,14 @@ message GetResponse {
   Org org = 1;
 }
 
+message BatchGetRequest {
```
efiShtain: why batch and not list as the other conventions ?

proto/autokitteh/orgs/v1/org.proto
```diff
@@ -13,6 +13,7 @@ message OrgMember {
   string user_id = 1;
   string org_id = 2;
   OrgMemberStatus status = 3;
+  repeated string roles = 4;
```
efiShtain, line 17: do we want / need multiple roles ?
is this future proofing stuff?

internal/kittehs/fs.go
```diff
@@ -57,6 +57,8 @@ func MapToMemFS(in map[string][]byte) (fs.FS, error) {
 	return memfs, nil
 }
 
+func TxtarStringToFS(s string) (fs.FS, error) { return TxtarToFS(txtar.Parse([]byte(s))) }
```
efiShtain: is this for testing only?
why do we need this wrapper?

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -583,14 +583,37 @@ func ParseOrg(r Org) (sdktypes.Org, error) {
 type OrgMember struct {
 	Base
 
-	OrgID  uuid.UUID `gorm:"primaryKey;type:uuid;not null"`
-	UserID uuid.UUID `gorm:"primaryKey;type:uuid;not null"`
-	Status int       `gorm:"index"`
+	OrgID  uuid.UUID      `gorm:"primaryKey;type:uuid;not null"`
+	UserID uuid.UUID      `gorm:"primaryKey;type:uuid;not null"`
+	Status int            `gorm:"index"`
+	Roles  datatypes.JSON // [str]
```
efiShtain, line 597: why is this a list ?

sdk/sdkclients/sdkorgsclient/client.go
itayd: wdym by limit? max num of entries in batch?

proto/autokitteh/orgs/v1/svc.proto
itayd: because we don't list, we get a specific set of entries by their ids.

proto/autokitteh/orgs/v1/org.proto
itayd: yes. this is definitely going to happen. we can discuss if you'd like.

internal/kittehs/fs.go
itayd: i'll remove.

proto/autokitteh/orgs/v1/org.proto
itayd: for example, if we'd like to allow all to invite others, we can have an "inviter" role.

internal/kittehs/fs.go
itayd: (used before, seems i stopped using it)

configs/opa_bundles/default/authz.rego
```diff
@@ -29,52 +37,44 @@ single_associated_project_id(pid) if {
...
 	input.action_type == "read"
-	is_org_member_of_resource
+	is_active_member_of_resource_org
 }
 
 #
 # Users
 #
 
+# A user can read its own user object.
```
daabr:
```suggestion
# A user can read their own user object.
```

internal/backend/db/dbgorm/orgs.go
```diff
@@ -42,6 +46,27 @@ func (gdb *gormdb) GetOrg(ctx context.Context, oid sdktypes.OrgID, n sdktypes.Sy
...
+	if err != nil {
+		return nil, translateError(err)
+	}
+
+	orgs, err := kittehs.TransformError(rs, scheme.ParseOrg)
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse orgs: %w", err)
+	}
+
+	if kittehs.ContainedIn(oids...)(authusers.DefaultOrg.ID()) {
```
efiShtain: if it's contained in oids, why wasn't it fetched from db?

internal/backend/db/dbgorm/orgs.go
```diff
@@ -107,62 +132,64 @@ func (gdb *gormdb) UpdateOrg(ctx context.Context, o sdktypes.Org, fm *sdktypes.F
...
-			Status: s,
-		}, nil
-	})
+	return kittehs.TransformError(ms, scheme.ParseOrgMember)
 }
 
-func (gdb *gormdb) AddOrgMember(ctx context.Context, oid sdktypes.OrgID, uid sdktypes.UserID, status sdktypes.OrgMemberStatus) error {
-	if oid == authusers.DefaultOrg.ID() {
+func (gdb *gormdb) AddOrgMember(ctx context.Context, m sdktypes.OrgMember) error {
+	if m.OrgID() == authusers.DefaultOrg.ID() {
```
efiShtain, line 154: not sure why this check is on the db level and not application level?

internal/backend/db/dbgorm/orgs.go
```diff
@@ -107,62 +132,64 @@ func (gdb *gormdb) UpdateOrg(ctx context.Context, o sdktypes.Org, fm *sdktypes.F
...
+		Status: int(m.Status().ToProto()),
+		Roles:  roles,
 	}
 
 	return translateError(gdb.db.WithContext(ctx).Create(&ou).Error)
 }
 
-func (gdb *gormdb) UpdateOrgMemberStatus(ctx context.Context, oid sdktypes.OrgID, uid sdktypes.UserID, s sdktypes.OrgMemberStatus) error {
-	if oid == authusers.DefaultOrg.ID() {
+func (gdb *gormdb) UpdateOrgMember(ctx context.Context, m sdktypes.OrgMember, fm *sdktypes.FieldMask) error {
```
efiShtain, line 173: why the field mask is propogated to the db layer ? 
it should be at the application layer or maybe even the tranport layer as a validation

internal/backend/orgsgrpcsvc/svc.go
```diff
@@ -201,14 +207,21 @@ func (s *server) ListMembers(ctx context.Context, req *connect.Request[orgsv1.Li
 		return nil, sdkerrors.AsConnectError(err)
 	}
 
+	var pbusers map[string]*usersv1.User
+	if msg.IncludeUsers {
```
efiShtain: can we move this down to the db layer and just do list members with a join to get the users ?

i think we should try to use as little logic as possible in this layer

internal/backend/orgsgrpcsvc/svc.go
```diff
@@ -201,14 +207,21 @@ func (s *server) ListMembers(ctx context.Context, req *connect.Request[orgsv1.Li
...
+		users, err := s.users.BatchGetByIDs(ctx, kittehs.Transform(ms, func(o sdktypes.OrgMember) sdktypes.UserID { return o.UserID() }))
```
efiShtain: can you rename IncludeUsers to something else ?
maybe "FullDetails" or something of that sort

The api is called list members, include users sounds like a membercan be something other than user

internal/backend/orgsgrpcsvc/svc.go
```diff
@@ -224,28 +237,26 @@ func (s *server) GetOrgsForUser(ctx context.Context, req *connect.Request[orgsv1
...
 	if err != nil {
 		return nil, sdkerrors.AsConnectError(err)
 	}
 
 	var pborgs map[string]*orgsv1.Org
 	if msg.IncludeOrgs {
-		pborgs = kittehs.ListToMap(orgs, func(o *sdkservices.OrgWithMemberStatus) (string, *orgsv1.Org) {
-			pborg := o.Org.ToProto()
-			return pborg.OrgId, pborg
+		orgs, err := s.orgs.BatchGetByIDs(ctx, kittehs.Transform(ms, func(o sdktypes.OrgMember) sdktypes.OrgID { return o.OrgID() }))
```
efiShtain: see my comment above, 
if we push this down to the db layer we can reuse the logic and get better performance (neglible, but cleaner)

internal/backend/orgsgrpcsvc/svc.go
```diff
@@ -266,18 +277,12 @@ func (s *server) GetMember(ctx context.Context, req *connect.Request[orgsv1.GetM
...
 	}
 
-	return connect.NewResponse(&orgsv1.GetMemberResponse{
-		Member: &orgsv1.OrgMember{
-			OrgId:  msg.OrgId,
-			UserId: msg.UserId,
-			Status: status.ToProto(),
-		},
-	}), nil
+	return connect.NewResponse(&orgsv1.GetMemberResponse{Member: m.ToProto()}), nil
```
efiShtain, line 240: do we even need get member in the API?

internal/backend/orgs/orgs.go
```diff
@@ -54,7 +56,9 @@ func (o *orgs) Create(ctx context.Context, org sdktypes.Org) (sdktypes.OrgID, er
 			return
 		}
 
-		if err = AddMember(ctx, tx, oid, authcontext.GetAuthnUserID(ctx), sdktypes.OrgMemberStatusActive); err != nil {
+		m := sdktypes.NewOrgMember(oid, authcontext.GetAuthnUserID(ctx)).WithStatus(sdktypes.OrgMemberStatusActive).WithRoles(OrgAdminRoleName)
```
efiShtain: NIT, can you make the line shorter?

proto/autokitteh/orgs/v1/org.proto
efiShtain: stating it again, 
this is future proofing and unrequired feature

proto/autokitteh/orgs/v1/svc.proto
efiShtain: semantic but ok

sdk/sdkclients/sdkorgsclient/client.go
efiShtain: yes, 
we can't let users ask for like 1000 objects

internal/backend/db/dbgorm/scheme/records.go
daabr: I for one support this. Supporting a list of roles instead of a single one is critical on day 1.

Although eventually they shouldn't be in this table but rather in a separate users-to-roles (many-to-many) mappings table, because we need to be able to mass-remove roles without searching and updating all users.

Also, role names should be modifiable, so again storing roles by name is problematic - but not urgent to fix.

But again, supporting multiple roles is critical.

internal/backend/gormkitteh/z.go
```diff
@@ -15,7 +15,7 @@ func OpenZ(z *zap.Logger, cfg *Config, f func(*gorm.Config)) (*gorm.DB, error) {
 	return Open(cfg, func(c *gorm.Config) {
 		l := &zapgorm2.Logger{
 			ZapLogger:                 z,
-			LogLevel:                  logger.Warn,            // prevent logging SQL queries by default
+			LogLevel:                  logger.Info,
```
daabr: This should remain `Warn` by default.

We do not want to log each and every DB query unless we're specifically debugging DB operations locally.

It leaks sensitive data, and it's very spammy.

sdk/sdkclients/sdkorgsclient/client.go
daabr: I would add this limit at the lowest (i.e. DB) layer

internal/backend/auth/authz/input.go
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	membershipsMap := kittehs.ListToMap(memberships, func(m sdktypes.OrgMember) (string, any) {
+		return m.OrgID().String(), map[string]any{
+			"status": m.Status().String(),
+			"roles":  kittehs.TransformToStrings(m.Roles()),
+		}
+	})
+
+	// TODO: Use https://docs.styra.com/opa/rego-by-example/builtins/regex/globs_match to match permissions.
```
daabr: Add a ticket ID

internal/backend/db/dbgorm/orgs.go
daabr: @efiShtain the RPC client is unreliable and optional, and the RPC server needs to be as lean as possible (if we split between the server and the DB, which we do). Checking auth on the lowest level actually makes sense to me as a last line of defense.

internal/backend/db/dbgorm/users.go
```diff
@@ -89,3 +93,24 @@ func (gdb *gormdb) UpdateUser(ctx context.Context, u sdktypes.User, fm *sdktypes
 			Error,
 	)
 }
+
+func (gdb *gormdb) BatchGetUsers(ctx context.Context, uids []sdktypes.UserID) ([]sdktypes.User, error) {
+	q := gdb.db.WithContext(ctx).Where("user_id in ?", kittehs.Transform(uids, func(uid sdktypes.UserID) uuid.UUID { return uid.UUIDValue() }))
+
+	var rs []scheme.User
+	err := q.Find(&rs).Error
```
daabr: This is where I would expect to see a hard limit on the number of elements that can be requested

sdk/sdkclients/sdkorgsclient/client.go
itayd: added in db

internal/backend/db/dbgorm/orgs.go
itayd: because maybe the oid isn't there

internal/backend/db/dbgorm/orgs.go
itayd: in the future i think i'll move default user and orgs to be seeded instead of being special.

internal/backend/db/dbgorm/orgs.go
itayd: the db needs to know what fields to update. this is a good representation as any?

internal/backend/orgsgrpcsvc/svc.go
itayd: why not?

itayd: i mean the q should be why yes, but sure - if you want to know only this specific member roles instead of listing everybody.

internal/backend/orgs/orgs.go
itayd: done

configs/opa_bundles/default/authz.rego
itayd: done

internal/backend/gormkitteh/z.go
itayd: done

internal/backend/db/dbgorm/users.go
itayd: done

tests/system/auth.go
```diff
@@ -29,11 +29,40 @@ var (
 	token = "INVALID_TOKEN"
 )
 
+func addUser(u sdktypes.User) string {
+	return fmt.Sprintf(
+		`insert into users(user_id,email,display_name,created_by,default_org_id) values (%q,%q,%q,%q,%q)`,
```
daabr: Never ever EVER use `fmt.Sprintf` to construct a DB query. Not even in test code like here. The potential for abuse is too much - even in test code.

The CodeQL failure below is not a false positive. Either use Gorm, or (less desirable, but acceptable in test code) the suggested fix below.

daabr (changes requested):
The CodeQL failure in this PR must be fixed. In many cases security risks are less or not relevant in test code, but this is not one of these cases. This is easy to abuse for SQL injection, potentially in a real DB too, beyond the expected scope of just system tests.

tests/system/auth.go
itayd: now using objects to populate the db

internal/backend/orgsgrpcsvc/svc.go
itayd: i follow the "include" pattern elsewhere... don't see much issues with it.

tests/system/server.go
```diff
@@ -21,12 +24,32 @@ const (
 	serverReadyTimeout = 20 * time.Second
 )
 
+func writeSeedObjects(t *testing.T) (string, error) {
+	for _, sobj := range seedObjects {
+		t.Logf("seed object: %s", sobj)
+	}
+
+	path := filepath.Join("/tmp" /*t.TempDir()*/, "autokitteh_seed_objects.json")
```
daabr: Bring back `t.TempDir()`

daabr (approved):

internal/backend/orgsgrpcsvc/svc.go
itayd: ended up getting rid of an includes all together and always returning the users/orgs. moved the logic for fetching to the db, but not using join yet - will do later on. this simplifies a lot of things as we don't need the BatchXXX methods anymore.

itayd:
merged with a based branch.

PR #976, opened 2025-01-11
feat(orgs): add globally unique org name

itayd:
if an org name is not given on creation, will automatically generate one.


internal/backend/db/dbgorm/orgs.go
```diff
@@ -14,21 +14,25 @@ import (
...
-		return sdktypes.InvalidOrg, sdkerrors.NewInvalidArgumentError("missing id")
+	if !oid.IsValid() && !n.IsValid() {
+		return sdktypes.InvalidOrg, sdkerrors.NewInvalidArgumentError("missing id or name")
 	}
 
 	if oid.IsValid() {
 		q = q.Where("org_id = ?", oid.UUIDValue())
 	}
 
+	if n.IsValid() {
```
tebeka, line 32: Shouldn't this be "else"? What happens if the user specifies non-matching org ID and name?

internal/backend/orgsgrpcsvc/svc.go
```diff
@@ -55,16 +55,32 @@ func (s *server) Get(ctx context.Context, req *connect.Request[orgsv1.GetRequest
...
+		return nil, sdkerrors.AsConnectError(err)
+	}
+
+	if oid.IsValid() && n.IsValid() {
+		return nil, sdkerrors.NewInvalidArgumentError("only one of org id or name must be provided")
+	}
+
+	var o sdktypes.Org
+
+	if oid.IsValid() {
```
tebeka: What is both specified and don't match?

tebeka (approved):
I'm approving. But in general I don't like guessing. I prefer that we'll force the user to specify a name.
Also, when getting a project, I prefer the user will tell us if it's a name or an ID.

itayd:
> I'm approving. But in general I don't like guessing. I prefer that we'll force the user to specify a name. Also, when getting a project, I prefer the user will tell us if it's a name or an ID.

We discussed it, this looks like the most sensible approach right now. Re the cli - we can discuss.

internal/backend/db/dbgorm/orgs.go
itayd: fixed

internal/backend/orgsgrpcsvc/svc.go
itayd: they can't be both specified, checked above

internal/backend/db/dbgorm/orgs.go
itayd: actually my mistake
if a user specifies non matching, it'll just be not found.

PR #977, opened 2025-01-12
fix: auto-drain messages shouldn't spam the log

daabr:
Also make the relevant code a little more readable.

### DIFF
--- a/internal/backend/deployments/deployments.go
+++ b/internal/backend/deployments/deployments.go
@@ -60,18 +60,17 @@ func (d *deployments) Autodrain() {
 	l.Info("periodically auto-deactivating drained deployments")
 
 	for {
-		t := d.cfg.AutoDrainingDeactivationInterval
-		t += time.Duration(rand.Float32() * float32(d.cfg.AutoDrainingDeactivationJitter))
-		time.Sleep(t)
+		jitter := rand.Float32() * float32(d.cfg.AutoDrainingDeactivationJitter)
+		time.Sleep(d.cfg.AutoDrainingDeactivationInterval + time.Duration(jitter))
 
 		n, err := d.db.DeactivateAllDrainedDeployments(context.Background())
 		if err != nil {
-			d.l.Error("auto-deactivation failed", zap.Error(err))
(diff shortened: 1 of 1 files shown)


internal/backend/deployments/deployments.go
```diff
@@ -53,25 +53,25 @@ func New(l *zap.Logger, cfg *Config, db db.DB, telemetry *telemetry.Telemetry) s
 }
 
 func (d *deployments) Autodrain() {
-	d.l.
-		With(zap.Duration("interval", d.cfg.AutoDrainingDeactivationInterval),
-			zap.Duration("jitter", d.cfg.AutoDrainingDeactivationJitter)).
-		Info("periodically autoinactivating drained deployments")
+	l := d.l.With(
```
itayd: i don't think this should appear in every log message

internal/backend/deployments/deployments.go
```diff
@@ -53,25 +53,25 @@ func New(l *zap.Logger, cfg *Config, db db.DB, telemetry *telemetry.Telemetry) s
...
+		zap.Duration("auto_drain_interval", d.cfg.AutoDrainingDeactivationInterval),
+		zap.Duration("auto_drain_jitter", d.cfg.AutoDrainingDeactivationJitter),
+	)
+	l.Info("periodically auto-deactivating drained deployments")
 
 	for {
-		t := d.cfg.AutoDrainingDeactivationInterval + time.Duration(rand.Float32()*float32(d.cfg.AutoDrainingDeactivationJitter))
-
-		d.l.Debug("waiting before next autodeactivation", zap.Duration("duration", t))
```
itayd, line 64: it's debug... but oh well

internal/backend/deployments/deployments.go
daabr: done

internal/backend/deployments/deployments.go
daabr: "Time is passing" is a category of log messages that not only don't add useful info, they dilute the value of logging because the extra noise makes it harder to find/browse/group useful info, not to mention they cost much more.

itayd (approved):

PR #979, opened 2025-01-12
bug(python): full_func_name should work on any callable

tebeka:
Fixes: ENG-1948


daabr (approved):

itayd (commented):
can you add a session test? also - how is it better/worse than https://github.com/autokitteh/autokitteh/pull/974?

tebeka:
> can you add a session test? 
There's a unit test.

> also - how is it better/worse than #974?
Missed that one.

PR #980, opened 2025-01-12
feat(slack): update mappings & TODOs of bot event handlers

daabr:
Remove events that we don't plan to support at this time, and add events that don't require any new code.

Testing for events: manual triggering via Slack website (e.g. https://api.slack.com/methods/conversations.leave/test). Specifically note that some events refer to the bot itself, not users, so they required 1-2 manual API calls.

Refs: INT-94, INT-142

### DIFF
--- a/integrations/slack/events/channel.go
+++ b/integrations/slack/events/channel.go
@@ -60,8 +60,12 @@ func (b *BoolOrInt) UnmarshalJSON(data []byte) error {
 }
 
 // https://api.slack.com/events/channel_archive
+// https://api.slack.com/events/channel_left
 // https://api.slack.com/events/channel_unarchive
 // https://api.slack.com/events/group_archive
+// https://api.slack.com/events/group_close
+// https://api.slack.com/events/group_deleted
+// https://api.slack.com/events/group_left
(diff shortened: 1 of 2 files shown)


mario99logic (commented):
I have a question not related to the specific code in the PR, but if you handled these events using ChannelCreatedRenameHandler why didn't we do the same for channel_created and channel_rename.

And another general question how did you test the events, I want to see if there is a more efficient way.

mario99logic (approved):

daabr:
Good questions.

> if you handled these events using ChannelCreatedRenameHandler why didn't we do the same for channel_created and channel_rename.

One key difference: the channel is a string in `ChannelGroupMemberEvent`, but a struct with sub-fields in `ChannelCreatedRenameEvent`.

> how did you test the events, I want to see if there is a more efficient way.

Added to the PR description. The most efficient way for events that can't be triggered automatically by user actions is using the Slack API reference website: each API call has a test UI (e.g. https://api.slack.com/methods/conversations.leave/test). You can copy the bot token from a connection, and make manual API calls easily.

PR #981, opened 2025-01-13
fix: reorder channel subscription

efiShtain:
in cases where user perform illegal acts
like calling the workflow function in the global
scope, we had an issue where the callback channels were not setup yet and that cause a deadlock which caused the workflow to keep restarting forever
this change refactor the callback channels so that we first subscribe to callback channels, then start the workflow and then subscribe to the done and health check channels as before.

### DIFF
--- a/runtimes/pythonrt/pythonrt.go
+++ b/runtimes/pythonrt/pythonrt.go
@@ -387,6 +387,108 @@ func (py *pySvc) call(ctx context.Context, val sdktypes.Value, args []sdktypes.V
 	}
 }
 
+func (py *pySvc) setupCallbackChannels(ctx context.Context) chan (error) {
+	callbackErrChan := make(chan error, 1)
+	go func() {
+		for {
+			select {
+			case r := <-py.channels.log:
+				py.log.Log(pyLevelToZap(r.level), r.message)
+				close(r.doneChannel)
+			case p := <-py.channels.print:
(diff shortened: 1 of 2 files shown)


itayd:
would love to see some tests around this
even session tests

runtimes/pythonrt/pythonrt.go
```diff
@@ -387,32 +387,60 @@ func (py *pySvc) call(ctx context.Context, val sdktypes.Value, args []sdktypes.V
 	}
 }
 
-// initialCall handles initial call from autokitteh, it does the message loop with Python.
-// We split it from Call since Call is also used to execute activities.
-func (py *pySvc) initialCall(ctx context.Context, funcName string, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error) {
-	if len(args) > 0 {
-		return sdktypes.InvalidValue, errors.New("initial call can't have positional args")
-	}
+func (py *pySvc) setupCallbackChannels(ctx context.Context) chan (error) {
```
tebeka: Find a better name? You're not setting up the channels,  your're starting the event loop.

tebeka (approved):
One minor comment

runtimes/pythonrt/pythonrt.go
efiShtain: fixed

efiShtain:
added one python test

PR #982, opened 2025-01-14
feat(integrations): add google drive events

pashafateev:
### **Changes**:
1. **Drive Changes API Integration**:
   - Changes from the Google Drive API are parsed and passed directly to users. No additional calls to the File API are made, as users can query further details as needed.
   - Notifications include events like file creation, modification, and deletion.

2. **Testing & Observations**:
   - Successfully tested by creating, modifying, and deleting files in Drive. Notifications were received and processed correctly.
   - File creation triggers two non-duplicate notifications, and currently, there is no clear way to filter these. This behavior might be a nuisance in workflows.
   - Changes are scoped to the authenticated connection (file-level scope for now).

---

### **Limitations**:
1. **Notification Volume**: Google Drive sends notifications for a variety of file operations, including opening files. These may be considered noise in certain workflows.
2. **Watch Expiration**: The watch expiration is limited to Google's maximum duration of one week. As a result, users will need to reauthenticate after a week unless the watch is manually renewed. This limitation will be addressed after the merge of **INT-156**, which adds watch renewal support.

---

### **Next Steps**:
1. **Enhancements**:
   - Implement watch renewal after **INT-156** merges.
   - Explore strategies for filtering noisy notifications.
   - Consider adding support for querying file metadata from the File API.

### DIFF
--- a/integrations/google/drive/client.go
+++ b/integrations/google/drive/client.go
@@ -1,7 +1,21 @@
 package drive
 
 import (
+	"context"
+	"fmt"
+	"os"
+	"time"
+
+	"go.uber.org/zap"
+	"golang.org/x/oauth2"
+	"golang.org/x/oauth2/google"
+	"google.golang.org/api/drive/v3"
+	"google.golang.org/api/googleapi"
+	googleoauth2 "google.golang.org/api/oauth2/v2"
+	"google.golang.org/api/option"
+
 	"go.autokitteh.dev/autokitteh/integrations/google/connections"
+	"go.autokitteh.dev/autokitteh/integrations/google/internal/vars"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/sdk/sdkintegrations"
 	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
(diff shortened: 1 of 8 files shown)


integrations/google/drive/events.go
```diff
@@ -0,0 +1,75 @@
...
+func ConstructEvents(ctx context.Context, vars sdkservices.Vars, cids []sdktypes.ConnectionID) ([]sdktypes.Event, error) {
+	l := extrazap.ExtractLoggerFromContext(ctx)
+	if len(cids) == 0 {
+		return nil, nil
+	}
+
+	a := api{logger: l, vars: vars, cid: cids[0]}
+	changes, err := a.listChanges(ctx)
+	if err != nil {
+		l.Error("Failed to list events", zap.Error(err))
```
daabr:
```suggestion
		l.Error("Failed to list Google Drive events", zap.Error(err))
```

integrations/google/drive/events.go
```diff
@@ -0,0 +1,75 @@
...
+		return nil, err
+	}
+
+	if len(changes) == 0 {
+		return nil, nil
+	}
```
daabr: Unnecessary check - the loop below has the same effect anyway.

integrations/google/drive/client.go
```diff
@@ -37,3 +57,261 @@ func New(cvars sdkservices.Vars) sdkservices.Integration {
...
+	if err != nil {
+		return nil, err
+	}
+
+	var src oauth2.TokenSource
+	if data.OAuthData != "" {
+		src, err = oauthTokenSource(ctx, data.OAuthData)
+	} // else {
+	// 	src, err = jwtTokenSource(ctx, data.JSON)
+	// }
```
daabr: ?

daabr (commented):
Great PR description!

daabr (approved):
Approved with a few tiny nits. Please test service accounts work before merging.

integrations/google/drive/client.go
pashafateev: Added support for service accounts

integrations/google/drive/events.go
pashafateev: Done

PR #983, opened 2025-01-15
feat(authz): add roles, simplify structure + some perm changes

itayd:
this simplifies the structure of the authz input to opa, as well as simplifies a bit the policy itself.
also add more tests.

additionally:
- permit any user to get any other user details.
- deny admins removing themselves from orgs.
- allow non-admins users removing themselves from orgs.

based on #975 .

Ref: ENG-1951.


internal/backend/auth/authz/input.go
```diff
@@ -73,56 +113,27 @@ func buildInput(ctx context.Context, db db.DB, id sdktypes.ID, action string, cf
...
-		"resource_id":         id.String(),  // resource id.
-		"resource_org_id":     oid.String(), // resource org id, if resource id is valid. else "".
-		"resource_project_id": pid.String(), // resource project id, if resource id is valid. else "".
-		"data":                cfg.data,     // aux data supplied by the caller.
-
-		// Explicit associations given as part of the check call.
-		"associated_org_ids":     slices.Collect(maps.Keys(oidsSet)), // all unique non-zero associated org ids and the resource org id.
-		"associated_project_ids": slices.Collect(maps.Keys(pidsSet)), // all unique non-zero associated project ids and the resource project id.
-		"associations":           associations,                       // name -> {"org_id": "xxx"} of associated resources.
+	// TODO: Use https://docs.styra.com/opa/rego-by-example/builtins/regex/globs_match to match permissions.
```
daabr, line 121: Linear ID, at least to investigate this. This is to document your current understanding of this task in detail, and known context - both of which you'll forget in a couple of days, and this comment will stay here forever.

configs/opa_bundles/default/authz.rego
```diff
@@ -6,29 +6,34 @@ import rego.v1
...
+	role in m.roles
 }
 
-is_org_admin(org_id) if has_active_role_in_org(org_id, "admin")
+is_active_org_member_of(org_id) if has_active_role_in_org(input.authn_user, org_id, "")
+
+is_org_admin(org_id) if has_active_role_in_org(input.authn_user, org_id, "admin")
 
-is_active_member_of_resource_org := is_active_org_member_of(input.resource_org_id)
+is_active_member_of_resource_org := is_active_org_member_of(input.resource.org_id)
```
daabr: "Resource" is too generic, it keeps confusing me when I try to read this file. Please rename to "subject" throughout this file.

configs/opa_bundles/default/authz.rego
```diff
@@ -6,21 +6,34 @@ import rego.v1
...
+has_active_role_in_org(user, org_id, "") if {
+	m := user.org_memberships[org_id]
+	m.status == "ACTIVE"
+}
+
+has_active_role_in_org(user, org_id, role) if {
+	m := user.org_memberships[org_id]
+	m.status == "ACTIVE"
+	role in m.roles
+}
```
daabr: It's confusing that both of these have the same name. Rename the 2nd and reuse the firs inside it, or keep only the 2nd one.

daabr (approved):

configs/opa_bundles/default/authz.rego
itayd: done

configs/opa_bundles/default/authz.rego
itayd: done

PR #985, opened 2025-01-15
feat(python): Add a retry library

tebeka:
This adds "tenacity" with have a "retry" decorator so users can do retries.

Fixes: ENG-1355

### DIFF
--- a/runtimes/pythonrt/runner/pyproject.toml
+++ b/runtimes/pythonrt/runner/pyproject.toml
@@ -25,6 +25,7 @@ all = [
 	"openai ~= 1.57",
 	"PyGithub @ git+https://github.com/PyGithub/PyGithub.git@b5f8f078f2b7fc0db6a041499c7a05b0bc7e750d",
 	"slack-sdk ~= 3.33",
+	"tenacity ~= 9.0",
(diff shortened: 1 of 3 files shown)


runtimes/pythonrt/runner/pyproject.toml
```diff
@@ -25,6 +25,7 @@ all = [
 	"openai ~= 1.57",
 	"PyGithub @ git+https://github.com/PyGithub/PyGithub.git@b5f8f078f2b7fc0db6a041499c7a05b0bc7e750d",
 	"slack-sdk ~= 3.33",
+	"tenacity ~= 9.0",
```
daabr, line 28: Nit: I think it should be near line 34 (the "General" packages section, rather than "Integrations"), or we can just merge them into a single block, but either way I like this addition.

daabr (approved):
Approved - see nit comment, choose whichever approach you prefer (merge as-is / change line / merge blocks).

runtimes/pythonrt/runner/pyproject.toml
tebeka: You're right. Fixed.

PR #986, opened 2025-01-16
feat: Add auth0 connTest function

mario99logic:
ConnTest function checks the response from the /api/v2/stats/active-users

### DIFF
--- a/integrations/auth0/client.go
+++ b/integrations/auth0/client.go
@@ -24,7 +24,7 @@ var (
 	clientIDName     = sdktypes.NewSymbol("client_id")
 	clientSecretName = sdktypes.NewSymbol("client_secret")
 	domainName       = sdktypes.NewSymbol("auth0_domain")
-	authToken        = sdktypes.NewSymbol("oauth_AccessToken")
+	authTokenName        = sdktypes.NewSymbol("oauth_AccessToken")
 )
 
 var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.IntegrationPB{
--- a/integrations/auth0/client.go
+++ b/integrations/auth0/client.go
@@ -97,7 +97,7 @@ func connTest(i *integration) sdkintegrations.OptFn {
 			return sdktypes.InvalidStatus, err
 		}
 
-		# TODO(INT-124): Use the refresh token to get a new access token.
+		// TODO(INT-124): Use the refresh token to get a new access token.
 		token := vs.Get(authTokenName).Value()
 		domain := vs.Get(domainName).Value()
 		// https://auth0.com/docs/api/management/v2/stats/get-active-users
--- a/integrations/auth0/client.go
+++ b/integrations/auth0/client.go
@@ -2,6 +2,8 @@ package auth0
 
 import (
 	"context"
+	"fmt"
+	"net/http"
 
 	"go.uber.org/zap"
 
@@ -22,6 +24,7 @@ var (
 	clientIDName     = sdktypes.NewSymbol("client_id")
 	clientSecretName = sdktypes.NewSymbol("client_secret")
 	domainName       = sdktypes.NewSymbol("auth0_domain")
+	authToken        = sdktypes.NewSymbol("oauth_AccessToken")
 )
 
 var desc = kittehs.Must1(sdktypes.StrictIntegrationFromProto(&sdktypes.IntegrationPB{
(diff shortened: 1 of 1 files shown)


integrations/auth0/client.go
```diff
@@ -22,6 +24,7 @@ var (
 	clientIDName     = sdktypes.NewSymbol("client_id")
 	clientSecretName = sdktypes.NewSymbol("client_secret")
 	domainName       = sdktypes.NewSymbol("auth0_domain")
+	authToken        = sdktypes.NewSymbol("oauth_AccessToken")
```
pashafateev:
```suggestion
	authTokenName        = sdktypes.NewSymbol("oauth_AccessToken")
```
I think it's confusing to name it `authToken` because it's not the token itself, it's just the symbol.
if you're going to add this, then I think you should also change line 74: 
```
token := vs.GetValueByString("oauth_AccessToken")
```
@daabr do we have a preference here?

daabr: I agree with @pashafateev , but I don't have a specific recommendation. `XxxName` or `XxxVarName` seems reasonable.

integrations/auth0/client.go
```diff
@@ -79,9 +82,42 @@ func connStatus(i *integration) sdkintegrations.OptFn {
 	})
 }
 
-// TODO(INT-113): Implement
-func connTest(_ *integration) sdkintegrations.OptFn {
+// connTest is an optional connection test provided by the integration
+// to AutoKitteh. It is used to verify that the connection is working
+// as expected. The possible results are "OK" and "error
```
daabr:
```suggestion
// as expected. The possible results are "OK" and "error.
```

integrations/auth0/client.go
```diff
@@ -79,9 +82,42 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+// https://auth0.com/docs/api/management/v2/stats/get-active-users
```
daabr: Given that this URL isn't connected to the function description above, maybe move it directly above line 102, or clarify in the function description that it's based on this API call.

integrations/auth0/client.go
```diff
@@ -79,9 +82,42 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+		if !cid.IsValid() {
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, "Init required"), nil
+		}
+
+		vs, err := i.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		token := vs.Get(authToken).Value()
```
daabr: What happens if the saved token is already expired?

daabr (commented):
Most of my comments are nits. The blocker for approval is the comment about this line:
```
token := vs.Get(authToken).Value()
```

integrations/auth0/client.go
mario99logic: If the token is expired then the check will fail because it's unauthorized, it responds with an HTTP 401 Unauthorized status code. and the output will be Invalid OAuth token

integrations/auth0/client.go
```diff
@@ -79,9 +82,42 @@ func connStatus(i *integration) sdkintegrations.OptFn {
 	})
 }
 
-// TODO(INT-113): Implement
-func connTest(_ *integration) sdkintegrations.OptFn {
+// connTest is an optional connection test provided by the integration
+// to AutoKitteh. It is used to verify that the connection is working
+// as expected. The possible results are "OK" and "error".
```
tebeka, line 88: > The possible results are "OK" and "error"

I don't understand this. Do you mean the possible statuses?

integrations/auth0/client.go
```diff
@@ -79,9 +82,42 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+		if err != nil {
+			return sdktypes.InvalidStatus, err
+		}
+
+		token := vs.Get(authTokenName).Value()
+		domain := vs.Get(domainName).Value()
+		// https://auth0.com/docs/api/management/v2/stats/get-active-users
+		url := fmt.Sprintf("https://%s/api/v2/stats/active-users", domain)
+
+		req, err := http.NewRequest(http.MethodGet, url, nil)
```
tebeka: Always have timeouts for HTTP requests. See https://pkg.go.dev/net/http#NewRequestWithContext

integrations/auth0/client.go
```diff
@@ -79,9 +82,42 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+func connTest(i *integration) sdkintegrations.OptFn {
 	return sdkintegrations.WithConnectionTest(func(ctx context.Context, cid sdktypes.ConnectionID) (sdktypes.Status, error) {
-		return sdktypes.NewStatus(sdktypes.StatusCodeUnspecified, "Not implemented"), nil
+		if !cid.IsValid() {
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, "Init required"), nil
+		}
+
+		vs, err := i.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+		if err != nil {
+			return sdktypes.InvalidStatus, err
```
tebeka, line 97: Why do you return error here and below nil error with error status? What's the difference?

tebeka (dismissed):
See comment on timeouts

integrations/auth0/client.go
mario99logic: yes

integrations/auth0/client.go
mario99logic: it verifies whether the retrieval of variables specific to the given connection (cid) was successful, If this is the case then there is a problem with AK. if there is an error below then the issue is with the HTTP request. 
@daabr am I correct?

integrations/auth0/client.go
mario99logic: Added timeout in the new version

integrations/auth0/client.go
daabr: The distinction is between an error in the server while trying to test the connection, and a connection test failure due to missing/stale auth data.

daabr (dismissed):

integrations/auth0/client.go
daabr: In that case it's a bug, you need to get a fresh the access token

integrations/auth0/client.go
daabr: See also INT-191

integrations/auth0/client.go
```diff
@@ -79,9 +83,45 @@ func connStatus(i *integration) sdkintegrations.OptFn {
...
+		}
+
+		token := vs.Get(authTokenName).Value()
```
daabr, line 101: See https://github.com/autokitteh/autokitteh/pull/986#discussion_r1923293521

The whole point of this check is to show that the connection is usable, not that it was created more than an hour ago.

daabr, line 101:
```suggestion
		# TODO(INT-124): Use the refresh token to get a new access token.
		token := vs.Get(authTokenName).Value()
```

daabr (approved):

PR #987, opened 2025-01-17
feat: add custom oauth to github

pashafateev:
- **Most notable change**: instead of checking for the `private_key` in Go, this code checks for it in the Python SDK for simplicity. Exposing the private key is acceptable since it has already been uploaded by the user and is not the private key for our GitHub app.

**Testing**
- Successfully authenticated connection with project vars.
- Received events and ran GitHub sample in kittehub.


integrations/github/save.go
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	// Read and parse POST request body.
+	if err := r.ParseForm(); err != nil {
+		l.Warn("Failed to parse incoming HTTP request", zap.Error(err))
+		c.AbortBadRequest("failed to parse form data")
+		return
+	}
+
+	clientID := r.FormValue("client_id")
```
tebeka: Instead of all these variables, maybe have a struct and pass it along?

```go
type AuthValues struct {
    clientID string
    clientSected string
    // ...
}

func formAuth(vals url.Values) AuthValues {
     return AuthValues {
         clientID: vals.Get("client_id"),
         clientSecret: vals.Get("client_secret"),
        // ...
}
```

You can also add a `Validate` method to this struct.

runtimes/pythonrt/py-sdk/autokitteh/github.py
```diff
@@ -47,7 +47,13 @@ def github_client(connection: str, **kwargs) -> Github:
     if not install_id:
         raise ConnectionInitError(connection)
 
-    app = GithubIntegration(auth=AppAuth(int(app_id), connection), **kwargs)
+    app = None
+    private_key = os.getenv(f"{connection}__private_key")
```
tebeka, line 51: Nit: Since Python 3.8 you can write:

```python
if (private_key = os.getenv(f"{connection}__private_key")):
    app = GithubIntegration(auth=Auth.AppAuth(int(app_id), private_key), **kwargs)
 else:
     app = GithubIntegration(auth=AppAuth(int(app_id), connection), **kwargs)
```

tebeka (commented):
I'd love to see some tests, at least for the utility functions.

web/github/connect/index.html
```diff
@@ -1,134 +1,146 @@
 <!DOCTYPE html>
```
efiShtain, line 1: @RonenMars 
Do we still need this page ? 
didn't we move everything to the web UI ?

runtimes/pythonrt/py-sdk/autokitteh/github.py
efiShtain: I think @pashafateev version is more readable

I'd remove the else and return earlier 
but that's just a preference

internal/backend/oauth/oauth.go
```diff
@@ -502,6 +505,29 @@ func (o *oauth) getConfigWithConnection(ctx context.Context, intg string, cid sd
 		optsCopy["audience"] = strings.Replace(opts["audience"], placeholder, domain, 1)
 	}
 
+	if intg == "github" {
```
efiShtain: can you extract this logic to a function?
not sure why we have such long functions with so many if conditions

integrations/github/webhooks/webhook.go
```diff
@@ -192,6 +197,34 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		l.Error("Failed to find connection IDs", zap.Error(err))
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		return "", err
+	}
+	if len(cids) == 0 {
+		l.Warn("Received GitHub event from app webhook, but no relevant connection found")
+		return "", nil
+	}
+	cid := cids[0]
```
efiShtain, line 216: what happens if there are multiple connections ?

integrations/github/webhooks/webhook.go
```diff
@@ -246,3 +279,11 @@ func (h handler) dispatchAsyncEventsToConnections(ctx context.Context, cids []sd
 		l.Debug("Event dispatched")
 	}
 }
+
+func (h handler) isCustomOAuth(ctx context.Context, cid sdktypes.ConnectionID) bool {
+	vs, err := h.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+	if err != nil {
+		return false
+	}
+	return vs.GetValueByString("client_secret") != ""
```
efiShtain: why does this say that it's custom oauth ?
can we make it more clear somehow?
also, not sure about "custom oauth", 
maybe userDefinedOauth or something along that way?
@daabr WDYT?

integrations/github/webhooks/webhook.go
```diff
@@ -192,6 +197,34 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		http.Error(w, "Internal Server Error", http.StatusInternalServerError)
+		return "", err
+	}
+	if len(cids) == 0 {
+		l.Warn("Received GitHub event from app webhook, but no relevant connection found")
+		return "", nil
+	}
+	cid := cids[0]
+	if h.isCustomOAuth(r.Context(), cid) {
+		vs, err := h.vars.Get(r.Context(), sdktypes.NewVarScopeID(cid))
```
efiShtain: isCustomOauth fetch all vars from db, just to check if it's custom oauth
if it is, the next line, you fetch the exact same vars from db

please fetch once and do the relevant checks

integrations/github/webhooks/webhook.go
```diff
@@ -192,6 +197,34 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	// Returning immediately without an error = acknowledgement of receipt.
 }
 
+func (h handler) getWebHookSecret(w http.ResponseWriter, r *http.Request) (string, error) {
+	l := extrazap.ExtractLoggerFromContext(r.Context())
+
+	s := os.Getenv(webhookSecretEnvVar)
```
efiShtain: we can probably do that once when we start this handler module, no need to redo everytime since it is a fixed value

integrations/github/oauth.go
```diff
@@ -92,7 +110,14 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	gh, err := newClientFromGitHubAppID(aid)
+	vs, err := h.vars.Get(r.Context(), sdktypes.NewVarScopeID(cid))
```
efiShtain: you get the same vs 3 times
in line 113
line 97
and in isCustomOauth

please fetch ones and reuse

integrations/github/oauth.go
```diff
@@ -92,7 +110,14 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
...
+	if err != nil {
+		l.Warn("failed to get vars", zap.Error(err))
+		c.AbortServerError("failed to get vars")
+		return
+	}
+
+	gh, err := NewClientFromGitHubAppID(aid, vs)
```
efiShtain: if the name of the functions says it accepts github app id, i think you should pass github app id and not the entire vs object
if you need more values change the signature or rename the function please

web/github/connect/index.html
daabr: UI code will exist only after the backend is already working, i.e. tested, so yes.

Are these HTML pages too complex? Yes, I'm making a prototype of a simpler and safer page in #1000.

Do we need the internal init UI after the web UI is ready? Yes, until INT-63 is fixed.

runtimes/pythonrt/py-sdk/autokitteh/github.py
```diff
@@ -47,7 +47,13 @@ def github_client(connection: str, **kwargs) -> Github:
...
+    if private_key:
+        # Exposing the private key is fine here as it's a custom OAuth and stored in connection variables.
```
daabr: The reader of this code doesn't need to know about AK's concept of "custom" OAuth, or connection vars, just why it's not a security risk.

```suggestion
        # Exposing the private key is fine here as it belongs to the user.
```

runtimes/pythonrt/py-sdk/autokitteh/github.py
daabr: Personally, I like @tebeka's version, but I agree with @efiShtain that it's less readable, so I'd go with the existing code too.

However, I don't think we should return early, because that would require duplicating the `return` line, which is a waste.

Unlike Go, the following is a very common practice in Python (in case `AAA` or `BBB` or `CCC` are not short enough to allow a single-line `a = BBB if AAA else CCC`, like here):

```py
if AAA:
   a = BBB
else:
   a = CCC
# Do something with a
```

internal/backend/oauth/oauth.go
```diff
@@ -59,6 +61,7 @@ func New(l *zap.Logger, vars sdkservices.Vars) sdkservices.OAuth {
 	l.Debug("Atlassian base URL for OAuth", zap.String("url", atlassianBaseURL))
 
 	// Determine GitHub base URL (to support GitHub Enterprise Server, i.e. on-prem).
+	// TODO: Add support for custom OAuth.
```
daabr: Create a ticket, and specify the ID here. Mention this PR in that ticket as a context link.

internal/backend/oauth/oauth.go
daabr: This would also be easily-testable, compared this its current location

integrations/github/webhooks/webhook.go
daabr: The "official" terms that I asked @RonenMars to use in the web UI are "OAuth v2/OAuth 2.0/GitHub Default App" and "... Custom App".

This is to clarify that it's not OAuth that is custom (which is sus), it's the app. It's not perfect, but it's unambiguous and concise, and that's really the important thing.

That said, in the backend code we did use "custom OAuth" when referring to this feature, just because it's short. But I think `"userDefinedApp"` is a better term to use **in the code** (accurate and not too long).

integrations/github/save.go
daabr: Alternatively, why read these values here pass them to `saveClientIDAndSecret`, instead of reading them inside it?

Personally, I wouldn't try to validate them (beyond the existing emptiness check) because right after you save them you start an OAuth flow - it anything is wrong the flow will simply fail, which is the best kind of black-box verification.

integrations/github/save.go
```diff
@@ -0,0 +1,87 @@
...
+		sdktypes.NewVar(vars.WebhookSecret).SetValue(webhookSecret).WithScopeID(scopeID).SetSecret(true),
+		sdktypes.NewVar(vars.EnterpriseURL).SetValue(enterpriseURL).WithScopeID(scopeID),
+		sdktypes.NewVar(vars.PrivateKey).SetValue(privateKey).WithScopeID(scopeID).SetSecret(true),
+	}
+
+	if err := h.vars.Set(ctx, vars...); err != nil {
+		return err
+	}
+
+	return nil
```
daabr:
```suggestion
	return h.vars.Set(ctx, vars...)
```

integrations/github/oauth.go
pashafateev: Done

integrations/github/webhooks/webhook.go
pashafateev: @efiShtain This check `vs.GetValueByString("client_secret") != ""` determines that it's a `userDefinedApp` because the only way a `client_secret` will be present in `vars` is if it was explicitly set by the user during the connection setup process. While a `client_secret` could theoretically be added manually through the internal web dashboard for other purposes, iiuc, this situation is only possible in self-hosted environments, and we don't care about that.

integrations/github/save.go
pashafateev: Done

integrations/github/webhooks/webhook.go
pashafateev: Are you sure? Depending on the `cid` this might be a user defined app. In that case this value will be replaced by `s = vs.GetValueByString("webhook_secret")`. I'm pretty sure it needs to checked with every request, unless I'm missing something.

integrations/github/webhooks/webhook.go
pashafateev: It won't matter because they're all part of the same app. it will either be our app or a user defined app, but it can't be both.
```
cids, err := h.vars.FindConnectionIDs(r.Context(), h.integrationID, vars.AppID, appID)
```

integrations/github/webhooks/webhook.go
pashafateev: Done

internal/backend/oauth/oauth.go

internal/backend/oauth/oauth.go

integrations/github/webhooks/webhook.go
pashafateev: @efiShtain Regarding "custom OAuth," I don’t have a strong preference for what it’s called, as long as the naming is consistent. The main reason I’ve been using "custom OAuth" is because it’s the term we’ve been using throughout the project so far. It’s already referenced in multiple places in the repository, beyond the scope of this PR. 

If we decide to change it here, I think we should update it everywhere to avoid confusion. However, if you feel this change is important, let me know, and I’ll make the update.

integrations/github/oauth.go
pashafateev: I generally agree with your point, but I’d like to push back in this specific case. The `vs` parameter is being passed solely because `getPrivateKey()` requires it to determine whether it’s a custom OAuth scenario. While I could retrieve the private key outside of `NewClientFromGitHubAppID`, I’m unsure if the effort is justified. What would we gain by doing so?

integrations/github/webhooks/webhook.go
efiShtain: So maybe limit the request to just one?
Just in case we have lots of connections so we don't fetch them for nothing
Should be supported in filter I think

integrations/github/webhooks/webhook.go
efiShtain: Can you add a specific indicator / symbol just for this feature so you can check and know it is custom oauth?

Instead of checking some other symbols exists

efiShtain: I prefer to rename
But if it's beyond the scope of this pr you can open a ticket and do it later

integrations/github/oauth.go
efiShtain: Readability - as I stated in the previous comment, signature is important and should explain what the function expects

Testability - you check concrete stuff and not unknown map, plus you don't need to handle the case the field is not in the vs object

Maintainability - if someone rename the var, it won't break

integrations/github/webhooks/webhook.go
daabr: @efiShtain I agree we should add a "FindConnectionID" function in the vars API, but not as a part of this PR. I created INT-197 for this.

integrations/github/webhooks/webhook.go
pashafateev: Both concerns are outside the scope of this PR because the `userDefinedApp` code affects all integrations, not just github:
INT-196 for rename
INT-198 for adding the new symbol

integrations/github/oauth.go
pashafateev: Done

integrations/github/webhooks/webhook.go
daabr: @pashafateev FYI, I added comments to both issues

integrations/github/jwt.go
```diff
@@ -112,7 +116,11 @@ func newClientFromGitHubAppID(appID int64) (*github.Client, error) {
 	return client, nil
 }
 
-func getPrivateKey() []byte {
+func getPrivateKey(privateKey string) []byte {
```
daabr, line 119: I'm fine with merging this as-is because we'll probably refactor it soon enough, but it's better to call `getPrivateKey` as early as possible (in/before line line 57) rather than "deep" in line 97.

In other words, the general principle is that when you're passing a value (e.g. a private key), it's better to know that you're passing the right one from the start or as soon as possible, rather than passing something which is tentative and may change just before it's being used.

It's not an absolute principle, but obviously it promotes less cognitive load on the reader of the code.

daabr (approved):

integrations/github/webhooks/webhook.go
efiShtain: Technically, the grpc service has findConnectionID (single, not plural) and it calls findConnectionIDS in the service. so it's a typo. 

I thought about adding a maxConnections or a limit argument to the function, but both options are fine

resolving since you opened 197

integrations/github/webhooks/webhook.go
efiShtain: it says env var, so I assume env var is not changed after the service is started

pashafateev: Ah, I misunderstood what you meant. Done.

PR #991, opened 2025-01-20
feat(jira): renew event watches

daabr:
Add to AK a Temporal worker that runs every 8 hours to perform internal maintenance tasks.

For now, this renews event watches in Jira OAuth connections, where the watch has less than 2 weeks to expire.

Also added placeholders for relevant Google event watches (INT-184) to be implemented separately.

Tested manually using the testing tip in INT-184, and commenting out the watch age check.

Refs: INT-156, INT-184


internal/backend/crontab/crontab.go
```diff
@@ -0,0 +1,180 @@
...
+func (ct *Crontab) workflow(wctx workflow.Context) error {
+	// Start multiple child workflows in parallel.
+	cwfs := []workflow.ChildWorkflowFuture{}
+	cwfs = append(cwfs, workflow.ExecuteChildWorkflow(wctx, ct.renewGoogleCalendarEventWatchesWorkflow))
+	cwfs = append(cwfs, workflow.ExecuteChildWorkflow(wctx, ct.renewGoogleDriveEventWatchesWorkflow))
+	cwfs = append(cwfs, workflow.ExecuteChildWorkflow(wctx, ct.renewJiraEventWatchesWorkflow))
+
+	// Report an error if any child workflow failed.
+	for _, future := range cwfs {
+		if err := future.Get(wctx, nil); err != nil {
```
itayd: use workflow selector

internal/backend/crontab/crontab.go
```diff
@@ -0,0 +1,180 @@
...
+
+	// Configure a worker for the internal maintenance workflow.
+	w := temporalclient.NewWorker(ct.logger, ct.temporal.Temporal(), taskQueueName, ct.cfg.Worker)
+	if w == nil {
+		return nil
+	}
+
+	w.RegisterWorkflowWithOptions(ct.workflow, workflow.RegisterOptions{Name: workflowName})
+
+	w.RegisterWorkflow(ct.renewGoogleCalendarEventWatchesWorkflow)
```
itayd: all these should be registered on from outside this component

internal/backend/crontab/crontab.go
```diff
@@ -0,0 +1,180 @@
...
+	Activity temporalclient.ActivityConfig `koanf:"activity"`
+}
+
+var (
+	Configs = configset.Set[Config]{
+		Default: &Config{},
+	}
+
+	scheduleSpec = client.ScheduleSpec{
+		Intervals: []client.ScheduleIntervalSpec{{Every: 8 * time.Hour}},
```
itayd: i think these should vary between handlers. we just need an abstraction on top of temporal, or something else.

daabr: The choice of a single 8-hour interval was very intentional for the tasks we have. I did consider multiple schedules, and making them configurable, but for now it would be bad for 2 reasons: (1) it's an early optimization (2) unjustified extra complexity and misconfiguration risk.

In the future we may need multiple schedules (maybe an hourly schedule for "micro-tasks"? Daily/weekly/monthly data stuff?), but not now or probably not soon.

Until then, we have to decrease abstraction and complexity, not increase it.

internal/backend/crontab/crontab.go
daabr: That was my initial intuition too, but...

`Start` is called when the server starts. Creating a plugin API that works before/during `Start` across FX would require too much work with no concrete value.

Also, having a Temporal worker with plugin workflows/activities can be a bad idea if there's isn't a clear need and a clear design - both are beyond the scope of this PR.

I suggest we wait with this until there's a clear need which is worth the required design time and the extra complexity.

internal/backend/crontab/crontab.go
daabr: I could be wrong, but based on my understanding of Temporal selectors, these are the things that our function needs or does, which `Select` does not provide:
- Wait for all futures, not just the first one that's ready
- Return an error from any of the futures, if there is one

internal/backend/crontab/crontab.go
efiShtain: One problem here is that you have an internal package that depends on something outside of internal
I think its problematic

internal/backend/crontab/crontab.go
```diff
@@ -0,0 +1,180 @@
...
+	"go.temporal.io/sdk/workflow"
+	"go.uber.org/zap"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/configset"
+	"go.autokitteh.dev/autokitteh/internal/backend/temporalclient"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+)
+
+const (
+	taskQueueName = "internal-scheduler-task-queue"
```
efiShtain: Schedules are by queue and not by namespace, so we need some prefix to the queue name which is configurable from outside so we can have multiple schedules over the same temporal cloud but for different environments
mainly for Paid customers with dedicated servers

internal/backend/crontab/crontab.go
efiShtain: I also think itay is correct in this case, each integration should register its own workflow to this schedule, or just register a simple callback function, it doesn't need to know that it runs in a workflow at all

internal/backend/crontab/crontab.go
efiShtain: https://community.temporal.io/t/what-is-the-best-way-to-wait-for-child-workflows-in-parent-workflow-in-golang/3247/4

seems @daabr implementation is what maxim recommended as well

internal/backend/crontab/google.go
```diff
@@ -0,0 +1,25 @@
...
+
+import (
+	"context"
+
+	"go.temporal.io/sdk/workflow"
+
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func (ct *Crontab) renewGoogleCalendarEventWatchesWorkflow(wctx workflow.Context) error {
```
efiShtain: I think you can have a list of registered callbacks 
and have one workflow implementation that just runs the callback

this would suppose integrations but any other maintenance task we might need in the future with basically 0 effort

internal/backend/crontab/jira.go
```diff
@@ -0,0 +1,118 @@
+package crontab
```
efiShtain: I prefere schedules or tasks or jobs, something more high level than crontab
but that's me

internal/backend/crontab/jira.go
```diff
@@ -0,0 +1,118 @@
...
+	"go.autokitteh.dev/autokitteh/internal/backend/temporalclient"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func (ct *Crontab) renewJiraEventWatchesWorkflow(wctx workflow.Context) error {
+	// Enumerate all Jira connections (there's no single connection var value
+	// that we're looking for, so we can't use "ct.vars.FindConnectionIDs").
+	ctx := authcontext.SetAuthnSystemUser(context.Background())
+	cs, err := ct.connections.List(ctx, sdkservices.ListConnectionsFilter{IntegrationID: jira.IntegrationID})
```
efiShtain: this might get out of hand pretty quickly if we have 1000 (or probably more) Jira connections and you'd try to renew all of them every 8 hours

if its valid for 2 weeks, no much sense in renewing the connections every 8 hours

also, there would probably be connections that were just created and are valid

I think we should extend connections object with valid_until or expire_at field, indicating when it is no more valid. then you can fetch only the relevant connections.


jira API has a retry-after header response if we reach rate limit, we should sleep the workflow for that time before retrying

internal/backend/crontab/jira.go
```diff
@@ -0,0 +1,118 @@
...
+	if err != nil {
+		ct.logger.Error("Failed to list Jira connections for event watch renewal", zap.Error(err))
+		return err
+	}
+
+	// Check each Jira connection, renew its event watch if needed, abort on the first error.
+	for _, c := range cs {
```
efiShtain: why abort on the first error?
let all of them update
better to have 99 happy users and 1 angry one than 100 angry ones assuming the first one fail for example

internal/backend/crontab/jira.go
```diff
@@ -0,0 +1,118 @@
...
+	twoWeeksFromNow := time.Now().UTC().AddDate(0, 0, 14)
+	expiration, err := time.Parse(time.RFC3339, vs.GetValue(jira.WebhookExpiration))
+	if err == nil && expiration.UTC().After(twoWeeksFromNow) {
+		// The event watch is still valid for at least 2 weeks,
+		// so there's no need to renew it.
+		return nil
+	}
+
+	// Load the Jira OAuth configuration, to get a fresh OAUth access token.
+	cfg, _, err := ct.oauth.Get(ctx, "jira")
```
efiShtain: we can probably do that once for all ?

efiShtain (commented):
tests ?
basic ones I guess

integrations/atlassian/jira/oauth.go
```diff
@@ -46,15 +48,15 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	url, err := apiBaseURL()
+	u, err := APIBaseURL()
```
tebeka, line 51: I sometimes go the other way to avoid name collision:

```go
import urllib "net/url"
// ...
url, err := urllib.JoinPath(url, "/ex/jira", res[0].ID)
```
```

File: integrations/atlassian/jira/oauth.go [hunk 11]
```diff
@@ -74,31 +76,37 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
 
 	// Register a new webhook to receive, parse, and dispatch
 	// Jira events, or extend the deadline of an existing one.
-	url += "/ex/jira/" + res[0].ID
+	u, err = url.JoinPath(u, "/ex/jira", res[0].ID)
```
tebeka, line 84: I don't see the code checking the res length is bigger than 0.

internal/backend/crontab/crontab.go
tebeka: Instead of adding a config option, we can use the project/user name as prefix.

internal/backend/crontab/crontab.go
```diff
@@ -0,0 +1,180 @@
...
+	cfg         *Config
+	logger      *zap.Logger
+	temporal    temporalclient.Client
+	connections sdkservices.Connections
+	vars        sdkservices.Vars
+	oauth       sdkservices.OAuth
+}
+
+func New(c *Config, l *zap.Logger, t temporalclient.Client) *Crontab {
+	return &Crontab{cfg: c, logger: l, temporal: t}
```
tebeka: Please validate config.

internal/backend/crontab/crontab.go
tebeka: I think that for first run, before we have any experience with how it goes - it's simpler to have a central location for all the workflows.

internal/backend/crontab/google.go
tebeka: I'd remove all of the non implemented ones for now and leave just the JIRA.

internal/backend/crontab/jira.go
tebeka: I agree that in the future we might have an overload, but currently it's OK for me as first go.

internal/backend/crontab/jira.go
tebeka: Agree, use `errors.Join` and return it after updating all of them.

integrations/atlassian/jira/oauth.go
daabr: Ack, but this is more idiomatic IMO: `urllib` is a Python name which is less recognizable in Go, and acronym var names are a style convention when they're obvious (`u = url.xxx()`).

integrations/atlassian/jira/oauth.go
daabr: You're right. Technically, the output of `accessibleResources` is guaranteed to have at least one result if there's no error, but it's good to verify it. Done.

internal/backend/crontab/crontab.go
daabr: I prefer to handle this in a separate PR, after we understand better how these tasks look like (except for Jira and Google) and how registration should/n't be modeled.

> you have an internal package that depends on something outside of internal
> I think its problematic

@efiShtain : packages under an `internal` directory are invisible to (i.e. can't be imported by) non-internal packages, but the other direction is fine. For example: internal packages use non-internal protos and SDK functions. There's no difference between that and public functions in the Jira integration.

But in a more general way, I see this comment as another way of saying that the concrete implementation of tasks may belong outside of the task scheduling mechanism. And I agree with this intuition, but I still think it's too early right now to do it.

internal/backend/crontab/crontab.go
daabr: @tebeka please clarify: validate what in `Config`? It just contains optional `temporalclient` structs.

internal/backend/crontab/google.go
daabr: @efiShtain making sure I understand your idea: you mean programmatic registration of callbacks or just a single place where you add hard-coded calls to concrete implementations elsewhere in the codebase?

FYI, programmatic registration would mean that the workflows are non-deterministic, and may require workflow versioning, which we don't use yet. OTOH, hard-coded callbacks would require more FX dependencies.

Either way, I'm still not sure what would be a **correct** signature for these callbacks: workflows or activities? which inputs that aren't too specific or too generic? how to avoid more FX dependencies for different checks, etc.

@tebeka ack, but this is going to happen right after this PR is merged, so it will save wasteful questions and comments on how/where to add these functions. In other words, these are not generic placeholders, they're a side-effect of splitting this to mini-tasks.

internal/backend/crontab/jira.go
daabr: I get your point. I'd prefer "scheduler" but we already have that (for project schedules), and I didn't find a reasonable way to add this code there, even though there are similarities. I think "tasks" and "jobs" are **too** generic to be clear names, but I'm OK with settling this terminology in the future.

internal/backend/crontab/jira.go
daabr: @efiShtain actually not exactly...

1. For now, the workflow spawns activities for all Jira connections, but if the connection doesn't have a watch, or the watch is still less than 2 weeks old, the activity ends immediately as a no-op
2. Even if we have 10K connections to actually refresh together, it probably won't take 8 hours
3. Even if we have 1M connections to actually refresh, and only a single server/worker, the current implementation will do whatever it can in each 8h interval, and the next one will start again **but** treat the previous successful refreshes as no-ops and skip to the remaining connections (see the last paragraph in the doc header in `crontab.go`).

A different approach would be to do the initial connection var checks in the workflow before spawning an activity - but these checks can technically fail, that's why I put them in activities (for retries).

We may decide in the future that reading connection vars from our DB is not considered a risk worth of retries, or that it's cost-efficient to reduce the number of Temporal activities - but this may change the whole mechanism considerably, and I don't expect it to be an urgent concern for a long time.

internal/backend/crontab/jira.go
daabr: Unfortunately no, if I understand your comment correctly.

Even if all connections use the same Jira app, i.e. same OAuth config in line 76, they may belong to different Jira domains (e.g. `autokitteh.atlassian.net`, `blah.atlassian.net`) and each domain obviously has different tokens (line 82), otherwise different companies could read each other's data just because they use the same Jira app.

Also, soon we will support custom OAuth Jira apps, so even the OAuth config in line 76 will be based on connection vars (see INT-48).

internal/backend/crontab/crontab.go
daabr: Not sure I understand...

The schedule is supposed to be a singleton per cloud instance / self-hosted server, i.e. a single schedule per Temporal namespace.

The task queue is also a singleton per cloud instance / self-hosted server, i.e. per namespace, so if you have a cloud with multiple servers (e.g. our prod) all of them will be able to handle the schedule's workflows and activities alike (because these maintenance tasks refers to their shared DB and to 3rd-party services, nothing server-specific).

In other words, wether or not the environment has a single or multiple tenants, and regardless of the number of servers in the environment, there should be a single schedule and a single TQ for its workflows.

internal/backend/crontab/jira.go
daabr: Done

internal/backend/crontab/jira.go
daabr: @efiShtain a couple of clarifications:

1. We **do** store the watch's expiration in a conn var, but we need to read this conn var, and we do it inside the activity - see my first response above about why, and the alternatives

2. Jira's 429 response: you're right, I forgot to handle this, but my approach is to end the activity that got this response (without returning an error) - so the workflow result will show only real errors, its total runtime won't be affected by the number of connections above the rate-limit, and the refreshing would still work in a staggered way every 8 hours (which is why I'm refreshing every 2 weeks instead of at the last moment of 3 weeks + 5 days). This approach will also cause the load to be spread evenly across the day and across the month, over time. Added INT-186.

internal/backend/crontab/crontab.go
itayd: ok

internal/backend/crontab/crontab.go
itayd: my problem with this approach is that now there is a coupling between "crontab" and integrations. but if we put a nice big TODO(ticket) on it for now I'm fine.

itayd: i'm not sure i understand how you're not failing on the internal here as the internal is specific to backend, and integrations are outside backend?

internal/backend/crontab/crontab.go
itayd: same as above, a nice big red colored TODO(ticket) is fine

internal/backend/crontab/jira.go
itayd: we already got scheduler. i'd just rename it to `cron` instead of crontab. crontab is based on a textual table.

internal/backend/crontab/jira.go
```diff
@@ -0,0 +1,120 @@
...
+	"go.autokitteh.dev/autokitteh/internal/backend/temporalclient"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func (ct *Crontab) renewJiraEventWatchesWorkflow(wctx workflow.Context) error {
+	// Enumerate all Jira connections (there's no single connection var value
+	// that we're looking for, so we can't use "ct.vars.FindConnectionIDs").
+	ctx := authcontext.SetAuthnSystemUser(context.Background())
+	cs, err := ct.connections.List(ctx, sdkservices.ListConnectionsFilter{IntegrationID: jira.IntegrationID})
```
itayd: this must run in an activity

internal/backend/crontab/jira.go
```diff
@@ -0,0 +1,120 @@
...
+		err := workflow.ExecuteActivity(actx, ct.renewJiraEventWatchActivity, c.ID()).Get(wctx, nil)
+		if err != nil {
+			errs = append(errs, err)
+		}
+	}
+
+	return errors.Join(errs...)
+}
+
+func (ct *Crontab) renewJiraEventWatchActivity(ctx context.Context, cid sdktypes.ConnectionID) error {
```
itayd: this needs to be in the integration

internal/backend/crontab/crontab.go
daabr: Added TODOs (INT-190) in `jira.go` and `google.com`

internal/backend/crontab/crontab.go
daabr: Added TODOs (INT-190) in `jira.go` and `google.com`

internal/backend/crontab/jira.go
daabr: I'll rename to cron before I merge, after approvals, to minimize chaos

internal/backend/crontab/jira.go
daabr: done

internal/backend/crontab/jira.go
daabr: INT-190

internal/backend/crontab/crontab.go
efiShtain: I'm fine with that

internal/backend/crontab/crontab.go
efiShtain: you are correct, I understood it wrong and this is fine
my bad

internal/backend/crontab/jira.go
efiShtain: you need to think about fetching 10 million rows from db just to ignore them if they are irrelevant
not only on jira's side

I understand we store the watch in the vars, but I suggest it should be a field on the connection instead (I don't like the vars over use anyway, but that's a different topic)

btw if you do it in one workflow, you have temporal limit on the workflow itself, you can't have more then 50K events in history, and each activity requires at least 2 events I think (start / finish, maybe more)
so you'd have to manage that as well

regarding rate limit from Jira, if we will spam them they will block us (I guess, I would if I was them) and we can't just assume we can retry forever on everything



All in all, I agree with @tebeka , this is fine for first go, but I think we should handle it in the near future

internal/backend/crontab/jira.go
efiShtain: ok

internal/backend/crontab/google.go
efiShtain: @daabr 
I mean just let the integrations subscribe on a "Tick" event. you should just schedule it. let them decide what to do when tick is called.

you can pass the cron service to the integrations using fx, they will subscribe to tick/schedule/on-run/what ever event. no need to pass any parameters. schedule doesn't know it content, only that it needs to start and call the registered callbacks.

regarding ordering, its not a must across different runs of the workflow, so the only problem that might occur is if we have a new integration added and a cron workflow is trying to reply and now the integrations list is different - I think this is an edge case which will resolve it self on the next schedule.

daabr: Subscribing to frequent and arbitrary "tick" events that anyone can attach themselves to is exactly the danger I'm trying to avoid. Also, waking up functions only for them to decide not to run yet is the opposite of scheduling.

But that's a longer discussion beyond the scope of this PR. I'll follow-up on it when we add more tasks that are not related to integrations.

internal/backend/crontab/jira.go
daabr: > you need to think about fetching 10 million rows from db just to ignore them if they are irrelevant not only on jira's side

As of the last commit, we load only 2 vars per connection in order to check if we need to renew that connection (watch ID + expiration), that's hardly a bottleneck/scalability problem.

> I suggest it should be a field on the connection instead (I don't like the vars over use anyway, but that's a different topic)

I understand your point, but watches are implemented in entirely different ways by different integrations, and most integrations don't have watches at all. But we can discuss this together offline.

> you have temporal limit on the workflow itself, you can't have more then 50K events in history, and each activity requires at least 2 events I think (start / finish, maybe more)

Ack, but we won't get to this point (refreshing ~20K Jira connections at the same time) until we have millions of users, as opposed to Jira rate-limiting which isn't hard to reach with only 10's of connections.

> regarding rate limit from Jira, if we will spam them they will block us (I guess, I would if I was them) and we can't just assume we can retry forever on everything

I agree, but my strategy is the opposite of spamming them, it's smoothing peaks by backing off for hours instead of seconds/minutes. Also, we can probably do batch requests instead - see INT-186 and INT-189.

itayd (approved):

PR #992, opened 2025-01-20
 use hard delete and detach foreign keys for sessions and events only

efiShtain:
this is a simplified version of this PR https://github.com/autokitteh/autokitteh/pull/965

it handles only sessions (and its dependents) and events

deleting trigger / connection does not delete events but remove the foreign key from events on trigger/connection

deleting event does not delete sessions but remove the foreign key from session on events

deleting session delete cascade the call attempts, log records and call specs


tebeka (approved):

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -170,10 +170,8 @@ type Event struct {
 	Seq       uint64 `gorm:"primaryKey;autoIncrement:true,index:idx_event_type_seq,priority:2"`
 
 	// enforce foreign keys
-	Connection *Connection
-	Trigger    *Trigger
-
-	DeletedAt gorm.DeletedAt `gorm:"index"`
+	Connection *Connection `gorm:"constraint:OnDelete:SET NULL"`
```
itayd, line 173: ahhh so gorm does the null setting automatically, interesting.

itayd (approved):

internal/backend/db/dbgorm/scheme/records.go
efiShtain: its not gorm, its postgres, but yes

itayd: so what happens to sqlite?

efiShtain: should work the same, it's basic sql (not only postgres)

PR #993, opened 2025-01-20
chore(infra): Bump Go to 1.23.5

tebeka:


itayd:
no issues with this, but q - should we go and bump every time? or should we try to stay on the lowest possible for compatibility?

itayd (approved):

daabr:
> no issues with this, but q - should we go and bump every time? or should we try to stay on the lowest possible for compatibility?

This update changes the patch number, which is guaranteed to be backward-compatible with all 1.23.x versions, so yeah.

As for minor version changes - I'm also in favor of that, but after a review of the cost/benefit, e.g. the update from 1.22 to 1.23 was important for us.

daabr (approved):

PR #996, opened 2025-01-21
fix(pythonrt): better wrapping of unhandled type

itayd:
- moved `safe_wrap` into `values.py`.
- now tries its best to wrap whatever it can, and what it cannot it wraps into an "unhandled" struct.

motivation: show python requests HTTP response objects.

### DIFF
--- a/runtimes/pythonrt/runner/values.py
+++ b/runtimes/pythonrt/runner/values.py
@@ -25,7 +25,7 @@ def wrap_unhandled(v: Any) -> pb.Value:
 
 
 def safe_wrap(v):
-    """Same as wrap, but does raise TypeError if a type is not supported.
+    """Same as wrap, but does not raise TypeError if a type is not supported.
 
     Instead, it returns a struct with the type and the string representation of the value.
     """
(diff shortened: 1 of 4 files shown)


runtimes/pythonrt/runner/values.py
```diff
@@ -12,6 +12,27 @@
...
+            fields={
+                "repr": wrap(repr(v)),
+                "type": wrap(str(type(v))),
+            },
+        )
+    )
+
+
+def safe_wrap(v):
+    """Same as wrap, but does raise TypeError if a type is not supported.
```
efiShtain: i guess you mean doesnt?

efiShtain (approved):
just fix the comment
LGTM

runtimes/pythonrt/runner/values.py
itayd: indeed

itayd:
```suggestion
    """Same as wrap, but does not raise TypeError if a type is not supported.
```

PR #997, opened 2025-01-21
fix(users): do not check status if not in field_mask

itayd:


RonenMars (approved):

configs/opa_bundles/default/authz.rego
```diff
@@ -60,7 +60,6 @@ allow if {
 	input.action.name == "update"
 	input.authn_user.id == input.subject.id
 	not "status" in input.data.field_mask
```
efiShtain, line 62: does this require that stats is not in the field mask ?

itayd: yup

PR #998, opened 2025-01-21
fix(pythonrt): values: deal with recursive values

itayd:


daabr (approved):

runtimes/pythonrt/runner/values.py
```diff
@@ -44,6 +44,25 @@ def wrap(v: Any, unhandled: Callable[[Any], pb.Value] = None) -> pb.Value:
     Namedtuples are wrapped as structs.
     """
 
+    history = history or []
```
tebeka, line 47: This is a bug. If you pass empty `history` to `wrap` this will create a new one. Use
```python
history = [] if history is None else history
```

PR #1001, opened 2025-01-22
feat(pythonrt): render request as a value

itayd:
only for display purposes

### DIFF
--- a/runtimes/pythonrt/runner/values.py
+++ b/runtimes/pythonrt/runner/values.py
@@ -7,6 +7,8 @@ from collections import namedtuple
 from datetime import UTC, datetime, timedelta
 from typing import Any, Callable
 
+import requests
+
 import pb.autokitteh.values.v1.values_pb2 as pb
 from google.protobuf.duration_pb2 import Duration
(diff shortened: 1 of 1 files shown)


runtimes/pythonrt/runner/values.py
```diff
@@ -106,6 +108,31 @@ def dive(vv):
         d.FromTimedelta(v)
         return pb.Value(duration=pb.Duration(v=d))
 
+    if isinstance(v, requests.Response):
+        json = text = None
+
+        if len(v.content) > 100 * 1024:
```
tebeka:
```suggestion
        max_size = 100 * 1024  # 100K
        if len(v.content) > max_size:
```

runtimes/pythonrt/runner/values.py
```diff
@@ -106,6 +108,31 @@ def dive(vv):
...
+            json = text = "<Response content too large to be included>"
```
tebeka:
```suggestion
            json = "<Response content too large to be included>"
            text = dive(v.text[:max_size] + "...")
```

tebeka (approved):
Two minor comments.

PR #1003, opened 2025-01-23
fix(pythonrt): wrap response object with invalid json now works

itayd:


itayd:
going to merge this in as it breaks main, @tebeka lmk if any issues

PR #1006, opened 2025-01-24
DRAFT: feat(events): next_event on child sessions

itayd:
this should probably be separated into two different PRs, but here for initial comments on the approach:
- when a child session is created using `start`, it sends a signal to the parent session when it finishes. that signal can be received using `next_event`. no subscription is required.
- `join` to supplement `next_event` to allow to wait for _all_ subscriptions to yield a result.

TODO:
- [ ] Schema is problematic because of cyclic dependencies. Need to figure out how to solve it.
- [ ] Probably can DRY up some code.
- [ ] Separate into two PRs: one for for the session signaling and one for the join.


internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -172,6 +173,7 @@ type Event struct {
 	// enforce foreign keys
 	Connection *Connection
 	Trigger    *Trigger
+	// Session    *Session
```
efiShtain, line 176: This removes the forein key, why ?

and why the event needs to know about the session ?
session has a triggering event
event might trigger multiple sessions
I think this is a problematic model

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -344,7 +341,7 @@ type Session struct {
 	Build      *Build
 	Deployment *Deployment
 	Project    *Project
-	Event      *Event `gorm:"references:EventID"`
+	// Event      *Event `gorm:"references:EventID"`
```
efiShtain, line 344: Session needs to reference the event 
or just hold the event content if we want and don't reference it anymore

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -483,25 +480,21 @@ type Signal struct {
...
-	switch {
-	case r.ConnectionID != nil:
-		dstID = sdktypes.NewEventDestinationID(sdktypes.NewIDFromUUIDPtr[sdktypes.ConnectionID](r.ConnectionID))
-	case r.TriggerID != nil:
-		dstID = sdktypes.NewEventDestinationID(sdktypes.NewIDFromUUIDPtr[sdktypes.TriggerID](r.TriggerID))
-	default:
-		return nil, sdkerrors.NewInvalidArgumentError("signal must have a connection or trigger")
+	dstID, err := parseEventDestinationID(r.ConnectionID, r.TriggerID, r.SessionID)
+	if err != nil {
+		return nil, fmt.Errorf("destination id: %w", err)
```
efiShtain, line 497:
```suggestion
		return nil, fmt.Errorf("parse destination id: %w", err)
```

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -629,3 +622,16 @@ type Ownership struct {
...
 	UserID string `gorm:"not null"`
 }
+
+func parseEventDestinationID(con, trg, ses *uuid.UUID) (sdktypes.EventDestinationID, error) {
+	switch {
+	case con != nil:
+		return sdktypes.NewEventDestinationID(sdktypes.NewIDFromUUIDPtr[sdktypes.ConnectionID](con)), nil
+	case trg != nil:
+		return sdktypes.NewEventDestinationID(sdktypes.NewIDFromUUIDPtr[sdktypes.TriggerID](trg)), nil
+	case ses != nil:
```
efiShtain, line 632: it makes no sesnse logic wise that an event has a destination which is a session

on a higher level, events has no destinations, at most they have subscribers. the event shouldn't know where it is going, not should it care. it can go no where or it can go everywhere.

internal/backend/sessions/sessionworkflows/activities.go
```diff
@@ -110,6 +122,54 @@ func (ws *workflows) deactivateDrainedDeploymentActivity(ctx context.Context, di
...
+		return temporalclient.TranslateError(err, "get session log for %v", session.ID())
+	}
+
+	rs := log.Log.Records()
+
+	if len(rs) == 0 {
+		return temporalclient.TranslateError(sdkerrors.ErrUnretryableUnknown, "no records")
+	}
+
+	s := rs[0].GetState()
```
efiShtain, line 144: do you really need to fetch all session log records (which might be hugh object) just to check if the first one is not in final state?

internal/backend/sessions/sessionworkflows/activities.go
```diff
@@ -110,6 +122,54 @@ func (ws *workflows) deactivateDrainedDeploymentActivity(ctx context.Context, di
 	return nil
 }
 
+func (ws *workflows) signalSessionActivity(ctx context.Context, session sdktypes.Session) error {
```
efiShtain, line 125: not sure i understand what you are trying to do with this function, the name is not clear

internal/backend/sessions/sessionworkflows/module.go
```diff
@@ -26,6 +26,7 @@ func (w *sessionWorkflow) newModule() sdkexecutor.Executor {
 		sdkmodule.ExportFunction("subscribe", w.subscribe, flags),
 		sdkmodule.ExportFunction("unsubscribe", w.unsubscribe, flags),
 		sdkmodule.ExportFunction("next_event", w.nextEvent, flags),
+		sdkmodule.ExportFunction("join", w.join, flags),
```
efiShtain, line 29: do we need a new special word ?
can't we just wait for all sessions in a loop ?
or maybe extend next_event with wait all events ?

internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -211,6 +228,99 @@ func (w *sessionWorkflow) nextEvent(ctx context.Context, args []sdktypes.Value,
...
+			if event == nil {
+				continue
+			}
+
+			if received[signalID.String()], err = sdktypes.WrapValue(event); err != nil {
+				return sdktypes.InvalidValue, err
+			}
+
+			delete(signals, signalID)
+			continue L
```
efiShtain, line 290: do we really needs this labels usage here ?

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -298,12 +298,18 @@ func (w *sessionWorkflow) createEventSubscription(wctx workflow.Context, filter
...
 	}
 
-	// generate a unique signal id.
 	var signalID uuid.UUID
-	if err := workflow.SideEffect(wctx, func(wctx workflow.Context) any {
-		return uuid.New()
-	}).Get(&signalID); err != nil {
-		return uuid.Nil, fmt.Errorf("generate signal ID: %w", err)
+
+	if did.IsSessionID() {
```
efiShtain, line 303: did is really hard to reason about and understand without looking for the definition in the code

```suggestion
	if destinationID.IsSessionID() {
```

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -267,7 +276,23 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
-	_ = workflow.ExecuteActivity(wctx, deactivateDrainedDeploymentActivityName, session.DeploymentID()).Get(wctx, nil)
+	if err := workflow.ExecuteActivity(
+		dwctx,
+		deactivateDrainedDeploymentActivityName,
+		session.DeploymentID(),
+	).Get(wctx, nil); err != nil {
+		l.Sugar().With("err", err).Errorf("deactivate drained deployment: %v", err)
+	}
+
+	if pid := session.ParentSessionID(); pid.IsValid() {
```
efiShtain, line 287: what's the purpose to this code ?

efiShtain (commented):
not sure I get the idea of the PR

changing the event to hold a reference to a sessions sounds off to me
adding a join keyword instead of reusing next_event in loop makes the system more complicated

in the context of limits, need to figure out how to limit the number of child sessions we allow and the rate of spawning them 

why are we even allowing child sessions right now? 
is it to allow parallelism ? if that's the case, let's support concurrent activities instead

configs/opa_bundles/default/authz.rego
```diff
@@ -316,6 +316,7 @@ allow if {
 	input.subject.kind == "ses"
 	input.action.name == "start"
 	is_active_member_of_single_assosicated_org_id
+	not input.data.session.id
```
tebeka, line 319: How is this change related to auth?

internal/backend/db/dbgorm/scheme/records.go
tebeka, line 176: Delete commented out code

internal/backend/db/dbgorm/scheme/records.go
tebeka, line 344: Delete commented code

internal/backend/db/dbgorm/scheme/records.go
tebeka: Add the signal ID as well to the error

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -483,25 +480,21 @@ type Signal struct {
 	DestinationID uuid.UUID  `gorm:"index;type:uuid;not null"`
 	ConnectionID  *uuid.UUID `gorm:"type:uuid"`
 	TriggerID     *uuid.UUID `gorm:"type:uuid"`
+	SessionID     *uuid.UUID `gorm:"type:uuid"`
```
tebeka, line 483: Why are these pointers? The uuid package has a `Nil` one.

internal/backend/sessions/sessions.go
```diff
@@ -45,7 +46,7 @@ func New(l *zap.Logger, config *Config, db db.DB, svcs sessionsvcs.Svcs, telemet
 	}
 }
 
-func (s *sessions) StartWorkers(ctx context.Context) error {
+func (s *sessions) StartWorkers(ctx context.Context, dispatch sdkservices.DispatchFunc) error {
```
tebeka, line 49: Why do we need to pass `dispatch`? Do we have more than one option?

internal/backend/sessions/sessionworkflows/activities.go
tebeka: He sets `PageSize` to 1

internal/backend/sessions/sessionworkflows/module.go
tebeka: From what I understand `join` is like `WaitGroup.Wait`. You don't get an event, only wait for events to be done.

internal/backend/sessions/sessionworkflows/syscalls.go
tebeka: We don't.

internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -211,6 +228,99 @@ func (w *sessionWorkflow) nextEvent(ctx context.Context, args []sdktypes.Value,
...
+
+			if event == nil {
+				continue
+			}
+
+			if received[signalID.String()], err = sdktypes.WrapValue(event); err != nil {
+				return sdktypes.InvalidValue, err
+			}
+
+			delete(signals, signalID)
```
tebeka, line 289: You're changing the map you are iterating over (see https://go.dev/ref/spec#For_range)
Either
```
sigs := slices.Collect(maps.Keys(signals))
```

And then iterate over signals. Or create a `processed` map or something.

internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -211,6 +228,99 @@ func (w *sessionWorkflow) nextEvent(ctx context.Context, args []sdktypes.Value,
...
+				break L
+			}
+
+			// get next event on this signal
+			event, err := w.getNextEvent(ctx, signalID)
+			if err != nil {
+				return sdktypes.InvalidValue, err
+			}
+
+			if received[signalID.String()], err = sdktypes.WrapValue(event); err != nil {
```
tebeka, line 311: This is confusing, I prefer:
```suggestion

			wrapped, err := sdktypes.WrapValue(event)
			if err != nil { .... }
			received[signalID.String()] = wrapped
```

internal/backend/sessions/sessionworkflows/syscalls.go
```diff
@@ -211,6 +228,99 @@ func (w *sessionWorkflow) nextEvent(ctx context.Context, args []sdktypes.Value,
...
+				return sdktypes.InvalidValue, err
+			}
+
+			delete(signals, signalID)
+
+			break
+		}
+	}
+
+	return sdktypes.NewDictValueFromStringMap(received), nil
```
tebeka, line 321: So many type conversions. We need to rethink the design.

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -220,8 +223,14 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
...
 	defer done()
 
-	sessionDurationHistogram.Record(metricsCtx, duration.Milliseconds(),
-		metric.WithAttributes(attribute.Bool("replay", isReplaying), attribute.Bool("success", err == nil)))
+	sessionDurationHistogram.Record(
+		metricsCtx,
+		duration.Milliseconds(),
+		metric.WithAttributes(
+			attribute.Bool("replay", isReplaying),
+			attribute.Bool("success", err == nil),
```
tebeka, line 231:
```suggestion
			attribute.String("error", fmt.Sprintf("%v", err)),
```

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -254,7 +263,7 @@ func (ws *workflows) sessionWorkflow(wctx workflow.Context, params *sessionWorkf
 				// No need to indicate the workflow as errored.
 				err = nil
 			} else {
-				l.Sugar().Errorf("session workflow error: %v", err)
+				l.Sugar().With("err", err).Errorf("session workflow error: %v", err)
```
tebeka, line 266: Why have err both as attribute and in the message?

runtimes/pythonrt/runner/syscalls.py
```diff
@@ -123,6 +123,38 @@ def ak_next_event(self, args, kw):
 
         return AttrDict(data) if isinstance(data, dict) else data
 
+    def ak_join(self, args, kw):
```
tebeka, line 126: This is old code, the new syscalls have regular Python signature.
```suggestion
    def ak_join(self,signal, *signals):
```

runtimes/pythonrt/runner/syscalls.py
```diff
@@ -123,6 +123,38 @@ def ak_next_event(self, args, kw):
...
+        req = pb.JoinRequest(runner_id=self.runner_id, signal_ids=ids)
+        if timeout:
+            if isinstance(timeout, int | float):
+                req.timeout_ms = int(timeout * 1000)
+            elif isinstance(timeout, timedelta):
+                req.timeout_ms = int(timeout.total_seconds() * 1000)
+            else:
+                raise TypeError(
+                    f"timeout {timeout!r} should be a timedelta or number of seconds"
+                )
```
tebeka, line 147: There's duplicate code in `ak_next_event`, move to a function.

sdk/sdktypes/session_state.go
```diff
@@ -107,3 +109,26 @@ func (p SessionState) Concrete() concreteSessionState {
...
+	case SessionStateError:
+		fields["value"] = s.GetProgramError().Value()
+	case SessionStateRunning:
+	case SessionStateStopped:
+		fields["reason"] = NewStringValue(s.Reason())
+	default:
+		return InvalidValue, fmt.Errorf("%w: unknown state type", sdkerrors.ErrUnretryableUnknown)
+	}
+
+	fields["type"] = NewStringValue(p.Type().String())
```
tebeka, line 131: Is this the one filling the state? Do comment.

tebeka (changes requested):
Some code comments, but:

- We can implement `join` in the Python SDK without any code changes to `ak`
- I looks like this is the path to creating workflow DAG, which I don't like. It is complicated (See [here](https://docs.celeryq.dev/en/stable/userguide/canvas.html) and [here](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dags.html)). I prefer to wait for user request before moving on and have a good design document on what we support.

PR #1007, opened 2025-01-24
fix: high performance sqlite

itayd:
the instructions given in https://kerkour.com/sqlite-for-servers seem to be working. tested under load, not a single BUSY was seen.

this essentially separates db connections into write and read connections, limiting writes to a single connection. also add a bunch of pragmas. for non-sqlite, there is no separation.

get rid of the sqlite lock.

also get rid of some unused functions.

Ref: ENG-190, ENG-648

### DIFF
--- a/internal/backend/db/db.go
+++ b/internal/backend/db/db.go
@@ -4,7 +4,6 @@ import (
 	"context"
 
 	"github.com/google/uuid"
-	"go.uber.org/zap"
 	"gorm.io/gorm"
 
 	"go.autokitteh.dev/autokitteh/internal/backend/health/healthreporter"
@@ -12,20 +11,6 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
-
-type Transaction interface {
-	DB
-	Commit() error
-
-	// Does nothing if already committed.
-	Rollback() error
-}
-
-func LoggedRollback(z *zap.Logger, tx Transaction) {
-	if err := tx.Rollback(); err != nil {
-		z.Error("rollback error", zap.Error(err))
-	}
-}
 
 type DB interface {
 	Connect(context.Context) error
@@ -36,10 +21,7 @@ type DB interface {
 
 	healthreporter.HealthReporter
 
-	GormDB() *gorm.DB
-
-	// Begina a transaction.
-	Begin(context.Context) (Transaction, error)
+	GormDB() (r, w *gorm.DB)
 
 	Transaction(context.Context, func(tx DB) error) error
 
--- a/internal/backend/db/dbgorm/dbgorm.go
+++ b/internal/backend/db/dbgorm/dbgorm.go
@@ -15,7 +15,6 @@ import (
 
 	"go.autokitteh.dev/autokitteh/internal/backend/db"
 	"go.autokitteh.dev/autokitteh/internal/backend/gormkitteh"
-	"go.autokitteh.dev/autokitteh/internal/kittehs"
 	"go.autokitteh.dev/autokitteh/migrations"
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 )
@@ -26,7 +25,7 @@ type gormdb struct {
 	z   *zap.Logger
 	cfg *Config
 
-	wdb, rdb *gorm.DB
+	writer, reader *gorm.DB
 }
 
 var _ db.DB = (*gormdb)(nil)
@@ -44,7 +43,7 @@ func New(z *zap.Logger, cfg *Config) (db.DB, error) {
 	return &gormdb{z: z, cfg: cfg}, nil
 }
 
-func (db *gormdb) GormDB() (r, w *gorm.DB) { return db.rdb, db.wdb }
+func (db *gormdb) GormDB() (r, w *gorm.DB) { return db.reader, db.writer }
 
 func connect(_ context.Context, z *zap.Logger, cfg *Config) (r *gorm.DB, w *gorm.DB, err error) {
 	gormCfgFn := func(cfg *gorm.Config) { cfg.SkipDefaultTransaction = true }
@@ -67,8 +66,9 @@ func connect(_ context.Context, z *zap.Logger, cfg *Config) (r *gorm.DB, w *gorm
 
 	sqlDB.SetMaxOpenConns(n)
 
-	// in memory db doesn't need any separation between writing and reading.
-	if cfg.Type != "sqlite" || strings.HasPrefix(cfg.DSN, ":memory:") {
+	// For in memory sqlite in memory database we will use the same connection for reads and writes
+	// since otherwise it will have two distinct databases for these.
+	if cfg.Type == "sqlite" && strings.HasPrefix(cfg.DSN, ":memory:") {
 		w = r
 		return
 	}
@@ -92,7 +92,7 @@ func connect(_ context.Context, z *zap.Logger, cfg *Config) (r *gorm.DB, w *gorm
 }
 
 func (db *gormdb) Connect(ctx context.Context) (err error) {
-	db.rdb, db.wdb, err = connect(ctx, db.z.Named("gorm"), db.cfg)
+	db.reader, db.writer, err = connect(ctx, db.z.Named("gorm"), db.cfg)
 	return
 }
 
@@ -139,13 +139,13 @@ func foreignKeys(gormdb *gormdb, enable bool) error {
 	}
 	stmt := fkStmtByDB[gormdb.cfg.Type][enable]
 
-	if err := gormdb.rdb.Exec(stmt).Error; err != nil {
-		return fmt.Errorf("exec %q failed on rdb: %w", stmt, err)
+	if err := gormdb.reader.Exec(stmt).Error; err != nil {
+		return fmt.Errorf("read exec: %q: %w", stmt, err)
 	}
 
-	if gormdb.wdb != gormdb.rdb {
-		if err := gormdb.wdb.Exec(stmt).Error; err != nil {
-			return fmt.Errorf("exec %q failed on wdb: %w", stmt, err)
+	if gormdb.writer != gormdb.reader {
+		if err := gormdb.writer.Exec(stmt).Error; err != nil {
+			return fmt.Errorf("write exec: %q: %w", stmt, err)
 		}
 	}
 
@@ -169,10 +169,12 @@ func initGoose(client *sql.DB, dialect string) (ver int64, err error) {
 func (db *gormdb) Migrate(ctx context.Context) error {
 	db.z.Info("migrating")
 
-	_, client := db.client(true)
-
-	_, err := initGoose(client, db.cfg.Type)
+	_, client, err := db.client(true)
 	if err != nil {
+		return err
+	}
+
+	if _, err = initGoose(client, db.cfg.Type); err != nil {
 		return err
 	}
 
@@ -185,7 +187,11 @@ func (db *gormdb) Migrate(ctx context.Context) error {
 }
 
 func (db *gormdb) MigrationRequired(ctx context.Context) (bool, int64, error) {
-	_, client := db.client(false)
+	_, client, err := db.client(false)
+	if err != nil {
+		return false, 0, err
+	}
+
 	dbversion, err := initGoose(client, db.cfg.Type)
 	if err != nil {
 		return false, 0, err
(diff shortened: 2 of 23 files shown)


internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -50,51 +45,59 @@ func New(z *zap.Logger, cfg *Config) (db.DB, error) {
...
-	client, err := gormkitteh.OpenZ(z, cfg, func(cfg *gorm.Config) {
-		cfg.SkipDefaultTransaction = true
-	})
-	if err != nil {
-		return nil, fmt.Errorf("opendb: %w", err)
+	sqlDB.SetMaxOpenConns(n)
+
+	// in memory db doesn't need any separation between writing and reading.
+	if cfg.Type != "sqlite" || strings.HasPrefix(cfg.DSN, ":memory:") {
+		w = r
```
efiShtain, line 72: it can be helpful to have postgres with read and write separation

we already always have in production read replicas, but we use them as hot swap instead of actually reading from them (except for grafana and dashboards)

I suggest leaving this on and the configuration can provide the same instance for read and writes or different instances

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -50,51 +45,59 @@ func New(z *zap.Logger, cfg *Config) (db.DB, error) {
...
-func connect(_ context.Context, z *zap.Logger, cfg *Config) (*gorm.DB, error) {
-	client, err := gormkitteh.OpenZ(z, cfg, func(cfg *gorm.Config) {
-		cfg.SkipDefaultTransaction = true
-	})
-	if err != nil {
-		return nil, fmt.Errorf("opendb: %w", err)
+	sqlDB.SetMaxOpenConns(n)
+
+	// in memory db doesn't need any separation between writing and reading.
+	if cfg.Type != "sqlite" || strings.HasPrefix(cfg.DSN, ":memory:") {
```
efiShtain: if we're going to have lots of differences between sqlite and postgres, maybe its a good time to split the dbs to two modules ?

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -134,12 +137,23 @@ var fkStmtByDB = map[string]map[bool]string{
 	},
 }
 
-func foreignKeys(gormdb *gormdb, enable bool) {
+func foreignKeys(gormdb *gormdb, enable bool) error {
```
efiShtain, line 136: this is for example sqlite only

internal/backend/db/dbgorm/dbgorm.go
itayd: we don't have a lot of differences... and they mostly just around the foreign keys and r/w separation that you think we should separate anyway for postgres. that said I sympathise, but that's going to be a lot of work... maybe in the future.

internal/backend/db/dbgorm/dbgorm.go
itayd: i think that should be done in a separate pr. i guess postgres wont require only a single writer?

internal/backend/db/dbgorm/dbgorm.go
itayd: i wonder if we can get rid of it, I'll check.

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -27,15 +28,9 @@ type Config = gormkitteh.Config
...
-	db  *gorm.DB
 	cfg *Config
 
-	// See https://github.com/mattn/go-sqlite3/issues/274.
-	// Used only for protecting writes when using sqlite.
-	//
-	// TODO(ENG-190): This is not... ideal. Really find a better
-	//                solution. Everything else I tried didn't work.
-	mu *sync.Mutex
+	wdb, rdb *gorm.DB
```
daabr: Instances of this struct are already named "db", which makes it more sensible to rename these fields to "read" and "write" for 3 reasons:

1. Fix non-idiomatic naming: "gdb.rdb" is awkward
2. Accuracy: even though the type is "DB" it's actually a DB connection, calling them "read DB" and "write DB" is misleading
3. "read" and "write" are much easier to understand than "r" and "w" at this level, for casual code readers, and still very short ("r" and "w" inside other functions is still good)

internal/backend/db/dbgorm/dbgorm.go
daabr: @itayd I'm ok with not doing this now, like you wrote, but FYI "foreign keys and r/w separation" is "a lot of differences". These are literally different implementations that even have different abstraction layers, so the code is overly complicated as long as you keep them squished together.

There are much better design patterns to handle this a common API over multiple concrete engines. DB access is even a canonical example for that.

internal/backend/db/dbgorm/dbgorm.go
daabr: I agree with both of you. Intelligent usage in PG (read replicas, separation between [lots of] read-only connections and [a few] read/write connections) is critical. But also not urgent right now.

@itayd PG definitely supports multiple concurrent write connections, but writing to PG DBs at scale requires careful design. The cost of fixing badly-designed R/W modeling in PG (which causes terrible bottleneck/scale issues and DB ops risks) is very high down the road.

@efiShtain I strongly agree that treating connections as read-only vs. write-only/RW should be a configuration rather than forcing any assumption in the code about the architecture of using read replicas. For now this is mostly theoretical, but if/when we use real DB clients instead of ORM it will become critical.

internal/backend/db/dbgorm/dbgorm.go
daabr: @itayd in the meantime - document that this function is only relevant for SQLite, to help code readers and future maintenance. Ideally also explain briefly why.

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -134,12 +137,23 @@ var fkStmtByDB = map[string]map[bool]string{
...
-func foreignKeys(gormdb *gormdb, enable bool) {
+func foreignKeys(gormdb *gormdb, enable bool) error {
 	if _, found := fkStmtByDB[gormdb.cfg.Type]; !found {
 		panic(fmt.Errorf("unknown DB type: %s", gormdb.cfg.Type))
 	}
 	stmt := fkStmtByDB[gormdb.cfg.Type][enable]
-	gormdb.db.Exec(stmt)
+
+	if err := gormdb.rdb.Exec(stmt).Error; err != nil {
+		return fmt.Errorf("exec %q failed on rdb: %w", stmt, err)
```
daabr: Nit for both of the new error messages - these are more readable:

`"read exec error: %q: %w"`
`"write exec error: %q: %w"`

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -428,11 +450,20 @@ func delete[T any](db *gorm.DB, ctx context.Context, where string, args ...any)
 	return nil
 }
 
-func (db *gormdb) client(debug bool) *sql.DB {
-	q := db.db
+func (db *gormdb) client(debug bool) (r, w *sql.DB) {
```
daabr: It's wrong for this function to panic instead of return an error, so I'd prefer you fix this, but it's not critical since your PR only extends the return type.

internal/backend/db/dbgorm/tx.go
```diff
@@ -10,59 +10,32 @@ import (
...
-
-	return tx.db.Rollback().Error
+func (db *gormdb) Transaction(ctx context.Context, f func(db db.DB) error) error {
+	return db.transaction(ctx, func(tx *gormdb) error { return f(tx) })
 }
 
-func (db *gormdb) Transaction(ctx context.Context, f func(db db.DB) error) error {
-	return db.transaction(ctx, func(tx *tx) error {
-		return f(tx)
+func (db *gormdb) transaction(ctx context.Context, f func(tx *gormdb) error) error {
```
daabr: `transaction` and `rtransaction` are not acceptable names, they're practically indistinguishable, so it's very confusing.

internal/backend/db/dbgorm/tx.go
```diff
@@ -10,59 +10,32 @@ import (
...
-					ctx: ctx,
-				},
-			)
-		})
+func (db *gormdb) rtransaction(ctx context.Context, f func(tx *gormdb) error) error {
+	return db.rdb.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		return f(
+			&gormdb{
+				z:   db.z.With(zap.String("txid", uuid.NewString())),
+				wdb: nil, // panic on writes.
```
daabr: This is a footgun. A panic is not an appropriate result for mistakenly using a read-only connection to the DB. AK should return errors on such transaction, and the code using transactions should handle errors.

daabr (commented):
I like this PR! See all my comments, the only ones that stopped me from approving are in `internal/backend/db/dbgorm/tx.go`.

internal/backend/db/db.go
```diff
@@ -13,20 +12,6 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-type Transaction interface {
```
tebeka, line 16: Maybe it's worth looking at https://github.com/Thiht/transactor

internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -50,51 +45,59 @@ func New(z *zap.Logger, cfg *Config) (db.DB, error) {
 		return nil, err
 	}
 
-	db := &gormdb{z: z, cfg: cfg}
+	return &gormdb{z: z, cfg: cfg}, nil
+}
+
+func (db *gormdb) GormDB() (r, w *gorm.DB) { return db.rdb, db.wdb }
```
tebeka: Why named output variables?

internal/backend/db/dbgorm/dbgorm.go
tebeka: I prefer to have utility function as as `IsSQLite` or `DBType`. There's code in `Setup` that only checks `db.cfg.Type`.

internal/backend/db/dbgorm/dbgorm.go
efiShtain: @daabr not sure why orm vs straight talking to the db makes a difference here

we have a connection pool right now working with PG, so it won't be a single writer anyway, and of course it's not a requirement

we have a read replica we can use in production, why not use it ?

all I suggest is not treating sqlite as special, just always have read and write separation, 
if you pass the same connection string, it would open double the connections to the write instance, that's fine. we can manage that. Otherwise it will open connection pool to the write and a different pool to the read. which is also fine. 
its total gain with zero lost

internal/backend/db/dbgorm/dbgorm.go
efiShtain: I'm fine with doing it later
resolving

internal/backend/db/dbgorm/dbgorm.go
itayd: will check if the condition can be removed

internal/backend/db/dbgorm/dbgorm.go
itayd: `reader` and `writer`

internal/backend/db/dbgorm/dbgorm.go
itayd: to make it clear which is writer and which is reader

internal/backend/db/dbgorm/dbgorm.go
itayd: done

internal/backend/db/dbgorm/dbgorm.go
itayd: done

internal/backend/db/dbgorm/tx.go
itayd: done

internal/backend/db/db.go
itayd: maybe in the future...

internal/backend/db/dbgorm/dbgorm.go
itayd: problem is with sqlite memory is that it open two different distinct databases, so we have to have this to make these the same. There is a comment about it a line above.

i changed the condition to be more specific to the case, and allow all other to have a separation.

internal/backend/db/dbgorm/dbgorm.go
itayd: looks like it's for pg as well, see a few lines above.

internal/backend/db/dbgorm/dbgorm.go
itayd: done

internal/backend/db/dbgorm/tx.go
itayd: tbh i don't know how to prevent this. the code itself should not do any writes in this case and i think it warrants a panic. in any case, how this can be solved?

internal/backend/db/dbgorm/dbgorm.go
itayd: this also checks the DSN, not just the db type.

daabr (approved):

itayd:
merging this. @efiShtain lmk if any issues.

PR #1008, opened 2025-01-24
refactor: default user is a normal user

itayd:
do not give any special treatment to default user - it is treated like any other normal user. only diff it cannot login and is populated in the users service on startup.


efiShtain:
Amazing! Finally!
good!

let's take it a step further

create a migration that generates the default user and default org always, no reason not having them in db anyway

if server is in mode "Default user" - just use that user always (email can be default@autokitteh.com or something we know)

otherwise, remove all references to special checks for default user

WDYT ?

itayd:
> if server is in mode "Default user" - just use that user always (email can be [default@autokitteh.com](mailto:default@autokitteh.com) or something we know)

I don't really like it - I want default to not be able to login, and I don't see why it should even have an email. Why have it if we don't need it?

> otherwise, remove all references to special checks for default user

That's... what this PR does? (I mean except for the seeding).

daabr (approved):
Noice!

Unlike Efi, I personally prefer special entities won't exist in prod DBs, so the server won't be only 1 step away from a problem just because of a misconfiguration. But that's a minor point beyond the scope of this PR.

PR #1013, opened 2025-01-27
feat: Block organization delete if it has projects

mario99logic:
Prevent organization deletion if it has any projects (active or inactive).
Used db.ListProjects() to check for existing projects. If there's a more optimized approach, please advise.

Tested it locally using Descope to log in.

### DIFF
--- a/internal/backend/db/db.go
+++ b/internal/backend/db/db.go
@@ -57,6 +57,8 @@ type DB interface {
 
 	ListProjects(context.Context, sdktypes.OrgID) ([]sdktypes.Project, error)
 
+	CountProjects(context.Context, sdktypes.OrgID) (int, error)
+
 	// Returns nill, nil if no resources are set.
 	GetProjectResources(context.Context, sdktypes.ProjectID) (map[string][]byte, error)
 
--- a/internal/backend/db/dbgorm/orgs.go
+++ b/internal/backend/db/dbgorm/orgs.go
@@ -44,6 +44,14 @@ func (gdb *gormdb) DeleteOrg(ctx context.Context, oid sdktypes.OrgID) error {
 		return sdkerrors.NewInvalidArgumentError("missing id")
 	}
 
+	projectCount, err := gdb.CountProjects(ctx, oid)
+	if err != nil {
+		return translateError(err)
+	}
+	if projectCount > 0 {
+		return errors.New("cannot delete an organization with projects")
+	}
+
 	return translateError(gdb.transaction(ctx, func(tx *tx) error {
 		err := tx.db.Where("org_id = ?", oid.UUIDValue()).Delete(&scheme.OrgMember{}).Error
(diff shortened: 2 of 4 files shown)


internal/backend/orgs/orgs.go
```diff
@@ -96,6 +97,14 @@ func (o *orgs) Delete(ctx context.Context, id sdktypes.OrgID) error {
 		return err
 	}
 
+	projects, err := o.db.ListProjects(ctx, id)
```
itayd: few points:
1. You should do this in a transaction together with DeleteOrg - otherwise there could be a race condition where someone might create a project after the list and before the delete, and we'll have it in a weird state. Check how transactions are used elsewhere.
2. You don't need to list the projects - that's too much. You need only to either count the projects (Which is much faster) or check if any exists.

mario99logic: Based on the second point, I chose to add the count projects function because it also allows for potential future use cases.

internal/backend/db/db.go
```diff
@@ -57,6 +57,8 @@ type DB interface {
 
 	ListProjects(context.Context, sdktypes.OrgID) ([]sdktypes.Project, error)
 
+	CountProjects(context.Context, sdktypes.OrgID) (int, error)
```
itayd, line 60: why do you export this function in the interface?

internal/backend/db/dbgorm/projects.go
```diff
@@ -147,6 +147,20 @@ func (gdb *gormdb) listProjects(ctx context.Context, oid sdktypes.OrgID) ([]sche
...
+
+	if oid.IsValid() {
+		q = q.Where("org_id = ?", oid.UUIDValue())
+	}
+
+	var count int64
+	if err := q.Model(&scheme.Project{}).Count(&count).Error; err != nil {
+		return 0, err
+	}
+	return int(count), nil
```
itayd, line 161: this should be inside the transaction in orgs.go.

internal/backend/db/dbgorm/orgs.go
```diff
@@ -44,6 +44,14 @@ func (gdb *gormdb) DeleteOrg(ctx context.Context, oid sdktypes.OrgID) error {
...
 	}
 
+	projectCount, err := gdb.CountProjects(ctx, oid)
+	if err != nil {
+		return translateError(err)
+	}
+	if projectCount > 0 {
+		return errors.New("cannot delete an organization with projects")
+	}
+
```
efiShtain, line 54: if the foreign keys exists (meaning project belongs to an organization), you can just try delete it and if you get foreign key violation error it means there are projects and you can't delete the organization

it reduces the count check
and make the code simpler (I think)

tebeka (changes requested):
See comments by @itayd

PR #1014, opened 2025-01-27
refactor jwt to async encryption

efiShtain:
update issuer
add aud claim
use ecdsa encryption

### DIFF
--- a/internal/backend/auth/authtokens/authjwttokens/jwt.go
+++ b/internal/backend/auth/authtokens/authjwttokens/jwt.go
@@ -1,12 +1,12 @@
 package authjwttokens
 
 import (
-	"encoding/hex"
 	"encoding/json"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
+
+	"crypto/ecdsa"
 
 	j "github.com/golang-jwt/jwt/v5"
 	"github.com/google/uuid"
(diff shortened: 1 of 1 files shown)


internal/backend/auth/authtokens/authjwttokens/jwt.go
```diff
@@ -18,46 +18,68 @@ import (
...
 
-const issuer = "ak"
+const (
+	issuer       = "https://api.autokitteh.cloud"
+	oneYear      = 365 * 24 * time.Hour
+	devPublicKey = `-----BEGIN PUBLIC KEY-----
+MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEpn+ugYKvxyjH1PP5M+TyI9AhxnWP
+3NtBzkt35Ppv9aX2YoBMXbbQcgUFDCwKY3QpsCmUILh3vno97lHkbgwjbQ==
+-----END PUBLIC KEY-----`
+	devPrivateKey = `-----BEGIN EC PRIVATE KEY-----
```
daabr, line 28: You can't use hard-coded default private keys, even in a dev server. It makes the dev server absolutely insecure, to the point I can't even run it on my work laptop.

Either way, you need to document in Docusaurus (https://github.com/autokitteh/docs.autokitteh.com/tree/main/docs/config) how to create and configure these keys, regardless of the server's mode, but especially for people who want to run in non-dev mode. Otherwise it's making AK unusable.

daabr (changes requested):

PR #1018, opened 2025-01-28
fix(forms): renew event watches

daabr:
Refs: INT-214

### DIFF
--- a/internal/backend/cron/google.go
+++ b/internal/backend/cron/google.go
@@ -34,7 +34,7 @@ func (cr *Cron) renewGoogleEventWatchesWorkflow(wctx workflow.Context, la listAc
 		return err
 	}
 
-	errs := make([]error, 0)
+	var errs []error
 	for _, cid := range cids {
 		if err := workflow.ExecuteActivity(actx, ra, cid).Get(wctx, nil); err != nil {
 			errs = append(errs, err)
@@ -82,7 +82,7 @@ func (cr *Cron) listGoogleDriveConnectionsActivity(ctx context.Context) ([]sdkty
 		return nil, err
 	}
 
-	cids := make([]sdktypes.ConnectionID, 0)
+	var cids []sdktypes.ConnectionID
 	for _, c := range cs {
 		cid := c.ID()
 		if cr.checkGoogleDriveEventWatch(ctx, cid) {
@@ -106,7 +106,7 @@ func (cr *Cron) listGoogleFormsConnectionsActivity(ctx context.Context) ([]sdkty
 		return nil, err
 	}
 
-	cids := make([]sdktypes.ConnectionID, 0)
+	var cids []sdktypes.ConnectionID
 	for _, c := range cs {
 		cid := c.ID()
 		if cr.checkGoogleFormsEventWatch(ctx, cid) {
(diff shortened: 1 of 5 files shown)


pashafateev (approved):

internal/backend/cron/google.go
```diff
@@ -2,20 +2,64 @@ package cron
...
+
+func (cr *Cron) renewGoogleEventWatchesWorkflow(wctx workflow.Context, la listActivity, ra renewActivity) error {
+	actx := temporalclient.WithActivityOptions(wctx, taskQueueName, cr.cfg.Activity)
+
+	var cids []sdktypes.ConnectionID
+	if err := workflow.ExecuteActivity(actx, la).Get(wctx, &cids); err != nil {
+		return err
+	}
+
+	errs := make([]error, 0)
```
itayd: why the 0?

itayd: you don't even need to make

```
var errs []error
```

append to nil works

internal/backend/cron/google.go
```diff
@@ -26,10 +70,70 @@ func (cr *Cron) listGoogleDriveConnectionsActivity(ctx context.Context) ([]sdkty
 	return nil, nil // INT-184: Implement.
 }
 
+func (cr *Cron) listGoogleFormsConnectionsActivity(ctx context.Context) ([]sdktypes.ConnectionID, error) {
+	ctx = authcontext.SetAuthnSystemUser(ctx)
+
+	// Enumerate all Google Forms connections (there's no single connection var
+	// value that we're looking for, so we can't use "cr.vars.FindConnectionIDs").
+	cs, err := cr.connections.List(ctx, sdkservices.ListConnectionsFilter{
```
itayd, line 101: that can be a very big list in the future, need to figure out pagination here

internal/backend/cron/jira.go
```diff
@@ -82,25 +82,25 @@ func (ct *Cron) checkJiraEventWatch(ctx context.Context, cid sdktypes.Connection
...
 	if err != nil {
-		l.Error("invalid Jira event watch expiration time during renewal check",
+		l.Warn("invalid Jira event watch expiration time during renewal check",
 			zap.String("expiration", e), zap.Error(err),
 		)
 		return false
 	}
 
 	// Update this event watch only if it's about to expire in less than 2 weeks.
+	twoWeeksFromNow := time.Now().UTC().AddDate(0, 0, 14)
```
itayd, line 94: don't you want this configurable?

itayd (approved):

itayd (approved):
appproved with comments, nothing blocking

internal/backend/cron/google.go
daabr: yes, but I need to declare `errs` in advance, and this is a good placeholder in case we want to optimize in the future, but today we don't, and that's why I'm using 0

internal/backend/cron/jira.go
daabr: no, I explicitly don't want users to be able to configure this on their own.

But even if we do add it in the future, it will be in an integration's config and not a cron config, and that's why a config struct isn't useful yet.

internal/backend/cron/google.go
daabr: on second thought, your suggestion is still better, changing now

internal/backend/cron/jira.go
itayd: not on their own, as ak config not user's

PR #1026, opened 2025-01-30
feat(descope): expose project id

itayd:
- allow a client to figure out the project id by accessing `<HOST>/auth/descope/project_id`.
- remove management key, unused.

### DIFF
--- a/internal/backend/auth/authloginhttpsvc/descope.go
+++ b/internal/backend/auth/authloginhttpsvc/descope.go
@@ -13,8 +13,9 @@ import (
 )
 
 const (
-	descopeRootPath  = "/auth/descope"
-	descopeLoginPath = descopeRootPath + "/login"
+	descopeRootPath      = "/auth/descope"
+	descopeProjectIDPath = descopeRootPath + "/project_id"
+	descopeLoginPath     = descopeRootPath + "/login"
 )
 
 func registerDescopeRoutes(mux *http.ServeMux, cfg descopeConfig, onSuccess func(context.Context, *loginData) http.Handler) error {
@@ -27,7 +28,7 @@ func registerDescopeRoutes(mux *http.ServeMux, cfg descopeConfig, onSuccess func
 		return err
 	}
 
-	mux.HandleFunc(descopeRootPath+"/project_id", func(w http.ResponseWriter, _ *http.Request) {
+	mux.HandleFunc(descopeLoginPath, func(w http.ResponseWriter, _ *http.Request) {
 		w.Header().Set("Content-Type", "application/json")
 		fmt.Fprintf(w, "%q", cfg.ProjectID)
 	})
--- a/internal/backend/auth/authloginhttpsvc/descope.go
+++ b/internal/backend/auth/authloginhttpsvc/descope.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	_ "embed"
 	"errors"
+	"fmt"
 	"net/http"
 
 	"github.com/descope/go-sdk/descope/client"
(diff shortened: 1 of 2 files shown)


RonenMars (dismissed):

RonenMars:
Continuing our discussion, I still don't think that the path that reveals the descopeProjectID should be public.
I'll give you the simplest example:
Currently, our Descope is handling the login of the users, adding them to the Descope Users list, and creating a contact in HubSpot.
If someone gets the ID from the production and uses it in their local development environment or any other instance - it will add the users who logged in to their app to our Descope Users list and HubSpot.
This is my main concern.

@itayd @efiShtain

efiShtain:
> Continuing our discussion, I still don't think that the path that reveals the descopeProjectID should be public. I'll give you the simplest example: Currently, our Descope is handling the login of the users, adding them to the Descope Users list, and creating a contact in HubSpot. If someone gets the ID from the production and uses it in their local development environment or any other instance - it will add the users who logged in to their app to our Descope Users list and HubSpot. This is my main concern.
> 
> @itayd @efiShtain

If this can happen, it means we didn't block things well enough in descope or their product is basically useless - I assume we didn't configure something correctly

the projectID is a public entity
letting the UI get it over an API is better for security then embedding it in the UI code - but it is basically public in both cases

internal/backend/auth/authloginhttpsvc/descope.go
```diff
@@ -4,26 +4,35 @@ import (
...
-	client, err := client.NewWithConfig(&client.Config{ProjectID: cfg.ProjectID, ManagementKey: cfg.ManagementKey})
+	client, err := client.NewWithConfig(&client.Config{ProjectID: cfg.ProjectID})
 	if err != nil {
 		return err
 	}
 
-	mux.Handle(descopeLoginPath, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	mux.HandleFunc(descopeRootPath+"/project_id", func(w http.ResponseWriter, _ *http.Request) {
+		w.Header().Set("Content-Type", "application/json")
+		fmt.Fprintf(w, "%q", cfg.ProjectID)
```
efiShtain, line 33: is this a valid json ?

internal/backend/auth/authloginhttpsvc/descope.go
```diff
@@ -4,26 +4,35 @@ import (
...
 	}
 
-	client, err := client.NewWithConfig(&client.Config{ProjectID: cfg.ProjectID, ManagementKey: cfg.ManagementKey})
+	client, err := client.NewWithConfig(&client.Config{ProjectID: cfg.ProjectID})
 	if err != nil {
 		return err
 	}
 
-	mux.Handle(descopeLoginPath, http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+	mux.HandleFunc(descopeRootPath+"/project_id", func(w http.ResponseWriter, _ *http.Request) {
```
efiShtain: add this as a const like the other path, makes it more clear

internal/backend/auth/authloginhttpsvc/descope.go
itayd: %q is quoted, so yes.

internal/backend/auth/authloginhttpsvc/descope.go
itayd: done

internal/backend/auth/authloginhttpsvc/descope.go
efiShtain: I'm not sure a quoted string is a valid json
it has to be wrapped in [] or {} and has to have "key":"value" representation

itayd: a quoted string is valid json

efiShtain (approved):

PR #1027, opened 2025-01-30
fix: github and slack custom oauth save - handling GET request

RonenMars:
Inside the [server "router"](https://github.com/autokitteh/autokitteh/blob/98169778d174b826fd2729a22029aff2668f6bae/integrations/github/server.go#L42C1-L43C1) we allow starting the GitHub custom OAuth flow with POST method for the internal connections UI and with GET method for the WebUI.

Inside the Custom OAuth Save Handler we are taking care only of the POST method, and this is what I fixed here, added a proper treatment for the GET method (for the WebUI).

### DIFF
--- a/integrations/github/save.go
+++ b/integrations/github/save.go
@@ -19,26 +19,32 @@ import (
 func (h handler) handleSave(w http.ResponseWriter, r *http.Request) {
 	c, l := sdkintegrations.NewConnectionInit(h.logger, w, r, desc)
 
-	// Check "Content-Type" header.
-	contentType := r.Header.Get(headerContentType)
-	if !strings.HasPrefix(contentType, contentTypeForm) {
-		c.AbortBadRequest("unexpected content type")
-		return
+	var form url.Values
+	if r.Method == "GET" {
+		form = r.URL.Query()
+	} else {
+		// Check "Content-Type" header.
+		contentType := r.Header.Get(headerContentType)
+		if !strings.HasPrefix(contentType, contentTypeForm) {
+			c.AbortBadRequest("unexpected content type")
+			return
+		}
+
+		// Read and parse POST request body.
+		if err := r.ParseForm(); err != nil {
+			l.Warn("Failed to parse incoming HTTP request", zap.Error(err))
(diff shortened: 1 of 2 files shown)


pashafateev:
@RonenMars Slack and Auth0 are implemented in the same way. Why is it only an issue here?

RonenMars:
> @RonenMars Slack and Auth0 are implemented in the same way. Why is it only an issue here?

@pashafateev I think I found the difference.
Sorry for the lame screenshots from the mobile, but the check is the method is POST missing in GitHub custom OAuth:

RonenMars:
Yep, that was the issue, fixed according to the same condition we used in the `auth0` integration.

pashafateev:
> Yep, that was the issue, fixed according to the same condition we used in the `auth0` integration.

Makes sense, but I have a follow up question unrelated to github. Do we have the same issue in Slack?

RonenMars:
No, since on custom OAuth from frontend we're sending the data to:
```
https://SERVER_ADDRESS/auth0/save?cid=CONN_ID&origin=web&auth_type=oauth&client_id=YYY&client_secret=ZZZ&auth0_domain=LALALA
```
or
```
https://SERVER_ADDRESS/github/custom-oauth?cid=CONN_ID&origin=web&auth_type=custom-oauth&client_id=XXX&client_secret=YYY&app_id=ZZZ&webhook_secret=LALALA&enterprise_url=URLRU&private_key=PRIVATEKEY
```

Then the backend saves the vars of the custom oauth app, and from there, it redirects to `https://SERVER_ADDRESS/oauth/start/github?cid=%s&origin=%s`

But in the case of oauth (not custom) - we already have the (env) vars of the applications - so in that case, the frontend skips the "vars-save" step, and goes straight to: `https://SERVER_ADDRESS/oauth/start/${integrationName}?cid=${oauthConnectionId}&origin=web`

pashafateev:
Worked out offline:
Slack custom oauth has not been incorporated into the front end yet, which is why the issue was not noticed there.

pashafateev (approved):

daabr (approved):

PR #1031, opened 2025-01-31
fix(google): more robust watch renewal

daabr:
Follow-up to Jira in #1025:
1. Google Forms: if the watched form was deleted (ID not found), forget it and its watches
2. Google Calendar & Forms: watches are optional, don't try renew if there's no expiration time
3. Better logging

Also tweaked Jira: if a watch ID is invalid and gets removed, treating it as a successful activity makes more sense than a non-retryable error. In contrast, server errors still require "loud" failures.

Refs: INT-219

### DIFF
--- a/internal/backend/cron/jira.go
+++ b/internal/backend/cron/jira.go
@@ -118,9 +118,7 @@ func (cr *Cron) renewJiraEventWatchActivity(ctx context.Context, cid sdktypes.Co
 	if err != nil {
 		l.Warn("invalid Jira event watch ID for renewal", zap.String("watch_id", wid), zap.Error(err))
 		cr.deleteInvalidWatchID(ctx, cid, wid)
-		return temporal.NewNonRetryableApplicationError(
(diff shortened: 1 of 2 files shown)


internal/backend/cron/jira.go
```diff
@@ -118,9 +118,7 @@ func (cr *Cron) renewJiraEventWatchActivity(ctx context.Context, cid sdktypes.Co
 	if err != nil {
 		l.Warn("invalid Jira event watch ID for renewal", zap.String("watch_id", wid), zap.Error(err))
 		cr.deleteInvalidWatchID(ctx, cid, wid)
-		return temporal.NewNonRetryableApplicationError(
-			"invalid Jira event watch ID: "+err.Error(), "NumError", err, cid.String(), wid,
-		)
+		return nil // No need to retry.
```
pashafateev, line 121: I think I understand everything but this.

pashafateev (approved):

internal/backend/cron/jira.go
daabr: explained offline

PR #1033, opened 2025-01-31
feat(microsoft): support more auth types + MS Teams fixes

daabr:
1. Add support for "daemon apps" (apps with their own identity, like private OAuth apps but without user-delegated permissions, i.e. bots)
2. Connection test for daemon apps (there's no "me" user identity)
3. Add OAuth scopes for MS Teams
4. `"isCustomOauth"`: add logic to detect private MS apps
5. Python SDK: add MS Team initialization function, support for private OAuth/daemon apps, and certificate-based private apps

Refs: INT-202, INT-227

### DIFF
--- a/integrations/microsoft/connection/checks.go
+++ b/integrations/microsoft/connection/checks.go
@@ -5,9 +5,12 @@ package connection
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
+	"strings"
 
 	"golang.org/x/oauth2"
 
@@ -39,6 +42,8 @@ func Status(v sdkservices.Vars) sdkintegrations.OptFn {
 			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
 		case integrations.OAuthDefault, integrations.OAuthPrivate:
 			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "Using OAuth 2.0"), nil
+		case integrations.DaemonApp:
+			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "Using daemon app"), nil
 		default:
 			return sdktypes.NewStatus(sdktypes.StatusCodeError, "Bad auth type"), nil
 		}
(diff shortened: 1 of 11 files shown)


integrations/microsoft/connection/checks.go
```diff
@@ -131,19 +137,64 @@ func GetUserInfo(ctx context.Context, t *oauth2.Token) (*UserInfo, error) {
...
+		// https://learn.microsoft.com/en-us/answers/questions/1853467/aadsts7000229-the-client-application-is-missing-se
+		return errors.New("missing tenant ID")
+	}
+	u := fmt.Sprintf("https://login.microsoftonline.com/%s/oauth2/v2.0/token", tenantID)
+
+	// TODO(INT-227): Add support for certificate-based authentication.
+	data := url.Values{}
+	data.Set("grant_type", "client_credentials")
+	data.Set("client_id", vs.GetValueByString("private_client_id"))
+	data.Set("client_secret", vs.GetValueByString("private_client_secret"))
```
pashafateev: nit: why not symbols?

integrations/microsoft/connection/checks.go
```diff
@@ -131,19 +137,64 @@ func GetUserInfo(ctx context.Context, t *oauth2.Token) (*UserInfo, error) {
...
+	data.Set("scope", "https://graph.microsoft.com/.default")
+
+	if data.Get("client_id") == "" || data.Get("client_secret") == "" {
```
pashafateev, line 159: nit: symbols?

integrations/microsoft/save.go
```diff
@@ -68,6 +68,22 @@ func (h handler) handleSave(w http.ResponseWriter, r *http.Request) {
 		}
 		startOAuth(w, r, c, l)
 
+	// Same as a private OAuth 2.0 app, but without the OAuth 2.0 flow
+	// (it uses application permissions instead of user-delegated ones).
```
pashafateev, line 72: I don't understand what an OAuth 2.0 app without OAuth flow means?

pashafateev (approved):

runtimes/pythonrt/py-sdk/autokitteh/microsoft.py
```diff
@@ -68,13 +104,78 @@ def __init__(self, connection: str) -> None:
...
+    client_id = os.getenv(connection + "__private_client_id")
+    client_secret = os.getenv(connection + "__private_client_secret")
+    if not client_id or not client_secret:
+        raise ConnectionInitError(connection)
+
+    return identity.ClientSecretCredential(
+        client_id=client_id,
+        client_secret=client_secret,
+        tenant_id=os.getenv(connection + "__private_tenant_id", "") or "common",
+    )
```
pashafateev, line 181: I don't really understand these different modes, but it's probably very specific to Microsoft

integrations/microsoft/connection/checks.go
daabr: good point, done

integrations/microsoft/connection/checks.go
daabr: not here, these are Go map values (even though they're based on connection vars)

integrations/microsoft/save.go
daabr: It's basically a bot app. It's configured like a regular user-delegated OAuth app (client ID, client secret, etc.), but doesn't actually use an OAuth 2.0 3-legged flow.

runtimes/pythonrt/py-sdk/autokitteh/microsoft.py
daabr: yes, the Docusaurus documentation should clarify this

PR #1034, opened 2025-02-01
fix(pythonrt): correctly interpret activity reply error

itayd:
instead of showing "activity reply not a Custom value", show the actual error.

@tebeka  - i am not sure this is the "right way (tm)". please advise.


itayd:
replaced by https://github.com/autokitteh/autokitteh/pull/1049

PR #1035, opened 2025-02-03
fix: Block connection deletion with active tiggers

mario99logic:
Implemented `hasTriggers` function  to check for existing triggers and integrated it into `deleteConnection`.

### DIFF
--- a/internal/backend/db/dbgorm/connections.go
+++ b/internal/backend/db/dbgorm/connections.go
@@ -3,6 +3,7 @@ package dbgorm
 import (
 	"context"
 	"errors"
+	"fmt"
 	"maps"
 
 	"github.com/google/uuid"
@@ -34,7 +35,7 @@ func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection
 func (gdb *gormdb) deleteConnectionsAndVars(ctx context.Context, what string, id uuid.UUID) error {
 	// should be transactional with context already applied
 	if what == "connection_id" {
-		hasTriggers, err := gdb.hasTriggers(ctx, id)
+		hasTriggers, err := gdb.doesConnectionHaveTriggers(ctx, id)
 		if err != nil {
 			return translateError(err)
 		}
@@ -56,6 +57,23 @@ func (gdb *gormdb) deleteConnectionsAndVars(ctx context.Context, what string, id
 	}
 
 	return nil
+}
+
+func (gdb *gormdb) doesConnectionHaveTriggers(ctx context.Context, connID uuid.UUID) (bool, error) {
+	var exists bool
+	q := gdb.db.WithContext(ctx)
+
+	q = q.Model(&scheme.Trigger{}).
+		Select("1").
+		Where("connection_id = ?", connID).
+		Where("deleted_at IS NULL").
+		Limit(1)
+
+	err := q.Find(&exists).Error
+	if err != nil {
+		return false, fmt.Errorf("checking active triggers: %w", err)
+	}
+	return exists, nil
 }
 
 func (gdb *gormdb) deleteConnection(ctx context.Context, id uuid.UUID) error {
--- a/internal/backend/db/dbgorm/connections.go
+++ b/internal/backend/db/dbgorm/connections.go
@@ -33,6 +33,15 @@ func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection
 
 func (gdb *gormdb) deleteConnectionsAndVars(ctx context.Context, what string, id uuid.UUID) error {
 	// should be transactional with context already applied
+	if what == "connection_id" {
+		hasTriggers, err := gdb.hasTriggers(ctx, id)
+		if err != nil {
+			return translateError(err)
+		}
+		if hasTriggers {
+			return errors.New("cannot delete a connection that has associated triggers")
+		}
+	}
 
 	var ids []uuid.UUID
 	q := gdb.db.WithContext(ctx).Model(&scheme.Connection{})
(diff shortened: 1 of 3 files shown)


internal/backend/db/dbgorm/connections.go
```diff
@@ -49,6 +50,13 @@ func (gdb *gormdb) deleteConnectionsAndVars(ctx context.Context, what string, id
 }
 
 func (gdb *gormdb) deleteConnection(ctx context.Context, id uuid.UUID) error {
+	hasTriggers, err := gdb.hasTriggers(ctx, id)
+	if err != nil {
+		return translateError(err)
+	}
```
efiShtain: not sure this check is atomic

a trigger can be created between checking there are no triggers and calling the delete connection

also someone can call the deleteConnectionsAndVars directly which bypass this check

so it should probably either move to the deleteConnectionAndVars

you can verify in the delete connection there are no non deleted triggers in one atomic way I think

mario99logic: Isn't this already atomic since `deleteConnection` runs in a transaction?
Also noticed `deleteConnectionsAndVars` is only called in two places:

- in deleteConnection (after trigger check)
- in deleteProjectAndDependents (where trigger checks aren't needed since we're deleting the whole project)

So when would someone call `deleteConnectionsAndVars` where we'd need the check?
Maybe we should either:

1. Keep the check in deleteConnection since it's specific to single connection deletions
2. Or just remove deleteConnection entirely and put the check in deleteConnectionsAndVars but then the check is not relevant if where deleting a project

mario99logic: Moved hasTrigger check to `deleteConnectionsAndVars`. Kept `deleteConnection` for now because it's used in `connections_test`.  Making it atomic requires changing `deleteConnectionsAndVars` and other dependent code, which isn't worth it right now as the race condition isn't critical.

internal/backend/db/dbgorm/triggers.go
```diff
@@ -30,6 +30,23 @@ func (gdb *gormdb) getTriggerByID(ctx context.Context, triggerID uuid.UUID) (*sc
 	return getOne[scheme.Trigger](gdb.db.WithContext(ctx), "trigger_id = ?", triggerID)
 }
 
+func (gdb *gormdb) hasTriggers(ctx context.Context, connID uuid.UUID) (bool, error) {
```
efiShtain: i think this should be in the connections file, it is related only to connections and would probably never be used by other services

internal/backend/db/dbgorm/connections.go
```diff
@@ -32,6 +33,15 @@ func (gdb *gormdb) createConnection(ctx context.Context, conn *scheme.Connection
 
 func (gdb *gormdb) deleteConnectionsAndVars(ctx context.Context, what string, id uuid.UUID) error {
 	// should be transactional with context already applied
+	if what == "connection_id" {
```
efiShtain, line 37: not sure i like this condition

does it work only for connection id or project id ?
if so, maybe we should split it to two dedicated functions instead of passing strings

deleteConntectionAndVars
deleteProjectConntectionsAndVars

mario99logic: We can use the function `deleteConnection` only for connection id since there is no use for it other than calling `deleteConnectionsAndVars`, I dont think there is the need to create another function, also we can use a transaction for it to make the process atomic and also use FOR(UPDATE) which blocks adding new rows (creating new triggers) to the db while making the check

internal/backend/db/dbgorm/triggers.go
mario99logic: similar to org and projects, list projects is in the projects file so I used the same logic here

internal/backend/db/dbgorm/connections.go
efiShtain: select for update won't block inserting other rows
it only block reading the same rows both with select for update

the condition bothers me, not the transaction/atomicity, that's a different issue

internal/backend/db/dbgorm/triggers.go
efiShtain: there is already listTriggers, which belongs to the trigger service cause it's not specific to any other model,
you can extend it with countOnly if you want

hasTriggers is specifically for connections and specific for one usage in connections and won't be used in any other use cases, so it's more related to connections than to triggers, as I see it

internal/backend/db/dbgorm/connections.go
efiShtain: just make sure you add tests if you keep this one here

internal/backend/db/dbgorm/connections_test.go
```diff
@@ -132,6 +132,14 @@ func TestDeleteConnectionAndVars(t *testing.T) {
 	// delete all connections for specific projectID
 	assert.NoError(t, f.gormdb.deleteConnectionsAndVars(ctx, "project_id", p.ProjectID))
 	f.assertConnectionDeleted(t, c2, c3)
+
+	trigger := f.newTrigger(c4, p)
+	f.createTriggersAndAssert(t, trigger)
+
+	// Try deleting a connection that has associated triggers
```
efiShtain: make it a dedicated test please
TestCantDeleteConnectionWithTriggers

daabr (commented):
LGTM but I'm not an expert on the DB's conventions. Please wait for @efiShtain 's approval too.

daabr (approved):

internal/backend/db/dbgorm/connections_test.go
```diff
@@ -134,6 +134,22 @@ func TestDeleteConnectionAndVars(t *testing.T) {
 	f.assertConnectionDeleted(t, c2, c3)
 }
 
+func TestCantDeleteConnectionWithTriggers(t *testing.T) {
```
efiShtain, line 137: Cant

efiShtain (approved):

PR #1039, opened 2025-02-04
fix-ish: get rid of SessionLog sdktypes

itayd:
it's annoying. we don't want it. this should also make @efiShtain happy(ish).


efiShtain:
Not sure I get what you tried to do here, 
its a breaking change for the API just to rename a field ?

itayd:
> Not sure I get what you tried to do here,

Removed the SessionLog object. When @RonenMars migrates away from it we could remove it from the API.

> its a breaking change for the API just to rename a field ?

Nothing breaking in this PR.

efiShtain (approved):

PR #1040, opened 2025-02-04
feat(sessions): get prints

itayd:
add another method to the sessions service: get prints. this will return only the session prints. the new prints methods returns prints as values, which for now will consist of only strings, but in the future would be any arbitrary value (think js's `console.log`).

in the future the prints will be pulled out from the session log and would be able to be retrieved only using get prints.


sdk/sdkservices/sessions.go
```diff
@@ -39,6 +39,16 @@ type GetLogResults struct {
 	sdktypes.PaginationResult
 }
 
+type SessionPrint struct {
```
efiShtain, line 42: why is this in sdkservices and not in sdktypes ?
I'm not sure where it should be

proto/autokitteh/sessions/v1/session.proto
```diff
@@ -114,7 +114,7 @@ message SessionLogRecord {
   string process_id = 2;
 
   // one of the following is required.
-  Print print = 10;
+  Print print = 10; // deprecated, avoid.
```
efiShtain, line 117: I think there is a standard way for grpc to deprecate fields

commenting this line out and adding a reserved 10 line so the number can't be used

but since you already break the API with renaming stuff, maybe we should just remove the field ?

proto/autokitteh/sessions/v1/svc.proto
```diff
@@ -118,6 +120,34 @@ message GetLogResponse {
   string next_page_token = 10;
 }
 
+message GetPrintsRequest {
+  string session_id = 1 [(buf.validate.field).string.min_len = 1];
+
+  bool ascending = 11;
+
+  int32 page_size = 20;
```
efiShtain, line 128: this one also has to be greater than 0

proto/autokitteh/sessions/v1/svc.proto
```diff
@@ -118,6 +120,34 @@ message GetLogResponse {
...
+}
+
+message GetPrintsResponse {
+  message Print {
+    values.v1.Value v = 1;
+    google.protobuf.Timestamp t = 2;
+  }
+
+  repeated Print prints = 1 [(buf.validate.field).repeated.items.required = true];
+  int64 count = 2;
```
efiShtain: we have this in some places, 
it's kind of dangerous in the naive implementation
assuming we have 10k prints, we don't want to count all of them in db just to get the 10k number and return 20 actual fields

and anyway the user won't show next page or number of pages, it is just polling the next token until nil is returned, so the count isn't really required.

so either we remove the count, or create some stats table to handle it

on the other hand, if you think that for the mid run we won't have prints list with more then let's say 500 prints, I think we can leave it as is

efiShtain (approved):
Pre approve, assuming you fix what we discussed online

proto/autokitteh/sessions/v1/svc.proto
itayd: removed count.

proto/autokitteh/sessions/v1/session.proto
itayd: we don't break the api here. thing is i also want to be json compatible, so for now i can leave it in. once we migrate UI away from this we can just remove this.

proto/autokitteh/sessions/v1/svc.proto
itayd: i did not see anywhere else this check, but i did limited it now to 100 by default as being done in GetLog.

PR #1049, opened 2025-02-05
fix(python): Better error from activity reply

tebeka:
Fixes: ENG-1990


itayd:
did you see https://github.com/autokitteh/autokitteh/pull/1034?

itayd (approved):

PR #1052, opened 2025-02-05
refactor(integrations): simpler descriptors

daabr:
Reuse a common function instead of copy-pasting the full protocol buffer in all integrations.

In a few integrations that define connection var names, rename them from "Xxx" to "XxxVar".


itayd (approved):

itayd:
just make sure the var names themselves did not change

PR #1054, opened 2025-02-05
feat(sessionsrpc): straight input of a json object

itayd:
instead of ingesting param by param which presents difficulties with jsonifying parameters, just accept a straight raw json object which must be a dictionary and process it as an input. all the "smart" conversion is done on the backend side.


proto/autokitteh/sessions/v1/svc.proto
```diff
@@ -19,6 +19,10 @@ message StartRequest {
 
   // Helper: if set, merged into and overwrites the session's inputs.
   map<string, string> json_inputs = 2;
+
+  // Helper: if set,  merged into and overwrites the session's inputs.
```
daabr:
```suggestion
  // Helper: if set, merged into and overwrites the session's inputs.
```

daabr (approved):
I don't mind updating the web UI signature on the way, just making sure that it's intentional.

Other than than, good unit tests, and I left a tiny nit suggestion.

itayd:
> I don't mind updating the web UI signature on the way, just making sure that it's intentional.

It's all backward compatible.

PR #1055, opened 2025-02-06
refactor(integrations): webhook handlers

daabr:
New reusable functions to register UI/save/OAuth webhook handlers in `integrations/common/server.go` - mostly for new-style integrations, but not just.

Most of the PR is tweaks related to that (using these new functions), or cosmetic tweaks (e.g. rename `muxes *muxes.Muxes` variable to `m` to be more idiomatic about Go var naming).

### DIFF
--- a/integrations/common/server.go
+++ b/integrations/common/server.go
@@ -1,0 +1,38 @@
+package common
+
+import (
+	"embed"
+	"fmt"
+	"net/http"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/muxes"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+// ServeStaticUI registers an integration's static web content to
+// AutoKitteh's internal user-authenticated HTTP server. User auth
+// isn't actually required for serving these files, but it's impossible
+// to use them (i.e. to initialize a connection) without authentication.
+func ServeStaticUI(m *muxes.Muxes, i sdktypes.Integration, fs embed.FS) {
+	pattern := fmt.Sprintf("GET %s/", i.ConnectionURL().Path)
+	m.Auth.Handle(pattern, http.FileServer(http.FS(fs)))
+}
+
+// RegisterSaveHandler registers a webhook to handle saving connection variables.
+// This is always the first step in initializing a connection, regardless of its
+// auth type. It's also the last step for non-OAuth authentication types. The
+// handler function requires an authenticated user context for database access.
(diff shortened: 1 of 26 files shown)


integrations/common/server.go
```diff
@@ -0,0 +1,38 @@
...
+	pattern := fmt.Sprintf("GET %s/", i.ConnectionURL().Path)
+	m.Auth.Handle(pattern, http.FileServer(http.FS(fs)))
+}
+
+// RegisterSaveHandler registers a webhook to handle saving connection variables.
+// This is always the first step in initializing a connection, regardless of its
+// auth type. It's also the last step for non-OAuth authentication types. The
+// handler function requires an authenticated user context for database access.
+func RegisterSaveHandler(m *muxes.Muxes, i sdktypes.Integration, h http.HandlerFunc) {
+	pattern := fmt.Sprintf(" %s/save", i.ConnectionURL().Path)
```
mario99logic, line 26:
```suggestion
	pattern := fmt.Sprintf("%s/save", i.ConnectionURL().Path)
```

integrations/common/server.go
```diff
@@ -0,0 +1,38 @@
...
+	m.Auth.HandleFunc(http.MethodGet+pattern, h)
+	m.Auth.HandleFunc(http.MethodPost+pattern, h)
+}
+
+// RegisterOAuthHandler registers a webhook to handle the last step in a 3-legged
+// OAuth 2.0 flow. This is the only step that requires an authenticated user context
+// for database access. The handler function receives an incoming redirect request
+// from AutoKitteh's generic OAuth service, which contains an OAuth token (if the
+// OAuth flow was successful) and form parameters for debugging and validation.
+func RegisterOAuthHandler(m *muxes.Muxes, i sdktypes.Integration, h http.HandlerFunc) {
```
mario99logic, line 36: question in general, are timeouts configured for these handlers?

mario99logic (approved):

integrations/common/server.go
daabr: No, see in lines 27-28 that I prepend "GET" or "POST" to this pattern string. Without the space it will be "GET/path" instead of "GET /path".

integrations/common/server.go
daabr: Timeouts in HTTP handlers are handled with the request's context (`r.Context()`). The client can decide to abort the request for any reason at any time, and this will cause the request's context to be cancelled.

Servers can also implement their own timeouts, e.g. "don't allow requests to linger for more than X seconds" - but: (A) it doesn't fit everyone, (B) even when implemented, it will simply cause the client request to be closed, which will have the same result as described above.

PR #1057, opened 2025-02-06
fix: redirect from Github marketplace

mario99logic:
Marketplace Installation Flow Enhancement

- Added handleGitHubMarketplace function to handle direct app installations from GitHub Marketplace
- Redirects users to GitHub app installations page after successful installation
- Tested functionality using ngrok for local development

The redirect currently points to `/settings/installations/{installation_id}`. We could alternatively redirect to `/apps/installations/{installation_id}` to maintain marketplace context.

Refs: INT-172

### DIFF
--- a/internal/backend/oauth/webhook.go
+++ b/internal/backend/oauth/webhook.go
@@ -84,6 +84,10 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if handled := s.handleGitHubMarketplace(w, r, intg, state); handled {
+		return
+	}
+
 	// Ensure the state parameter is well-formed.
 	sub := regexp.MustCompile(`^(.+)_([a-z]+)$`).FindStringSubmatch(state)
 	if len(sub) != 3 {
@@ -154,6 +158,15 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
 	}
 
 	redirect(w, r, intg, sub[1], sub[2], "oauth", oauthData)
+}
+
+func (s *svc) handleGitHubMarketplace(w http.ResponseWriter, r *http.Request, intg string, state string) bool {
+	if intg == "github" && state == "" {
+		// User came from Marketplace (no AK-generated state)
+		http.Redirect(w, r, "https://github.com/settings/installations/"+r.FormValue("installation_id"), http.StatusFound)
+		return true
+	}
+	return false
 }
 
 func abort(w http.ResponseWriter, r *http.Request, intg, cid, origin, err string) {
--- a/internal/backend/oauth/webhook.go
+++ b/internal/backend/oauth/webhook.go
@@ -87,7 +87,7 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
 
 	handled, err := s.redirectToGitHub(w, r, intg, state)
 	if err != nil {
-		l.Error("Failed to redirect to GitHub", zap.Error(err))
+		l.Warn("Failed to redirect to GitHub", zap.Error(err))
 		return
 	}
 	if handled {
--- a/internal/backend/oauth/webhook.go
+++ b/internal/backend/oauth/webhook.go
@@ -85,7 +85,12 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
-	if handled := s.redirectToGitHub(w, r, intg, state); handled {
+	handled, err := s.redirectToGitHub(w, r, intg, state)
+	if err != nil {
+		l.Error("Failed to redirect to GitHub", zap.Error(err))
+		return
+	}
+	if handled {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


internal/backend/oauth/webhook.go
```diff
@@ -156,6 +160,15 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
 	redirect(w, r, intg, sub[1], sub[2], "oauth", oauthData)
 }
 
+func (s *svc) handleGitHubMarketplace(w http.ResponseWriter, r *http.Request, intg string, state string) bool {
```
daabr: 1. "handle" is too generic for this function's name, in general try to avoid "get", "handle", "do", etc. This function does something specific: it redirects the user to GitHub instead of AutoKitteh, if the "state" parameter if missing (because that means the user installed the app from GitHub Apps and not from AutoKitteh).
2. "Marketplace" is too specific. This applies also if you install the app from GitHub, not just the GitHub Marketplace.

--> So I suggest you rename this function to "redirectToGitHub", and also add Go documentation to this function that explains what it does and why (basically what I wrote in 1 above). Be careful to do this in a way to conforms to Go style: https://go.dev/doc/comment#func

internal/backend/oauth/webhook.go
```diff
@@ -156,6 +160,15 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
...
+	if intg == "github" && state == "" {
+		// User came from Marketplace (no AK-generated state)
```
daabr: This is a sentence, use punctuation. Try to avoid "AK" in our code, prefer "AutoKitteh".

But anyway, this comment line shouldn't be here at all - it should document the whole function. Remember:
- Comments about a function describe what it does and why, rarely how - for the benefits of callers and readers who are not familiar with it
- Comments inside a function describe internal implementation logic (in case it's long / not easy to understand it) - for the benefit of code maintainers who are expected to modify it

internal/backend/oauth/webhook.go
```diff
@@ -156,6 +160,15 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
...
+		http.Redirect(w, r, "https://github.com/settings/installations/"+r.FormValue("installation_id"), http.StatusFound)
```
daabr: This URL is incorrect. You're missing the org ID:

- https://github.com/settings/installations/59430286 --> 404
- https://github.com/organizations/autokitteh/settings/installations/59430286 is correct

But anyway, please use this URL instead: `https://github.com/apps/{app_name}` (e.g. https://github.com/apps/autokitteh).

This is where the user started the install process, so it's also a good place to finish, instead of the app's settings page which is where they redirected from before they got to AutoKitteh's OAuth handler.

daabr:
> Tested functionality using ngrok for local development

Please explain how you tested this successfully, but the code is using a wrong URL.

> The redirect currently points to `/settings/installations/{installation_id}`. We could alternatively redirect to `/apps/installations/{installation_id}` to maintain marketplace context.

1. Marketplace URLs are not `https://github.com/apps/installations/{installation_id}`, they're `https://github.com/marketplace/{app_name}`; for example: https://github.com/marketplace/autokitteh
2. Where did you see `/apps/installations/{installation_id}` URLs?
3. This PR should support GitHub Apps and Marketplace installations alike (neither of them has a state param), but it's safer to redirect to the GitHub Apps URL: all apps have an Apps link, but not all apps have a Marketplace listing
4. See my comment in the PR about the desired URL (`/apps/{app_name}`)

mario99logic:
Tested the GitHub installation redirect logic by:

1. Creating a test GitHub App (non-Marketplace)
2. Installing the app from settings and confirming it correctly redirects to /settings/installations/{installation_id}
3. This matches the redirect URL used in the redirectToGitHub function, validating that our implementation handles app installations properly and redirects to the URL in the function when there is no state parameter

internal/backend/oauth/webhook.go
```diff
@@ -156,6 +161,30 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
...
+// redirectToGitHub redirects users to the GitHub App installation page when they
+// initiate OAuth without a state parameter, typically after installing from the
+// GitHub Marketplace. This ensures a smoother user experience.
+func (s *svc) redirectToGitHub(w http.ResponseWriter, r *http.Request, intg string, state string) bool {
+	if intg != "github" || state != "" {
+		return false
+	}
+
+	githubAppName := os.Getenv("GITHUB_APP_NAME")
+	if githubAppName == "" {
```
daabr: FYI - I started writing a comment that there's no need to guess the name if the env var is missing, because according to [this](https://docs.autokitteh.com/integrations/github/config#environment-variables) it's required, but actually the point of this PR is to prefer a nice UX in case of an error, so this is good. 👍

internal/backend/oauth/webhook.go
```diff
@@ -156,6 +161,30 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
...
+		githubAppName = "autokitteh"
+	}
+
+	enterpriseURL := os.Getenv("GITHUB_ENTERPRISE_URL")
+	if enterpriseURL == "" {
+		enterpriseURL = "https://github.com"
+	}
+
+	redirectURL := fmt.Sprintf("%s/apps/%s", enterpriseURL, githubAppName)
```
daabr: Use https://pkg.go.dev/net/url#JoinPath instead of `Sprintf` - it requires an error check (which can happen only if the URL in the env var is invalid) but it's safer, because you don't have to assume/check if the URL ends with "/" or not

internal/backend/oauth/webhook.go
```diff
@@ -156,6 +161,30 @@ func (s *svc) exchange(w http.ResponseWriter, r *http.Request) {
...
 
+// redirectToGitHub redirects users to the GitHub App installation page when they
+// initiate OAuth without a state parameter, typically after installing from the
+// GitHub Marketplace. This ensures a smoother user experience.
+func (s *svc) redirectToGitHub(w http.ResponseWriter, r *http.Request, intg string, state string) bool {
+	if intg != "github" || state != "" {
+		return false
+	}
+
+	githubAppName := os.Getenv("GITHUB_APP_NAME")
```
daabr: Keep var names short: `github` is obvious and redundant in a function called `redirectToGitHub`. `appName` is clear enough.

daabr (approved):
Approved with 2 small comments

PR #1058, opened 2025-02-07
feat(py-sdk): oauth2 requests session for height and zoom

itayd:


runtimes/pythonrt/py-sdk/autokitteh/height.py
```diff
@@ -0,0 +1,24 @@
+"""Initialize an Height client, based on an AutoKitteh connection."""
```
daabr:
```suggestion
"""Initialize a Height client, based on an AutoKitteh connection."""
```

runtimes/pythonrt/py-sdk/autokitteh/oauth2_session.py
```diff
@@ -0,0 +1,58 @@
...
+from os import getenv
+
+from .connections import refresh_oauth
+from .connections import check_connection_name
+from .errors import ConnectionInitError
+
+
+class OAuth2Session(requests.Session):
+    """
+    Encapsulates arequests session, based on an AutoKitteh connection.
```
daabr:
```suggestion
    """Encapsulates a requests session, based on an AutoKitteh connection.

```

runtimes/pythonrt/py-sdk/autokitteh/oauth2_session.py
```diff
@@ -0,0 +1,58 @@
+import requests
+from os import getenv
+
+from .connections import refresh_oauth
+from .connections import check_connection_name
```
daabr, line 5: FYI, this won't pass Ruff's formatting, but we're not enforcing that in the autokitteh repo yet, so no need to fix it now.

daabr (approved):
Approved with a few cosmetic nits. Your idea that we discussed earlier feels better - a generic OAuth request initializer based on AK connections, instead of a "client", but this is good and needed now anyway.

runtimes/pythonrt/py-sdk/autokitteh/height.py
itayd: done

runtimes/pythonrt/py-sdk/autokitteh/oauth2_session.py
itayd: done

runtimes/pythonrt/py-sdk/autokitteh/oauth2_session.py
itayd: this is a result of ruff...

PR #1059, opened 2025-02-09
feat(zoom): support server-to-server auth

daabr:
Tested only the saving of credentials, not their usage - that should be done in INT-247.

Refs: INT-241

### DIFF
--- a/integrations/zoom/vars.go
+++ b/integrations/zoom/vars.go
@@ -27,8 +27,11 @@ func newOAuthData(t *oauth2.Token) oauthData {
 	}
 }
 
-// privateOAuth contains the user-provided details of a private Zoom OAuth 2.0 app.
-type privateOAuth struct {
+// privateApp contains the user-provided details of a private
+// Zoom OAuth 2.0 app or Service-to-Service internal app.
+type privateApp struct {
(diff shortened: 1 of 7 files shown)


integrations/zoom/vars.go
```diff
@@ -27,8 +27,11 @@ func newOAuthData(t *oauth2.Token) oauthData {
 	}
 }
 
-// privateOAuth contains the user-provided details of a private Zoom OAuth 2.0 app.
-type privateOAuth struct {
+// privateApp contains the user-provided details of a private
+// Zoom OAuth 2.0 app or Service-to-Service internal app.
+type privateApp struct {
```
mario99logic, line 32: Ive seen in the Linear integration for example the name of the struct is `privateOAuth`, is there a reason why name of the struct is `privateApp`

mario99logic (approved):

integrations/zoom/vars.go
daabr: Good question. Some integrations use the same configuration details for OAuth and non-OAuth authentication modes - I'm using "private app" in these case (examples: Zoom, MS Teams). If it's OAuth vs. something completely different (e.g. API key, JSON) then I'm using "private OAuth".

PR #1060, opened 2025-02-10
fix(integrations): don't prepend "connect" to legacy webhooks

daabr:
Regression caused by #1055.

Solution: don't use `LegacyDescriptor` (which added `/connect/` after the integration's URL path prefix, and move all relevant static web content accordingly (`xxx/connect/*` --> `xxx/*`).

Tested: Google Sheets, AWS, Asana, Slack.

Also removed the obsolete "HTTP" connection UI - replaced by webhook triggers.

Refs: INT-257


integrations/google/server.go
```diff
@@ -26,28 +27,26 @@ const (
...
-	uiPath := "GET " + desc.ConnectionURL().Path + "/"
-
-	// Connection UI.
-	m.NoAuth.Handle(uiPath, http.FileServer(http.FS(static.GoogleWebContent)))
+	common.ServeStaticUI(m, desc, static.GoogleWebContent)
+	uiPath := fmt.Sprintf("GET %s/", desc.ConnectionURL().Path)
 
 	urlPath := strings.ReplaceAll(uiPath, "google", "gmail")
-	m.NoAuth.Handle(urlPath, http.FileServer(http.FS(static.GmailWebContent)))
+	m.Auth.Handle(urlPath, http.FileServer(http.FS(static.GmailWebContent)))
```
pashafateev, line 34: why did this change to `Auth`?

daabr: See the rationale in https://github.com/autokitteh/autokitteh/blob/main/integrations/common/server.go#L12-L15

pashafateev (approved):

PR #1062, opened 2025-02-11
feat(sessions): get history from temporal, not from db

itayd:
motivation:
- do not do extra activities on any single activity, saves history length and size.
- be more aligned with temporal to ease future handling of concurrency and retries.

changes:
- instead of getting workflow history from the db, get it directly from temporal, translating its history into autokitteh history.
- prints go to the db, and are reported only on GetPrints
- print now can report any arbitrary value, but currently implemented only string

notes:
- this will only show actual temporal activities in the history for calls, so next_event, sleep etc would not show. 
- for migration purposes (deploying while workflows are running), old workflows would still be persisting data to the log via activities, but all new ones will not.


internal/backend/sessions/sessionworkflows/history.go
```diff
@@ -0,0 +1,385 @@
...
+
+			var v sdktypes.Value
+			if err := v.UnmarshalJSON(payloads[1].GetData()); err != nil {
+				return sdktypes.InvalidSessionLogRecord, temporalclient.TranslateError(err, "unmarshal print text")
+			}
+
+			return sdktypes.NewPrintSessionLogRecord(t, v, 0), nil
+
+		default:
+			return sdktypes.InvalidSessionLogRecord, nil
```
daabr, line 275: Log debugging data to warn us about the existence of this unexpected activity type

internal/backend/sessions/sessionworkflows/history.go
```diff
@@ -0,0 +1,385 @@
...
+			),
+		), nil
+
+	case *historypb.HistoryEvent_ActivityTaskFailedEventAttributes, *historypb.HistoryEvent_ActivityTaskTimedOutEventAttributes:
+		// do not report this, as it is always for infrastructure failures.
+		// we retry these indefinitely, so they are not interesting.
+		return sdktypes.InvalidSessionLogRecord, nil
+
+	default:
+		return sdktypes.InvalidSessionLogRecord, nil
```
daabr, line 367: Log debugging data to warn us about the existence of this unexpected attribute type

daabr (approved):
Approved with one minor comment X2

internal/backend/sessions/sessionworkflows/history.go
itayd: no need - this is not unexpected,and not an error, just something we don't care about (yet?)

internal/backend/sessions/sessionworkflows/history.go
itayd: no need - this is not unexpected,and not an error, just something we don't care about (yet?)

PR #1063, opened 2025-02-11
ci: go tests and session tests on big instances

itayd:


tebeka:
What prompted this change?

PR #1068, opened 2025-02-12
DRAFT: feat: notification service proto

itayd:
will be used to send messages to orgs and users. this is a proposal, and we'll discuss.


proto/autokitteh/notifications/v1/notification.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package autokitteh.notifications.v1;
+
+message Notification {
+  string id = 1;
+  string recipient_id = 2; // org or user id.
+  string type = 3;
+  map<string, string> data = 4;
```
tebeka, line 9: Why a map and not a string?

proto/autokitteh/notifications/v1/svc.proto
```diff
@@ -0,0 +1,38 @@
...
+
+message SendRequest {
+  Notification notification = 1 [(buf.validate.field).required = true];
+}
+
+message SendResponse {
+  string notification_id = 1 [(buf.validate.field).string.min_len = 1];
+}
+
+message ListRequest {
```
tebeka, line 16:
```suggestion
message ListRequests {
```

proto/autokitteh/notifications/v1/notification.proto
```diff
@@ -0,0 +1,10 @@
+syntax = "proto3";
+
+package autokitteh.notifications.v1;
+
+message Notification {
+  string id = 1;
+  string recipient_id = 2; // org or user id.
+  string type = 3;
```
tebeka, line 8: What are the types? Why do we need that?

proto/autokitteh/notifications/v1/svc.proto
```diff
@@ -0,0 +1,38 @@
...
+  string recipient_id = 1 [(buf.validate.field).string.min_len = 1]; // org or user id.
+  string type = 2; // optional, if provided, only notifications of this type will be returned.
```
tebeka, line 18: Why do we need a filter? Can't the UI do it?
IMO only read/unread filter, it at all.

PR #1072, opened 2025-02-12
fix(python): A better way to pickle errors.

tebeka:
We try to pickle and un-pickle error, if this doesn't work we pickle the result of __reduce__ and re-create the error from it.
    
Fixes: ENG-1996


runtimes/pythonrt/runner/main.py
```diff
@@ -140,6 +140,31 @@ def set_exception_args(err):
 Result = namedtuple("Result", "value error traceback")
 
 
+def is_pickleable(err):
```
efiShtain, line 144: I guess who ever calls this would try to pickle the object later

maybe try_pickle(...) which returns the pickled object if it succeeds or None otherwise is more optimistic ?

efiShtain (approved):

runtimes/pythonrt/runner/main.py
```diff
@@ -305,7 +330,12 @@ def ActivityReply(
             )
 
         if result.error:
-            call.fut.set_exception(result.error)
+            try:
+                error = restore_error(result.error)
+            except (TypeError, ValueError) as err:
+                log.exception("can't restore error: %r", err)
+                error = Exception(repr(result.error))
```
itayd: AutokittehException or some such

runtimes/pythonrt/runner/main.py
tebeka: I pickle a `Result` that contains the error, so it won't help.

runtimes/pythonrt/runner/main.py
tebeka: Done.

PR #1075, opened 2025-02-13
fix(python): Don't patch module level calls

tebeka:
Don't patch module level calls (such as creating namedtuple). Changed the AST transformer to keep track of indentation level and work top-down.

Fixes: ENG-1994


runtimes/pythonrt/runner/loader.py
```diff
@@ -23,14 +23,33 @@ class Transformer(ast.NodeTransformer):
     def __init__(self, file_name, src):
         self.file_name = file_name
         self.code_lines = src.splitlines()
+        # Indent of class/function definition. -1 means top level code
+        self.fn_indent = -1
+
+    def visit(self, node):
+        # Visit AST nodes. We keep track of functions and indent, in order not to patch
+        # module level calls.
```
efiShtain, line 31: can you give some examples of module level calls ? 
not sure I understand what it means

tebeka: See the test:

```python
from collections import namedtuple

User = namedtuple('User', 'login groups')  # <- Module level call
```

Or do you mean to add this explanation in the doc?

efiShtain: why is named tuple module level call ?
what does it mean ? how is that different than any other function call ?

itayd: @tebeka it's not clear to me as well - if we get a namedtuple as a result from an activity, will that break anything?

tebeka: Currently, we patch *every* call, including top level calls like the `namedtuple` - this is causing issues with namedtuples and makes the `User` looks like it's coming from the `call` module and not the user module - which is why every time we call `User('elliot', ['wheel', 'elliot'])` is in an activity.

This PR changes, so we *don't* patch top level calls at all.

daabr: I think that's Efi's question: some module-level function calls may need to be activities. It's not reasonable to assume that module-level functions calls never fail.

itayd (approved):

runtimes/pythonrt/runner/loader.py
tebeka: We do not support activities at module load time.

PR #1076, opened 2025-02-13
feat: session start and dispatch start to use same inputs format

itayd:
this makes session run from ui work by just copying and pasting the inputs.

### DIFF
--- a/internal/backend/dispatcher/workflow.go
+++ b/internal/backend/dispatcher/workflow.go
@@ -3,7 +3,6 @@ package dispatcher
 import (
 	"errors"
 	"fmt"
-	"maps"
 
 	"github.com/google/uuid"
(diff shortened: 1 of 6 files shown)


internal/backend/dispatcher/workflow.go
```diff
@@ -3,7 +3,6 @@ package dispatcher
 import (
 	"errors"
 	"fmt"
-	"maps"
```
tebeka, line 6: GitHub says the variables below are unused.

tebeka (approved):
See comments from GH about unused variables.

internal/backend/dispatcher/workflow.go
itayd: huh i don't see it in my repo, probably not most up to date

PR #1078, opened 2025-02-13
feat: Add soft delete for org and block org deletion if it has associated projects

mario99logic:
Added soft deletion functionality for organizations and implements safeguards around org deletion:

Key changes:
- Added `deleted_at` column to organizations table for soft delete 
- Prevents org deletion if it has associated projects
- Added row-level locking during project creation to prevent race conditions
- Renames org to its ID before soft deletion

Technical details:
- Uses row-level `FOR UPDATE` lock when creating projects to prevent race conditions during org deletion
- Added validation check `doesOrgHasProject` to ensure orgs with projects cannot be deleted
- Migration added for `deleted_at` column

Testing:
- tried deleting an org with associated projects
- deleting an org with no projects
- checking the name of the org and deleted_at column in postgres after deletion


internal/backend/db/dbgorm/orgs.go
```diff
@@ -45,7 +45,22 @@ func (gdb *gormdb) DeleteOrg(ctx context.Context, oid sdktypes.OrgID) error {
 	}
 
 	return translateError(gdb.transaction(ctx, func(tx *tx) error {
-		err := tx.db.Where("org_id = ?", oid.UUIDValue()).Delete(&scheme.OrgMember{}).Error
+		hasProject, err := gdb.doesOrgHasProject(ctx, oid)
```
efiShtain: you need to lock the organization for update here as well

so no one can create projects on it if you are in the middle of deleting it

efiShtain: you can simplify this change using this psuedo sql code
```sql
-- start transaction
begin 
-- lock org row
select org_id from organizations where org_id = '' for update
-- if no org - return error not found

-- assume org is found, try setting deleted at
-- only 
update orgs set 
deleted_at = NOW() from projects 
where not exists (
    select 1 from projects p 
    where p.org_id = orgs.org_id and p.deleted_at is null
);


-- if deleted delete org members
commit
```



this will lock the org row first and will set the org to deleted only if there are no projects. you can check the number of updated rows in the result, if nothing is updated - there are active projects, if one row is updated, the organization is now deleted.

if it succeeded, delete the org members as well


In my opinion this is better, 
if you think it is more complicated, just add the row lock at the start and stay with current code

mario99logic: Quick question about the row locking - since we're using a single atomic SQL statement to check projects and update the org, do we really need the FOR UPDATE lock? It seems redundant but I might be missing something.

internal/backend/db/dbgorm/orgs.go
```diff
@@ -45,12 +46,36 @@ func (gdb *gormdb) DeleteOrg(ctx context.Context, oid sdktypes.OrgID) error {
...
 	return translateError(gdb.transaction(ctx, func(tx *tx) error {
-		err := tx.db.Where("org_id = ?", oid.UUIDValue()).Delete(&scheme.OrgMember{}).Error
+		var org scheme.Org
+		err := tx.db.Set("gorm:for_update", true).
+			Model(&scheme.Org{}).
+			Where("org_id = ?", oid.UUIDValue()).
+			First(&org).Error
+		if err != nil {
+			return translateError(err)
+		}
```
efiShtain: if think, if you'll check here if the org is not deleted it makes the code more readable
update where org_id = & deleted_at is not null

that way if nothing returned, just return not found

the next update query just reference the org_id. since you lock for update, no one can change it.

what do you think?

efiShtain (approved):

efiShtain (changes requested):
use gorm standard locking please 

// Basic FOR UPDATE lock
db.Clauses(clause.Locking{Strength: "UPDATE"}).Find(&users)
// SQL: SELECT * FROM `users` FOR UPDATE

internal/backend/db/dbgorm/orgs.go
efiShtain: is you mea this sql it is not atomic as a single instruction, it is atomic if it runs in a transaction but stuff can happen between the subquery and the outer query

```sql
update orgs set 
deleted_at = NOW() from projects 
where not exists (
    select 1 from projects p 
    where p.org_id = orgs.org_id and p.deleted_at is null
);
```

internal/backend/db/dbgorm/orgs.go
mario99logic: Agree, if its not Null from the beginning its irrelevant.

internal/backend/db/dbgorm/orgs.go
```diff
@@ -45,7 +47,38 @@ func (gdb *gormdb) DeleteOrg(ctx context.Context, oid sdktypes.OrgID) error {
...
+		if result.RowsAffected == 0 {
+			return errors.New("cannot delete an organization that has associated projects")
+		}
+
+		err = tx.writer.Where("org_id = ?", oid.UUIDValue()).Delete(&scheme.OrgMember{}).Error
+		if err != nil {
+			return translateError(err)
+		}
+
+		err = tx.writer.Model(&scheme.Org{}).
```
efiShtain: you already to it in row 62 no need to do it again

efiShtain (approved):

PR #1081, opened 2025-02-14
refactor(salesforce): retrieve `instance_url` from token exchange response instead of user input

pashafateev:
**testing**: successfully authenticated salesforce and ran a simple workflow using our python sdk

### DIFF
--- a/internal/backend/oauth/oauth.go
+++ b/internal/backend/oauth/oauth.go
@@ -758,6 +758,22 @@ func (o *oauth) Exchange(ctx context.Context, integration string, cid sdktypes.C
 	if err != nil {
 		return nil, fmt.Errorf("failed to exchange code: %w", err)
 	}
+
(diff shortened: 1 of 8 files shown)


internal/backend/oauth/oauth.go
```diff
@@ -758,6 +758,22 @@ func (o *oauth) Exchange(ctx context.Context, integration string, cid sdktypes.C
 	if err != nil {
 		return nil, fmt.Errorf("failed to exchange code: %w", err)
 	}
+
+	if integration == "salesforce" {
```
daabr: This belongs in salesforce's own `oauth.go` - see line 57 there

pashafateev: Moved to webhook.go and url.go

daabr (approved):

PR #1083, opened 2025-02-17
fix(history): properly decode activity data

itayd:
- use temporal's data convertor instead of naively using json. this takes care of proper codec/encryption.
- get rid of LazyTemporalClient, just use the temporalclient.Client.TemporalClient instead. This is a mouthful and I hate it, if you have other naming suggestions, feel free and I'll take care of it in a future change.

### DIFF
--- a/internal/backend/cron/cron.go
+++ b/internal/backend/cron/cron.go
@@ -81,7 +81,7 @@ func (cr *Cron) Start(ctx context.Context, c sdkservices.Connections, v sdkservi
 	cr.oauth = o
 
 	// Configure a worker for the internal maintenance workflow.
-	w := temporalclient.NewWorker(cr.logger, cr.temporal.Temporal(), taskQueueName, cr.cfg.Worker)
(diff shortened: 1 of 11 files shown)


daabr (approved):
1. Re your PR description: the common Go rule of thumb is to call a client in the `foo` package `foo.Client` and not `foo.FooClient` - see https://google.github.io/styleguide/go/decisions.html#repetition
2. Why do you even need an interface here? And if you do, don't call it `Client`
3. Unnecessary indentation: a `type` block with just a single definition

itayd:
> 1. Re your PR description: the common Go rule of thumb is to call a client in the `foo` package `foo.Client` and not `foo.FooClient` - see https://google.github.io/styleguide/go/decisions.html#repetition
> 2. Why do you even need an interface here? And if you do, don't call it `Client`
> 3. Unnecessary indentation: a `type` block with just a single definition

fixed 3. 1+2 will be done in another pr. https://linear.app/autokitteh/issue/ENG-2011/refactor-backendtemporalclient.

PR #1085, opened 2025-02-18
fix(sessions): empty created_at

itayd:
currently when views by ui, created_at appears empty. it turns out there was a double created_at in sessions. there is no need on the created_at, and we already sort by the session_id which is time ordered.

from https://uuid7.com/:
> Natural Sorting: Traditional databases often require additional timestamp columns to sort records based on creation time. With UUIDv7, you can achieve this sorting using the UUID itself, eliminating the need for extra columns.


itayd:
feel free to merge on approval, will not be in today's sync

RonenMars:
@itayd  does it mean that we won't have the start time on the list?

efiShtain:
This PR fixes the issue and keep the created_at that we need

https://github.com/autokitteh/autokitteh/pull/1086

Closing this PR

PR #1093, opened 2025-02-20
chore(slack): migrate old connections

daabr:
Follow-up for INT-264.

Refs: INT-275

### DIFF
--- a/integrations/common/vars.go
+++ b/integrations/common/vars.go
@@ -67,8 +67,8 @@ func SaveAuthType(r *http.Request, vars sdkservices.Vars, vsid sdktypes.VarScope
 	return authType
 }
 
-// RenameVar renames a variable in the given connection scope.
-// This is useful for non-trivial data migrations.
+// RenameVar renames a variable in the given connection scope. It does nothing if
+// the variable doesn't already exist. This is useful for non-trivial data migrations.
 func RenameVar(ctx context.Context, v sdkservices.Vars, vsid sdktypes.VarScopeID, old, new sdktypes.Symbol) error {
 	vs, err := v.Get(ctx, vsid, old)
 	if err != nil {
--- a/integrations/common/vars.go
+++ b/integrations/common/vars.go
@@ -66,3 +66,24 @@ func SaveAuthType(r *http.Request, vars sdkservices.Vars, vsid sdktypes.VarScope
 	_ = vars.Set(r.Context(), v)
 	return authType
 }
+
+// RenameVar renames a variable in the given connection scope.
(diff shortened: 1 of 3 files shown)


integrations/common/vars.go
```diff
@@ -66,3 +66,24 @@ func SaveAuthType(r *http.Request, vars sdkservices.Vars, vsid sdktypes.VarScope
...
+
+	o := vs.Get(old)
+	if !o.IsValid() {
+		return nil
+	}
+
+	n := sdktypes.NewVar(new).SetValue(o.Value()).SetSecret(o.IsSecret())
+	if err := v.Set(ctx, n.WithScopeID(vsid)); err != nil {
+		return err
+	}
```
pashafateev: if there's no error, it's guaranteed that the new var has been saved?

integrations/slack/integration.go
```diff
@@ -95,3 +98,38 @@ func reopenExistingWebSocketConnections(ctx context.Context, l *zap.Logger, v sd
 		h.OpenWebSocketConnection(appID, appToken, botToken)
 	}
 }
+
+// TODO(INT-167): Remove this after all cloud envs have v0.14.4+ deployed.
+func migrateOldConnectionVars(l *zap.Logger, v sdkservices.Vars) {
+	ctx := context.Background()
+	iid := sdktypes.NewIntegrationIDFromName(desc.UniqueName().String())
+	cids, err := v.FindConnectionIDs(ctx, iid, common.AuthTypeVar, "")
```
pashafateev, line 106: why empty string? Would this mean of any `auth_type`?

integrations/common/vars.go
daabr: yes

integrations/slack/integration.go
daabr: exactly, if you specify a non-empty string it uses it as a filter, otherwise it just checks that the var exists, and in this is the one var that always exists (otherwise there's nothing to migrate anyway).

pashafateev (approved):

efiShtain:
how do you make sure all old connections were migrated ? 

do we really need this migration in code ? 
can't we do it offline ? assuming current version is backward compatible

daabr:
> how do you make sure all old connections were migrated ?

This code runs when the server starts, and covers all existing Slack connections.
Once we upgrade all environments to run this code, I'll remove it.

> do we really need this migration in code ? can't we do it offline ? assuming current version is backward compatible

I would have preferred to use a regular data migration. Unfortunately this requires more logic than I'm comfortable putting in SQL queries.

PR #1094, opened 2025-02-20
refactor(google drive): differentiate between new connection and existing for watch renewal

pashafateev:
**Testing**: Manually tested by running a Google Drive project. Created a new file using a workflow, which initiated a new Drive watch. Set the cron job to run every minute and configured the expiration to 24 hours to ensure renewal was triggered each time.

The behavior was as expected—no errors related to an existing channel were triggered.

### DIFF
--- a/integrations/google/drive/client.go
+++ b/integrations/google/drive/client.go
@@ -154,11 +154,10 @@ func (a api) watchEvents(ctx context.Context, connID sdktypes.ConnectionID, user
 	addr := os.Getenv("WEBHOOK_ADDRESS")
 	watchID := connID.String() + "/events"
 	req := client.Changes.Watch(startToken.StartPageToken, &drive.Channel{
-		Id:      watchID,
-		Token:   userEmail + "/" + watchID,
-		Address: fmt.Sprintf("https://%s/googledrive/notif", addr),
-		Type:    "web_hook",
-		// Expiration: time.Now().Add(time.Hour*24*7).Unix() * 1000,
+		Id:         watchID,
+		Token:      userEmail + "/" + watchID,
+		Address:    fmt.Sprintf("https://%s/googledrive/notif", addr),
+		Type:       "web_hook",
 		Expiration: time.Now().Add(time.Hour*24).Unix() * 1000,
 	})
 
--- a/integrations/google/drive/client.go
+++ b/integrations/google/drive/client.go
@@ -10,7 +10,6 @@ import (
 	"golang.org/x/oauth2"
 	"golang.org/x/oauth2/google"
 	"google.golang.org/api/drive/v3"
-	"google.golang.org/api/googleapi"
 	googleoauth2 "google.golang.org/api/oauth2/v2"
 	"google.golang.org/api/option"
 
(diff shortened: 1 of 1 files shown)


integrations/google/drive/client.go
```diff
@@ -152,39 +152,39 @@ func (a api) watchEvents(ctx context.Context, connID sdktypes.ConnectionID, user
...
-		return resp, nil
-	}
-
-	gerr, ok := err.(*googleapi.Error)
-	a.logger.Warn("Google Drive watch channel creation error", zap.Any("googleApiError", gerr))
-	if !ok || gerr.Code != 400 || len(gerr.Errors) != 1 {
-		return nil, err
-	}
-	if gerr.Errors[0].Reason != "channelIdNotUnique" {
+	// check if the channel already exists
```
daabr: 1. Capitalize and put punctuate full-sentence comments
2. It seems this block (lines 164-179) belongs before the definition of `req`, in other words `req` is defined in line 157 but used only in line 181

integrations/google/drive/client.go
```diff
@@ -152,39 +152,39 @@ func (a api) watchEvents(ctx context.Context, connID sdktypes.ConnectionID, user
...
+	if err != nil {
 		return nil, err
 	}
-
-	// If the channel already exists, stop and recreate it.
-	a.logger.Info("Google Drive watch channel already exists - stopping and recreating")
-	if err := a.stopWatch(ctx, connID); err != nil {
-		return nil, fmt.Errorf("stop existing watch channel: %w", err)
+	wid := vs.Get(vars.DriveEventsWatchID).Value()
+	if wid != "" {
```
daabr: 1. Use `GetValue()` instead of `Get().Value()`
2. The var assignment and condition are both short, combine them into a single line

integrations/google/drive/client.go
```diff
@@ -152,39 +152,39 @@ func (a api) watchEvents(ctx context.Context, connID sdktypes.ConnectionID, user
...
+		// sanity check
+		if watchID != wid {
```
daabr, line 161: Given that the watchID is based on the connection ID which can't change, what's the point of this sanity check?

integrations/google/drive/client.go
```diff
@@ -152,39 +152,39 @@ func (a api) watchEvents(ctx context.Context, connID sdktypes.ConnectionID, user
...
+		if err := a.stopWatch(ctx, connID); err != nil {
+			return nil, fmt.Errorf("stop existing watch channel: %w", err)
+		}
 	}
 
-	resp, err = req.Do()
+	resp, err := req.Do()
 	if err != nil {
-		return nil, fmt.Errorf("recreate watch channel: %w", err)
+		gerr, _ := err.(*googleapi.Error)
```
daabr: What's the point of this type cast?

pashafateev: use zap.Error and remove 183

daabr (approved):

integrations/google/drive/client.go
pashafateev: Done

integrations/google/drive/client.go
pashafateev: Done

integrations/google/drive/client.go
pashafateev: Done

PR #1095, opened 2025-02-20
DRAFT: bug(python): test for relative imports

itayd:
add a test to reproduce relative import error

Ref: ENG-2022


tebeka:
That's how Python works, not an issue with ak :)

PR #1096, opened 2025-02-20
feat(session-test): test that sessions are invoking specific actions

itayd:
if a `calls.txt` is present in the txtar, compare a list of run call names to given list

also:
- speed up polling for session logs by decreasing the poll interval. should speed up ci.
- allow a space in code location string, this allows vscode to color better txtar stuff

PR #1097, opened 2025-02-20
feat: prevent some sdk functions from running as activities

itayd:
no need to run any of the sdk client init functions or the package install functions as activities, even if called in non-global scope.

this introduces a `@_noactivity` decorator for internal SDK use to achieve that.

based on https://github.com/autokitteh/autokitteh/pull/1096.

### DIFF
--- a/runtimes/pythonrt/py-sdk/autokitteh/decorators.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/decorators.py
@@ -17,3 +17,14 @@ def activity(fn: callable) -> callable:
     """
     setattr(fn, ACTIVITY_ATTR, True)
     return fn
+
+
+def _noactivity(fn: callable) -> callable:
+    """Decorator to mark a function as not a Temporal activity.
+
+    For SDK internal use only - as this does not prevent from a function
+    called from outside the sdk to run as an activity or starting other activities.
+    """
+
+    setattr(fn, ACTIVITY_ATTR, False)
+    return fn
--- a/runtimes/pythonrt/py-sdk/autokitteh/sdktest.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/sdktest.py
@@ -1,0 +1,28 @@
+from random import random
+
+from .decorators import activity, noactivity
+
+
+def test_default(*args, **kwargs):
+    return kwargs.get("ret")
+
+
+@activity
+def test_activity(*args, **kwargs):
+    return kwargs.get("ret")
+
+
+@noactivity
+def test_noactivity(*args, **kwargs):
+    return kwargs.get("ret")
+
+
+@noactivity
+def test_noactivity_internal_call(*args, **kwargs):
+    return test_default(*args, **kwargs)
+
+
+@noactivity
+def test_noactivity_external_call(*args, **kwargs):
+    _ = random()
+    return kwargs.get("ret")
--- a/runtimes/pythonrt/runner/call.py
+++ b/runtimes/pythonrt/runner/call.py
@@ -19,9 +19,18 @@ AK_FUNCS = {
 }
 
 
-def is_marked_activity(fn):
-    """Return true if function is marked as an activity."""
-    return getattr(fn, decorators.ACTIVITY_ATTR, False)
+def get_activity_attr(fn):
+    """Return true if function is marked as an activity.
+
+    Returns:
+    None - not explicitly set.
+    True - explicitly set as activity.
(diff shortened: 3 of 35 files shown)


tebeka:
In most cases, these functions are called at module level and then they're not activities.
In other cases, when they are called inside a function, IMO they should be an activity.

If we do go that path, then I prefer the `AKCall` will detect at runtime that these function come from the `autokitteh` module and not run them as activities.

runtimes/pythonrt/py-sdk/autokitteh/decorators.py
```diff
@@ -17,3 +17,14 @@ def activity(fn: callable) -> callable:
     """
     setattr(fn, ACTIVITY_ATTR, True)
     return fn
+
+
+def _noactivity(fn: callable) -> callable:
```
tebeka, line 22: I prefer not to place this in `decorators.py` but some internal module that start with an `_` (say `_activity.py`), and then the name can start without a `_`.

runtimes/pythonrt/py-sdk/autokitteh/sdktest.py
tebeka, line 1: Tests to in the `tests` directory. Or if we want then next to code then name them `decorators_test.py` so it'll be next to `decorators.py` when you look at the directory.

runtimes/pythonrt/py-sdk/autokitteh/sdktest.py
```diff
@@ -0,0 +1,28 @@
+from random import random
+
+from .decorators import activity, noactivity
```
tebeka, line 3:
```suggestion
from .decorators import activity, _noactivity
```

runtimes/pythonrt/py-sdk/autokitteh/sdktest.py
```diff
@@ -0,0 +1,28 @@
...
+
+
+def test_default(*args, **kwargs):
```
tebeka, line 6: What does that test?

runtimes/pythonrt/runner/call.py
```diff
@@ -19,9 +19,18 @@
...
+
+    Returns:
+    None - not explicitly set.
+    True - explicitly set as activity.
+    False - explicitly set as non-activity.
+    """
+    if hasattr(fn, decorators.ACTIVITY_ATTR):
+        return getattr(fn, decorators.ACTIVITY_ATTR)
+
+    return None
```
tebeka, line 33:
```suggestion
return getattr(fn, decorators.ACTIVITY_ATTR, None)
```

runtimes/pythonrt/runner/call.py
```diff
@@ -83,8 +92,9 @@ def should_run_as_activity(self, fn):
         if self.in_activity or self.loading:
             return False
 
-        if is_marked_activity(fn):
-            return True
+        a = get_activity_attr(fn)
+        if a is not None:
+            return a
```
tebeka, line 97:
```suggestion
        if (mark := get_activity_attr(fn)) is not None:
            return mark
```

tests/sessions/python_sdk_activity.txtar
```diff
@@ -0,0 +1,32 @@
+oink
+meow
+woof
+moo
+hiss
+
+-- main.py main --
+from random import random
+
+from autokitteh.decorators import noactivity
```
tebeka, line 10:
```suggestion
from autokitteh.decorators import _noactivity
```

tests/sessions/python_sdk_activity.txtar
```diff
@@ -0,0 +1,32 @@
...
+from autokitteh.sdktest import *
```
tebeka, line 11: Don't `import *` :)
Also, place `sdktest.py` in the txtar file and remove it from the py-sdk code.

tebeka (changes requested):

runtimes/pythonrt/py-sdk/autokitteh/sdktest.py
daabr: +1, per current convention rename to `something_test.py`, but more importantly I don't see any actual unit tests here, so remove this file, or move and implement it.

daabr (changes requested):
I strongly dislike private decorators: it's a self contradiction. But that's easy to fix beyond this PR.

The blocking problem is adding a fake test file, which is actually a harmful practice for unit testing, and for readability of a public library. If you **really** want it, create a module that has "fake" in its name and use that for testing purposes.

PR #1099, opened 2025-02-20
fix: skip flakey python tests

efiShtain:


runtimes/pythonrt/local_runner_test.go
```diff
@@ -60,6 +60,9 @@ func Test_createVEnv(t *testing.T) {
 var tarData []byte
 
 func TestRunner_Start(t *testing.T) {
+	// The test is flakey, so we skip it for now
```
tebeka, line 63: Please add the linear issue here

tebeka (approved):

runtimes/pythonrt/local_runner_test.go
efiShtain: done

PR #1100, opened 2025-02-20
feat(slack,github): improve handling of event signatures

daabr:
tl;dr - need more & better logging for signature verification errors.

Refactor `getCustomOAuthSigningSecret()` to handle all OAuth signing secrets. Ensure that all related methods pass `appID`, `enterpriseID`, `teamID` (in this order).

When reading connection vars from the DB for this, read only the signing secret, not all vars.

In case of signature verification errors, log these details: received signature, whether or not the signing secret is an empty string (server or connection config error), app ID, enterprise ID, team ID. None of these details is sensitive.

Cosmetic tweaks:

- New function for standard reporting of HTTP errors
- Changing log messages to be lower-cased, unlike comments - to standardize logging style across AK
- Rename `getCustomOAuthSigningSecret()` ("get" not idiomatic, "custom" is an old term)

Refs: INT-286

### DIFF
--- a/integrations/slack/webhooks/webhook.go
+++ b/integrations/slack/webhooks/webhook.go
@@ -111,7 +111,7 @@ func (h handler) checkRequest(w http.ResponseWriter, r *http.Request, l *zap.Log
 		return nil
 	}
 
-	signingSecret, aid, eid, tid, err := h.oauthSigningSecret(r.Context(), l, body, wantContentType)
+	signingSecret, info, err := h.oauthSigningSecret(r.Context(), l, body, wantContentType)
 	if err != nil {
 		l.Error("failed to get signing secret", zap.Error(err))
 		common.HTTPError(w, http.StatusInternalServerError)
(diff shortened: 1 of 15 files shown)


integrations/slack/webhooks/webhook.go
```diff
@@ -205,78 +199,84 @@ func (h handler) dispatchAsyncEventsToConnections(ctx context.Context, cids []sd
...
+	return p.APIAppID, p.Enterprise.ID, p.Team.ID, nil
 }
 
-// getCustomOAuthSigningSecret gets the signing secret from the connection's
-// variables for Custom OAuth.
-func (h handler) getCustomOAuthSigningSecret(ctx context.Context, body []byte, wantContentType string, l *zap.Logger) (string, error) {
-	appID, teamID, _, err := h.extractIDs(body, wantContentType, l)
+// oauthSigningSecret reads the signing secret from the connection's
+// variables, or uses the signing secret of the server's default Slack app.
+func (h handler) oauthSigningSecret(ctx context.Context, l *zap.Logger, body []byte, wantContentType string) (secret, appID, enterpriseID, teamID string, err error) {
```
pashafateev: struct instead of 5 return types?

pashafateev (approved):

integrations/slack/webhooks/webhook.go
daabr: good idea, done

PR #1101, opened 2025-02-20
fix: dispatch Jira events based on domain + webhook id

pashafateev:
### tl;dr:  
This update modifies how Jira events are dispatched by incorporating both the **Jira domain** and **webhook ID** when identifying connections.  

#### Key Changes:  
1. **Introduced `webhookKey`**:  
   - A new function `webhookKey(domain, webhookID string) string` creates a unique key by combining the Jira domain with the webhook ID.  

2. **Updated Webhook Key Handling**:  
   - Added a new `WebhookKeySymbol` in `vars.go` to represent this composite key.  

3. **Refactored Issue URL Handling**:  
   - Introduced `transformIssueURL` to normalize the Jira issue URL before constructing the webhook key.  

4. **Modified Event Dispatch Logic**:  
   - Instead of using `webhookID` alone, event dispatch now uses `webhookKey` for identifying connections (`FindConnectionIDs` now queries using `WebhookKeySymbol`).  

### Testing
Manually tested by initializing a new Jira connection and successfully dispatching events.


integrations/atlassian/jira/event.go
```diff
@@ -106,10 +107,17 @@ func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
 		return int(f)
 	})
 
+	u, err := transformIssueURL(jiraEvent, l)
+	if err != nil {
+		l.Error("Failed to transform issue URL", zap.Error(err))
```
daabr: 1. Not all Jira events relate to issues
2. It's easier to understand the initial meaning/context of this log message in Datadog when you specify it's related to Jira
```suggestion
		l.Error("Failed to transform Jira URL", zap.Error(err))
```

integrations/atlassian/jira/event.go
```diff
@@ -122,6 +130,30 @@ func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
 	// Returning immediately without an error = acknowledgement of receipt.
 }
 
+// transformIssueURL is used to transform the issue URL in the Jira event
+// to a normalized URL that can be used for the webhook key.
+func transformIssueURL(jiraEvent map[string]any, l *zap.Logger) (string, error) {
+	issue, ok := jiraEvent["issue"].(map[string]any)
```
daabr, line 137: What if the event isn't related to issues?

pashafateev: [These are the only supported events](https://developer.atlassian.com/cloud/jira/platform/webhooks/#registering-a-webhook-using-the-rest-api--for-connect-and-oauth-2-0-apps-
) for an OAuth 2.0 app. 

The request headers don't have the URL and the only option I see is using issue from the event

daabr (approved):

PR #1106, opened 2025-02-22
WIP: feat(temporalclient): support for large payloads in data converter

itayd:
currently supporting inmem, file and s3 using gokv.

note: we'll have to set prod env to specify s3 bucket.

TODO: we need to figure out when to cleanup stuff. For that best if we associate a session id and/or timestamps to each blob. Will figure out in the future.


internal/backend/temporalclient/config.go
```diff
@@ -77,7 +86,13 @@ var (
 			EnableHelperRedirect:      true,
 		},
 		Test: &Config{
-			Monitor:              defaultMonitorConfig,
+			Monitor: defaultMonitorConfig,
+			DataConverter: DataConverterConfig{
+				LargePayload: LargePayloadConfig{
+					ThresholdSize: 2 * 1024 * 1024, // 2MB
```
daabr: Don't we want a lower threshold for testing? Not a blocker, just wondering.

internal/backend/temporalclient/dataconverter.go
```diff
@@ -84,7 +86,24 @@ func NewDataConverter(l *zap.Logger, cfg *DataConverterConfig, parent converter.
 		codecs = append(codecs, converter.NewZlibCodec(converter.ZlibCodecOptions{AlwaysEncode: true}))
 	}
 
-	return converter.NewCodecDataConverter(parent, codecs...), nil
+	loCodec, err := newLargePayloadCodec(l.Named("lo"), cfg.LargePayload)
```
daabr: Nit: "lo" is a too-short acronym that can mean anything. "lob" or "blob" are common abbreviations for "[binary] large object".

internal/backend/temporalclient/config.go
```diff
@@ -62,10 +62,19 @@ var (
 				Encryption: DataConverterEncryptionConfig{
 					Encrypt: true,
 				},
+				LargePayload: LargePayloadConfig{
```
daabr, line 65: Optional: until we have D3 configured, why not specify in prod one of the other storage options, temporarily, just in case?

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,178 @@
+package temporalclient
+
+import (
+	"errors"
+	"maps"
+	"path/filepath"
+	"strconv"
+
+	"github.com/philippgille/gokv"
```
daabr: A KV abstraction is an unnecessary layer for this feature, it just adds complication and risk.

All we need is S3, and in the distant future we'll see if/how we need to support other clouds based on business needs.

Moreover, this specific package doesn't seem to be actively maintained, or even that popular. Even worse, it's so old that it uses insecure crypto deps (see unreviewed bot PRs), and an ancient AWS package instead of v2 which we already depend on.

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,178 @@
...
+
+type LargePayloadConfig struct {
+	StoreType string `koanf:"store_type"` // "none" to disable.
+
+	// ThresholdSize is the maximum size of a payload that will be stored in Temporal.
+	// If a payload is larger than this size, it will be stored in a separate store.
+	ThresholdSize int `koanf:"threshold_size"`
+
+	FileStoreRootPath string `koanf:"file_store_root_path"` // if store_type == "file".
+	S3StoreBucketName string `koanf:"s3_store_bucket_name"` // if store_type == "s3".
```
daabr: Nit: S3 = "Simple Storage Service". So "S3 store" is like saying "ATM machine".

Actually, why not use just one string for the local-files root path and cloud-storage bucket URL?

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,178 @@
...
+)
+
+type LargePayloadConfig struct {
+	StoreType string `koanf:"store_type"` // "none" to disable.
+
+	// ThresholdSize is the maximum size of a payload that will be stored in Temporal.
+	// If a payload is larger than this size, it will be stored in a separate store.
+	ThresholdSize int `koanf:"threshold_size"`
+
+	FileStoreRootPath string `koanf:"file_store_root_path"` // if store_type == "file".
```
daabr: Nit: a "file store" is actually anything that stores files, even S3 can be considered a file store. But the meaning is clear, so whatever.

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,178 @@
...
+			l.Error("failed to store payload in external store", zap.Error(err))
+			return nil, err
+		}
+
+		md := maps.Clone(in.Metadata)
+		if md == nil {
+			md = make(map[string][]byte, 2)
+		}
+		md[largeObjectKeyMetadataKey] = []byte(key)
+		md[largeObjectSizeMetadataKey] = []byte(strconv.Itoa(size))
```
daabr, line 147: Maybe also add a checksum? OK to leave as a distant TODO (but with a ticket ID)

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,178 @@
...
+	for _, in := range payloads {
+		md := maps.Clone(in.GetMetadata())
+
+		key, ok := md[largeObjectKeyMetadataKey]
+		if !ok {
+			out = append(out, in)
+			continue
+		}
+
+		size, _ := strconv.Atoi(string(md[largeObjectSizeMetadataKey]))
```
daabr: You can't ignore this error, you need to check it and `return nil, err` if needed.

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,178 @@
...
+
+		l := c.l.With(zap.Int("size", size), zap.String("key", string(key)))
+
+		l.Debug("payload is stored in external store, fetching")
+
+		if size == 0 {
+			size = c.cfg.ThresholdSize
+		}
```
daabr: If you check `Atoi` for errors above, this can't happen. Also, you should never make assumptions about the size - it'll always result in data corruption with unexpected impact on the server, and also that's a classic vector for buffer attacks.

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,178 @@
...
+
+		data := make([]byte, size)
+		found, err := c.store.Get(string(key), &data)
```
daabr: Another problem with using `gokv`: who's in charge of cleanup? is there even cleanup? That's a critical requirement after successful retrieval.

internal/backend/temporalclient/config.go
itayd: why?

internal/backend/temporalclient/dataconverter.go
itayd: changed

internal/backend/temporalclient/config.go
itayd: changed

internal/backend/temporalclient/largepayloads_codec.go
itayd: Disagree - this is a very simple package, merely an interface with some implementations. This allows also to work locally with blobs.

internal/backend/temporalclient/largepayloads_codec.go
itayd: changed names, but i think explicit fields are better.

internal/backend/temporalclient/largepayloads_codec.go
itayd: i think it's clear enough.

internal/backend/temporalclient/largepayloads_codec.go
itayd: this is needed only to make an allocation, if there is an error we allocate anyway with a minimal threshold size.

itayd: removed.

internal/backend/temporalclient/largepayloads_codec.go
itayd: ok, will remove.

internal/backend/temporalclient/largepayloads_codec.go
itayd: it's not gokv's concern. i forgot to mention in the description that is TBD.

internal/backend/temporalclient/largepayloads_codec.go
itayd: added sha.

internal/backend/temporalclient/largepayloads_codec.go
itayd: brought kv in and redid the s3

go.mod
```diff
@@ -38,6 +38,7 @@ require (
 	github.com/flosch/pongo2/v6 v6.0.0
 	github.com/fullstorydev/grpcurl v1.9.2
 	github.com/glebarez/sqlite v1.11.0
+	github.com/go-test/deep v1.1.0
```
tebeka: `testify` already has good diffs, why do we need this one?

internal/backend/temporalclient/dataconverter.go
```diff
@@ -84,7 +86,24 @@ func NewDataConverter(l *zap.Logger, cfg *DataConverterConfig, parent converter.
...
+		return nil, fmt.Errorf("new large payload codec: %w", err)
+	}
+
+	if loCodec != nil {
+		codecs = append(codecs, loCodec)
+	}
+
+	cvt := converter.NewCodecDataConverter(parent, codecs...)
+
+	if loCodec != nil {
```
tebeka, line 100: Do you mean `if cvt != nil`? You already check for `loCodec != nil` above.
If you *do* mean to check `loCodec`, the maybe flip the logic:
```
if loCodec == nil {
    continue
}
// ...
```

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+func newLargePayloadCodec(l *zap.Logger, cfg LargePayloadConfig) (converter.PayloadCodec, error) {
+	var store kvstore.Store
+
+	switch cfg.StoreType {
+	case "":
+		return nil, errors.New("store type not specified")
+	case "none":
+		l.Warn("large payloads store is disabled")
```
tebeka:
```suggestion
		l.Info("large payloads store is disabled")
```

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,201 @@
...
+		return nil, nil
+	case "inmem":
+		store = gomap.NewStore(gomap.Options{Codec: encoding.Gob})
+		l.Warn("using in-memory store for large payloads")
```
tebeka, line 71:
```suggestion
		l.Info("using in-memory store for large payloads")
```

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,201 @@
...
+
+		l.Info("using S3 store for large payloads", zap.String("bucket", cfg.S3BucketName))
+	default:
+		return nil, errors.New("unknown store type")
+	}
+
+	return &largePayloadCodec{cfg: cfg, l: l, store: store}, nil
+}
+
+func (c *largePayloadCodec) Encode(payloads []*commonpb.Payload) ([]*commonpb.Payload, error) {
```
tebeka, line 109: In general: It bothers me that protobuf is all over the system, if should be limited to API levels.
But, since this is storage layer - maybe :)

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,201 @@
...
+	store kvstore.Store
+}
+
+func newLargePayloadCodec(l *zap.Logger, cfg LargePayloadConfig) (converter.PayloadCodec, error) {
+	var store kvstore.Store
+
+	switch cfg.StoreType {
+	case "":
+		return nil, errors.New("store type not specified")
+	case "none":
```
tebeka: Use constants for the names. Validate the config when loading.

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,201 @@
...
+		if store, err = s3.NewClient(s3.Options{
+			BucketName: cfg.S3BucketName,
+			Codec:      encoding.Gob,
+		}); err != nil {
+			return nil, err
+		}
+
+		l.Info("using S3 store for large payloads", zap.String("bucket", cfg.S3BucketName))
+	default:
+		return nil, errors.New("unknown store type")
```
tebeka:
```suggestion
		return nil, fmt.Errorf("unknown store type: %#v", cfg.StorageType)
```

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,201 @@
...
+		l := c.l.With(zap.Int("size", size))
+
+		if size < c.cfg.ThresholdSize {
+			l.Debug("payload is below threshold size, storing in Temporal", zap.Int("size", size), zap.Int("threshold", c.cfg.ThresholdSize))
+			out = append(out, in)
+			continue
+		}
+
+		if c.cfg.MaxSize != 0 && size > c.cfg.MaxSize {
+			l.Error("payload is above maximum size")
```
tebeka, line 125: Add more info in the error, size and `in.Metadata`

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,201 @@
...
+	"go.autokitteh.dev/autokitteh/internal/kvstore/encoding"
+	"go.autokitteh.dev/autokitteh/internal/kvstore/file"
+	"go.autokitteh.dev/autokitteh/internal/kvstore/gomap"
+	"go.autokitteh.dev/autokitteh/internal/kvstore/s3"
+	"go.autokitteh.dev/autokitteh/internal/xdg"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+const (
+	prefix                     = "ak_blob_"
```
tebeka, line 27: Use distinct type to avoid name collision in the metadata (same as what we do in context).

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,201 @@
...
+		key, ok := md[largeObjectKeyMetadataKey]
+		if !ok {
+			out = append(out, in)
+			continue
+		}
+
+		expectedSHA, ok := md[largeObjectHashMetadataKey]
+		if !ok {
+			c.l.Error("payload hash not found in metadata")
+			return nil, errors.New("payload hash not found in metadata")
```
tebeka, line 174:
```suggestion
			return nil, fmt.Errorf("payload hash not found in metadata (key=%#v)", key)
```

internal/backend/temporalclient/largepayloads_codec.go
```diff
@@ -0,0 +1,201 @@
...
+			return nil, err
+		}
+
+		if !found {
+			l.Error("payload not found in external store")
+			return nil, err
+		}
+
+		sha := sha256.Sum256(data)
+		if !bytes.Equal(sha[:], expectedSHA) {
```
tebeka, line 194:
```suggestion
		if !bytes.Equal(sha, expectedSHA) {
```

internal/kvstore/encoding/codec.go
tebeka, line 1: Why do we need codecs? Can't we use `gob` and that's it?

internal/kvstore/encoding/gob.go
```diff
@@ -0,0 +1,28 @@
...
+)
+
+// GobCodec encodes/decodes Go values to/from gob.
+// You can use encoding.Gob instead of creating an instance of this struct.
+type GobCodec struct{}
+
+// Marshal encodes a Go value to gob.
+func (c GobCodec) Marshal(v any) ([]byte, error) {
+	buffer := new(bytes.Buffer)
+	encoder := gob.NewEncoder(buffer)
```
tebeka:
```suggestion
	var buffer bytes.Buffer
	encoder := gob.NewEncoder(&buffer)
```

internal/kvstore/file/file.go
```diff
@@ -0,0 +1,201 @@
...
+	"os"
+	"path/filepath"
+	"sync"
+
+	"go.autokitteh.dev/autokitteh/internal/kvstore"
+	"go.autokitteh.dev/autokitteh/internal/kvstore/encoding"
+	"go.autokitteh.dev/autokitteh/internal/kvstore/util"
+)
+
+var defaultFilenameExtension = "json"
```
tebeka, line 15: Why JSON is the default?

internal/kvstore/file/file.go
```diff
@@ -0,0 +1,201 @@
...
+
+// store is a gokv.store implementation for storing key-value pairs as files.
+type store struct {
+	// For locking the locks map
+	// (no two goroutines may create a lock for a filename that doesn't have a lock yet).
+	locksLock *sync.Mutex
+	// For locking file access.
+	fileLocks         map[string]*sync.RWMutex
```
tebeka, line 23: IMO we can start with a single lock or even without a lock since linux/osx can handle multiple writers to a single file. Also, there might be more than a single server running and then these locks won't help.

internal/kvstore/file/file.go
```diff
@@ -0,0 +1,201 @@
...
+	filenameExtension string
+	directory         string
+	codec             encoding.Codec
+}
+
+// Set stores the given value for the given key.
+// Values are automatically marshalled to JSON or gob (depending on the configuration).
+// The key must not be "" and the value must not be nil.
+func (s *store) Set(_ context.Context, k string, v any) error {
```
tebeka, line 32: I don't like we're ignoring context. On AWS files can be mounted on NFS and then it can block forever.
Same for `Get`

internal/kvstore/file/file.go
```diff
@@ -0,0 +1,201 @@
...
+	return os.WriteFile(filePath, data, 0o600)
+}
+
+// Get retrieves the stored value for the given key.
+// You need to pass a pointer to the value, so in case of a struct
+// the automatic unmarshalling can populate the fields of the object
+// that v points to with the values of the retrieved object's values.
+// If no value is found it returns (false, nil).
+// The key must not be "" and the pointer must not be nil.
+func (s *store) Get(_ context.Context, k string, v any) (found bool, err error) {
```
tebeka, line 65:
```suggestion
func (s *store) Get[T](_ context.Context, k string, v *T) (found bool, err error) {
```

tebeka, line 65: Instead of returning `found` and `err`, return just `err` and use `fs.ErrNotExist` for not found.

internal/kvstore/file/file.go
```diff
@@ -0,0 +1,201 @@
...
+	filePath := filepath.Clean(s.directory + "/" + filename)
+
+	// File lock and file handling.
+	lock.Lock()
+	defer lock.Unlock()
+	err := os.Remove(filePath)
+	if os.IsNotExist(err) {
+		return nil
+	}
+	return err
```
tebeka: All of this code repeats in `Get` and `Set`. Have a `withFile` method that will get the mutator.

internal/kvstore/file/file.go
```diff
@@ -0,0 +1,201 @@
...
+
+// prepFileLock returns an existing file lock or creates a new one
+func (s *store) prepFileLock(escapedKey string) *sync.RWMutex {
+	s.locksLock.Lock()
+	lock, found := s.fileLocks[escapedKey]
+	if !found {
+		lock = new(sync.RWMutex)
+		s.fileLocks[escapedKey] = lock
+	}
+	s.locksLock.Unlock()
```
tebeka: Move to defer in line 135.

internal/kvstore/file/file.go
tebeka, line 1: Instead of using file based, why not sqlite with a blobs? It'll take care of locking and it's a single file.

internal/kvstore/redis/redis_test.go
```diff
@@ -0,0 +1,162 @@
...
+		defer client.Close()
+		err := client.Set(context.Background(), "foo", nil)
+		if err == nil {
+			t.Error("Expected an error")
+		}
+	})
+
+	// Test passing nil or pointer to nil value for retrieval
+
+	createTest := func(codec encoding.Codec) func(t *testing.T) {
```
tebeka: Why is this internal function? Make it external.

internal/kvstore/redis/redis_test.go
```diff
@@ -0,0 +1,162 @@
...
+	t.Run("get with nil / nil value parameter", createTest(encoding.Gob))
+}
+
+// TestClose tests if the close method returns any errors.
+func TestClose(t *testing.T) {
+	client := createClient(t, encoding.JSON)
+	err := client.Close()
+	if err != nil {
+		t.Error(err)
+	}
```
tebeka:
```suggestion
	assert.NoError(t, err)
```
Nit: I think that since we use testify in some tests, we'll use it all over.

internal/kvstore/s3/s3.go
```diff
@@ -0,0 +1,163 @@
...
+}
+
+// Close closes the client.
+// In the S3 implementation this doesn't have any effect.
+func (c Client) Close() error {
+	return nil
+}
+
+// Options are the options for the S3 client.
+type Options struct {
```
tebeka, line 110: Do we also need access keys?

internal/kvstore/s3/s3.go
```diff
@@ -0,0 +1,163 @@
...
+	// Encoding format.
+	// Optional (encoding.JSON by default).
+	Codec encoding.Codec
+}
+
+// DefaultOptions is an Options object with default values.
+// Region: "" (use shared config file or environment variable),
+// AWSaccessKeyID: "" (use shared credentials file or environment variable),
+// AWSsecretAccessKey: "" (use shared credentials file or environment variable),
+// CustomEndpoint: "", UsePathStyleAddressing: false, Codec: encoding.JSON
```
tebeka: These are not found in `Options`

internal/kvstore/test/test.go
```diff
@@ -0,0 +1,372 @@
...
+	"strconv"
+	"sync"
+	"testing"
+
+	"github.com/go-test/deep"
+
+	"go.autokitteh.dev/autokitteh/internal/kvstore"
+)
+
+// Foo is just some struct for common tests.
```
tebeka, line 16: Nit: Delete all `foo` and `bar`

internal/kvstore/util/util.go
```diff
@@ -0,0 +1,29 @@
+package util
```
tebeka: `util` is an anti-pattern

internal/kvstore/util/util.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"errors"
+)
+
+// CheckKeyAndValue returns an error if k == "" or if v == nil
+func CheckKeyAndValue(k string, v any) error {
+	if err := CheckKey(k); err != nil {
+		return err
+	}
+	return CheckVal(v)
```
tebeka:
```suggestion
	return errors.Join(CheckKey(k), CheckVal(v))
```

internal/kvstore/util/util.go
```diff
@@ -0,0 +1,29 @@
...
+func CheckKey(k string) error {
+	if k == "" {
+		return errors.New("The passed key is an empty string, which is invalid")
+	}
+	return nil
+}
+
+// CheckVal returns an error if v == nil
+func CheckVal(v any) error {
+	if v == nil {
```
tebeka: Check that it's a pointer. See above how to catch it in compile time.

tebeka (changes requested):
Apart from code specific comments.
- I don't think we need that many backends, maybe just file and s3
- I don't think we need codecs at all
- I'd love to see a test that gets a store and tests it, then run this tests on all types

itayd:
> * I don't think we need that many backends, maybe just file and s3

gomap is for testing.
i think i can remove redis for now.

> * I don't think we need codecs at all

huh? that's the way temporal handles this stuff. there is no other way?

> * I'd love to see a test that gets a store and tests it, then run this tests on all types

there is one on gomap and file. no idea how to do it on s3 locally?

itayd:
@tebeka most if not all of your comments refer to code that i brought in from gokv. will address them anyway.

go.mod
itayd: done

internal/backend/temporalclient/dataconverter.go
itayd: can't - i need to wrap the converter only when lo is enabled, and i can pass the codec only before creating the dataconverter.

internal/backend/temporalclient/largepayloads_codec.go
itayd: done

internal/backend/temporalclient/largepayloads_codec.go
itayd: i agree in principal but no choice here - that's part of the temporal API and that's how they expect us to work.

internal/backend/temporalclient/largepayloads_codec.go
itayd: done

internal/backend/temporalclient/largepayloads_codec.go
itayd: done

internal/backend/temporalclient/largepayloads_codec.go
itayd: done

internal/backend/temporalclient/largepayloads_codec.go
itayd: i don't think it's relevant here - the metadata is just a `map[string][]byte`.

internal/backend/temporalclient/largepayloads_codec.go
itayd: these two are unrelated - this adds no more info.

internal/backend/temporalclient/largepayloads_codec.go
itayd:
```
cannot use sha (variable of type [32]byte) as []byte value in argument to bytes.Equalcompiler[IncompatibleAssign](https://pkg.go.dev/golang.org/x/tools/internal/typesinternal#IncompatibleAssign)
```

internal/kvstore/encoding/codec.go
itayd: i think i might use it in the future for other stuff, this is how gokv is implemented. i'd rather keep it for now. removing it is too much trouble.

internal/kvstore/encoding/gob.go
itayd: done

internal/kvstore/file/file.go
itayd: that's what gokv chooses. i override it anyway.

internal/kvstore/file/file.go
itayd: is there a way doing io with context in go?

internal/kvstore/file/file.go
itayd: this is an implementation of the store interface, you can't use generics here.

itayd: this is not necessarily over an fs, so i hesitate involving the `fs.` package here...

internal/kvstore/file/file.go
itayd: done

internal/kvstore/redis/redis_test.go
itayd: removed redis

internal/kvstore/redis/redis_test.go
itayd: removed redis

internal/kvstore/file/file.go
itayd: that's the implementation that was available in gokv. didn't think it's worth it to go overboard rn. maybe will do in the future.

internal/kvstore/s3/s3.go
itayd: we're just using the default keys.

internal/kvstore/s3/s3.go
itayd: removed

internal/kvstore/test/test.go
itayd: not sure i follow?

internal/kvstore/util/util.go
itayd: changed to `checks`.

internal/kvstore/util/util.go
itayd: done

internal/kvstore/util/util.go
itayd: since the interface is using any, this can't be done.

PR #1107, opened 2025-02-22
feat: add session_id to parameters when starting a session

itayd:
this can work now:

```
def on_trigger(event):
  print(event.session_id)
  print(event.data)
```

sometimes the user might want to correlate the session id with data that is sent elsewhere. we might also can associate more data to the event other than the input data.

### DIFF
--- a/cmd/ak/cmd/sessions/test.go
+++ b/cmd/ak/cmd/sessions/test.go
@@ -31,6 +31,9 @@ var (
 
 	// File "/opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result`
 	pyLibRe = regexp.MustCompile(`File ".*/lib/python3\.\d+/(.*\.py)", line (\d+), in (.*)`)
+
+	// Some python version like to put an annoying ^^^^ marker to show where the error is.
+	pyAnnoyingErrorLocationMarkerRe = regexp.MustCompile(`^\s*~*\^+$`)
 )
 
 func normalizePath(p string) string {
@@ -151,6 +154,11 @@ var testCmd = common.StandardCommand(&cobra.Command{
 			s := bufio.NewScanner(strings.NewReader(ps))
 			for s.Scan() {
 				line := normalizePath(s.Text())
+
+				if pyAnnoyingErrorLocationMarkerRe.MatchString(line) {
+					continue
+				}
+
 				prints.WriteString(line)
 				prints.WriteRune('\n')
 			}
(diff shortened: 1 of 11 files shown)


cmd/ak/cmd/sessions/test.go
```diff
@@ -31,6 +31,9 @@ var (
 
 	// File "/opt/hostedtoolcache/Python/3.12.8/x64/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result`
 	pyLibRe = regexp.MustCompile(`File ".*/lib/python3\.\d+/(.*\.py)", line (\d+), in (.*)`)
+
+	// Some python version like to put an annoying ^^^^ marker to show where the error is.
```
daabr, line 35: Correction :)
```suggestion
	// Some Python versions put a helpful ^^^^ marker to show exactly where the error is.
```

cmd/ak/cmd/sessions/test.go
```diff
@@ -31,6 +31,9 @@ var (
...
+	pyAnnoyingErrorLocationMarkerRe = regexp.MustCompile(`^\s*~*\^+$`)
```
daabr, line 36: What's the meaning of (0 or more) `~` characters here?

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -574,7 +574,12 @@ func (w *sessionWorkflow) run(wctx workflow.Context, l *zap.Logger) (prints []sd
 			return prints, err
 		}
 
-		if retVal, err = run.Call(ctx, callValue, nil, w.data.Session.Inputs()); err != nil {
+		inputs := map[string]sdktypes.Value{
```
daabr, line 577: Not blocking, but if possible: can you also add `event_type`?

daabr (approved):

cmd/ak/cmd/sessions/test.go
itayd: no

cmd/ak/cmd/sessions/test.go
itayd: soemtimes it likes to do ~~~~ or some thing donno why

daabr: I never saw that, but in that case I guess a more accurate regex might be:
```
`^\s*[~^]+$`
```
(Be careful: `[~^]` means "`~` or `^`", but `[^~]` means "everything except `~`")

internal/backend/sessions/sessionworkflows/workflow.go
itayd: in a different pr

PR #1108, opened 2025-02-22
refactor(CI): improve build time & isolation

daabr:
Part 1 of refactoring CI: don't run any more or less, just segment it differently, enable caching, and avoid duplication.

- Go:
  - Don't use `make bin/ak` - it blocks caching (due to timestamp-based build flags)
  - Split between unit tests and long-running system tests
  - Don't run session tests in unit-tests job, already running in other jobs
  - Don't run `go mod download` - useful in the past, but not anymore with better caching
- Python: move Python runtime's Go tests to Python workflow (due to Python dependency)

Stats:

- Go in main: 12m32s
  - https://github.com/autokitteh/autokitteh/actions/runs/13466518794/job/37633353273
- Go in this PR: 2m52s + 7m42s = 10m34s
  - https://github.com/autokitteh/autokitteh/actions/runs/13477279622/job/37657993992?pr=1108
  - https://github.com/autokitteh/autokitteh/actions/runs/13477279622/job/37658023378?pr=1108

Refs: ENG-2014

PR #1114, opened 2025-02-23
refactor(CI): bring back "go mod download"

daabr:
It does have some potential time-saving impact in the first time you run checks for a new PR: somewhere between 5 and 50 seconds.

Also update golangci-lint version on the way.

Ref: ENG-2014, #1108

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -86,7 +86,7 @@ golangci_lint=$(shell which golangci-lint)
 $(OUTDIR)/tools/golangci-lint:
 	mkdir -p $(OUTDIR)/tools
 ifeq ($(golangci_lint),)
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(OUTDIR)/tools" v1.63.1
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(OUTDIR)/tools" v1.64.5
 else
 	ln -fs $(golangci_lint) $(OUTDIR)/tools/golangci-lint
 endif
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -86,7 +86,7 @@ golangci_lint=$(shell which golangci-lint)
 $(OUTDIR)/tools/golangci-lint:
 	mkdir -p $(OUTDIR)/tools
 ifeq ($(golangci_lint),)
-	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(OUTDIR)/tools" v1.63.1
+	curl -sSfL https://raw.githubusercontent.com/golangci/golangci-lint/master/install.sh | sh -s -- -b "$(OUTDIR)/tools" v1.64.5
```
tebeka, line 89: Security: Don't love running shell scrips off the web. We can use our own, something like:
```bash
#!/bin/bash

version=1.64.5
os=$(go env GOOS)
arch=$(go env GOARCH)
url=https://github.com/golangci/golangci-lint/releases/download/v${version}/golangci-lint-${version}-${os}-${arch}.tar.gz

curl -sL ${url} |
	tar \
		-C ${OUTDIR}/tools \
		--strip-components=1 \
		-xzf - golangci-lint-${version}-${os}-${arch}/golangci-lint
```

tebeka (approved):
Approved with one security note.

Makefile
daabr: I understand your hesitation, but this is the official way to install this tool for generic CI, see here:

https://golangci-lint.run/welcome/install/#other-ci

I updated the link in line 83

PR #1116, opened 2025-02-23
fix: optimize withProjectOrgID join so events listing is faster

efiShtain:
no need for distinct since the results always has some id in each row use right join on projects so the optimizer knows to use the correct index instead of fetching both tables and join later


tebeka (approved):

itayd:
obsoleted by https://github.com/autokitteh/autokitteh/pull/1131

PR #1118, opened 2025-02-23
fix: connstatus false positive

mario99logic:
Introduces a common function to validate that an OAuth access token isn’t empty, confirming successful OAuth flow. A separate check was added for Google due to differing variable names. 
I haven’t tested the connStatus for every integration (lack of OAuth apps), but since it worked on Slack, I assume it works for the rest. 

refs: INT-225

### DIFF
--- a/integrations/common/vars.go
+++ b/integrations/common/vars.go
@@ -38,6 +38,11 @@ func EncodeOAuthData(t *oauth2.Token) OAuthData {
 	}
 }
 
+// IsOAuthTokenEmpty returns true if the OAuth 2.0 token is empty.
+func IsOAuthTokenEmpty(vs sdktypes.Vars) bool {
+	return vs.GetValue(OAuthAccessTokenVar) == ""
+}
+
 // ReadConnectionVars returns the connection's variables, or
 // an error if the connection is not initialized or accessible.
 func ReadConnectionVars(ctx context.Context, vars sdkservices.Vars, cid sdktypes.ConnectionID) (sdktypes.Vars, sdktypes.Status, error) {
--- a/integrations/common/vars.go
+++ b/integrations/common/vars.go
@@ -18,6 +18,8 @@ var (
 	OAuthExpiryVar       = sdktypes.NewSymbol("oauth_expiry")
 	OAuthRefreshTokenVar = sdktypes.NewSymbol("oauth_refresh_token")
 	OAuthTokenTypeVar    = sdktypes.NewSymbol("oauth_token_type")
+
+	LegacyOAuthAccessTokenVar = sdktypes.NewSymbol("oauth_AccessToken")
 )
 
 // OAuthData contains OAuth 2.0 token details.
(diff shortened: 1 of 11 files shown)


integrations/common/vars.go
```diff
@@ -38,6 +38,11 @@ func EncodeOAuthData(t *oauth2.Token) OAuthData {
 	}
 }
 
+// IsOAuthTokenEmpty returns true if the OAuth 2.0 token is empty.
+func IsOAuthTokenEmpty(vs sdktypes.Vars) bool {
+	return vs.GetValue(OAuthAccessTokenVar) == ""
```
tebeka: Are there any other limitation we can check for validity? I assume `"x"` is not a valid token.

integrations/common/vars.go
```diff
@@ -38,6 +38,11 @@ func EncodeOAuthData(t *oauth2.Token) OAuthData {
 	}
 }
 
+// IsOAuthTokenEmpty returns true if the OAuth 2.0 token is empty.
+func IsOAuthTokenEmpty(vs sdktypes.Vars) bool {
```
daabr: 1. Weakness: this function depends on getting a `Vars` instance that *loaded* this var if it exists - this is not documented, and can't be checked

2. This function is too trivial, but it's easy to to make it better: notice that all the calls to it are:

```py
if common.IsOAuthTokenEmpty(vs) {
	return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
}
return sdktypes.NewStatus(sdktypes.StatusCodeOK, "Using OAuth 2.0"), nil
```

So a better thing would be to wrap this entire block as a function instead of just the check.

integrations/common/vars.go
daabr: Unfortunately there are no rules about the content of an OAuth access token - each service provider has their own conventions.

integrations/google/connections/connections.go
```diff
@@ -40,6 +42,9 @@ func ConnStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
 		case integrations.JSONKey:
 			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "Using JSON key"), nil
 		case integrations.OAuth:
+			if vs.GetValue(AccessToken) == "" {
```
daabr: See the comment in `integrations/common/vars.go`. You can create a similar function for all the old-style integrations that use `oauthAccessToken` instead of `oauth_access_token`.

Also note that you can do the same for any auth type, by specifying the name (as a symbol or a string) of the variable that needs to be non-empty.

integrations/common/vars.go
mario99logic: I considered combining the token check and status return, but kept them separate because:

1. The token check appears in a switch statement where all cases return NewStatus with consistent formatting. Having a separate boolean check keeps the switch cases clean and uniform
2. This gives us flexibility if we need to handle empty tokens differently in future code paths

I will make the changes if you think it's cleaner.

integrations/google/connections/connections.go
mario99logic: I created another function for the old-style integrations, the old token var name is `oauth_AccessToken`

integrations/common/vars.go
```diff
@@ -28,6 +30,24 @@ type OAuthData struct {
...
+// It depends on sdktypes.Vars being preloaded with OAuthAccessTokenVar, which isn't validated.
+func CheckOAuthToken(vs sdktypes.Vars) (sdktypes.Status, error) {
+	if vs.GetValue(OAuthAccessTokenVar) == "" {
+		return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
+	}
+	return sdktypes.NewStatus(sdktypes.StatusCodeOK, "Using OAuth 2.0"), nil
+}
+
+// CheckOAuthToken returns a warning status if LegacyOAuthAccessTokenVar is missing in sdktypes.Vars; otherwise, it returns OK.
+// It depends on sdktypes.Vars being preloaded with LegacyOAuthAccessTokenVar, which isn't validated.
```
daabr: Incorrect function name + https://go.dev/doc/comment#doclinks
```suggestion
// CheckLegacyOAuthToken returns a warning status if [LegacyOAuthAccessTokenVar] is missing in [sdktypes.Vars]; otherwise,
// it returns OK. It depends on [sdktypes.Vars] being preloaded with [LegacyOAuthAccessTokenVar], which isn't validated.
```

integrations/common/vars.go
```diff
@@ -28,6 +30,24 @@ type OAuthData struct {
 	TokenType    string `var:"oauth_token_type"`
 }
 
+// CheckOAuthToken returns a warning status if OAuthAccessTokenVar is missing in sdktypes.Vars; otherwise, it returns OK.
+// It depends on sdktypes.Vars being preloaded with OAuthAccessTokenVar, which isn't validated.
```
daabr: https://go.dev/doc/comment#doclinks + cosmetic tweak (shorter 1st line)
```suggestion
// CheckOAuthToken returns a warning status if [OAuthAccessTokenVar] is missing in [sdktypes.Vars]; otherwise,
// it returns OK. It depends on [sdktypes.Vars] being preloaded with [OAuthAccessTokenVar], which isn't validated.
```

daabr (approved):

PR #1119, opened 2025-02-23
refactor(CI): path-based triggers for OPA

daabr:
OPA tests (like PostgreSQL tests) are limited to the root directory where all the code is.

This applies only to the PR review phase. All tests run when merging to main.

Refs: ENG-2014

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -6,13 +6,6 @@ name: CI - Go
 
 on:
   pull_request:
-    paths:
-      - .github/workflows/go.yml
-      - .golangci.yml
-      - go.mod
-      - tests/**
-      - "**.go"
-      - "**.py"
   push:
     branches: [main]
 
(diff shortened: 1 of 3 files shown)


itayd:
i agree re opa
all the other ones i strongly disagree - we had cases where python code was not touched, but because go code changed it caused python to fail.

tebeka (dismissed):
I agree with @itayd  that we need to test Go when Python changes

daabr:
> I agree with @itayd that we need to test Go when Python changes

I'm OK with that, but I'm interested in the details: we already consider all Go tests in `pythonrt` as part of the Python CI.

Is there any other Go code that is relevant for Python too?
Any Python code that affects the Go-based AK server?

Anyway, I entangled Python and Go: a change in either will trigger both CIs. But if a PR doesn't touch either Go or Python code/tests/configs we don't run these workflows.

itayd (approved):

PR #1120, opened 2025-02-23
refactor(CI): Docker images

daabr:
1. Split Docker jobs out of the Go workflow: they're not related to Go, and we won't see an unexecuted job in every PR and merge-to-main now

2. Note that the job to publish the image when the PR is merged to main wasn't triggered as expected: the `if` means that the job got triggered when the `needs` jobs end, regardless of their success or failure.

   ```yaml
   needs:
     - go-unit-tests
     - go-session-tests
     - go-system-tests
     - static-analysis
   if: github.ref == 'refs/heads/main'
   ```

   (If we want to block, we can use AutoKitteh - we have a sample exactly for the scenario of 
   triggering a workflow when a group of other workflows are successful)

3. Remove `with: fetch-depth: 0` from `checkout` - it means full history, not just the latest commit (which is the default behavior) - it's needed only in Go Releaser

4. `release.yml`: changed trigger from `push: tags: "v*"` to `release: types: [published]` (see https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#release)

5. `WORKING_DIRECTORY` removed: not needed since we use the root dir of a monorepo

6. `${GITHUB_REF#refs/tags/}` --> `${GITHUB_REF_NAME}` (same thing, without string substitution, which doesn't even work if the ref isn't a tag)

7. Change timeout from 30 minutes to 10 (it always takes 5 minutes to finish in all modes)

8. Misc version updates of GitHub Actions and cosmetic tweaks

Refs: ENG-2014

### DIFF
--- a/.github/workflows/docker-pr.yml
+++ b/.github/workflows/docker-pr.yml
@@ -15,15 +15,6 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
-      - name: Setup version info
-        run: |
-          # These are consumed in the Dockerfile.
-          echo "${GITHUB_REF_NAME}" > .version
-          echo "${GITHUB_SHA}" > .commit
-
-      - name: Fetch web platform
-        run: make webplatform
-
       - name: Build but don't push
         uses: docker/build-push-action@v6
         with:
--- a/.github/workflows/docker-pr.yml
+++ b/.github/workflows/docker-pr.yml
@@ -1,0 +1,34 @@
+# See also "docker-main.yml", "manual-docker-image.yml", and "release.yml".
+# https://github.com/marketplace/actions/build-and-push-docker-images
+
+name: CI - Docker (PRs)
+
+on:
+  pull_request:
+
+jobs:
+  test-docker-build:
+    name: Test Docker build
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup version info
+        run: |
+          # These are consumed in the Dockerfile.
+          echo "${GITHUB_REF_NAME}" > .version
+          echo "${GITHUB_SHA}" > .commit
+
+      - name: Fetch web platform
+        run: make webplatform
+
+      - name: Build but don't push
+        uses: docker/build-push-action@v6
+        with:
+          context: .
+          platforms: linux/amd64
+          tags: autokitteh:latest
+          push: false
+          provenance: false
--- a/.github/workflows/manual-docker-image.yml
+++ b/.github/workflows/manual-docker-image.yml
@@ -16,7 +16,7 @@ on:
         description: version to push
 
 jobs:
-  publish-docker-image:
+  publish_docker_image:
     name: Publish Docker image
     runs-on: ubuntu-latest
     timeout-minutes: 10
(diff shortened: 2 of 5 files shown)


tebeka (approved):
I'm OK with the changes.
My main concern is that the CI is very complex for this stage in our lives (> 500 lines of YAML). Is there a way to drop some things or simplify it?

daabr:
> I'm OK with the changes. My main concern is that the CI is very complex for this stage in our lives (> 500 lines of YAML). Is there a way to drop some things or simplify it?

Specifically in the context of Docker, I don't think we use the manual workflow or the "latest" images for non-release merges, but I'll let @efiShtain decide what he wants to do with that.

If there's anything else you want to change or remove, feel free to suggest it. My personal perspective is that CI impact is measured in time (spent vs. saved) and not LOC.

.github/workflows/docker-pr.yml
```diff
@@ -0,0 +1,34 @@
...
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup version info
+        run: |
+          # These are consumed in the Dockerfile.
+          echo "${GITHUB_REF_NAME}" > .version
+          echo "${GITHUB_SHA}" > .commit
+
+      - name: Fetch web platform
```
efiShtain: we don't need this part
it does the fetch and build in the dockerfile

.github/workflows/docker-pr.yml
```diff
@@ -0,0 +1,34 @@
...
+    timeout-minutes: 10
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup version info
+        run: |
+          # These are consumed in the Dockerfile.
+          echo "${GITHUB_REF_NAME}" > .version
+          echo "${GITHUB_SHA}" > .commit
```
efiShtain: don't really care about these two if we don't publish the image

.github/workflows/manual-docker-image.yml
```diff
@@ -15,32 +17,35 @@ on:
...
       id-token: write
       contents: read
     steps:
       - name: Checkout
         uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
+
+      - name: Fetch web platform
+        run: make webplatform
```
efiShtain: not needed

.github/workflows/release.yml
```diff
@@ -19,55 +23,57 @@ jobs:
...
       - name: Setup version info
         run: |
           # These are consumed in the Dockerfile.
-          echo "${GITHUB_REF#refs/tags/}" > .version
+          echo "${GITHUB_REF_NAME}" > .version
           echo "${GITHUB_SHA}" > .commit
+
+      - name: Fetch web platform
+        run: make webplatform
+
```
efiShtain: no need

.github/workflows/docker-pr.yml
daabr: done

.github/workflows/manual-docker-image.yml
daabr: done

.github/workflows/release.yml
daabr: done

.github/workflows/docker-pr.yml
daabr: done

PR #1121, opened 2025-02-24
refactor(CI): minor cosmetic facelift for remaining workflows

daabr:
Refs: ENG-2014

### DIFF
--- a/.github/workflows/manifest.yml
+++ b/.github/workflows/manifest.yml
@@ -1,24 +1,23 @@
-name: Validate Manifest Schema is Up-to-Date
+name: CI - Manifest
 
 on:
-  push:
-    branches:
-      - main
-    paths:
-      - ".github/workflows/ci-manifest.yml"
-      - "manifest.schema.yaml"
-      - "internal/manifest/manifest.go"
-      - "docs/autokitteh.yaml"
   pull_request:
     paths:
-      - ".github/workflows/ci-manifest.yml"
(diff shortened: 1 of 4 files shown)


tebeka:
> Refs: [INT-2014](https://linear.app/autokitteh/issue/INT-2014)

Wrong number?

.github/workflows/manifest.yml
```diff
@@ -28,8 +27,10 @@ jobs:
         uses: actions/setup-go@v5
         with:
           go-version-file: go.mod
+
       - name: Build AK
-        run: make bin/ak
+        run: go build -trimpath -o bin/ak ./cmd/ak
```
tebeka, line 32: Why?

tebeka (approved):

daabr:
> > Refs: [INT-2014](https://linear.app/autokitteh/issue/INT-2014)
> 
> Wrong number?

Oops, ENG-2014, will fix in all relevant PRs

.github/workflows/manifest.yml
daabr: Noted in other PRs related to ENG-2014: `make bin/ak` uses timestamp-based flags which conflict with caching, so whenever we build AK in CI we use `go build`.

https://github.com/search?q=repo%3Aautokitteh%2Fautokitteh+%22build+ak%22+path%3Ayml&type=code

PR #1123, opened 2025-02-24
refactor: kill almost all modules and syscalls

itayd:
This adds to `RunCallbacks` explicit functions to invoke instead of calling stuff in the module `ak` exported by the sessions. This also kills all modules exported by the sessions except testtools, which I might kill later.


internal/backend/sessions/sessionworkflows/callbacks.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return sdktypes.InvalidSessionID, err
+	}
+
+	w.l.Info("session started", zap.Any("sid", sid))
+
+	return sid, nil
+}
+
+func (w *sessionWorkflow) subscribe(ctx context.Context, _ sdktypes.RunID, name, filter string) (string, error) {
```
tebeka: Why the `RunID` parameter if we don't use it?

internal/backend/sessions/sessionworkflows/callbacks.go
```diff
@@ -0,0 +1,159 @@
...
+	_, connection := kittehs.FindFirst(w.data.Connections, func(c sdktypes.Connection) bool {
+		return c.Name().String() == name
+	})
+
+	_, trigger := kittehs.FindFirst(w.data.Triggers, func(t sdktypes.Trigger) bool {
+		return t.Name().String() == name
+	})
+
+	if connection.IsValid() && trigger.IsValid() {
```
tebeka: This can be in the switch below.

internal/backend/sessions/sessionworkflows/callbacks.go
```diff
@@ -0,0 +1,159 @@
...
+	return signalID.String(), nil
+}
+
+func (w *sessionWorkflow) unsubscribe(ctx context.Context, rid sdktypes.RunID, signalID string) error {
+	uuid, err := uuid.Parse(signalID)
+	if err != nil {
+		return err
+	}
+
+	w.removeEventSubscription(ctx, uuid)
```
tebeka: We does `removeEventSubscritpion` does hide the underlying error?

internal/backend/sessions/sessionworkflows/callbacks.go
```diff
@@ -0,0 +1,159 @@
...
+
+	If event exists, return it - this will increment this event's sequence counting
+
+If no event found, wait on first signal in signals list
+Get the next event on this signal (has to exists since we got a signal on it)
+return this event
+*/
+func (w *sessionWorkflow) nextEvent(ctx context.Context, rid sdktypes.RunID, signals []string, timeout time.Duration) (sdktypes.Value, error) {
+	if len(signals) == 0 {
+		return sdktypes.InvalidValue, errors.New("expecting at least one signal")
```
tebeka: Design question: Why is this an error and not a null operation?

internal/backend/sessions/sessionworkflows/callbacks.go
```diff
@@ -0,0 +1,159 @@
...
+	if did := w.data.Session.DeploymentID(); did.IsValid() {
+		d, err := w.ws.svcs.Deployments.Get(authcontext.SetAuthnSystemUser(ctx), did)
+		if err != nil {
+			return false, err
+		}
+
+		state = d.State()
+	}
+
+	return state == sdktypes.DeploymentStateActive, nil
```
tebeka:
```suggestion
		return d.State() == sdktypes.DeploymentStateActive, nil
	}

	return false, nil
```

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -510,10 +488,39 @@ func (w *sessionWorkflow) run(wctx workflow.Context, l *zap.Logger) (prints []sd
 			if err != nil {
 				l.With(zap.String("text", text)).Sugar().Warnf("failed to add print session record: %v", err)
 			}
+
+			return err
+		},
+		Now: func(nowCtx context.Context, runID sdktypes.RunID) (time.Time, error) {
```
tebeka, line 491: Why do we need `Now`?

runtimes/pythonrt/pythonrt.go
```diff
@@ -409,7 +384,9 @@ func (py *pySvc) setupCallbacksListeningLoop(ctx context.Context) chan (error) {
 				py.log.Log(pyLevelToZap(r.level), r.message)
 				close(r.doneChannel)
 			case p := <-py.channels.print:
-				py.cbs.Print(ctx, py.runID, p.message)
+				if err := py.cbs.Print(ctx, py.runID, p.message); err != nil {
+					callbackErrChan <- err
```
tebeka: Add an error log here?

runtimes/pythonrt/pythonrt.go
```diff
@@ -593,7 +566,7 @@ func (py *pySvc) drainPrints(ctx context.Context) {
 		case r := <-py.channels.log:
 			py.log.Log(pyLevelToZap(r.level), r.message)
 		case r := <-py.channels.print:
-			py.cbs.Print(ctx, py.runID, r.message)
+			_ = py.cbs.Print(ctx, py.runID, r.message)
```
tebeka, line 552: Don't ignore the error

runtimes/pythonrt/pythonrt.go
```diff
@@ -647,7 +620,7 @@ func (py *pySvc) Call(ctx context.Context, v sdktypes.Value, args []sdktypes.Val
 				py.log.Log(pyLevelToZap(r.level), r.message)
 				close(r.doneChannel)
 			case p := <-py.channels.print:
-				py.cbs.Print(ctx, py.runID, p.message)
+				_ = py.cbs.Print(ctx, py.runID, p.message)
```
tebeka: Don't ignore the error

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -184,61 +185,63 @@ func (s *workerGRPCHandler) Activity(ctx context.Context, req *userCode.Activity
...
-		successChannel: callbackChan,
-		errorChannel:   errorChannel,
+		name: name,
+		fn:   fn,
+		ch:   make(chan callbackResponse, 1),
+	}
+
+	runner.channels.callback <- msg
+
+	l.Debug("callback sent", zap.Duration("duration", time.Since(startedAt)))
```
tebeka, line 211:
```suggestion
	l.Debug("callback sent", zap.Duration("duration", time.Since(startedAt)), zap.String("name", name))
```

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -257,140 +260,110 @@ func (s *workerGRPCHandler) StartSession(ctx context.Context, req *userCode.Star
...
-
-		resp := userCode.NextEventResponse{
-			Event: &userCode.Event{
-				Data: data,
-			},
-		}
-		return &resp, nil
 	}
+
+	out, err := sdktypes.ValueWrapper{SafeForJSON: true}.Unwrap(resp.value.(sdktypes.Value))
```
tebeka, line 332: The type assertion might panic, use ", ok" first.

sdk/sdkservices/runtimes.go
```diff
@@ -46,22 +47,31 @@ type Runtime interface {
 }
 
 type (
-	RunLoadFunc  = func(ctx context.Context, rid sdktypes.RunID, path string) (map[string]sdktypes.Value, error)
-	RunCallFunc  = func(ctx context.Context, rid sdktypes.RunID, v sdktypes.Value, args []sdktypes.Value, kwargs map[string]sdktypes.Value) (sdktypes.Value, error)
-	RunPrintFunc = func(ctx context.Context, rid sdktypes.RunID, text string)
-	NewRunIDFunc = func() sdktypes.RunID
+	RunLoadFunc            = func(ctx context.Context, rid sdktypes.RunID, path string) (map[string]sdktypes.Value, error)
```
tebeka: Why do we need these types, you can specify the function signature in `RunCallbacks` and that's it.

sdk/sdkservices/runtimes.go
```diff
@@ -46,22 +47,31 @@ type Runtime interface {
...
-	// Returns sdktypes.ProgramErrorAsError if not internal error.
-	Load RunLoadFunc
-
-	// Returns sdktypes.ProgramErrorAsError if not internal error.
-	Call RunCallFunc
-
-	Print RunPrintFunc
-
-	NewRunID NewRunIDFunc
+	Load               RunLoadFunc // Returns sdktypes.ProgramErrorAsError if not internal error.
```
tebeka: Is this just for starlark?

sdk/sdkservices/runtimes.go
```diff
@@ -88,12 +98,73 @@ func (rc *RunCallbacks) SafeCall(ctx context.Context, rid sdktypes.RunID, v sdkt
...
-		return
+		return nil
+	}
+
+	return rc.Print(ctx, rid, text)
+}
+
+func (rc *RunCallbacks) SafeNewRunID() (sdktypes.RunID, error) { return sdktypes.NewRunID(), nil }
+
+func (rc *RunCallbacks) SafeSleep(ctx context.Context, rid sdktypes.RunID, d time.Duration) error {
```
tebeka: What's the purpose of all these `Safe*` functions?

tebeka (commented):
I general I'm in favor of reducing 675LOC, but at this time I think the risk is too high. After M2?

Also, a lot of changes in starlark, I think it's time to delete it.

itayd:
> I general I'm in favor of reducing 675LOC, but at this time I think the risk is too high. After M2?

-1348 +675, more signifacnt :)

I don't think the risk here is high - it's a refactoring that I am pretty confident in, and I want to build on this.

> 
> Also, a lot of changes in starlark, I think it's time to delete it.

Disagree, as you might have guessed :P

internal/backend/sessions/sessionworkflows/callbacks.go
itayd: This is the same code as before, I just moved it around. @efiShtain ?

internal/backend/sessions/sessionworkflows/workflow.go
itayd: for starlark for example - need to hook time to here. tbh, I think python should do the same, it's much less costly than activity. Also for sleep - where python should call the explicit function, because it keeps track using hearbeats and does not redo the entire sleep on a replay.

runtimes/pythonrt/worker_grpc_handler.go
itayd: it can't - it's always an `sdktypes.Value` here.

sdk/sdkservices/runtimes.go
itayd: yes for now.

sdk/sdkservices/runtimes.go
itayd: if callbacks are nil or the specific callback is nil, does not crash. I think i'll remove them.

daabr (approved):

internal/backend/sessions/sessionworkflows/callbacks.go
itayd: we might - good to know which run id it came from. might log in the future.

internal/backend/sessions/sessionworkflows/callbacks.go
itayd: done

internal/backend/sessions/sessionworkflows/callbacks.go
itayd: agree, changed.

internal/backend/sessions/sessionworkflows/callbacks.go
itayd: done

runtimes/pythonrt/pythonrt.go
itayd: done

runtimes/pythonrt/pythonrt.go
itayd: done

runtimes/pythonrt/worker_grpc_handler.go
itayd: it's already at line 189, autoattached to all logs in this function.

sdk/sdkservices/runtimes.go
itayd: removed.

sdk/sdkservices/runtimes.go
itayd: removed

PR #1124, opened 2025-02-24
bug(python): Catch BaseException instead of Exception

tebeka:
Some function (sys.exit for example) raise exceptions that do not inherit from Exception.

Fixes: ENG-2028

### DIFF
--- a/runtimes/pythonrt/runner/main.py
+++ b/runtimes/pythonrt/runner/main.py
@@ -412,13 +412,14 @@ class Runner(pb.runner_rpc.RunnerService):
         value = error = tb = None
         try:
             value = fn(*args, **kw)
-            if isinstance(value, Exception):
-                set_exception_args(value)
-        except Exception as err:
+        except BaseException as err:
             log.error("%s raised: %s", func_name, err)
             tb = TracebackException.from_exception(err)
             error = err
             set_exception_args(error)
+
+        if isinstance(value, Exception):
+            set_exception_args(value)
 
         if not is_pickleable(error):
             log.info("non pickleable: %r", error)


runtimes/pythonrt/runner/main.py
```diff
@@ -412,14 +412,15 @@ def _call(self, fn, args, kw):
         value = error = tb = None
         try:
             value = fn(*args, **kw)
-            if isinstance(value, Exception):
-                set_exception_args(value)
-        except Exception as err:
+        except BaseException as err:
```
efiShtain, line 415: so BaseException also handle Exception?

runtimes/pythonrt/runner/main.py
```diff
@@ -412,14 +412,15 @@ def _call(self, fn, args, kw):
...
             log.error("%s raised: %s", func_name, err)
             tb = TracebackException.from_exception(err)
             error = err
             set_exception_args(error)
 
+        if isinstance(value, Exception):
```
efiShtain, line 421: what would set value to be of excption type?
wouldn't it be just None ?

runtimes/pythonrt/runner/main.py
tebeka: Yes, see [here](https://docs.python.org/3/library/exceptions.html#exception-hierarchy)

runtimes/pythonrt/runner/main.py
tebeka: Assume the following code:

```python
from urllib.request import URLError

err = URLError("can't connect to host")
```

The call to `URLError` is an activity and the return value is an exception. See

efiShtain (approved):

PR #1125, opened 2025-02-24
feat: Delete events when soft deleting project

mario99logic:
This PR addresses the issue where an event could still be accessed after its associated project had been deleted. The fix ensures that when a project is deleted, all related events are also removed.

Introduced `deleteProjectEvents` to delete sessions and events when soft deleting a project.

refs: ENG-1949

### DIFF
--- a/internal/backend/db/dbgorm/projects.go
+++ b/internal/backend/db/dbgorm/projects.go
@@ -48,6 +48,16 @@ func (gdb *gormdb) deleteProject(ctx context.Context, projectID uuid.UUID) error
 	return gdb.writeTransaction(ctx, func(tx *gormdb) error { return tx.deleteProjectAndDependents(ctx, projectID) })
 }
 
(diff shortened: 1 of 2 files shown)


internal/backend/db/dbgorm/projects.go
```diff
@@ -48,6 +48,16 @@ func (gdb *gormdb) deleteProject(ctx context.Context, projectID uuid.UUID) error
 	return gdb.writeTransaction(ctx, func(tx *gormdb) error { return tx.deleteProjectAndDependents(ctx, projectID) })
 }
 
+func (gdb *gormdb) deleteSessionsAndEvents(ctx context.Context, projectID uuid.UUID) error {
```
efiShtain: look at the code

delete deployment already delete the sessions

so only deleting events is needed here

efiShtain (approved):

PR #1126, opened 2025-02-24
fix(projects): Export all variables

tebeka:
Fixes: ENG-2032

### DIFF
--- a/internal/backend/projects/projects.go
+++ b/internal/backend/projects/projects.go
@@ -308,14 +308,15 @@ func (ps *Projects) exportManifest(ctx context.Context, projectID sdktypes.Proje
 	}
 
 	for _, v := range vars {
-		if v.IsSecret() {
-			continue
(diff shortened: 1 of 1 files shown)


internal/backend/projects/projects.go
```diff
@@ -308,14 +308,15 @@ func (ps *Projects) exportManifest(ctx context.Context, projectID sdktypes.Proje
...
+			Name:   v.Name().String(),
+			Secret: v.IsSecret(),
 		}
-		v := manifest.Var{
-			Name:  v.Name().String(),
-			Value: v.Value(),
+
+		if !v.IsSecret() {
+			mv.Value = v.Value()
 		}
```
RonenMars: We don't need values of the variables in any case for the export. Only the variables' names.

tebeka: Fixed.

daabr (approved):

PR #1127, opened 2025-02-24
fix(CI): obsolete bits in system tests

daabr:
1. It's a waste of time to rebuild AK **inside** the system test suite, because both `make test-system` and the relevant GitHub CI workflows already build AK in advance, and taking advantage of the local/GitHub build cache

2. Remove the `version.txtar` system test - no longer relevant in the new CI where we don't set any version attributes in the binary file

Refs: ENG-2014

### DIFF
--- a/tests/system/ak_test.go
+++ b/tests/system/ak_test.go
@@ -21,6 +21,8 @@ import (
 	"context"
 	"fmt"
 	"io/fs"
+	"os"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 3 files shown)


tests/system/ak_test.go
```diff
@@ -90,12 +90,20 @@ func TestSuite(t *testing.T) {
...
 
-	return akPath
+	wd, err := os.Getwd()
+	if err != nil {
+		t.Fatalf("failed to get current working directory: %v", err)
+	}
+
+	path, err := filepath.Abs(filepath.Join(wd, "..", "..", "bin", "ak"))
+	if err != nil {
+		t.Fatalf("failed to construct ak binary file path: %v", err)
```
itayd: failed to find?

itayd (approved):

tests/system/ak_test.go
daabr: no, just construct the abs path (it should never fail, but theoretically it can) - but that's a good point, I should have added `Stat()` to check that it is found.

PR #1129, opened 2025-02-25
BUG-REPRO: python subscribe is stuck

itayd:
i modified the events system test to use python instead of starlark. looks like when `subscribe` is called, python runner treats it as a regular call and not the overridden call. this causes a deadlock inside the runtime (it tries to send a syscall callback when it's stuck processing the call in the select).

to repro:
```
$ make bin
$ go test -v ./tests/system -run 'TestSuite/sessions/events.txtar'
```


tests/system/testdata/sessions/events.txtar
```diff
@@ -49,16 +49,18 @@ project:
...
--- program.star --
-def on_http_start():
-  s1 = subscribe("http2")
-  s2 = subscribe("http1")
+-- program.py --
+from autokitteh import subscribe, next_event
+
+def on_http_start(_):
+  s1 = subscribe("http2", "true")
+  s2 = subscribe("http1", "true")
```
tebeka: Can comment on lines below, but subscribe works with one argument that can be a `str` or `list[str]`

```python
  print(4, next_event([s1, s2], timeout=2)['body']['bytes'])
  print(5, next_event([s1, s2], timeout=2)['body']['bytes'])
```

tebeka (approved):

tests/system/testdata/sessions/events.txtar
itayd: done

PR #1131, opened 2025-02-25
fix: index event org for better lists

itayd:


itayd:
@efiShtain for some reason RIGHT JOIN produces some empty rows. only INNER JOIN does the trick. let's discuss.

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -158,6 +158,7 @@ type Event struct {
 	Base
 
 	ProjectID     uuid.UUID  `gorm:"index;type:uuid"` // TODO(authz-migration): not null.
+	OrgID         uuid.UUID  `gorm:"index;type:uuid"` // use only for list.
```
efiShtain: if we want to have the foreign key constraint, you should also add a field 
Org of type Org

migrations/postgres/20250225055027_event_org_id.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- modify "events" table
+ALTER TABLE "events" ADD COLUMN "org_id" uuid NULL;
```
efiShtain: I think we should make sure it is not null
but we need to do it in steps, 
this migration, back fill the data, then alter the column to be not null
events must belong to an org

internal/backend/db/dbgorm/helpers.go
```diff
@@ -45,7 +45,7 @@ func withProjectOrgID(q *gorm.DB, oid sdktypes.OrgID, table string) *gorm.DB {
 		return q
 	}
 
-	return q.Joins(fmt.Sprintf("INNER JOIN projects ON %s.project_id = projects.project_id AND projects.org_id = ?", table), oid.UUIDValue()).Distinct()
+	return q.Joins(fmt.Sprintf("INNER JOIN projects ON %s.project_id = projects.project_id AND projects.org_id = ?", table), oid.UUIDValue())
```
efiShtain, line 48: once you have the org id, why do we need this join at all ?

we do  need a composite index on (org_id, seq desc) so we can list events backward by org id

itayd: there are queries that still require a join with projects, we just obviated that for events.

will add the index.

migrations/postgres/20250225055027_event_org_id.sql
itayd: yeah the steps thing is why i haven't done it yet as non-null. will do that in the future.

internal/backend/db/dbgorm/helpers.go
itayd: found an issue with event_type_seq index, fixed that as well.

tests/system/server.go
```diff
@@ -63,6 +63,7 @@ func startAKServer(t *testing.T, ctx context.Context, akPath string, userCfg map
 		"http.addr_filename":                  serverHTTPAddrFile, // In the test's temporary directory.
 		"authhttpmiddleware.use_default_user": "false",
 		"svc.seed_objects_path":               seedObjectsPath,
+		"db.debug":                            "true",
```
daabr: No, this is too noisy for normal system tests, which are too noisy already. It's ok to use it in local testing, not in CI.

itayd: fixed

daabr (approved):

PR #1132, opened 2025-02-25
feat(python): autokitteh module functions should not run in an activity by default

itayd:
motivation: when initializing clients, don't have to do it in the module globals

### DIFF
--- a/runtimes/pythonrt/py-sdk/autokitteh/__init__.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/__init__.py
@@ -7,10 +7,6 @@ from .events import next_event, subscribe, unsubscribe, start
 from . import errors
 
 
-def nop():
-    pass
-
-
 __all__ = [
     "AttrDict",
     "Event",
@@ -20,5 +16,4 @@ __all__ = [
     "start",
     "subscribe",
     "unsubscribe",
-    "nop",
 ]
--- a/runtimes/pythonrt/py-sdk/autokitteh/__init__.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/__init__.py
@@ -7,6 +7,10 @@ from .events import next_event, subscribe, unsubscribe, start
 from . import errors
 
 
+def nop():
+    pass
+
+
 __all__ = [
     "AttrDict",
     "Event",
@@ -16,4 +20,5 @@ __all__ = [
     "start",
     "subscribe",
     "unsubscribe",
+    "nop",
 ]
--- a/runtimes/pythonrt/runner/call.py
+++ b/runtimes/pythonrt/runner/call.py
@@ -81,7 +81,7 @@ class AKCall:
             return mark
 
         fnmod = fn.__module__
-        if fnmod == ak_mod_name or fnmod.startswith(ak_mod_name + "."):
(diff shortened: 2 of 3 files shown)


runtimes/pythonrt/py-sdk/autokitteh/__init__.py
```diff
@@ -7,6 +7,10 @@
 from . import errors
 
 
+def nop():
```
tebeka: The `py-sdk` is for external users, not for testing. This should not be here.
You can use `check_connection_name` or another function from `connections` or another function like that.

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
```diff
@@ -16,4 +20,5 @@
     "start",
     "subscribe",
     "unsubscribe",
+    "nop",
```
tebeka: Remove :)

runtimes/pythonrt/runner/call.py
```diff
@@ -76,6 +77,9 @@ def should_run_as_activity(self, fn):
         if is_marked_activity(fn):
             return True
 
+        if fn.__module__ == autokitteh.__name__:
```
tebeka: Need to check for sub modules as well, something like:
```python
ak_mod = autokitteh.__module__
ak_mod_prefix = f"{ak_mod}."



def is_ak_fn(fn):
    if fn.__module__ == ak_mod:
        return True

    # Sub module
    return fn.__module__.startswith(ak_mod_prefix)
```

tests/sessions/python_autokitteh_funcs.txtar
```diff
@@ -0,0 +1,10 @@
+-- main.py main --
+
+from autokitteh import nop
+
+nop()
+
+def main(_):
+    nop()
```
tebeka:
```suggestion
from autokitteh.connections import check_connection_name

check_connection_name("Grumpy")

def main(_):
    check_connection_name("Garfield")
```

tebeka (changes requested):

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
itayd: how would i test it then?

tests/sessions/python_autokitteh_funcs.txtar
itayd: `check_connection_name` is an internal sdk function, but i guess i can access it somehow?

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
itayd: done

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
itayd: done

tests/sessions/python_autokitteh_funcs.txtar
itayd: done

runtimes/pythonrt/runner/call.py
itayd: done

tebeka (approved):

PR #1133, opened 2025-02-25
feat(signals): support temporal signals in sessions

itayd:
`next_event` is great, but it is best kept for events. added here is `signal` and `next_signal` which work natively with temporal signals. additionally, child sessions send a completion signal to their parents.

motivation: now signals can work in session_tests, which next_event doesn't since it requires the event. additonally, in tempokitteh (autokitteh over temporal without upper layers and all that jazz) can work with signals.

TODO:
- [ ] documentation (docs.autokitteh.com) (separate pr)
- [ ] kittehub examples (separate pr)
- [x] python support (separate pr)


efiShtain:
I see temporal as an implementation detail,
everything we are not abstracting will make us depends on temporal even more
in cases we might want to run without temporal, this signals won't be supported at all and require lots of knowledge from the user

in autokitteh, any external input is currently a trigger, 
not sure we should break it (but I'm not strongly against it)

maybe we should introduce ak_event of type session_completed / session_started etc...
it keeps working with the same constructs as today

what do you think?

internal/backend/sessions/sessionworkflows/callbacks_signals.go
```diff
@@ -0,0 +1,94 @@
...
+
+		if cancelled {
+			return nil, wctx.Err()
+		}
+
+		if !signal.Source.IsValid() {
+			return nil, nil
+		}
+
+		if !signal.Payload.IsValid() {
```
efiShtain, line 88: maybe an error is better in this case?

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -62,7 +62,7 @@ func runWorkflow(
 	l *zap.Logger,
 	ws *workflows,
 	params *sessionWorkflowParams,
-) (prints []sdkservices.SessionPrint, err error) {
+) (prints []sdkservices.SessionPrint, rv sdktypes.Value, err error) {
```
efiShtain, line 65: can we switch to a struct response ?
not sure why runWorkflow returns an rv value now
what is rv? return value?

efiShtain (approved):
I'm fine with this with the current context we discussed offline

please see my comments and refactor a bit if you think it fits

internal/backend/sessions/sessionworkflows/callbacks_signals.go
itayd: i don't think so - it shouldn't happen, but if it does i think it's ok.

internal/backend/sessions/sessionworkflows/workflow.go
itayd: you're probably right, but i think we should do it in another pr.

PR #1135, opened 2025-02-25
feat(python): allow to inhibit activities using a decorator

itayd:
```
"""Decorator to inhibit the execution of Temporal activities.
    Functions using this decorator will not spawn activities even if otherwise
    activities should have been launched. This is useful for performing operations
    that are required to run even on replay (such as various clients creation) and
    are completely deterministic. The function results are not cached and would be
    rerun in case of replay.
    CAVEAT: Do not use this on functions that take a long time to run (more than a second),
    as they will cause the workflow to timeout.
    """
```

NOTE: There are oddities in python rn that prevent proper testing of this. See https://github.com/autokitteh/autokitteh/pull/1134.

### DIFF
--- a/runtimes/pythonrt/py-sdk/autokitteh/decorators.py
+++ b/runtimes/pythonrt/py-sdk/autokitteh/decorators.py
@@ -1,7 +1,7 @@
 """Decorator to mark a function as a Temporal activity."""
 
 ACTIVITY_ATTR = "__activity__"
-INHIBIT_ACTIVITIES_ATTR = "__inhibit_activities__"
+INHIBIT_ACTIVITIES_ATTR = "__ak_inhibit_activities__"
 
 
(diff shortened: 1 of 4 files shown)


runtimes/pythonrt/py-sdk/autokitteh/decorators.py
```diff
@@ -1,6 +1,7 @@
 """Decorator to mark a function as a Temporal activity."""
 
 ACTIVITY_ATTR = "__activity__"
+INHIBIT_ACTIVITIES_ATTR = "__inhibit_activities__"
```
tebeka:
```suggestion
INHIBIT_ACTIVITIES_ATTR = "__ak_inhibit_activities__"
```

tebeka (approved):

PR #1139, opened 2025-02-26
fix(CI): deflake long-running Python tests

daabr:
Changes:

1. Replace shell scripts with Go tests (Go/Python session tests + Starlark runtime tests)
2. Move all `.txtar` files of Python session tests to a separate subdirectory (just rename, no file changes)
3. Create a reusable venv with `uv` in the beginning for all test-cases (with `PYTHONPATH` env var to use the codebase's `py-sdk` instead of PyPI)
4. Reuse some Go code across `/tests`
5. Disable flaky system tests (rename to `.txtar.disabled`)

Note: **using `uv` instead of `pip` is critical**: `pip` takes 1 minute to install, while `uv` takes less than 5 seconds. (Not) saving a full minute in each execution in 3 jobs (Go system tests, Go session tests, Python session tests) adds up very quickly and has dramatic effects on productivity and CI cost.

In light of this, I recommend we also use `uv` in regular runner operations. Availability during restarts and upgrades is also critical.

What this fixes:

1. Python sessions require ~1m to build a runner (i.e. the `lazy venv` server config doesn't apply): https://github.com/autokitteh/autokitteh/actions/runs/13533601682/job/37821006787?pr=1139

2. Worse: system tests and old (shell) Python session tests are not isolated, so reruns run much faster without having to build a new runner
   - System tests not isolated: `pyhttp` test should always take ~60s
   - Shell-based Python session tests not isolated: each test should take ~60s

Refs: ENG-2037, ENG-1984, ENG-2043


itayd:
i'm ok on this PR, but I need @tebeka to approve because his strong feeling about uv.

.github/workflows/go.yml
```diff
@@ -68,16 +75,25 @@ jobs:
...
 
       - name: Test
-        run: gotestsum -f github-actions -- -trimpath ./tests/system/...
+        run: gotestsum -f github-actions -- -timeout 0 -trimpath ./tests/system/...
 
   go-static-analysis:
     name: Go static analysis
     runs-on: ubuntu-latest
-    timeout-minutes: 10
+    timeout-minutes: 15
```
tebeka, line 96: That's a long timeout for static check. Maybe we can switch to something faster? (staticcheck, revive ...)

.github/workflows/go.yml
```diff
@@ -96,4 +112,4 @@ jobs:
           # See: https://github.com/golangci/golangci-lint/releases
           version: v1.64.5
           # https://github.com/golangci/golangci-lint-action/issues/308
-          args: --timeout=7m
+          args: --timeout=10m
```
tebeka, line 115: Why here it's 10m and above 15?

tests/client.go
tebeka, line 1: Since this file doesn't have a `_test.go` suffix, it *might* be built into the `ak` binary, need to be careful with it. Maybe a build tag?

tests/client.go
```diff
@@ -0,0 +1,49 @@
...
+	if akAddr != "" {
+		args = append(args, "--config", "http.service_url=http://"+akAddr)
+	}
+
+	if userToken != "" {
+		args = append(args, "--token", userToken)
+	}
+
+	cmd := exec.CommandContext(ctx, akPath, args...)
+	cmd.WaitDelay = time.Second // Kill the process on timeout.
```
tebeka, line 37: Why not use `timeout`?

tests/client.go
```diff
@@ -0,0 +1,49 @@
...
+	output, err := cmd.CombinedOutput()
+
+	r := &AKResult{
+		Output:     strings.TrimSpace(string(output)),
+		ReturnCode: cmd.ProcessState.ExitCode(),
+	}
+
+	if ee := new(exec.ExitError); errors.As(err, &ee) {
```
tebeka, line 47: Move to a function `IsExitError` or comment what you're doing here.

tests/files.go
```diff
@@ -0,0 +1,99 @@
...
+	"path/filepath"
+	"testing"
+)
+
+var runnerPath = filepath.Join("..", "..", "runtimes", "pythonrt", "runner")
+
+// AKPath returns the absolute path to the AK binary file. It also checks
+// that the current working directory is the calling test's directory, and
+// the AK binary file exists in the "bin" directory in the repository's root.
+func AKPath(t *testing.T) string {
```
tebeka, line 15: Make sure `ak` is in `PATH` on test setup instead of this?

tests/files.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return path
+}
+
+// SwitchToTempDir creates a temporary directory, for test isolation.
+func SwitchToTempDir(t *testing.T, venvPath string) string {
+	tmpPath := t.TempDir()
+
+	origPath, err := os.Getwd()
```
tebeka, line 37: 1.24 added [t.Chdir](https://pkg.go.dev/testing#T.Chdir), maybe we should switch?
Or add a `// TODO` comment on this fact?

tests/server.go
```diff
@@ -0,0 +1,133 @@
...
+	cmd.Env = append(cmd.Env, "AK_WEBPLATFORM__PORT=0")
+
+	// Capture stdout and stderr in a log file.
+	log, err := os.OpenFile(logFilename, os.O_CREATE|os.O_WRONLY|os.O_APPEND, 0o644)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create AK server log file: %w", err)
+	}
+
+	cmd.Stdout = log
+	cmd.Stderr = log
```
tebeka, line 56:
```suggestion
	cmd.Stdout = io.MultiWriter(log, os.Stdout)
	cmd.Stderr = io.MultiWriter(log, os.Stderr)
```

tests/server.go
```diff
@@ -0,0 +1,133 @@
...
+	}
+
+	server.Addr = addr
+	return server, nil
+}
+
+// waitForAddress waits for the AK server to write its address to a file.
+// It polls the file every 0.1s, and returns an empty string if it times out.
+func waitForAddress() string {
+	done := make(chan string)
```
tebeka, line 78: IMO this is an overkill, I usually write:
```go
until := time.Now().Add(startTimeout)
for time.Now() <= until {
    if addr, err := os.ReadFile(addrFilename); err == nil {
        return string(addr)
   }
   time.Sleep(100 * time.Millisecond)
}
return ""
```

I'd also like to get the error log from `ak` if we fail.

tests/server.go
```diff
@@ -0,0 +1,133 @@
...
+	logFilename  = "ak_server.log"
+	startTimeout = 10 * time.Second
+)
+
+// StartAKServer starts the AK server as a subprocess.
+// If the server started but isn't responding, we still return
+// the process details, so the caller can kill the entire process group.
+// It is assumed that the working directory is temporary and isolated.
+func StartAKServer(akPath, akMode string) (*AKServer, error) {
+	cmd := exec.Command(akPath, "up", "--mode", akMode)
```
tebeka, line 31: Add a comment that some CI step is building `ak`

tebeka (approved):
Some comments, nothing major

.github/workflows/go.yml
daabr: This (10m) is the internally-defined timeout just for this step (running `golangci-lint`), the timeout above (15m) is for the entire job. To be clear, the average runtime of all the other steps is usually 1m, the rest is for safety.

tests/files.go
daabr: Good idea, I'll add a TODO and a tracking ticket

.github/workflows/go.yml
daabr: From ENG-1933: `golangci-lint` is a linting framework, not a linting tool. It already runs (a copy of) `staticcheck` by default, and we can add revive if we want: https://golangci-lint.run/usage/linters/#disabled-by-default

Anyway, this check *usually* runs much faster then 10 minutes, but I don't want it to fail even in extreme cases. I haven't actually seen this check taking 10m, but because it can rarely get close to 7m, I'd rather not take the chance.

tests/client.go
daabr: I trust our reviewers to blocks any PR that tries to import anything from `/tests` into other packages.

Anyway, I wouldn't add a `_test` suffix to this file because it's a test helper and not a test in itself.

I don't understand your suggestion about a build tag: I don't know a Go directive comment that applies only to tests or non-tests. See https://pkg.go.dev/cmd/compile

tests/client.go
daabr: a `WaitDelay` is not exactly a timeout in itself, it's the amount of time we allow the process to end on its own *after* its `ctx` is canceled, before Go kills it forcefully. So any non-zero value will do (0 = wait forever, which is not what we want because the process never kills itself). I'm using a second because 1 nano is too little and flaky, and 1 minute is too much.

tests/files.go
daabr: We shouldn't use the `ak` that we have in the `PATH`, but rather a binary that we built just before running this test. In other words, this comment is cancelled out by a different comment below: https://github.com/autokitteh/autokitteh/pull/1139#discussion_r1977023158

But I'll add a comment about that.

tests/files.go
daabr: INT-312

tests/server.go
daabr: No, it's intentional that I'm not writing the server log to stdout/stderr by default, only if there's an error: see `PrintLog` in `server.go`.

tests/server.go
daabr: Using `select` is the right way to implement contextless operations in Go, for process stability and performance. The extra "complexity" is negligible in this case. IMO using `select` is actually more readable, even if not shorter.

Print server log if `ak` fails - we do that, see all the calls to `PrintLog()`.

tests/server.go
daabr: Done

tests/client.go
daabr: Done

PR #1143, opened 2025-02-26
fix: connStatus false positive for pat connections 

mario99logic:


integrations/google/connections/connections.go
```diff
@@ -39,6 +39,10 @@ func ConnStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
 
 		switch at.Value() {
 		case integrations.JSONKey:
+			_, err := google.JWTConfigFromJSON([]byte(vs.GetValue(vars.JSON)))
```
daabr: No need to generate a JWT from the JSON conn var - that's what the conn test function does. It's enough to test that the JSON conn var isn't empty, i.e. available for testing.

daabr (commented):
Why didn't you extend this to all non-OAuth connection status checks?

integrations/google/connections/connections.go
mario99logic: In the Google connection, when attempting to save the connection and then navigating back, the JSONKey variable remains set and is not empty. Therefore, this approach will not resolve the issue.

mario99logic:
I did not extend this to all non-OAuth connections because, aside from Google and GitHub, none of the other integrations save the connection when navigating back after attempting to save it.

integrations/google/creds.go
```diff
@@ -141,39 +151,53 @@ func (h handler) finalize(ctx context.Context, c sdkintegrations.ConnectionInit,
 		return
 	}
 
+	prevVs, err := h.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+	if err != nil {
+		l.Error("Previous JSON key retrieval error", zap.Error(err))
+		c.AbortServerError("previous JSON key retrieval error")
+		return
+	}
+	prevkey := prevVs.Get(vars.JSON).Value()
```
daabr:
```suggestion
	prevKey := prevVs.GetValue(vars.JSON)
```

integrations/google/creds.go
```diff
@@ -141,39 +151,53 @@ func (h handler) finalize(ctx context.Context, c sdkintegrations.ConnectionInit,
...
+
 	// Unique step for Google integrations (specifically Calendar, Forms, and
 	// Gmail): save the auth data before creating/updating event watches.
 	vsl := kittehs.TransformMapToList(vs.ToMap(), func(_ sdktypes.Symbol, v sdktypes.Var) sdktypes.Var {
 		return v.WithScopeID(sdktypes.NewVarScopeID(cid))
 	})
 
+	handleWatchError := func(err error, watchName string) {
```
daabr: This shouldn't be an inner function inside `finalize()`, it should be a regular function.

integrations/google/creds.go
```diff
@@ -141,39 +151,53 @@ func (h handler) finalize(ctx context.Context, c sdkintegrations.ConnectionInit,
...
+		l.Error(fmt.Sprintf("Google %s watches creation error", watchName), zap.Error(err))
```
daabr: Zap also gas a "sugared" variant that allows `Sprintf` syntax... But it's annoying to use when you add attributes after the message, so this is fine, just FYI.

integrations/google/creds.go
```diff
@@ -141,39 +151,53 @@ func (h handler) finalize(ctx context.Context, c sdkintegrations.ConnectionInit,
...
+		c.AbortServerError(watchName + " watches creation error")
+		jsonErr := h.restoreJSONKey(ctx, cid, prevkey)
```
daabr:
```suggestion
		jsonErr := h.restoreJSONKey(ctx, cid, prevKey)
```

integrations/google/creds.go
```diff
@@ -128,6 +128,16 @@ func (h handler) saveFormID(ctx context.Context, c sdkintegrations.ConnectionIni
 	return nil
 }
 
+// restoreJSONKey updates the JSON key value after a connection setup fails.
+// If a previous JSON key existed, it's restored; otherwise, it's set to empty.
+func (h handler) restoreJSONKey(ctx context.Context, cid sdktypes.ConnectionID, prevKey string) error {
+	v := sdktypes.NewVar(vars.JSON).SetValue(prevKey).WithScopeID(sdktypes.NewVarScopeID(cid))
```
daabr: Also need to add `SetSecret(true)` after `SetValue`

daabr (approved):
Approved with minor things to fix

PR #1144, opened 2025-02-26
fix(python): Dedent bug when traversing AST

tebeka:
Fixes: ENG-2036


itayd:
@tebeka it would really help if you would explain the description what the problem is and how it was solved...

itayd (approved):

PR #1148, opened 2025-02-27
fix(sessions): properly create a child workflow

itayd:
so far when a parent calls `start` a new session is created with the session interface, but it would happen every time on replay. this fixes it to also do a proper child session.

also use `SignalChildWorkflows` to signal child workflows.


internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -101,41 +100,77 @@ func (ws *workflows) StartWorkers(ctx context.Context) error {
...
+		memo["session__"+k] = v
+	}
+
+	return memo
+}
+
+func (ws *workflows) StartChildWorkflow(wctx workflow.Context, childSession sdktypes.Session, parentSessionData *sessiondata.Data) (workflow.ChildWorkflowFuture, error) {
+	childSessionID := childSession.ID()
+
+	l := ws.l.Sugar().With("session_id", childSessionID)
```
daabr: `Sugar()` isn't needed here, in fact it only makes line 157 worse - call `With()` before `Sugar()`

daabr (commented):
Almost all the functions in this PR require documentation to explain how they fit in the bigger picture. Without the context provided by Copilot I wouldn't be able to review this PR, and this context will be lost when someone looks at the merged code and not this PR.

daabr (approved):

internal/backend/sessions/sessionworkflows/workflows.go
itayd: done

PR #1151, opened 2025-02-27
fix: improve timestamp parsing

daabr:
Improvement in parsing of Go timestamps (`time.Time.String()`). Related to INT-277, but not a fix for it.

Also added unit tests for some shared integrations functions related to data migrations.

For additional context, see https://github.com/autokitteh/autokitteh/pull/1151#discussion_r1976560850 and the full files being modified.

### DIFF
--- a/integrations/common/time.go
+++ b/integrations/common/time.go
@@ -1,0 +1,15 @@
+package common
+
+import (
+	"regexp"
+	"time"
+)
+
+// ParseGoTimestamp parses a [time.Time.String] string.
+// It ignores unnecessary suffixes: sub-seconds and extra timezone details.
+// Local example: "2025-02-28 10:04:21.024 -0800 PST m=+3759.281638293".
+func ParseGoTimestamp(ts string) (time.Time, error) {
+	ts = regexp.MustCompile(` [A-Z].*`).ReplaceAllString(ts, "")
+	ts = regexp.MustCompile(`\.\d+`).ReplaceAllString(ts, "")
+	return time.Parse("2006-01-02 15:04:05 -0700", ts)
+}
--- a/integrations/common/time.go
+++ b/integrations/common/time.go
@@ -9,7 +9,10 @@ import (
 // It ignores unnecessary suffixes: sub-seconds and extra timezone details.
 // Local example: "2025-02-28 10:04:21.024 -0800 PST m=+3759.281638293".
 func ParseGoTimestamp(ts string) (time.Time, error) {
+	// Get rid of the " PST m=..." suffix (keep the numeric timezone).
 	ts = regexp.MustCompile(` [A-Z].*`).ReplaceAllString(ts, "")
+	// Get rid of the timestamp's sub-second suffix
+	// (it's both unnecessary and nondeterministic).
 	ts = regexp.MustCompile(`\.\d+`).ReplaceAllString(ts, "")
 	return time.Parse("2006-01-02 15:04:05 -0700", ts)
 }
--- a/integrations/common/migrations.go
+++ b/integrations/common/migrations.go
@@ -72,7 +72,9 @@ func MigrateDateTimeToRFC3339(ctx context.Context, v sdkservices.Vars, vsid sdkt
 		return nil
 	}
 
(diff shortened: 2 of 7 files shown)


itayd:
please add tests.

integrations/common/time.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"regexp"
+	"time"
+)
+
+// ParseGoTimestamp parses a [time.Time.String] string.
+// It ignores unnecessary suffixes: sub-seconds and extra timezone details.
+// Local example: "2025-02-28 10:04:21.024 -0800 PST m=+3759.281638293".
+func ParseGoTimestamp(ts string) (time.Time, error) {
```
efiShtain, line 11: this feels really off, 
time parsing is one of the hardest things to do
I would be very surprised if go has no existing solution for this stuff

why do we need to manipulate string regex here ?

daabr: We've been using `time.Time.String()` for OAuth (and a few other) timestamps in connection vars. This was a mistake exactly because there really isn't an existing solution in Go for its own `String()` output, it was never meant for storage and parsing.

See: https://pkg.go.dev/time#pkg-constants and https://pkg.go.dev/time#Time.String

This code will be used in order to convert these values to RFC-3339 which does have easy and standard parsing in Go and Python.

The regular expressions are needed in order to discard non-important / non-deterministic / hard-to-parse parts of the string:

- Sub-seconds: some platforms show ms, some ns, some don't have it at all; anyway it doesn't matter for our purposes
- Timezone name: almost impossible to parse correctly, and we already have the TZ shift which is better because it's an accurate number
- Monotonic clock reading: https://stackoverflow.com/questions/52809661/what-is-m-in-go-time-string#:~:text=Explanation%20from%20time.,a%20decimal%20number%20of%20seconds.

tebeka: Are these existing values in the DB? If so:
- Write "one-off" script to convert then
- Store time in TIMESTAMP column

To create a good string from time use t.MarshalText

This PR seems like it's fixing the wrong issue. But maybe I'm missing something.

integrations/common/time.go
```diff
@@ -0,0 +1,15 @@
...
+	ts = regexp.MustCompile(` [A-Z].*`).ReplaceAllString(ts, "")
+	ts = regexp.MustCompile(`\.\d+`).ReplaceAllString(ts, "")
```
tebeka: Add a comment with example of what this regexp does.

integrations/common/time.go
```diff
@@ -0,0 +1,15 @@
...
+	return time.Parse("2006-01-02 15:04:05 -0700", ts)
```
tebeka, line 17: Or use UnmarshalText

tebeka (commented):
I'm confused about what this PR solves, what is the actual issue?

integrations/common/time.go
daabr: I can't write a one-off script because:

1. There is no reliable way to convert Go timestamp strings using SQL
2. Even if there was, these values (and their assumed format) are used in *a lot* of places across our Go and Python code
3. The vars DB table doesn't (and shouldn't) support timestamp data types, only strings

So to prevent the risk of outages and data corruption, we have to migrate this slowly and carefully.

I do intend to use RFC3339 (which is what `MarshalText` uses) but that's beyond the scope of this PR.

integrations/common/time.go
daabr: Impossible by definition: https://pkg.go.dev/time#Time.UnmarshalText requires RFC3339, but these timestamp strings have a different format.

integrations/common/time.go
daabr: There's already an example in the function doc.

itayd (commented):
talked to daniel, this is a temporary fix until we get rid of the parsing. it just gets rid of the extra stuff, not the -0000, but the PST m=....

itayd (approved):

integrations/common/time.go
daabr: added more detailed comments

PR #1153, opened 2025-02-28
refactor: runtime start should be session id structure agnostic

itayd:
`start` should not care about how a session id looks. motivation: in tempokitteh there is no session id, but a temporal workflow id.

### DIFF
--- a/internal/backend/sessions/sessionworkflows/callbacks_misc.go
+++ b/internal/backend/sessions/sessionworkflows/callbacks_misc.go
@@ -10,8 +10,8 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-func (w *sessionWorkflow) start(wctx workflow.Context) func(context.Context, sdktypes.RunID, sdktypes.CodeLocation, map[string]sdktypes.Value, map[string]string) (sdktypes.SessionID, error) {
-	return func(ctx context.Context, rid sdktypes.RunID, loc sdktypes.CodeLocation, inputs map[string]sdktypes.Value, memo map[string]string) (sdktypes.SessionID, error) {
+func (w *sessionWorkflow) start(wctx workflow.Context) func(context.Context, sdktypes.RunID, sdktypes.CodeLocation, map[string]sdktypes.Value, map[string]string) (string, error) {
+	return func(ctx context.Context, rid sdktypes.RunID, loc sdktypes.CodeLocation, inputs map[string]sdktypes.Value, memo map[string]string) (string, error) {
 		l := w.l.With(zap.Any("rid", rid), zap.Any("loc", loc), zap.Any("inputs", inputs), zap.Any("memo", memo))
 
 		l.Info("child session start requested")
@@ -23,12 +23,12 @@ func (w *sessionWorkflow) start(wctx workflow.Context) func(context.Context, sdk
 			WithNewID()
 
 		if err := workflow.ExecuteActivity(wctx, createSessionActivityName, session).Get(wctx, nil); err != nil {
-			return sdktypes.InvalidSessionID, err
+			return "", err
 		}
 
 		f, err := w.ws.StartChildWorkflow(wctx, session, w.data)
 		if err != nil {
-			return sdktypes.InvalidSessionID, err
+			return "", err
 		}
 
 		sid := session.ID()
@@ -37,7 +37,7 @@ func (w *sessionWorkflow) start(wctx workflow.Context) func(context.Context, sdk
 
 		w.l.Info("child session started", zap.Any("child", sid), zap.Any("parent", w.data.Session.ID()))
 
-		return sid, nil
+		return sid.String(), nil
 	}
 }
 
--- a/internal/backend/sessions/sessionworkflows/callbacks_signals.go
+++ b/internal/backend/sessions/sessionworkflows/callbacks_signals.go
@@ -18,13 +18,18 @@ const userSignalNamePrefix = "user_"
 func userSignalName(name string) string               { return userSignalNamePrefix + name }
 func sessionSignalName(sid sdktypes.SessionID) string { return sid.String() }
 
-func (w *sessionWorkflow) signal(wctx workflow.Context) func(context.Context, sdktypes.RunID, sdktypes.SessionID, string, sdktypes.Value) error {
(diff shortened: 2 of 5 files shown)


internal/backend/sessions/sessionworkflows/callbacks_misc.go
```diff
@@ -10,8 +10,8 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-func (w *sessionWorkflow) start(wctx workflow.Context) func(context.Context, sdktypes.RunID, sdktypes.CodeLocation, map[string]sdktypes.Value, map[string]string) (sdktypes.SessionID, error) {
-	return func(ctx context.Context, rid sdktypes.RunID, loc sdktypes.CodeLocation, inputs map[string]sdktypes.Value, memo map[string]string) (sdktypes.SessionID, error) {
+func (w *sessionWorkflow) start(wctx workflow.Context) func(context.Context, sdktypes.RunID, sdktypes.CodeLocation, map[string]sdktypes.Value, map[string]string) (string, error) {
```
daabr, line 13: This function should have a clear https://go.dev/doc/comment#func that explains what it does, and why it's returning a string, i.e. the 2 use-cases of calling this function: for starting regular AutoKitteh sessions, and "lightweight" wrappers for Temporal workflows.

The comment should also explain the meaning of `""`.

internal/backend/sessions/sessionworkflows/callbacks_signals.go
```diff
@@ -18,14 +18,19 @@ const userSignalNamePrefix = "user_"
 func userSignalName(name string) string               { return userSignalNamePrefix + name }
 func sessionSignalName(sid sdktypes.SessionID) string { return sid.String() }
 
-func (w *sessionWorkflow) signal(wctx workflow.Context) func(context.Context, sdktypes.RunID, sdktypes.SessionID, string, sdktypes.Value) error {
-	return func(_ context.Context, _ sdktypes.RunID, sid sdktypes.SessionID, name string, v sdktypes.Value) error {
+func (w *sessionWorkflow) signal(wctx workflow.Context) func(context.Context, sdktypes.RunID, string, string, sdktypes.Value) error {
```
daabr, line 21: Same comment as above: how/when is this function used, what does this function return? Significant values? This is even more important because you return an anonymous function with multiple fields that contains arbitrary values which can be anything (2 strings and a `skdtypes.Value`).

In fact, consider also defining a type for this returned function, and document that type. This will also simplify this function's signature and make it more readable.

internal/backend/sessions/sessionworkflows/callbacks_signals.go
```diff
@@ -18,14 +18,19 @@ const userSignalNamePrefix = "user_"
...
+	return func(_ context.Context, _ sdktypes.RunID, dst string, name string, v sdktypes.Value) error {
```
daabr, line 22: Why does this function receive a context and a run ID if it doesn't need them? What is `dst` (destination of what)? What do `name` and `value` expected to contain?

internal/backend/sessions/sessionworkflows/callbacks_signals.go
```diff
@@ -18,14 +18,19 @@ const userSignalNamePrefix = "user_"
...
 		if !v.IsValid() {
 			v = sdktypes.Nothing
 		}
 
 		var f workflow.Future
 
+		sid, err := sdktypes.ParseSessionID(dst)
```
daabr, line 29: What's the point of accepting a string instead of a session ID object, if you return an error when the string isn't a valid sessions ID?

internal/backend/sessions/sessionworkflows/callbacks_signals.go
```diff
@@ -93,10 +98,16 @@ func (w *sessionWorkflow) nextSignal(wctx workflow.Context) func(context.Context
 			return nil, wctx.Err()
 		}
 
-		if signal == nil {
+		if signal == nil || signal.Name == "" {
```
daabr, line 101: What does it mean when a signal is `nil`? What does it mwan when it's not `nil` but has an empty name?

internal/backend/sessions/sessionworkflows/callbacks_signals.go
```diff
@@ -93,10 +98,16 @@ func (w *sessionWorkflow) nextSignal(wctx workflow.Context) func(context.Context
...
 			return nil, nil
 		}
 
+		if sid, err := sdktypes.ParseSessionID(signal.Name); err == nil {
```
daabr, line 105: How come lightweight wrappers of Temporal workflows don't needs this? (i.e. why only regular AK sessions?)

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -279,7 +279,7 @@ func (s *workerGRPCHandler) StartSession(ctx context.Context, req *userCode.Star
 		return &userCode.StartSessionResponse{Error: err.Error()}, nil
 	}
 
-	return &userCode.StartSessionResponse{SessionId: resp.value.(sdktypes.SessionID).String()}, nil
+	return &userCode.StartSessionResponse{SessionId: resp.value.(string)}, nil
```
daabr, line 282: `value`'s type is `any`, you can't just assume it's a string withtout checking the type first.

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -461,7 +456,7 @@ func (s *workerGRPCHandler) Signal(ctx context.Context, req *userCode.SignalRequ
 	}
 
 	if resp.err != nil {
-		err = status.Errorf(codes.Internal, "signal(%s,%s) -> %s", pbsig.Name, sid.String(), err)
+		err = status.Errorf(codes.Internal, "signal(%s,%s) -> %s", pbsig.Name, req.SessionId, err)
```
daabr, line 459:
```suggestion
		err = status.Errorf(codes.Internal, "signal(%q, %q) -> %s", pbsig.Name, req.SessionId, err)
```

runtimes/pythonrt/worker_grpc_handler.go
tebeka: Agree, the type assertion can panic.

PR #1154, opened 2025-02-28
prep sessions for tempokitteh

itayd:
bundled minor changes as prep for tempokitteh:
- add accessors to `BuildExport`
- specify parent policy for temporal, currently to not terminate child workflows when parent completes
- prep sessioncalls to be reused without need for various services

### DIFF
--- a/internal/backend/sessions/sessioncalls/calls.go
+++ b/internal/backend/sessions/sessioncalls/calls.go
@@ -7,6 +7,7 @@ import (
 	"sync"
 
 	enumspb "go.temporal.io/api/enums/v1"
+	"go.temporal.io/sdk/client"
 	"go.temporal.io/sdk/temporal"
 	"go.temporal.io/sdk/worker"
 	"go.temporal.io/sdk/workflow"
@@ -31,8 +32,9 @@ type CallParams struct {
 }
 
 type Calls interface {
-	StartWorkers(context.Context) error
+	StartWorkers(context.Context, client.Client) error
 	Call(ctx workflow.Context, params *CallParams) (sdktypes.SessionCallAttemptResult, error)
+	RegisterSessionCallActivity(r worker.ActivityRegistry)
 }
 
 type calls struct {
@@ -62,7 +64,7 @@ type calls struct {
 
 const generalTaskQueueName = "session_call_activities"
 
-func uniqueWorkerCallTaskQueueName() string {
+func UniqueWorkerCallTaskQueueName() string {
 	return fmt.Sprintf("%s_%s", generalTaskQueueName, fixtures.ProcessID())
 }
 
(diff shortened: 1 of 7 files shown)


sdk/sdktypes/build_export.go
```diff
@@ -42,3 +43,9 @@ func (r BuildExport) WithLocation(loc CodeLocation) BuildExport {
 func (r BuildExport) WithSymbol(sym Symbol) BuildExport {
 	return BuildExport{r.forceUpdate(func(pb *BuildExportPB) { pb.Symbol = sym.String() })}
 }
+
+func (r BuildExport) Location() CodeLocation {
+	return kittehs.Must1(CodeLocationFromProto(r.read().Location))
```
daabr, line 48: Don't `panic`, return an invalid location if parsing fails.

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -403,7 +403,7 @@ func (w *sessionWorkflow) removeEventSubscription(wctx workflow.Context, signalI
 	delete(w.lastReadEventSeqForSignal, signalID)
 }
 
-func (w *sessionWorkflow) run(wctx workflow.Context, l *zap.Logger) (prints []sdkservices.SessionPrint, retVal sdktypes.Value, err error) {
+func (w *sessionWorkflow) run(wctx workflow.Context, l *zap.Logger, globals map[string]sdktypes.Value) (prints []sdkservices.SessionPrint, retVal sdktypes.Value, err error) {
```
daabr, line 409: Document the meaning and required/expected/optional keys of `globals`

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -75,10 +77,13 @@ func New(l *zap.Logger, config Config, svcs *sessionsvcs.Svcs) Calls {
 	}
 }
 
-func (cs *calls) StartWorkers(ctx context.Context) error {
-	tc := cs.svcs.Temporal.TemporalClient()
-	cs.generalWorker = temporalclient.NewWorker(cs.l.Named("sessionscallsworker"), tc, generalTaskQueueName, cs.config.GeneralWorker)
-	cs.uniqueWorker = temporalclient.NewWorker(cs.l.Named("sessionscallsworker"), tc, uniqueWorkerCallTaskQueueName(), cs.config.UniqueWorker)
+func (cs *calls) StartWorkers(ctx context.Context, temporal client.Client) error {
```
daabr: Don't use a variable name that obscures an import name

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -31,8 +32,9 @@ type CallParams struct {
 }
 
 type Calls interface {
-	StartWorkers(context.Context) error
+	StartWorkers(context.Context, client.Client) error
```
efiShtain, line 35: why?

internal/backend/sessions/sessioncalls/calls.go
```diff
@@ -112,7 +117,7 @@ func (cs *calls) Call(wctx workflow.Context, params *CallParams) (sdktypes.Sessi
 
 	var attempt uint32 // relevant only if not using temporal for session logs.
 
-	if !params.UseTemporalForSessionLogs {
+	if !params.UseTemporalForSessionLogs && cs.svcs != nil && cs.svcs.DB != nil {
```
efiShtain, line 120: why?

efiShtain (commented):
not sure I get why this changes are needed at all

sdk/sdktypes/build_export.go
itayd: if you call this function, it means that the state of BuildExport must be valid. InvalidLocation would mean none, but it must never error.

internal/backend/sessions/sessionworkflows/workflow.go
itayd: commented

internal/backend/sessions/sessioncalls/calls.go
itayd: renamed

internal/backend/sessions/sessioncalls/calls.go
itayd: because the activities require the DB. the db will not be present when using tempokitteh. going to remove these activities anyway in a different pr.

internal/backend/sessions/sessioncalls/calls.go
itayd: explicitly set which client to use when starting workers. when using tempokitteh, svcs are not there.

efiShtain: I don't like the way this goes

everything is already convoluted, adding another layer of indirection / args makes stuff even harder

internal/backend/sessions/sessioncalls/calls.go
efiShtain: the condition is not readable, 
I can't look at it and understand what you try to achieve and why

tebeka:
@itayd Can we close this one?

PR #1157, opened 2025-03-03
feat: Add Zoom connTest function 

mario99logic:
- Implements OAuth and Server-to-Server authentication for Zoom API
- Adds connection tests to verify API access
- Includes token refresh logic for OAuth authentication

### DIFF
--- a/integrations/zoom/auth.go
+++ b/integrations/zoom/auth.go
@@ -3,6 +3,7 @@ package zoom
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
@@ -10,11 +11,12 @@ import (
 	"strings"
 	"time"
 
+	"go.autokitteh.dev/autokitteh/integrations/common"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
 var (
-	acountID     = sdktypes.NewSymbol("private_account_id")
+	accountID    = sdktypes.NewSymbol("private_account_id")
 	clientID     = sdktypes.NewSymbol("private_client_id")
 	clientSecret = sdktypes.NewSymbol("private_client_secret")
 )
@@ -56,44 +58,67 @@ func serverToken(ctx context.Context, app *privateApp) (string, error) {
 	return tokenData.AccessToken, nil
 }
 
-// ServerToServerTest handles the Server-to-Server authentication test for Zoom
-// (Based on: https://developers.zoom.us/docs/internal-apps/s2s-oauth/).
-func ServerToServerTest(ctx context.Context, vs sdktypes.Vars) (sdktypes.Status, error) {
-	id := vs.GetValue(acountID)
-	clientID := vs.GetValue(clientID)
-	secret := vs.GetValue(clientSecret)
-
-	app := privateApp{
-		AccountID:    id,
-		ClientID:     clientID,
-		ClientSecret: secret,
-	}
-
-	// Get a server-to-server token
-	token, err := serverToken(ctx, &app)
-	if err != nil {
-		return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
-	}
-
-	url := "https://api.zoom.us/v2/users"
+func makeAPICall(ctx context.Context, token string, url string) (*http.Response, error) {
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
 	if err != nil {
-		return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
+		return nil, fmt.Errorf("Failed to create request: %v", err)
 	}
 
 	req.Header.Set("Authorization", "Bearer "+token)
 	req.Header.Set("Content-Type", "application/json")
 
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+	}
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("API request failed: %v", err)
+	}
+	return resp, nil
+}
+
+// refreshTokenReq sends a request to Zoom's OAuth endpoint to refresh an expired access token.
+func refreshTokenReq(ctx context.Context, refreshToken string, clientID string, clientSecret string, vs sdktypes.Vars) (string, error) {
+	data := url.Values{}
+	data.Set("grant_type", "refresh_token")
+	data.Set("refresh_token", refreshToken)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost,
+		"https://zoom.us/oauth/token", strings.NewReader(data.Encode()))
+	if err != nil {
+		return "", fmt.Errorf("failed to create refresh request: %v", err)
+	}
+
+	req.SetBasicAuth(clientID, clientSecret)
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
 	client := &http.Client{Timeout: 10 * time.Second}
 	resp, err := client.Do(req)
 	if err != nil {
-		return sdktypes.NewStatus(sdktypes.StatusCodeError, "API request failed"), nil
+		return "", fmt.Errorf("refresh token request failed: %v", err)
 	}
 	defer resp.Body.Close()
 
-	if resp.StatusCode == http.StatusOK {
-		return sdktypes.NewStatus(sdktypes.StatusCodeOK, "Server-to-Server connection successful"), nil
+	if resp.StatusCode != http.StatusOK {
+		return "", errors.New("failed to refresh token")
 	}
 
-	return sdktypes.NewStatus(sdktypes.StatusCodeError, "Failed to connect to Zoom API"), nil
+	var tokenResp struct {
+		AccessToken  string `json:"access_token"`
+		RefreshToken string `json:"refresh_token"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&tokenResp); err != nil {
+		return "", fmt.Errorf("failed to decode token response: %v", err)
+	}
+
+	if err := vs.Set(common.OAuthAccessTokenVar, tokenResp.AccessToken, true); err != nil {
+		return "", errors.New("Failed to save new access token")
+	}
+	if err := vs.Set(common.OAuthRefreshTokenVar, tokenResp.RefreshToken, true); err != nil {
+		return "", errors.New("Failed to save new refresh token")
+	}
+
+	return tokenResp.AccessToken, nil
 }
--- a/integrations/zoom/auth.go
+++ b/integrations/zoom/auth.go
@@ -8,39 +8,21 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"strings"
-	"time"
 
 	"go.autokitteh.dev/autokitteh/integrations/common"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-var (
-	accountID    = sdktypes.NewSymbol("private_account_id")
-	clientID     = sdktypes.NewSymbol("private_client_id")
-	clientSecret = sdktypes.NewSymbol("private_client_secret")
-)
-
 // serverToken retrieves a Server-to-Server (2-legged OAuth) token, using the connection's
 // internal app details (based on: https://developers.zoom.us/docs/internal-apps/s2s-oauth/).
-func serverToken(ctx context.Context, app *privateApp) (string, error) {
+func serverToken(ctx context.Context, app privateApp) (string, error) {
 	data := url.Values{}
 	data.Set("grant_type", "account_credentials")
 	data.Set("account_id", app.AccountID)
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost,
-		"https://zoom.us/oauth/token", strings.NewReader(data.Encode()))
+	resp, err := httpPost(ctx, data, app.ClientID, app.ClientSecret)
 	if err != nil {
-		return "", fmt.Errorf("failed to create token request: %v", err)
-	}
-
-	req.SetBasicAuth(app.ClientID, app.ClientSecret)
-	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
-
-	client := &http.Client{Timeout: 10 * time.Second}
-	resp, err := client.Do(req)
-	if err != nil {
-		return "", fmt.Errorf("token request failed: %v", err)
+		return "", fmt.Errorf("refresh token request failed: %v", err)
 	}
 	defer resp.Body.Close()
 
@@ -49,8 +31,7 @@ func serverToken(ctx context.Context, app *privateApp) (string, error) {
 		return "", fmt.Errorf("failed to get token: status %d: %s", resp.StatusCode, string(body))
 	}
 
-	var tokenData tokenResp
-
+	var tokenData common.OAuthData
 	if err := json.NewDecoder(resp.Body).Decode(&tokenData); err != nil {
 		return "", fmt.Errorf("failed to decode token response: %v", err)
 	}
@@ -58,43 +39,13 @@ func serverToken(ctx context.Context, app *privateApp) (string, error) {
 	return tokenData.AccessToken, nil
 }
 
-func makeAPICall(ctx context.Context, token string, url string) (*http.Response, error) {
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
-	if err != nil {
-		return nil, fmt.Errorf("Failed to create request: %v", err)
-	}
-
-	req.Header.Set("Authorization", "Bearer "+token)
-	req.Header.Set("Content-Type", "application/json")
-
-	client := &http.Client{
-		Timeout: 10 * time.Second,
-	}
-
-	resp, err := client.Do(req)
-	if err != nil {
-		return nil, fmt.Errorf("API request failed: %v", err)
-	}
-	return resp, nil
-}
-
-// refreshTokenReq sends a request to Zoom's OAuth endpoint to refresh an expired access token.
-func refreshTokenReq(ctx context.Context, refreshToken string, clientID string, clientSecret string, vs sdktypes.Vars) (string, error) {
+// refreshToken sends a request to Zoom's OAuth endpoint to refresh an expired access token.
+func refreshToken(ctx context.Context, refreshT string, clientID string, clientSecret string, vs sdktypes.Vars) (string, error) {
 	data := url.Values{}
 	data.Set("grant_type", "refresh_token")
-	data.Set("refresh_token", refreshToken)
+	data.Set("refresh_token", refreshT)
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost,
-		"https://zoom.us/oauth/token", strings.NewReader(data.Encode()))
-	if err != nil {
-		return "", fmt.Errorf("failed to create refresh request: %v", err)
-	}
-
-	req.SetBasicAuth(clientID, clientSecret)
-	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
-
-	client := &http.Client{Timeout: 10 * time.Second}
-	resp, err := client.Do(req)
+	resp, err := httpPost(ctx, data, clientID, clientSecret)
 	if err != nil {
 		return "", fmt.Errorf("refresh token request failed: %v", err)
 	}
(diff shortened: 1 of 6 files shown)


integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
 // serverToken retrieves a Server-to-Server (2-legged OAuth) token, using the connection's
 // internal app details (based on: https://developers.zoom.us/docs/internal-apps/s2s-oauth/).
-func serverToken(_ context.Context, _ *privateApp) (*oauth2.Token, error) {
-	return nil, nil // TODO(INT-247): Implement.
+func serverToken(ctx context.Context, app *privateApp) (string, error) {
+	data := url.Values{}
+	data.Set("grant_type", "account_credentials")
+	data.Set("account_id", app.AccountID)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost,
```
tebeka: Are you sure this is how it works? Usually `url.Values` are used as URL query parameters.

```go
	query := url.Values{}
	query.Set("grant_type", "account_credentials")
	query.Set("account_id", "mcscrooge")
	u, _ := url.Parse("https://zoom.us/oauth/token")
	u.RawQuery = query.Encode()
        authURL := u.String()
```

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+		"https://zoom.us/oauth/token", strings.NewReader(data.Encode()))
+	if err != nil {
+		return "", fmt.Errorf("failed to create token request: %v", err)
+	}
+
+	req.SetBasicAuth(app.ClientID, app.ClientSecret)
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	client := &http.Client{Timeout: 10 * time.Second}
```
tebeka: Why 10 seconds? Also, the recommended way to set a timeout is to use `http.NewRequestWithContext`

```go
	ctx, cancel := context.WithTimeout(context.Background(), 10*time.Second)
	defer cancel()

	req, err := http.NewRequestWithContext(ctx, http.MethodGet, "http://example.com", nil)
	if err != nil {
		return err
	}
	resp, err := http.DefaultClient.Do(req)
	if err != nil {
		return err
	}
	fmt.Println(resp.StatusCode)
	return nil

```

integrations/zoom/vars.go
```diff
@@ -8,3 +8,10 @@ type privateApp struct {
 	ClientSecret string `var:"private_client_secret,secret"`
 	SecretToken  string `var:"private_secret_token,secret"`
 }
+
+type tokenResp struct {
```
tebeka: Since you use it only once to parse JSON, use an anonymous struct in `serverToken`

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+	}
+
+	req.Header.Set("Authorization", "Bearer "+token)
+	req.Header.Set("Content-Type", "application/json")
+
+	client := &http.Client{
+		Timeout: 10 * time.Second,
+	}
+
+	resp, err := client.Do(req)
```
tebeka: You can create one client globally and then use it all over.
Also: Why use pointer and not value?

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+		return nil, fmt.Errorf("API request failed: %v", err)
+	}
+	return resp, nil
+}
+
+// refreshTokenReq sends a request to Zoom's OAuth endpoint to refresh an expired access token.
+func refreshTokenReq(ctx context.Context, refreshToken string, clientID string, clientSecret string, vs sdktypes.Vars) (string, error) {
+	data := url.Values{}
+	data.Set("grant_type", "refresh_token")
+	data.Set("refresh_token", refreshToken)
```
tebeka: See comment about on `url.Values`

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+	"net/url"
+	"strings"
+	"time"
 
-	"golang.org/x/oauth2"
+	"go.autokitteh.dev/autokitteh/integrations/common"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var (
```
daabr: Connection variables should be defined in the `vars.go` file

integrations/zoom/auth.go
daabr: @tebeka Yes, afaik This Is The Way to construct a URL-encoded web form in a POST request: https://stackoverflow.com/questions/19253469/make-a-url-encoded-post-request-using-http-newrequest

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
 // internal app details (based on: https://developers.zoom.us/docs/internal-apps/s2s-oauth/).
-func serverToken(_ context.Context, _ *privateApp) (*oauth2.Token, error) {
-	return nil, nil // TODO(INT-247): Implement.
+func serverToken(ctx context.Context, app *privateApp) (string, error) {
+	data := url.Values{}
+	data.Set("grant_type", "account_credentials")
+	data.Set("account_id", app.AccountID)
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost,
+		"https://zoom.us/oauth/token", strings.NewReader(data.Encode()))
```
daabr: No need to break this line. If you want, you can use a helper var for the body.

integrations/zoom/vars.go
daabr: @mario99logic there are connection vars, so it makes sense to define the in this file,  but `tokenResp` is not a meaningful scope for these vars in this integration.

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+	var tokenData tokenResp
+
+	if err := json.NewDecoder(resp.Body).Decode(&tokenData); err != nil {
+		return "", fmt.Errorf("failed to decode token response: %v", err)
+	}
+
+	return tokenData.AccessToken, nil
+}
+
+func makeAPICall(ctx context.Context, token string, url string) (*http.Response, error) {
```
daabr: 1. This function is just a generic wrapper for an HTTP GET request, there's nothing specific for API calls, so the name is misleading - please rename to `httpGet`
2. There's nothing specifically related to authentication (other than using a bearer token), and it's not used anywhere in this file, so the location of this function is not right - please move to `checks.go`
3. With all the token types being used in this integration, it's worthwhile being explicit about the type/purpose/usage of the `token` argument to avoid confusion - please rename it to `bearerToken` even though it's longer (usually being verbose is bad, but in this case it's helpful)
4. This function isn't documented, see https://tip.golang.org/doc/comment#func

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("API request failed: %v", err)
+	}
+	return resp, nil
+}
+
+// refreshTokenReq sends a request to Zoom's OAuth endpoint to refresh an expired access token.
+func refreshTokenReq(ctx context.Context, refreshToken string, clientID string, clientSecret string, vs sdktypes.Vars) (string, error) {
```
daabr: 1. `Req` in the need is redundant
2. You're using a `privateApp` above, why not here too?

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+	data := url.Values{}
```
daabr: Notice that you're doing something very similar above - you can extract the common code into an `httpPost` helper function

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+	accountID    = sdktypes.NewSymbol("private_account_id")
+	clientID     = sdktypes.NewSymbol("private_client_id")
+	clientSecret = sdktypes.NewSymbol("private_client_secret")
 )
 
 // serverToken retrieves a Server-to-Server (2-legged OAuth) token, using the connection's
 // internal app details (based on: https://developers.zoom.us/docs/internal-apps/s2s-oauth/).
-func serverToken(_ context.Context, _ *privateApp) (*oauth2.Token, error) {
-	return nil, nil // TODO(INT-247): Implement.
+func serverToken(ctx context.Context, app *privateApp) (string, error) {
```
daabr: There's no reason to pass a reference to `privateApp` - it can't (i.e. it must not) be `nil`, and you're not modifying it

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
+		return "", fmt.Errorf("failed to get token: status %d: %s", resp.StatusCode, string(body))
+	}
+
+	var tokenData tokenResp
+
+	if err := json.NewDecoder(resp.Body).Decode(&tokenData); err != nil {
```
daabr: Merely declaring an empty var shouldn't be a separate block of code
```suggestion
	var tokenData tokenResp
	if err := json.NewDecoder(resp.Body).Decode(&tokenData); err != nil {
```

integrations/zoom/vars.go
daabr: Instead of defining this, you can reuse `common.OAuthData` - but you need to add the `json:"..."` annotations to `AccessToken`, `RefreshToken`, `Expiry` ("expires_in" in JSON).

See https://github.com/autokitteh/autokitteh/blob/main/integrations/microsoft/connection/vars.go for an example on how to do both `json` and `var` annotations.

No need to add JSON annotations to the other fields not mentioned out - you don't use them anyway.

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+	if err != nil {
+		return "", fmt.Errorf("refresh token request failed: %v", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", errors.New("failed to refresh token")
+	}
+
+	var tokenResp struct {
```
daabr: Same comment as the one about `tokenResp` in `vars.go`.

Regardless, there's no logic in defining this struct twice, in different files, both inside and outside functions that use it.

integrations/zoom/auth.go
```diff
@@ -2,12 +2,123 @@ package zoom
...
+		AccessToken  string `json:"access_token"`
+		RefreshToken string `json:"refresh_token"`
+	}
+
+	if err := json.NewDecoder(resp.Body).Decode(&tokenResp); err != nil {
+		return "", fmt.Errorf("failed to decode token response: %v", err)
+	}
+
+	if err := vs.Set(common.OAuthAccessTokenVar, tokenResp.AccessToken, true); err != nil {
```
daabr: Whenever you save a new access token, you also need to save its expiration timestamp

integrations/zoom/checks.go
```diff
@@ -46,13 +49,96 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+// oauthTest verifies the OAuth authentication with Zoom API using the "me" context.
+// (Based on: https://developers.zoom.us/docs/integrations/oauth/#the-me-context).
+func oauthTest(ctx context.Context, vs sdktypes.Vars) (sdktypes.Status, error) {
+	token := vs.GetValue(common.OAuthAccessTokenVar)
+	refreshToken := vs.GetValue(common.OAuthRefreshTokenVar)
+	clientID := os.Getenv("ZOOM_CLIENT_ID")
+	clientSecret := os.Getenv("ZOOM_CLIENT_SECRET")
+	url := "https://api.zoom.us/v2/users/me"
+
+	// First attempt with current token
```
daabr: Remember: if a comment is a full sentence, use punctuation, i.e. end with "."

integrations/zoom/checks.go
```diff
@@ -46,13 +49,96 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+	if err != nil {
+		return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusOK {
+		return sdktypes.NewStatus(sdktypes.StatusCodeOK, "OAuth connection successful"), nil
+	}
+
+	// Check if token expired
```
daabr: Discussed offline: compare the OAuth expiration to the current timestamp, instead of making the same API call twice

integrations/zoom/checks.go
```diff
@@ -46,13 +49,96 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+			return sdktypes.NewStatus(sdktypes.StatusCodeOK, "OAuth connection successful after token refresh"), nil
+		}
+
+		return sdktypes.NewStatus(sdktypes.StatusCodeError, "Failed to connect to Zoom API after token refresh"), nil
+	}
+
+	return sdktypes.NewStatus(sdktypes.StatusCodeError, "Failed to connect to Zoom API"), nil
+}
+
+// ServerToServerTest handles the Server-to-Server authentication test for Zoom
```
daabr: Don't use the word "handle" unless you really have to, 99% of the time there are more accurate and descriptive verbs.

integrations/zoom/checks.go
```diff
@@ -46,13 +49,96 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+// (Based on: https://developers.zoom.us/docs/internal-apps/s2s-oauth/).
+func ServerToServerTest(ctx context.Context, vs sdktypes.Vars) (sdktypes.Status, error) {
+	id := vs.GetValue(accountID)
+	clientID := vs.GetValue(clientID)
+	secret := vs.GetValue(clientSecret)
+
+	app := privateApp{
```
daabr: Discussed offline: use `vs.Decode`

integrations/zoom/checks.go
```diff
@@ -46,13 +49,96 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+	}
+
+	// Get a server-to-server token
+	token, err := serverToken(ctx, &app)
+	if err != nil {
+		return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
+	}
+
+	url := "https://api.zoom.us/v2/users"
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
```
daabr: Why not use `makeAPICall` (or `httpGet` according to my comment about this function name)

integrations/zoom/auth.go
mario99logic: I passed a reference to avoid modifying the original function, as it is used in the Save.go file. I will update it in the next push to demonstrate what I mean.

integrations/zoom/checks.go
mario99logic: changed it just kept the os.Getenv

tebeka:
BTW: The PR title is not relevant.

mario99logic:
notes for the changes:

- Implemented the fix for the Decode function that we discussed offline.
- In `oauthTest`, I use Decode, but I still retrieve `clientID` and `clientSecret` from environment variables since we don’t receive them from the client.
- In `refreshToken`, I introduced a new struct instead of using `common.OAuthData` because the expiry field in `common.OAuthData` is a string, whereas the JSON response provides it as an integer.
- Although the JSON response currently returns the same `expires_in` value each time, I prefer to retrieve it from the response to accommodate any potential future changes.

sdk/sdktypes/vars.go
```diff
@@ -113,12 +113,18 @@ func (vs Vars) Decode(out any) {
 		fv := v.Field(i)
 		ft := t.Field(i)
 
+		tag := strings.Split(ft.Tag.Get("var"), ",")
```
daabr: FYI - I already copied this code (and added unit tests) in https://github.com/autokitteh/autokitteh/pull/1164/files#diff-4f8b0dd502d810a388ec87745edc8ac98e0d5b734258a4420dbc27bdecf24fd9

You can remove the changes in this file in this PR, and rebase with the main branch.

integrations/zoom/checks.go
```diff
@@ -46,13 +52,116 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("API request failed: %v", err)
+	}
+	return resp, nil
+}
+
+// httpPost sends an HTTP POST request with a URL-encoded body and basic authentication.
+func httpPost(ctx context.Context, data url.Values, clientID, clientSecret string) (*http.Response, error) {
+	timeoutCtx, cancel := context.WithTimeout(ctx, 20*time.Second)
```
daabr: Nit: you're not using `ctx` anywhere else in this function, so you can just overwrite `ctx`: this will replace the input, but only inside this function, which is what you want.

Anyway, try to model this function more or less like a simplified, Zoom-specific copy of `httpRquest` from https://github.com/autokitteh/autokitteh/blob/main/integrations/common/http.go#L98

integrations/zoom/checks.go
```diff
@@ -46,13 +52,116 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
 	})
 }
+
+// oauthTest verifies the OAuth authentication with Zoom API using the "me" context.
+// (Based on: https://developers.zoom.us/docs/integrations/oauth/#the-me-context).
+func oauthTest(ctx context.Context, vs sdktypes.Vars) (sdktypes.Status, error) {
+	data := common.OAuthData{}
+	vs.Decode(&data)
+	clientID := os.Getenv("ZOOM_CLIENT_ID")
+	clientSecret := os.Getenv("ZOOM_CLIENT_SECRET")
```
daabr: Notice that this means you don't support private OAuth here yet.
Add a TODO comment with a ticket reference to fix that later.

integrations/zoom/checks.go
```diff
@@ -46,13 +52,116 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+	url := "https://api.zoom.us/v2/users/me"
+
+	// Check if the token is expired and refresh it.
```
daabr, line 69: Instead of this logic, use the new function `FreshOAuthToken` from https://github.com/autokitteh/autokitteh/blob/main/integrations/common/vars.go#L84

integrations/zoom/checks.go
```diff
@@ -46,13 +52,116 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+
+	if resp.StatusCode == http.StatusOK {
+		return sdktypes.NewStatus(sdktypes.StatusCodeOK, "Server-to-Server connection successful"), nil
+	}
+
+	return sdktypes.NewStatus(sdktypes.StatusCodeError, "Failed to connect to Zoom API"), nil
+}
+
+// httpGet sends an HTTP GET request to the specified URL with a Bearer token for authentication.
+func httpGet(ctx context.Context, bearerToken string, url string) (*http.Response, error) {
```
daabr: Please replace `httpGet` here with `common.HTTPGet` (but not `httpPost` below, because it doesn't support basic auth). See also INT-337 (for a future PR).

integrations/zoom/auth.go
```diff
@@ -2,12 +2,75 @@ package zoom
...
+	var tokenData common.OAuthData
+	if err := json.NewDecoder(resp.Body).Decode(&tokenData); err != nil {
+		return "", fmt.Errorf("failed to decode token response: %v", err)
+	}
+
+	return tokenData.AccessToken, nil
+}
+
+// refreshToken sends a request to Zoom's OAuth endpoint to refresh an expired access token.
+func refreshToken(ctx context.Context, refreshT string, clientID string, clientSecret string, vs sdktypes.Vars) (string, error) {
```
daabr: Don't refresh directly, instead use `common.FreshOAuthData` from https://github.com/autokitteh/autokitteh/blob/main/integrations/common/vars.go#L84

integrations/zoom/checks.go
```diff
@@ -46,13 +53,74 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
-		return sdktypes.NewStatus(sdktypes.StatusCodeOK, ""), nil
 	})
 }
+
+// oauthTest verifies the OAuth authentication with Zoom API using the "me" context.
+// (Based on: https://developers.zoom.us/docs/integrations/oauth/#the-me-context).
+func oauthTest(ctx context.Context, v sdkservices.Vars, o sdkservices.OAuth, vs sdktypes.Vars) (sdktypes.Status, error) {
+	// TODO(INT-338): Support private OAuth.
+	// Check if the token is expired and refresh it.
+	desc := common.Descriptor("zoom", "", "")
```
daabr: No need to redefine `desc` here, this file already has access to the `desc` global var in `integration.go`

integrations/zoom/checks.go
```diff
@@ -46,13 +53,74 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+	if err != nil {
+		return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
+	}
+
+	return sdktypes.NewStatus(sdktypes.StatusCodeOK, "OAuth connection successful"), nil
+}
+
+// ServerToServerTest validates the Server-to-Server authentication test for Zoom.
+// (Based on: https://developers.zoom.us/docs/internal-apps/s2s-oauth/).
+func ServerToServerTest(ctx context.Context, vs sdktypes.Vars) (sdktypes.Status, error) {
```
daabr: Why is this function exported? Is it used in any other package?

daabr (approved):
Approved with 2 easy-to-fix comments

integrations/zoom/checks.go
mario99logic: no need to export it, changed it.

PR #1161, opened 2025-03-04
fix: better error handling in pythonrt so runner wont stuck

efiShtain:
- added more try except blocks around pickling of errors

- made sure runner will exit even if it couldn't call worker.Done for some reason

There is still an exception with pickling we should figure out, but this will at least make the code always end

### DIFF
--- a/runtimes/pythonrt/runner/main.py
+++ b/runtimes/pythonrt/runner/main.py
@@ -464,7 +464,6 @@ class Runner(pb.runner_rpc.RunnerService):
             self.worker.Done(req)
         except Exception as err:
             log.error("on_event: done send error: %r", err)
-        
 
     @mark_no_activity
     def ak_print(self, *objects, sep=" ", end="\n", file=None, flush=False):
--- a/runtimes/pythonrt/runner/main.py
+++ b/runtimes/pythonrt/runner/main.py
@@ -454,7 +454,7 @@ class Runner(pb.runner_rpc.RunnerService):
                 req.result.custom.data = data
                 req.result.custom.value.CopyFrom(values.safe_wrap(result.value))
         except (TypeError, pickle.PickleError) as err:
-                    req.error = f"can't pickle {result.value} - {err}"
+            req.error = f"can't pickle {result.value} - {err}"
         except Exception as err:
             req.error = f"unexpected error: {err}"
             log.exception("on_event: %r", err)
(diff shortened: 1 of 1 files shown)


runtimes/pythonrt/runner/main.py
```diff
@@ -444,21 +444,25 @@ def on_event(self, fn, event):
...
+                req.error = result_error(result.error)
+                tb = pb_traceback(result.traceback)
+                req.traceback.extend(tb)
+            else:
+                try:
+                    data = pickle.dumps(result)
+                    req.result.custom.data = data
+                    req.result.custom.value.CopyFrom(values.safe_wrap(result.value))
+                except (TypeError, pickle.PickleError) as err:
+                    req.error = f"can't pickle {result.value} - {err}"
```
daabr: Nested `try` are pointless and very inefficient: move the inner `except` to the same level of the new one (make sure `except Exception` is always last)

```suggestion
        try:
            if result.error:
                req.error = result_error(result.error)
                tb = pb_traceback(result.traceback)
                req.traceback.extend(tb)
            else:
                data = pickle.dumps(result)
                req.result.custom.data = data
                req.result.custom.value.CopyFrom(values.safe_wrap(result.value))
        except (TypeError, pickle.PickleError) as err:
            req.error = f"can't pickle {result.value} - {err}"
```

runtimes/pythonrt/runner/main.py
```diff
@@ -444,21 +444,25 @@ def on_event(self, fn, event):
...
+        except Exception as err:
+            req.error = f"unexpected error: {err}"
+            log.exception("on_event: %r", err)
 
         try:
             self.worker.Done(req)
-        except grpc.RpcError as err:
+        except Exception as err:
```
daabr, line 464: Better to keep specific error types, and distinguish them from unexpected ones:

```suggestion
        except grpc.RpcErr as err:
            log.error("on_event: RPC error when sending done: %r", err)
        except Exception as ex:
            log.error("on_event: unexpected error when sending done: %r", ex)
```

daabr (approved):
nice - see comments for better error reporting

tebeka (changes requested):
See @daabr comment about nesting. Code a bit too defensive IMO but OK with merging this.

runtimes/pythonrt/runner/main.py
efiShtain: there can be lots of other exception types in this case, 
we don't really case in this scenario, 
we're not retrying and it's the end of the runner

runtimes/pythonrt/runner/main.py
efiShtain: technically since this will run only once and end I don't really mind
left it with nesting since the error handling was like that before

but sure, refactored

tebeka (approved):

PR #1162, opened 2025-03-04
feat(salesforce): add support for Salesforce events

pashafateev:
#### OAuth and Authentication Enhancements
- **Token Management:**
  - Introduced the `oauthToken` function in `integrations/salesforce/auth.go` to automatically refresh stale tokens.
  - Added `bearerToken` to generate the correct authorization header.
- **User Information Retrieval:**
  - Implemented `getUserInfo` in `integrations/salesforce/checks.go` to retrieve user details from Salesforce, extract the organization ID, and simultaneously test the connection.

#### Event Handling and Pub/Sub Integration
- **New Event Handling:**
  - Added `integrations/salesforce/events.go` to process incoming Salesforce events, transform them into AutoKitteh events, and dispatch them asynchronously.
- **Pub/Sub Subscription via gRPC:**
  - Introduced `integrations/salesforce/grpc.go` to establish a secure gRPC connection
  - Created `integrations/salesforce/subscription.go` to manage subscriptions, handle message streaming, renew subscriptions, and decode event payloads using Avro.
  - Implemented `renewSubscription` to request additional messages from the Salesforce Pub/Sub API after the initial batch of events has been received.

### Testing and Validation
- Manually tested by authenticating a Salesforce project, triggering an event in the Salesforce UI, and verifying that the event correctly triggered a workflow in AutoKitteh with the expected event data.

### Known Issues
- If an existing Salesforce connection fails to establish a gRPC connection (for example, due to an expired authentication token), the only current workaround is to create a new connection
- While the addition of the refresh token and token source helps mitigate this risk, further improvements are needed to handle reconnection scenarios more gracefully.


internal/backend/oauth/oauth.go
```diff
@@ -489,7 +489,7 @@ func New(l *zap.Logger, vars sdkservices.Vars) sdkservices.OAuth {
 				},
 				RedirectURL: redirectURL + "salesforce",
 				// https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oauth_tokens_scopes.htm
-				Scopes: []string{""},
+				Scopes: []string{"api", "refresh_token"},
```
daabr, line 492: Wasn't this only for testing? Or do you want to keep this list?

integrations/salesforce/vars.go
```diff
@@ -1,5 +1,13 @@
 package salesforce
 
+import "go.autokitteh.dev/autokitteh/sdk/sdktypes"
+
+var (
+	instanceURLVar      = sdktypes.NewSymbol("instance_url")
+	oauthAccessTokenVar = sdktypes.NewSymbol("oauth_access_token")
```
daabr: Use `common.OAuthAccessTokenVar` in your code instead of redefining it here

integrations/salesforce/auth.go
```diff
@@ -0,0 +1,61 @@
...
+	"go.autokitteh.dev/autokitteh/integrations"
+	"go.autokitteh.dev/autokitteh/integrations/common"
+	"go.autokitteh.dev/autokitteh/internal/backend/auth/authcontext"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+// oauthToken returns the OAuth token stored in the
+// connection variables. If it's stale, we refresh it first.
+func oauthToken(ctx context.Context, vs sdktypes.Vars, o sdkservices.OAuth) *oauth2.Token {
```
daabr, line 18: FYI, this function definitely feels like it should in `common` (and with unit testing), but we'll do it later to finish with this PR quickly.

integrations/salesforce/oauth.go
```diff
@@ -56,12 +56,20 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	accessToken := data.Token.AccessToken
+	instanceURL := data.Extra["instance_url"].(string)
```
daabr, line 65: Cosmetic nit: the comment below refers to an entire block of code, not just a single line, so it makes sense and is more readable to define `accessToken` and `instanceURL` below it rather than above it.

integrations/salesforce/oauth.go
```diff
@@ -56,12 +56,20 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
...
-	ctx := r.Context()
+	userInfo, err := getUserInfo(r.Context(), instanceURL, accessToken)
+	if err != nil {
+		l.Error("failed to get user info", zap.Error(err))
+		c.AbortServerError("failed to get user info")
+		return
+	}
+	orgID := userInfo["organization_id"].(string)
 
+	ctx := r.Context()
```
daabr: Better: move this line above `getUserInfo()` and use `ctx` there too

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+
+import (
+	"context"
+	"sync"
+
+	pb "github.com/developerforce/pub-sub-api/go/proto"
+	"github.com/linkedin/goavro/v2"
+	"go.uber.org/zap"
+
+	// "go.autokitteh.dev/autokitteh/integrations/common"
```
daabr: remove unused imports

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	conn, err := initConn(accessToken, instanceURL, orgID, h.logger)
+	if err != nil {
+		h.logger.Error("failed to create gRPC connection", zap.Error(err))
+		return
+	}
+
+	client := pb.NewPubSubClient(conn)
+	pubSubClients[instanceURL] = &client
+
+	for _, eventType := range supportedEventTypes() {
```
daabr: If we only support 1 generic and hard-coded event type, there's no need for a loop, and for a helper function that returns it (but move the reference link from there to here - that's good).

integrations/salesforce/oauth.go
```diff
@@ -76,10 +84,12 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
 	}
 
 	http.Redirect(w, r, urlPath, http.StatusFound)
+
+	h.Subscribe(instanceURL, orgID, h.bearerToken(ctx, l, cid), cid)
```
daabr: Doesn't it make more sense to call `bearerToken()` inside `Subscribe`?

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+var (
+	// Key = Salesforce instance URL (to ensure one gRPC client per app).
+	// Writes happen during server startup, and when a user
+	// creates a new Salesforce connection in the UI.
+	pubSubClients = make(map[string]*pb.PubSubClient)
+
+	mu = &sync.Mutex{}
+)
+
+func (h handler) Subscribe(instanceURL, orgID, accessToken string, cid sdktypes.ConnectionID) {
```
daabr: 1. Question: why is this function exported?
2. Please add a function comment

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+
+	client := pb.NewPubSubClient(conn)
+	pubSubClients[instanceURL] = &client
+
+	for _, eventType := range supportedEventTypes() {
+		go h.handleSalesForceSubscription(ctx, client, eventType, cid)
+	}
+}
+
+func (h handler) handleSalesForceSubscription(ctx context.Context, client pb.PubSubClient, topicName string, cid sdktypes.ConnectionID) {
```
daabr: 1. Missing function doc: what it does, mention that it uses Pub/Sub based on streaming gRPC (with ref links?)
2. Naming: "Salesforce" is a single word, not two
3. Naming: is "Salesfoce" needed at all inside the "salesforce" package? There are no other "non-Salesforce" subscriptions here (see https://google.github.io/styleguide/go/decisions.html#repetition)
4. Avoid "handle" whenever possible (like avoiding "get" - see https://google.github.io/styleguide/go/decisions.html#getters), it's usually too generic - for example, what about "eventLoop"? (just a suggestion)

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	const defaultBatchSize = int32(100)
+	// var latestReplayId []byte
```
daabr: Obsolete?

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+
+	err := stream.Send(fetchReq)
+	if err != nil {
+		l.Error("failed to request more messages", zap.Error(err))
+		return 0, err
+	}
+	return int(defaultBatchSize), nil
+}
+
+func decodePayload(schema *pb.SchemaInfo, payload []byte, l *zap.Logger) (map[string]any, error) {
```
daabr: Convention: loggers should be the first parameter, or among the first (e.g. if you pass a `ctx`, it must always be the first parameter)

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+				h.logger.Error("failed to decode event", zap.Error(err))
+				continue
+			}
+
+			h.handleSalesForceEvent(data, topicName)
+		}
+	}
+}
+
+func renewSubscription(stream pb.PubSub_SubscribeClient, defaultBatchSize int32, topicName string, l *zap.Logger) (int, error) {
```
daabr: Convention: loggers should be the first parameter, or among the first (e.g. if you pass a ctx, it must always be the first parameter)

integrations/salesforce/checks.go
```diff
@@ -55,3 +59,41 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
 		return sdktypes.NewStatus(sdktypes.StatusCodeError, "Not implemented"), nil
 	})
 }
+
+func getUserInfo(ctx context.Context, instanceURL, accessToken string) (map[string]interface{}, error) {
+	// Create the request
```
daabr: General comment for the entire PR: https://google.github.io/styleguide/go/decisions.html#comment-sentences

integrations/salesforce/grpc.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"context"
+	"crypto/tls"
+
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+type salesforceAuth struct {
```
daabr: 1. If this struct must implement some interface, document this
2. Isn't "grpcAuth" a more accurate name? Also avoid repetition with the package name

integrations/salesforce/events.go
```diff
@@ -0,0 +1,84 @@
...
+// Transform the received Salesforce event into an AutoKitteh event.
+func transformEvent(l *zap.Logger, salesforceEvent any, eventType string) (sdktypes.Event, error) {
+	l = l.With(
+		zap.String("eventType", eventType),
+		zap.Any("event", salesforceEvent),
+	)
+
+	wrapped, err := sdktypes.WrapValue(salesforceEvent)
+	if err != nil {
+		l.Error("Failed to wrap Salesforce event", zap.Error(err))
```
daabr: Cosmetic nit for this entire file: we don't treat log messages (and error messages) as independent sentences. Don't capitalize.

integrations/salesforce/grpc.go
```diff
@@ -0,0 +1,45 @@
...
+	accessToken string
+	instanceURL string
+	tenantID    string
+}
+
+func initConn(accessToken, instanceURL, orgID string, l *zap.Logger) (*grpc.ClientConn, error) {
+	conn, err := grpc.NewClient(
+		"api.pubsub.salesforce.com:443",
+		grpc.WithTransportCredentials(credentials.NewTLS(&tls.Config{})),
```
daabr, line 23: Question: does this mean "some kind of basic/generic/default TLS config" or "no TLS"? We need this connection to be encrypted in some minimal way.

integrations/salesforce/grpc.go
```diff
@@ -0,0 +1,45 @@
...
+		grpc.WithPerRPCCredentials(&salesforceAuth{
+			accessToken: accessToken,
+			instanceURL: instanceURL,
+			tenantID:    orgID,
+		}),
+	)
+	if err != nil {
+		l.Error("failed to create gRPC connection", zap.Error(err))
```
daabr: If we see this in Datadog, it would to be a little more specific
```suggestion
		l.Error("failed to create gRPC connection for Salesforce events", zap.Error(err))
```

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	defer mu.Unlock()
+
+	if _, ok := pubSubClients[instanceURL]; ok {
+		return
+	}
+
+	ctx := context.Background()
+	conn, err := initConn(accessToken, instanceURL, orgID, h.logger)
+	if err != nil {
+		h.logger.Error("failed to create gRPC connection", zap.Error(err))
```
daabr: We already logged this inside `initConn`, no need to log it again.

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	// Key = Salesforce instance URL (to ensure one gRPC client per app).
+	// Writes happen during server startup, and when a user
+	// creates a new Salesforce connection in the UI.
+	pubSubClients = make(map[string]*pb.PubSubClient)
+
+	mu = &sync.Mutex{}
+)
+
+func (h handler) Subscribe(instanceURL, orgID, accessToken string, cid sdktypes.ConnectionID) {
+	// Ensure multiple users don't reference the same app at the same time.
```
daabr: More precisely, we're avoiding here duplication due to race conditions, not concurrent references

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	pb "github.com/developerforce/pub-sub-api/go/proto"
+	"github.com/linkedin/goavro/v2"
+	"go.uber.org/zap"
+
+	// "go.autokitteh.dev/autokitteh/integrations/common"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var (
+	// Key = Salesforce instance URL (to ensure one gRPC client per app).
```
daabr, line 17: Good comment

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func (h handler) handleSalesForceSubscription(ctx context.Context, client pb.PubSubClient, topicName string, cid sdktypes.ConnectionID) {
+	const defaultBatchSize = int32(100)
+	// var latestReplayId []byte
+	numLeftToReceive := 0
+
+	stream, err := client.Subscribe(ctx)
+	if err != nil {
+		h.logger.Error("failed to create stream", zap.Error(err))
```
daabr: For easier triage/debugging in Datadog:
```suggestion
		h.logger.Error("failed to create gRPC stream for Salesforce events", zap.Error(err))
```

daabr: Much better than using `h.logger` directly throughout this function: `l := h.logger.With(zap.String("connection_id", cid.String())` (the CID is an important debugging details, we shouldn't add it explicitly in every logging statement)

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+		return
+	}
+
+	// Start receiving messages
+	for {
+		if numLeftToReceive <= 0 {
+			n, err := renewSubscription(stream, defaultBatchSize, topicName, h.logger)
+			if err != nil {
+				h.logger.Error("failed to renew subscription", zap.Error(err))
```
daabr:
```suggestion
				h.logger.Error("failed to renew Salesforce events subscription", zap.Error(err))
```

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+				continue
+			}
+			numLeftToReceive = n
+		}
+
+		// Assumes that the stream will abort after 270 seconds of inactivity.
+		// https://developer.salesforce.com/docs/platform/pub-sub-api/references/methods/subscribe-rpc.html#subscribe-keepalive-behavior
+		msg, err := stream.Recv()
+		if err != nil {
+			h.logger.Error("error receiving message", zap.Error(err))
```
daabr:
```suggestion
			h.logger.Error("error receiving Salesforce event", zap.Error(err))
```

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+		}
+
+		// TODO(INT-314): Save the latest replay ID.
+		// latestReplayId = msg.GetLatestReplayId()
+
+		// Process the received message.
+		eventCount := len(msg.Events)
+		if eventCount > 0 {
+			numLeftToReceive -= eventCount
+		}
```
daabr: Unnecessary `if` - the count can't be `< 0`, and `numLeftToReceive -= 0` is not a problem

```suggestion
		numLeftToReceive -= len(msg.Events)
```

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+
+		for _, event := range msg.Events {
+			schema, err := client.GetSchema(ctx, &pb.SchemaRequest{
+				SchemaId: event.Event.SchemaId,
+			})
```
daabr: Short enough to fit in a single line
```suggestion
			schema, err := client.GetSchema(ctx, &pb.SchemaRequest{SchemaId: event.Event.SchemaId})
```

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+			if err != nil {
+				h.logger.Error("failed to get schema", zap.Error(err))
```
daabr: Specify a more meaningful (though not too verbose) error message triage/debugging in Datadog, also in line 97 below

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+				continue
+			}
+			data, err := decodePayload(schema, event.Event.Payload, h.logger)
+			if err != nil {
+				h.logger.Error("failed to decode event", zap.Error(err))
```
daabr: No need to log this, already done inside `decodePayload`

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+		l.Error("failed to request more messages", zap.Error(err))
+		return 0, err
+	}
+	return int(defaultBatchSize), nil
+}
+
+func decodePayload(schema *pb.SchemaInfo, payload []byte, l *zap.Logger) (map[string]any, error) {
+	codec, err := goavro.NewCodec(schema.SchemaJson)
+	if err != nil {
+		l.Error("failed to create codec", zap.Error(err))
```
daabr: Specify in this function more meaningful (though not too verbose) error messages for triage/debugging in Datadog

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+		return nil, err
+	}
+
+	var data map[string]any
+	// Payload may contain multiple records
```
daabr: I wanted to ask to move this comment above `var data`, but what does it mean anyway? it's it obvious that a map may contain multiple keys?

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	native, _, err := codec.NativeFromBinary(payload)
+	if err != nil {
+		l.Error("failed to decode event", zap.Error(err))
+		return nil, err
+	}
+	data, ok := native.(map[string]any)
+	if !ok {
+		l.Error("failed to cast decoded event to map[string]any")
```
daabr, line 140: Do you think we should log `data` with `zap.Any`? Or does it contain potentially sensitive data?

integrations/salesforce/subscription.go
pashafateev: done

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+				continue
+			}
+
+			h.handleSalesForceEvent(data, topicName)
+		}
+	}
+}
+
+func renewSubscription(stream pb.PubSub_SubscribeClient, defaultBatchSize int32, topicName string, l *zap.Logger) (int, error) {
+	l.Info("requesting more messages", zap.Int32("batchSize", defaultBatchSize))
```
daabr: Is this a useful log? I expect this would happen a lot over a period of times, so trying to minimize spam in Datadog, unless there's a purpose to knowing whenever this happens.

integrations/salesforce/auth.go
pashafateev: Should I open a ticket for that?

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/events.go
```diff
@@ -0,0 +1,84 @@
...
+	"context"
+
+	"go.uber.org/zap"
+
+	"go.autokitteh.dev/autokitteh/integrations/internal/extrazap"
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+func (h handler) handleSalesForceEvent(decodedEventPayload map[string]any, topicName string) {
```
daabr:
```suggestion
func (h handler) dispatchEvent(payload map[string]any, topic string) {
```

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/events.go
```diff
@@ -0,0 +1,84 @@
...
+	cids, err := h.vars.FindConnectionIDs(context.Background(), h.integrationID, instanceURLVar, "")
+	if err != nil {
+		h.logger.Error("Failed to find connection IDs", zap.Error(err))
+		return
+	}
+
+	h.dispatchAsyncEventsToConnections(cids, akEvent)
+}
+
+// Transform the received Salesforce event into an AutoKitteh event.
```
daabr:
```suggestion
// transformEvent transforms the received Salesforce event payload into an AutoKitteh event.
```

integrations/salesforce/events.go
```diff
@@ -0,0 +1,84 @@
...
+func transformEvent(l *zap.Logger, salesforceEvent any, eventType string) (sdktypes.Event, error) {
```
daabr: If you intend to attach attributes to the logger *before* calling this function it makes sense to pass the logger as an argument. If not, you can attach this function to the `handler` receiver like in the function above, and call `l := h.logger.With(...)` - whichever you think makes more sense.

integrations/salesforce/events.go
```diff
@@ -0,0 +1,84 @@
...
+	l = l.With(
+		zap.String("eventType", eventType),
```
daabr:
```suggestion
		zap.String("event_type", eventType),
```

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/events.go
```diff
@@ -0,0 +1,84 @@
...
+
+	akEvent, err := sdktypes.EventFromProto(&sdktypes.EventPB{
+		EventType: eventType,
+		Data:      kittehs.TransformMapValues(data, sdktypes.ToProto),
+	})
+	if err != nil {
+		l.Error("Failed to convert protocol buffer to SDK event",
+			zap.Any("data", data),
+			zap.Error(err),
+		)
```
daabr: You already added the payload to `l`
```suggestion
		l.Error("failed to convert protocol buffer to SDK event", zap.Error(err))
```

integrations/salesforce/checks.go
```diff
@@ -55,3 +59,41 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, instanceURL+"/services/oauth2/userinfo", nil)
```
daabr: Slightly longer (due to having to handle the error case), but better: https://pkg.go.dev/net/url#JoinPath

integrations/salesforce/checks.go
```diff
@@ -55,3 +59,41 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Check for non-200 status codes
+	if resp.StatusCode != http.StatusOK {
+		return nil, errors.New("Salesforce API error: " + string(body))
+	}
+
+	var userInfo map[string]interface{}
```
daabr:
```suggestion
	var userInfo map[string]any
```

integrations/salesforce/integration.go
```diff
@@ -27,23 +30,50 @@ func New(v sdkservices.Vars) sdkservices.Integration {
 func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.DispatchFunc) {
 	common.ServeStaticUI(m, desc, static.SalesforceWebContent)
 
-	h := newHTTPHandler(l, v, o, d)
+	iid := sdktypes.NewIntegrationIDFromName(desc.UniqueName().String())
```
daabr: No need to initialize `iid` here, you can do it in `newHTTPHandler` too because `desc` is a module-level var.

internal/backend/oauth/oauth.go
pashafateev: The only reason for keeping it is that it represents the minimal functionality required for events to work as they currently do. If the client app does not activate at least these permissions and the scopes are left blank, there will be no indication that something went wrong. However, including the scopes results in an error that may be difficult for a general user to understand.

For this PR, I'm fine with either.

integrations/salesforce/integration.go
```diff
@@ -27,23 +30,50 @@ func New(v sdkservices.Vars) sdkservices.Integration {
...
+	h := newHTTPHandler(l, v, o, d, iid)
+
 	common.RegisterSaveHandler(m, desc, h.handleSave)
 	common.RegisterOAuthHandler(m, desc, h.handleOAuth)
 
 	// TODO: Event webhooks (no AutoKitteh user authentication by definition, because
 	// these asynchronous requests are sent to us by third-party services).
+
```
daabr: The TODO comment above is obsolete

integrations/salesforce/integration.go
```diff
@@ -27,23 +30,50 @@ func New(v sdkservices.Vars) sdkservices.Integration {
...
+	integrationID sdktypes.IntegrationID
 }
 
-func newHTTPHandler(l *zap.Logger, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.DispatchFunc) handler {
-	return handler{logger: l, oauth: o, vars: v, dispatch: d}
+func newHTTPHandler(l *zap.Logger, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.DispatchFunc, i sdktypes.IntegrationID) handler {
+	return handler{logger: l, oauth: o, vars: v, dispatch: d, integrationID: i}
+}
+
+func reopenExistingPubSubConnections(ctx context.Context, l *zap.Logger, v sdkservices.Vars, h handler) {
```
daabr: Why not attach this function to the handler instead of passing all the args?

integrations/salesforce/integration.go
```diff
@@ -27,23 +30,50 @@ func New(v sdkservices.Vars) sdkservices.Integration {
...
+	iid := sdktypes.NewIntegrationIDFromName(desc.UniqueName().String())
+	h := newHTTPHandler(l, v, o, d, iid)
+
 	common.RegisterSaveHandler(m, desc, h.handleSave)
 	common.RegisterOAuthHandler(m, desc, h.handleOAuth)
 
 	// TODO: Event webhooks (no AutoKitteh user authentication by definition, because
 	// these asynchronous requests are sent to us by third-party services).
+
+	reopenExistingPubSubConnections(context.Background(), l, v, h)
```
daabr: If you're just passing a new `context.Background()`, why not do `ctx := context.Background()` inside `reopen...()`?

integrations/salesforce/integration.go
```diff
@@ -27,23 +30,50 @@ func New(v sdkservices.Vars) sdkservices.Integration {
...
+	cids, err := v.FindConnectionIDs(ctx, iid, instanceURLVar, "")
+	if err != nil {
+		l.Error("failed to list Salesforce connection IDs", zap.Error(err))
+		return
+	}
+
+	for _, cid := range cids {
+		data, err := v.Get(ctx, sdktypes.NewVarScopeID(cid))
+		if err != nil {
+			l.Error("can't restart Salesforce PubSub connection", zap.Error(err))
```
daabr: Also log `connection_id`

integrations/salesforce/vars.go
pashafateev: Done

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go
pashafateev: Not sure I understand this comment

integrations/salesforce/subscription.go
pashafateev: Done and done

integrations/salesforce/subscription.go
pashafateev: I think this check is necessary because this is where I receive empty keepalive messages. I don't want to count those as events because we will be renewing subscriptions too often.

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go
pashafateev: Good point. Done.

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go
pashafateev: No, it was useful for testing

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go
pashafateev: It's related more to payload than to data. `codec.NativeFromBinary()`'s return type is `(interface{}, []byte, error)`.

> On success, it returns the decoded datum, a byte slice containing the remaining undecoded bytes, and a nil error value.

It reads a certain amount of bytes and then returns a byte slice with the remaining (unread) bytes. When I checked what's inside for the custom event it looked identical to the first decoded bytes. I think it's worth investigating in the future, but I don't think it's urgent for now.

integrations/salesforce/subscription.go
pashafateev: It has sensitive data, but it's encoded. If it failed to cast, then it should still be encoded? Not sure if that is considered sensitive at that point.

integrations/salesforce/subscription.go

integrations/salesforce/integration.go
pashafateev: Done

integrations/salesforce/integration.go
pashafateev: Done

integrations/salesforce/integration.go
pashafateev: Done.

integrations/salesforce/auth.go
pashafateev: INT-316

integrations/salesforce/checks.go
pashafateev: Done

integrations/salesforce/checks.go
pashafateev: Done

integrations/salesforce/events.go
pashafateev: Done

integrations/salesforce/events.go
pashafateev: Done

integrations/salesforce/events.go
pashafateev: done

integrations/salesforce/events.go
pashafateev: Done

integrations/salesforce/events.go
pashafateev: Done

integrations/salesforce/subscription.go

integrations/salesforce/events.go
pashafateev: Done

integrations/salesforce/grpc.go
pashafateev: Done

integrations/salesforce/grpc.go
pashafateev: > WithTransportCredentials returns a DialOption which configures a connection level security credentials (e.g., TLS/SSL). This should not be used together with WithCredentialsBundle.

Is this sufficient?

integrations/salesforce/grpc.go
pashafateev: Done

integrations/salesforce/subscription.go

integrations/salesforce/oauth.go
```diff
@@ -88,6 +110,63 @@ func (h handler) saveConnection(ctx context.Context, vsid sdktypes.VarScopeID, t
...
+
+	formData := url.Values{
+		"token":           {accessToken},
+		"token_type_hint": {"access_token"},
+		"client_id":       {vs.GetValueByString("private_client_id")},
+		"client_secret":   {vs.GetValueByString("private_client_secret")},
+	}
+	encodedFormData := formData.Encode()
+
+	u, err := url.JoinPath(instanceURL, "services", "oauth2", "introspect")
```
daabr: In contrast to OS path, URLs can use only "/" as a separator, so no need to split elements in the path
```suggestion
	u, err := url.JoinPath(instanceURL, "services/oauth2/introspect")
```

integrations/salesforce/oauth.go
```diff
@@ -88,6 +110,63 @@ func (h handler) saveConnection(ctx context.Context, vsid sdktypes.VarScopeID, t
 	for k, v := range extra {
 		vs = vs.Append(sdktypes.NewVar(sdktypes.NewSymbol(k)).SetValue(fmt.Sprintf("%v", v)))
 	}
+	vs = vs.Append(sdktypes.NewVar(orgIDVar).SetValue(orgID))
+	vs = vs.Append(sdktypes.NewVar(common.OAuthExpiryVar).SetValue(exp.Format(time.RFC3339)))
 
 	return h.vars.Set(ctx, vs.WithScopeID(vsid)...)
 }
+
+func (h handler) accessTokenExpiration(ctx context.Context, instanceURL, accessToken string, cid sdktypes.ConnectionID) (time.Time, error) {
```
daabr: Why not pass a reference to a token object, and just update the time inside it (or return an error)?

This should also simplify the `saveConnection` function - there's no need to pass a token object as well as a separate expiration time for it.

integrations/salesforce/oauth.go
```diff
@@ -88,6 +110,63 @@ func (h handler) saveConnection(ctx context.Context, vsid sdktypes.VarScopeID, t
...
+		return time.Time{}, err
+	}
+
+	formData := url.Values{
+		"token":           {accessToken},
+		"token_type_hint": {"access_token"},
+		"client_id":       {vs.GetValueByString("private_client_id")},
+		"client_secret":   {vs.GetValueByString("private_client_secret")},
+	}
+	encodedFormData := formData.Encode()
```
daabr: You can skip this line and just use `formData.Encode()` in line 138

integrations/salesforce/oauth.go
```diff
@@ -88,6 +110,63 @@ func (h handler) saveConnection(ctx context.Context, vsid sdktypes.VarScopeID, t
...
+		return time.Time{}, err
+	}
+	defer resp.Body.Close()
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return time.Time{}, err
+	}
+
+	var tokenInfo map[string]interface{}
```
daabr:
```suggestion
	var tokenInfo map[string]any
```

integrations/salesforce/oauth.go
```diff
@@ -88,6 +110,63 @@ func (h handler) saveConnection(ctx context.Context, vsid sdktypes.VarScopeID, t
...
+	if err := json.Unmarshal(body, &tokenInfo); err != nil {
+		return time.Time{}, errors.New("failed to parse token info")
+	}
+
+	// Extract the expiration timestamp
```
daabr: https://google.github.io/styleguide/go/decisions#comment-sentences

integrations/salesforce/oauth.go
```diff
@@ -55,13 +60,28 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
...
+	instanceURL := data.Extra["instance_url"].(string)
+	userInfo, err := getUserInfo(ctx, instanceURL, accessToken)
+	if err != nil {
+		l.Error("failed to get user info", zap.Error(err))
+		c.AbortServerError("failed to get user info")
+		return
+	}
+	orgID := userInfo["organization_id"].(string)
+
+	// get the access token expiration time
```
daabr: https://google.github.io/styleguide/go/decisions#comment-sentences

integrations/salesforce/oauth.go
```diff
@@ -88,6 +110,63 @@ func (h handler) saveConnection(ctx context.Context, vsid sdktypes.VarScopeID, t
 	for k, v := range extra {
 		vs = vs.Append(sdktypes.NewVar(sdktypes.NewSymbol(k)).SetValue(fmt.Sprintf("%v", v)))
 	}
+	vs = vs.Append(sdktypes.NewVar(orgIDVar).SetValue(orgID))
+	vs = vs.Append(sdktypes.NewVar(common.OAuthExpiryVar).SetValue(exp.Format(time.RFC3339)))
```
daabr: No, this should be handled by `common.EncodeOAuthData`

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	const defaultBatchSize = int32(100)
+	numLeftToReceive := 0
+
+	stream, err := client.Subscribe(ctx)
+	if err != nil {
+		l.Error("failed to create gRPC stream for Salesforce events", zap.Error(err))
+		return
+	}
+
+	// Start receiving messages
```
daabr: https://google.github.io/styleguide/go/decisions#comment-sentences

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	}
+}
+
+func renewSubscription(l *zap.Logger, stream pb.PubSub_SubscribeClient, defaultBatchSize int32, topicName string) (int, error) {
+	fetchReq := &pb.FetchRequest{
+		TopicName:    topicName,
+		NumRequested: defaultBatchSize,
+	}
+
+	// TODO(INT-314): Use the latest replay ID if available for resumption
```
daabr: https://google.github.io/styleguide/go/decisions#comment-sentences

integrations/salesforce/integration.go
```diff
@@ -27,23 +30,44 @@ func New(v sdkservices.Vars) sdkservices.Integration {
 func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.DispatchFunc) {
 	common.ServeStaticUI(m, desc, static.SalesforceWebContent)
 
-	h := newHTTPHandler(l, v, o, d)
+	h := newHTTPHandler(l, v, o, d, sdktypes.NewIntegrationIDFromName(desc.UniqueName().String()))
```
daabr: I meant that you don't need to call `sdktypes.NewIntegrationIDFromName` here, you can call it just before line 52, and that would make this line and line 51 shorter and simpler.

integrations/salesforce/integration.go
```diff
@@ -27,23 +30,44 @@ func New(v sdkservices.Vars) sdkservices.Integration {
...
+	cids, err := h.vars.FindConnectionIDs(ctx, h.integrationID, instanceURLVar, "")
+	if err != nil {
+		h.logger.Error("failed to list Salesforce connection IDs", zap.Error(err))
+		return
+	}
+
+	for _, cid := range cids {
+		data, err := h.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+		if err != nil {
+			h.logger.With(zap.String("connection_id", cid.String())).Error("can't restart Salesforce PubSub connection", zap.Error(err))
```
daabr: The point of using `With` is to initialize (in order to use multiple times) a new logger, based on `h.logger` but with additional situation-specific attributes.

In this case, if it doesn't make sense to pass `l` to `h.subscribe`, you don't need `With`, just add the attributes after the message and before `zap.Error`.

integrations/salesforce/grpc.go
```diff
@@ -0,0 +1,46 @@
...
+)
+
+// https://developer.salesforce.com/docs/platform/pub-sub-api/guide/supported-auth.html
+type grpcAuth struct {
+	accessToken string
+	instanceURL string
+	tenantID    string
+}
+
+func initConn(accessToken, instanceURL, orgID string, l *zap.Logger) (*grpc.ClientConn, error) {
```
daabr: `l` should be the first parameter

integrations/salesforce/grpc.go
```diff
@@ -0,0 +1,46 @@
...
+import (
+	"context"
+	"crypto/tls"
+
+	"go.uber.org/zap"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+// https://developer.salesforce.com/docs/platform/pub-sub-api/guide/supported-auth.html
```
daabr, line 12:
```suggestion
// https://developer.salesforce.com/docs/platform/pub-sub-api/guide/supported-auth.html
// https://pkg.go.dev/google.golang.org/grpc/credentials#PerRPCCredentials
```

integrations/salesforce/auth.go
```diff
@@ -0,0 +1,72 @@
...
+		AccessToken:  vs.GetValue(common.OAuthAccessTokenVar),
+		RefreshToken: vs.GetValue(common.OAuthRefreshTokenVar),
+		TokenType:    vs.GetValue(common.OAuthTokenTypeVar),
+	}
+
+	if expiryStr := vs.GetValue(common.OAuthExpiryVar); expiryStr != "" {
+		if expiry, err := time.Parse(time.RFC3339, expiryStr); err == nil {
+			t1.Expiry = expiry
+		} else {
+			// TODO: Question: Should we force a refresh here?
```
daabr: Yes. This is called (indirectly) in `subscribe` which doesn't necessarily happen only when the connection is created.

Also, in the latest version of this PR, all connections should be initialized with a valid expiry. So you can move this parsing before initializing `t1` and add the expiry timestamp to it like the other fields.

If the parsing fails, set the expiry to "now - 1 minute". Also, no need for having `expiryStr` and checking if it's empty or not.

integrations/salesforce/auth.go
```diff
@@ -0,0 +1,72 @@
...
+
+	vs, err := h.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+	if err != nil {
+		l.Error("failed to read connection vars",
+			zap.String("connection_id", cid.String()), zap.Error(err),
+		)
+		return ""
+	}
+
+	switch authType := common.ReadAuthType(vs); authType {
```
daabr: The prefix of a bearer token in the Authorization header should always be `"Bearer "` regardless of auth type. So this switch is redundant, even if/when we'll support more than 1 auth type.

integrations/salesforce/auth.go
pashafateev: Done

integrations/salesforce/auth.go
pashafateev: I was mostly keeping it for the default case

integrations/salesforce/grpc.go
pashafateev: Done

integrations/salesforce/integration.go
pashafateev: Ah nice. Done.

integrations/salesforce/integration.go
pashafateev: Done

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go

integrations/salesforce/auth.go
pashafateev: Done

integrations/salesforce/auth.go
```diff
@@ -0,0 +1,62 @@
...
+	"go.autokitteh.dev/autokitteh/internal/backend/auth/authcontext"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+// oauthToken returns the OAuth token stored in the
+// connection variables. If it's stale, we refresh it first.
+func oauthToken(ctx context.Context, vs sdktypes.Vars, o sdkservices.OAuth) *oauth2.Token {
+	expiryStr := vs.GetValue(common.OAuthExpiryVar)
+	exp, err := time.Parse(time.RFC3339, expiryStr)
```
daabr:
```suggestion
	exp, err := time.Parse(time.RFC3339, vs.GetValue(common.OAuthExpiryVar))
```

integrations/salesforce/auth.go
```diff
@@ -0,0 +1,62 @@
...
+	if err != nil {
+		exp = time.Now().Add(-time.Minute)
```
daabr: I'm not 100% it's needed, but for safety convert to UTC (`Now().UTC()...`)

integrations/salesforce/checks.go
```diff
@@ -55,3 +60,41 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
 		return sdktypes.NewStatus(sdktypes.StatusCodeError, "Not implemented"), nil
 	})
 }
+
+func getUserInfo(ctx context.Context, instanceURL, accessToken string) (map[string]interface{}, error) {
+	u, err := url.JoinPath(instanceURL, "services", "oauth2", "userinfo")
```
daabr:
```suggestion
	u, err := url.JoinPath(instanceURL, "services/oauth2/userinfo")
```

integrations/salesforce/checks.go
```diff
@@ -55,3 +60,41 @@ func test(v sdkservices.Vars) sdkintegrations.OptFn {
...
+	if err != nil {
+		return nil, err
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u, nil)
```
daabr:
```suggestion
	req, err := http.NewRequestWithContext(ctx, http.MethodGet, u, http.NoBody)
```

integrations/salesforce/integration.go
```diff
@@ -28,22 +31,43 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, o sdkservices.OAut
 	common.ServeStaticUI(m, desc, static.SalesforceWebContent)
 
 	h := newHTTPHandler(l, v, o, d)
+
```
daabr:
```suggestion
```

integrations/salesforce/integration.go
```diff
@@ -28,22 +31,43 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, o sdkservices.OAut
...
+	logger        *zap.Logger
+	vars          sdkservices.Vars
+	oauth         sdkservices.OAuth
+	dispatch      sdkservices.DispatchFunc
+	integrationID sdktypes.IntegrationID
 }
 
 func newHTTPHandler(l *zap.Logger, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.DispatchFunc) handler {
-	return handler{logger: l, oauth: o, vars: v, dispatch: d}
+	return handler{logger: l, oauth: o, vars: v, dispatch: d, integrationID: sdktypes.NewIntegrationIDFromName(desc.UniqueName().String())}
```
daabr: Cleaner
```suggestion
	iid := sdktypes.NewIntegrationIDFromName(desc.UniqueName().String())
	return handler{logger: l, oauth: o, vars: v, dispatch: d, integrationID: iid}
```

integrations/salesforce/oauth.go
```diff
@@ -76,10 +95,12 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
 	}
 
 	http.Redirect(w, r, urlPath, http.StatusFound)
+
+	h.subscribe(instanceURL, orgID, cid)
```
daabr, line 96: Nit: even though the effect of `http.Redirect` happens only after we exit this function, it's easier to understand when it's also the last statement - please reverse the order with `h.subscribe`.

integrations/salesforce/oauth.go
```diff
@@ -88,6 +109,59 @@ func (h handler) saveConnection(ctx context.Context, vsid sdktypes.VarScopeID, t
...
+
+	body, err := io.ReadAll(resp.Body)
+	if err != nil {
+		return err
+	}
+
+	var tokenInfo map[string]any
+	if err := json.Unmarshal(body, &tokenInfo); err != nil {
+		return errors.New("failed to parse token info")
+	}
```
daabr: Nit: you can combine them together with `json.NewDecoder(resp.Body).Decode(tokenInfo)` - a few less lines of code, one less error check

integrations/salesforce/subscription.go
```diff
@@ -0,0 +1,143 @@
...
+	// Prevent duplication due to race conditions.
+	mu.Lock()
+	defer mu.Unlock()
+
+	if _, ok := pubSubClients[instanceURL]; ok {
+		return
+	}
+
+	ctx := context.Background()
+	conn, err := initConn(h.logger, h.bearerToken(ctx, h.logger, cid), instanceURL, orgID)
```
daabr: Instead of passing `h.logger`, use the same as in line 52 - this will help a lot to debug RPC issues

daabr (approved):

integrations/salesforce/auth.go
pashafateev: What's your thought process behind that?

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/oauth.go
pashafateev: Done

integrations/salesforce/auth.go
daabr: RFC3339 was designed to be machine-readable, so even if you have a timezone in your timestamp (due to using local time) it shouldn't be a problem, but regardless stored timestamps are almost always in UTC to ensure absolute consistency and safety: even if the data doesn't move between timezones, the local timezone still changes twice a year due to daylight savings.

Concrete example: the times just before and just after a daylight saving time switch don't *look* like they almost the same, unless they're in UTC.

integrations/salesforce/subscription.go
pashafateev: done

PR #1163, opened 2025-03-05
chore(go): Update to go 1.24.1 and bump dependencies

tebeka:


go.mod
```diff
@@ -93,44 +93,43 @@ require (
 )
 
 require (
-	ariga.io/atlas-go-sdk v0.6.5 // indirect
-	cel.dev/expr v0.19.2 // indirect
-	cloud.google.com/go v0.118.2 // indirect
-	cloud.google.com/go/ai v0.10.0 // indirect
-	cloud.google.com/go/auth v0.14.1 // indirect
+	ariga.io/atlas-go-sdk v0.2.3 // indirect
```
daabr: Why are all the versions of the indirect deps **older** than before?

tebeka: Good question. I deleted everything from `go.mod` and ran `go mod tidy`.
I couldn't reproduce with a smaller project (just `cel-go`)

Also, it's not all (for example `github.com/googleapis/enterprise-certificate-proxy`)

tebeka: Updated indirects manually to latest versions as well.

efiShtain (approved):

PR #1164, opened 2025-03-06
feat(linear): events webhook

daabr:
Refs: INT-284, INT-322

PR #1166, opened 2025-03-08
refactor(integration): misc alignments after #1164

daabr:
Refactoring follow-ups across all new-style integrations, following #1164.

See list of commits for a more granular view of changes: https://github.com/autokitteh/autokitteh/pull/1166/commits

Refs: INT-284

### DIFF
--- a/integrations/microsoft/connection/checks.go
+++ b/integrations/microsoft/connection/checks.go
@@ -24,7 +24,7 @@ import (
 // ensures that the connection is at least theoretically usable.
 func Status(v sdkservices.Vars) sdkintegrations.OptFn {
 	return sdkintegrations.WithConnectionStatus(func(ctx context.Context, cid sdktypes.ConnectionID) (sdktypes.Status, error) {
-		vs, errStatus, err := common.ReadConnectionVars(ctx, v, cid)
+		vs, errStatus, err := common.ReadVarsWithStatus(ctx, v, cid)
 		if errStatus.IsValid() || err != nil {
 			return errStatus, err
 		}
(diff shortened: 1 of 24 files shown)


integrations/microsoft/connection/checks.go
```diff
@@ -108,10 +108,6 @@ func GetUserInfo(ctx context.Context, t *oauth2.Token) (*UserInfo, error) {
 	return &user, nil
 }
 
-type orgInfoWrapper struct {
```
pashafateev, line 111: why did you choose to remove the type?

integrations/salesforce/integration.go
```diff
@@ -40,29 +40,30 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, o sdkservices.OAut
...
 func newHTTPHandler(l *zap.Logger, v sdkservices.Vars, o sdkservices.OAuth, d sdkservices.DispatchFunc) handler {
-	iid := sdktypes.NewIntegrationIDFromName(desc.UniqueName().String())
-	return handler{logger: l, oauth: o, vars: v, dispatch: d, integrationID: iid}
+	l = l.With(zap.String("integration", desc.UniqueName().String()))
+	return handler{logger: l, oauth: o, vars: v, dispatch: d}
 }
 
 func (h handler) reopenExistingPubSubConnections(ctx context.Context) {
-	cids, err := h.vars.FindConnectionIDs(ctx, h.integrationID, instanceURLVar, "")
+	cids, err := h.vars.FindConnectionIDs(ctx, desc.ID(), instanceURLVar, "")
```
pashafateev, line 55: they're the same?

integrations/microsoft/connection/checks.go
daabr: I'm using an anonymous struct instead - see line 136

integrations/salesforce/integration.go
daabr: Yes - see https://github.com/autokitteh/autokitteh/blob/main/integrations/common/integration.go#L16

pashafateev (approved):

PR #1167, opened 2025-03-09
refactor(integration): common HTTP request logic

daabr:
Refactoring follow-ups across all integrations, following #1164 and #1166.

All the changes in this PR are effectively the same:

- Replace direct usage of `http.NewRequestWithContext` with `common.HTTP*` for code reuse and standardization of best practices
- Move timeout definitions from the `http.Client` to `ctx`
- Remove duplicated `consts` related to HTTP headers
- `jira/event.go`: use `common.DispatchEvent` (should have been in #1166)
- `salesforce/oauth.go`: save one reading of client ID connection variable
- `salesforce/oauth.go`: simpler parsing of expiration timestamps
- Merge simplified `salesforce/auth.go` into `/salesforce/subscription.go`

Refs: INT-284


integrations/atlassian/confluence/webhooks.go
```diff
@@ -110,23 +111,27 @@
...
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, base+restPath, http.NoBody)
 	if err != nil {
 		l.Warn("Failed to construct HTTP request to list Confluence webhooks", zap.Error(err))
 		return 0, false
 	}
 
 	req.SetBasicAuth(user, key)
 
-	resp, err := httpClient.Do(req)
+	resp, err := http.DefaultClient.Do(req)
```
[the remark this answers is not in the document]
daabr: base URL intentionally received from Atlassian Confluence webhook - that's how their API works.

integrations/common/http.go
```diff
@@ -19,30 +20,84 @@ const (
 	ContentTypeForm            = "application/x-www-form-urlencoded"
 	ContentTypeJSON            = "application/json"                // Accept
 	ContentTypeJSONCharsetUTF8 = "application/json; charset=utf-8" // Content-Type
+
+	HTTPTimeout = 3 * time.Second
```
mario99logic, line 25: Is the `HTTPTimeout` appropriate for all requests? 

Should we allow custom timeouts?

integrations/common/http.go
```diff
@@ -19,30 +20,84 @@ const (
 	ContentTypeForm            = "application/x-www-form-urlencoded"
 	ContentTypeJSON            = "application/json"                // Accept
 	ContentTypeJSONCharsetUTF8 = "application/json; charset=utf-8" // Content-Type
+
```
mario99logic, line 24: Would it be better to make an HTTPClient struct with configurable timeouts and headers?

integrations/atlassian/confluence/oauth.go
```diff
@@ -93,30 +95,16 @@ type resource struct {
...
-	defer resp.Body.Close()
-
-	if resp.StatusCode != http.StatusOK {
-		logWarnIfNotNil(l, "Unexpected response on accessible resources", zap.Int("status", resp.StatusCode))
-		return nil, fmt.Errorf("accessible resources: unexpected status code %d", resp.StatusCode)
-	}
-
 	var res []resource
-	if err := json.NewDecoder(resp.Body).Decode(&res); err != nil {
+	if err := json.Unmarshal(resp, &res); err != nil {
```
mario99logic, line 107: Doesn't `Decode()` handle incomplete JSON more gracefully?

mario99logic (approved):

integrations/atlassian/confluence/client.go
```diff
@@ -145,7 +148,7 @@ func apiTokenConnTest(ctx context.Context, l *zap.Logger, vs sdktypes.Vars) erro
 	req.SetBasicAuth(email, token)
 	req.Header.Set("Accept", "application/json")
 
-	resp, err := httpClient.Do(req)
+	resp, err := http.DefaultClient.Do(req)
```
tebeka, line 151: Why no use the functions from common?

integrations/atlassian/confluence/event.go
```diff
@@ -55,9 +46,9 @@ func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
 	// (Confluence doesn't recognize secrets when creating webhooks).
 
 	// Check the "Content-Type" header.
-	contentType := r.Header.Get(headerContentType)
-	if !strings.HasPrefix(contentType, contentTypeJSON) {
-		l.Warn("Incoming Atlassian event with bad header", zap.String(headerContentType, contentType))
+	contentType := r.Header.Get(common.HeaderContentType)
+	if !strings.HasPrefix(contentType, common.ContentTypeJSON) {
```
tebeka: Create `IsJSONRequest` function?

integrations/atlassian/confluence/oauth.go
tebeka: I don't think so, it's the same underlying code (https://github.com/golang/go/blob/master/src/encoding/json/decode.go)

integrations/atlassian/confluence/webhooks.go
```diff
@@ -110,23 +112,27 @@ var webhookEvents = map[string][]string{
 // https://developer.atlassian.com/server/confluence/webhooks/
 // https://confluence.atlassian.com/doc/managing-webhooks-1021225606.html
 func getWebhook(ctx context.Context, l *zap.Logger, base, user, key, category string) (int, bool) {
+	ctx, cancel := context.WithTimeout(ctx, common.HTTPTimeout)
+	defer cancel()
+
 	// TODO(ENG-965): Support pagination.
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, base+restPath, nil)
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, base+restPath, http.NoBody)
```
tebeka, line 119: I find `nil` more readable than `http.NoBody`

integrations/atlassian/confluence/webhooks.go
```diff
@@ -110,23 +112,27 @@ var webhookEvents = map[string][]string{
...
+	resp, err := http.DefaultClient.Do(req)
 	if err != nil {
 		l.Warn("Failed to list Confluence webhooks", zap.Error(err))
 		return 0, false
 	}
 	defer resp.Body.Close()
 
-	body, err := io.ReadAll(resp.Body)
+	// Read the response's body, up to 1 MiB.
+	body, err := io.ReadAll(io.LimitReader(resp.Body, 1<<20))
```
tebeka: Use [http.MaxBytesReader](https://pkg.go.dev/net/http#MaxBytesReader)
Should size be configurable?

integrations/atlassian/jira/save.go
```diff
@@ -8,21 +8,18 @@ import (
...
-
 // handleAuth saves a new AutoKitteh connection with user-submitted data.
 func (h handler) handleSave(w http.ResponseWriter, r *http.Request) {
 	c, l := sdkintegrations.NewConnectionInit(h.logger, w, r, desc)
 
 	// Check "Content-Type" header.
-	contentType := r.Header.Get(headerContentType)
-	if !strings.HasPrefix(contentType, contentTypeForm) {
+	contentType := r.Header.Get(common.HeaderContentType)
+	if !strings.HasPrefix(contentType, common.ContentTypeForm) {
```
tebeka: Create `IsFormRequest` function.

integrations/atlassian/jira/webhooks.go
```diff
@@ -1,19 +1,17 @@
...
-	"io"
-	"net/http"
 	"os"
 	"slices"
 	"strings"
 	"time"
 
 	"go.uber.org/zap"
 
+	"go.autokitteh.dev/autokitteh/integrations/common"
```
tebeka, line 14: In general I don't like "common" packages

integrations/atlassian/jira/webhooks.go
```diff
@@ -84,39 +82,17 @@ type webhookListResponse struct {
 // https://developer.atlassian.com/server/jira/platform/webhooks/
 func getWebhook(ctx context.Context, l *zap.Logger, baseURL, token string) (int, bool) {
 	// TODO(ENG-965): Support pagination.
-	req, err := http.NewRequestWithContext(ctx, http.MethodGet, baseURL+"/rest/api/3/webhook", nil)
+	u := baseURL + "/rest/api/3/webhook"
+	resp, err := common.HTTPGet(ctx, u, "Bearer "+token)
```
tebeka, line 86: Maybe have an `HTTPClient` in `common` with a `Get` method (also `Post` ...) and `Do`? (like the `http.DeaultClient`)

integrations/atlassian/jira/webhooks.go
```diff
@@ -84,39 +82,17 @@ type webhookListResponse struct {
...
-			zap.Int("status", resp.StatusCode),
-			zap.ByteString("body", body),
-		)
+		l.Warn("Failed to request Jira webhook", zap.Error(err))
 		return 0, false
 	}
 
 	var list webhookListResponse
-	if err := json.Unmarshal(body, &list); err != nil {
+	if err := json.Unmarshal(resp, &list); err != nil {
```
tebeka, line 93: Limit size

integrations/atlassian/jira/webhooks.go
```diff
@@ -225,58 +179,24 @@ func registerWebhook(ctx context.Context, l *zap.Logger, baseURL, token string)
 		},
 	}
 
-	body, err := json.Marshal(r)
+	u := baseURL + "/rest/api/3/webhook"
+	resp, err := common.HTTPPostJSON(ctx, u, "Bearer "+token, req)
```
tebeka, line 183: Thinking about this API, maybe use something line gin's [Context.JSON](https://pkg.go.dev/github.com/gin-gonic/gin#Context.JSON)

integrations/auth0/oauth.go
```diff
@@ -70,38 +70,12 @@ func (h handler) handleOAuth(w http.ResponseWriter, r *http.Request) {
...
-
-	// Check response status code.
-	if resp.StatusCode != http.StatusOK {
-		l.Error("Auth0 API request failed",
-			zap.Int("status_code", resp.StatusCode),
-			zap.String("response", string(body)))
-		c.AbortBadRequest("Auth0 API request failed")
+	// Tests Auth0's Management API.
+	url := fmt.Sprintf("https://%s/api/v2/roles", d)
+	auth := "Bearer " + oauthToken.AccessToken
```
tebeka, line 75: This is coming around a log, maybe a `GenAuth` function

integrations/common/http.go
```diff
@@ -19,30 +20,84 @@ const (
...
+	HTTPTimeout = 3 * time.Second
+	HTTPMaxSize = 1 << 23 // 2^23 bytes = 8 MiB
```
tebeka, line 26:
```suggestion
	MaxHTTPRequestBodySize = 1 << 23 // 2^23 bytes = 8 MiB
```

integrations/common/http.go
```diff
@@ -19,30 +20,84 @@ const (
...
 )
 
+func HTTPDeleteJSON(ctx context.Context, u, auth string, payload any) ([]byte, error) {
+	if s, ok := payload.(string); ok {
```
tebeka, line 40: This is risky, I don't like guessing.
I prefer to explicit declare that you want to convert the value into JSON. Maybesome something in the lines of:
```go
type JSON struct {
	v any
}

func JSONed(v any) JSON {
	return JSON{v}
}

func encode(v any) {
	if _, ok := v.(JSON); ok {
		println("json")
	} else {
		println("not json")
	}
}

func main() {
	encode("a")
	encode(JSONed("a"))
}
```

Or pass `io.Reader`

integrations/common/http.go
```diff
@@ -19,30 +20,84 @@ const (
...
+	}
+
+	body, err := json.Marshal(payload)
+	if err != nil {
+		return nil, fmt.Errorf("failed to marshal JSON payload: %w", err)
+	}
+	return httpRequest(ctx, http.MethodDelete, u, auth, ContentTypeJSONCharsetUTF8, body)
+}
+
+func HTTPGet(ctx context.Context, u, auth string) ([]byte, error) {
```
tebeka, line 51: Don't see a lot of value in this convenience function.

integrations/common/http.go
```diff
@@ -19,30 +20,84 @@ const (
...
+	return httpGet(ctx, u, auth, "", nil)
+}
+
+func HTTPGetJSON(ctx context.Context, u, auth string, payload any) ([]byte, error) {
+	if s, ok := payload.(string); ok {
```
tebeka, line 56: See comment above on guessing

integrations/common/http.go
```diff
@@ -62,8 +117,8 @@ func HTTPPost(ctx context.Context, u, auth, contentType string, body []byte) ([]
 	}
 	defer resp.Body.Close()
 
-	// Read the response's body, up to 8 MiB.
-	payload, err := io.ReadAll(io.LimitReader(resp.Body, 1<<23))
+	// Read the response's body.
+	payload, err := io.ReadAll(io.LimitReader(resp.Body, HTTPMaxSize))
```
tebeka: Use `HTTPMaxBytesReader`

integrations/microsoft/connection/auth.go
```diff
@@ -35,39 +32,23 @@ func DaemonToken(ctx context.Context, vs sdktypes.Vars) (*oauth2.Token, error) {
...
-	data := url.Values{}
-	data.Set("grant_type", "client_credentials")
-	data.Set("client_id", vs.GetValue(privateClientIDVar))
-	data.Set("client_secret", vs.GetValue(privateClientSecretVar))
-	data.Set("scope", "https://graph.microsoft.com/.default")
+	form := url.Values{}
+	form.Set("grant_type", "client_credentials")
+	form.Set("client_id", vs.GetValue(privateClientIDVar))
+	form.Set("client_secret", vs.GetValue(privateClientSecretVar))
+	form.Set("scope", "https://graph.microsoft.com/.default")
```
tebeka:
```suggestion
	form := url.Values{
		"grant_type":    {"client_credentials"},
		"client_id":     {vs.GetValue(privateClientIDVar)},
		"client_secret": {vs.GetValue(privateClientSecretVar)},
		"scope":         {"https://graph.microsoft.com/.default"},
	}
```

tebeka (approved):
A lot of cosmetic comment, up to you ...

integrations/common/http.go
daabr: We should/will allow it in the future, but it's not needed for now

integrations/common/http.go
daabr: Different requests use different values - so we can't add them all to the client, but that's why we have the HTTP* functions to wrap this.

integrations/atlassian/confluence/client.go
daabr: In this specific case the request uses simple auth instead of a bearer token. I could support it too in the underlying common function, but it would make it more complex just for this one-off, so I preferred not to do it.

integrations/atlassian/confluence/event.go
daabr: good idea, done

integrations/atlassian/confluence/webhooks.go
daabr: `http.MaxBytesReader` - thanks!
Configurability: in the future probably, but for now it's premature optimization.

integrations/atlassian/jira/save.go
daabr: done

integrations/atlassian/jira/webhooks.go
daabr: I agree, it's actually noted as a bad practice in the Go style guide, but I don't have a better name yet.

integrations/atlassian/jira/webhooks.go
daabr: no need, the body that was read (by the new common function) is already size-limited

integrations/auth0/oauth.go
daabr: We have a common function for hadling OAuth tokens being used in HTTP requests: `FreshOAuthToken`.

As for prepending `"Bearer "`: while it's a common and standard-based pattern, unfortunately it's not absolute, and replacing `a + b` with `a(b)` actually obscures the meaning, even if it's simple.

integrations/common/http.go
daabr: That's a whole lot of code to replace a simple "if string return bytes, else convert to JSON".

integrations/common/http.go
daabr: done

integrations/common/http.go
daabr: done in all instances of `LimitReader`

integrations/atlassian/jira/client.go
```diff
@@ -122,7 +123,7 @@ func oauthConnTest(vs sdktypes.Vars) error {
 
 	// TODO(INT-173): Create & use a new access token using the refresh token.
 	token := vs.GetValueByString("oauth_AccessToken")
-	_, err = accessibleResources(nil, baseURL, token)
+	_, err = accessibleResources(context.Background(), nil, baseURL, token)
```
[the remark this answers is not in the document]
daabr: ack, will address in a separate PR that refactors Atlassian integrations

PR #1169, opened 2025-03-10
feat: Add Zoom event handler 

mario99logic:
refs: INT-242

### DIFF
--- a/integrations/zoom/events.go
+++ b/integrations/zoom/events.go
@@ -10,7 +10,6 @@ import (
 	"io"
 	"net/http"
 	"os"
-	"strings"
 
 	"go.uber.org/zap"
 
@@ -87,8 +86,8 @@ func (h handler) checkRequest(w http.ResponseWriter, r *http.Request) map[string
 	l := h.logger.With(zap.String("url_path", r.URL.Path))
 
 	// Check the request's HTTP headers.
-	ct := r.Header.Get("Content-Type")
-	if !strings.HasPrefix(ct, "application/json") {
+	if common.PostWithoutJSONContentType(r) {
+		ct := r.Header.Get(common.HeaderContentType)
 		l.Warn("incoming event: unexpected content type", zap.String("content_type", ct))
 		common.HTTPError(w, http.StatusBadRequest)
 		return nil
--- a/integrations/zoom/events.go
+++ b/integrations/zoom/events.go
@@ -160,11 +160,9 @@ func (h handler) signingSecret() (string, error) {
 // header, using HMAC-SHA256 for integrity verification.
 // (based on https://developers.zoom.us/docs/api/webhooks/#verify-with-zooms-header)
 func (h handler) checkSignature(signature, timestamp, secret string, body []byte) bool {
-	message := fmt.Sprintf("v0:%s:%s", timestamp, string(body))
-
 	// Create HMAC SHA-256 hash using the webhook secret token.
 	mac := hmac.New(sha256.New, []byte(secret))
-	mac.Write([]byte(message))
+	mac.Write([]byte(fmt.Sprintf("v0:%s:%s", timestamp, string(body))))
 	expectedSignature := "v0=" + hex.EncodeToString(mac.Sum(nil))
 
 	return hmac.Equal([]byte(signature), []byte(expectedSignature))
--- a/integrations/zoom/events.go
+++ b/integrations/zoom/events.go
@@ -1,7 +1,6 @@
 package zoom
 
 import (
-	"bytes"
 	"crypto/hmac"
 	"crypto/sha256"
 	"encoding/hex"
@@ -85,10 +84,7 @@ func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
 // required headers, and signature and returns the parsed JSON payload if valid.
 // Otherwise it returns nil with an HTTP error response.
 func (h handler) checkRequest(w http.ResponseWriter, r *http.Request) map[string]any {
-	l := h.logger.With(
-		zap.String("url_path", r.URL.Path),
-		zap.String("event_type", r.Header.Get("Zoom-Event")),
-	)
+	l := h.logger.With(zap.String("url_path", r.URL.Path))
 
 	// Check the request's HTTP headers.
 	ct := r.Header.Get("Content-Type")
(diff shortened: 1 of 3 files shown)


integrations/zoom/events.go
```diff
@@ -0,0 +1,170 @@
...
+	r.Body = io.NopCloser(bytes.NewBuffer(body))
+
+	secret, err := h.signingSecret()
+	if err != nil {
+		l.Error("incoming event: signing secret not found", zap.Error(err))
+		common.HTTPError(w, http.StatusInternalServerError)
+		return nil
+	}
+
+	if secret == "" {
```
pashafateev, line 120: why is this check necessary when you check it in `signingSecret()`?
I.e. if `secret == ""` it will never make it to line 125

daabr: Good question. Mario asked me about this too :)

`signingSecret()` (in Linear and here) currently supports only default OAuth. When we'll support private OAuth apps there:

- If the relevant connections don't contain a secret token, return an error, just like the existing check for default OAuth apps
- If there are no relevant connections, we should silently ignore the event, not report an error - so return `"", nil`.

integrations/zoom/events.go
```diff
@@ -0,0 +1,170 @@
...
+	if err := json.Unmarshal(body, &payload); err != nil {
+		l.Warn("incoming event: failed to parse JSON body", zap.Error(err))
+		common.HTTPError(w, http.StatusBadRequest)
+		return nil
+	}
+
+	return payload
+}
+
+// retrieves the Zoom secret token from environment variables.
```
daabr:
```suggestion
// signingSecret retrieves the Zoom secret token from environment variables.
// TODO(INT-332): Support all AutoKitteh connection auth types.
```

integrations/zoom/events.go
```diff
@@ -0,0 +1,170 @@
...
+// checkSignature compares the computed hash of the request body, using a preconfigured
+// Zoom webhook secret token, to the signature provided in the request's `x-zm-signature`
+// header, using HMAC-SHA256 for integrity verification.
+// (based on https://developers.zoom.us/docs/api/webhooks/#verify-with-zooms-header)
+func (h handler) checkSignature(signature, timestamp, secret string, body []byte) bool {
+	message := fmt.Sprintf("v0:%s:%s", timestamp, string(body))
+
+	// Create HMAC SHA-256 hash using the webhook secret token.
+	mac := hmac.New(sha256.New, []byte(secret))
+	mac.Write([]byte(message))
```
daabr:
```suggestion
	// Create HMAC SHA-256 hash using the webhook secret token.
	mac := hmac.New(sha256.New, []byte(secret))
	mac.Write([]byte(fmt.Sprintf("v0:%s:%s", timestamp, string(body))))
```

integrations/zoom/events.go
```diff
@@ -0,0 +1,170 @@
...
+		return "", errors.New("missing Zoom secret token")
+	}
+	return secret, nil
+}
+
+// checkSignature compares the computed hash of the request body, using a preconfigured
+// Zoom webhook secret token, to the signature provided in the request's `x-zm-signature`
+// header, using HMAC-SHA256 for integrity verification.
+// (based on https://developers.zoom.us/docs/api/webhooks/#verify-with-zooms-header)
+func (h handler) checkSignature(signature, timestamp, secret string, body []byte) bool {
```
daabr: `handler` isn't used in this function, so it's not needed as a received in this function.

Note that the same can be said about `signingSecret()` above, but that will change as soon as we implement INT-332.

integrations/zoom/events.go
```diff
@@ -0,0 +1,170 @@
...
+// required headers, and signature and returns the parsed JSON payload if valid.
+// Otherwise it returns nil with an HTTP error response.
+func (h handler) checkRequest(w http.ResponseWriter, r *http.Request) map[string]any {
+	l := h.logger.With(
+		zap.String("url_path", r.URL.Path),
+		zap.String("event_type", r.Header.Get("Zoom-Event")),
+	)
+
+	// Check the request's HTTP headers.
+	ct := r.Header.Get("Content-Type")
```
daabr: See new change in https://github.com/autokitteh/autokitteh/pull/1167/files#diff-b31fccecc7fd296ae2a58699fde7718f6dd63ae88011c5e28642fa72cc3fe41a

integrations/zoom/events.go
```diff
@@ -0,0 +1,170 @@
...
+		return nil
+	}
+
+	body, err := io.ReadAll(r.Body)
+	if err != nil {
+		l.Error("incoming event: failed to read request body", zap.Error(err))
+		common.HTTPError(w, http.StatusInternalServerError)
+		return nil
+	}
+	r.Body = io.NopCloser(bytes.NewBuffer(body))
```
daabr: 1. Don't modify the `request` reference we were given
2. I think `NopCloser` has no useful affect *after* you've already consumed the body in line 110?

integrations/zoom/events.go
```diff
@@ -0,0 +1,170 @@
...
+	signature := r.Header.Get("x-zm-signature")
+	timestamp := r.Header.Get("x-zm-request-timestamp")
+
+	if signature == "" || timestamp == "" {
+		l.Warn("incoming event: missing Zoom verification headers")
+		common.HTTPError(w, http.StatusBadRequest)
+		return nil
+	}
+
+	body, err := io.ReadAll(r.Body)
```
daabr:
```suggestion
	// Read the request's JSON body, up to 8 MiB, to prevent DDoS attacks.
	body, err := io.ReadAll(http.MaxBytesReader(nil, r.Body, 1<<23))
```

integrations/zoom/events.go
```diff
@@ -0,0 +1,170 @@
...
+func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+
+	// Check the request's headers and parse its body.
+	zoomEvent := h.checkRequest(w, r)
+	if zoomEvent == nil {
+		return
+	}
+
+	eventType, ok := zoomEvent["event"].(string)
```
daabr, line 34: What's the difference between this and the HTTP header in line 90?

mario99logic: I forgot to remove the HTTP header.
In zoom request there is no header about the event type, only in the body.

daabr (approved):

PR #1170, opened 2025-03-11
fix(salesforce): ignore self-triggering events

pashafateev:
testing: successfully received event when using Salesforce UI and successfully ignored events coming from API 

refs: INT-327

### DIFF
--- a/integrations/salesforce/subscription.go
+++ b/integrations/salesforce/subscription.go
@@ -99,6 +99,16 @@ func (h handler) eventLoop(ctx context.Context, client pb.PubSubClient, subscrib
 				continue
 			}
 
+			// TODO(INT-329): Temporary filter to ignore self-triggered events.
+			changeOrigin, ok := data["ChangeOrigin"].(string)
+			if !ok {
+				l.Error("ChangeOrigin is not a string in event data")
+				continue
+			}
+			if !strings.Contains(changeOrigin, "SfdcInternalAPI") {
+				continue
+			}
+
 			// Extract changed entity name for the event type.
 			header, ok := data["ChangeEventHeader"].(map[string]any)
 			if !ok {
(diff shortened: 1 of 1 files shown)


integrations/salesforce/subscription.go
```diff
@@ -105,6 +105,17 @@ func (h handler) eventLoop(ctx context.Context, client pb.PubSubClient, subscrib
 				l.Error("ChangeEventHeader is not a map in event data")
 				continue
 			}
+
+			// TODO(INT-329): Temporary filter to ignore self-triggered events.
+			changeOrigin, ok := header["changeOrigin"].(string)
```
daabr: notice that `ok` here tells you that they is found:
```
changeOrigin, ok := header["changeOrigin"]
```
and `ok` here tells that you it's successfully cast into a string:
```
changeOrigin, ok := header["changeOrigin"].(string)
```
But what happens in the latter case if the key is not found? Will the cast still work? Or is the key guaranteed to exist? (Which would make sense, I guess)

integrations/salesforce/subscription.go
```diff
@@ -105,6 +105,17 @@ func (h handler) eventLoop(ctx context.Context, client pb.PubSubClient, subscrib
...
+			if !ok {
+				l.Error("changeOrigin is not a string in ChangeEventHeader")
```
daabr: Since this log message is specific to Salesforce, mention Salesforce somehwhere in this message, so we know what it means when we see it in Datadog

daabr (approved):
Approved with 2 small comments

integrations/salesforce/subscription.go

integrations/salesforce/subscription.go
pashafateev: I'm not sure, so I'll play it safe and add a check.

PR #1171, opened 2025-03-11
fix(sessions): properly return history events when doing descending filter

itayd:


internal/backend/sessions/sessionworkflows/history.go
```diff
@@ -57,6 +57,10 @@ func (ws *workflows) GetWorkflowLog(ctx context.Context, filter sdkservices.Sess
 		enumspb.HISTORY_EVENT_FILTER_TYPE_ALL_EVENT,
 	)
 
+	// Keep track of all the events and their ids in this map so we can easily
```
tebeka, line 60: I don't understand this comment.

internal/backend/sessions/sessionworkflows/history.go
```diff
@@ -70,6 +74,10 @@ func (ws *workflows) GetWorkflowLog(ctx context.Context, filter sdkservices.Sess
 			}
 
 			riter.rs = append(riter.rs, event)
+
+			// On a descending query, we need to prepopulate the events so they'll be available when
+			// we bump into a "later" event before the "earlier" event.
+			events[event.EventId] = event
```
tebeka, line 84: Why not sort events at the end?

internal/backend/sessions/sessionworkflows/history.go
itayd: you know how much time it took me to phrase it. i honestly don't know how to do it better. thing is that in reverse order we get the "future" before the "past" so we need to repopulate the events map in order to "see the future".

internal/backend/sessions/sessionworkflows/history.go
itayd: it's unrelated to sort.

internal/backend/sessions/sessionworkflows/history.go
daabr: General notes for writing better comments:

1. When describing a data collection, you should describe **it**, usually not what happens elsewhere thanks to it (unless it's really important/tricky) - this description is basically "this is a map of A's and B's that tracks A's and B's" - which is a just a generic tautology
2. Abstract and/or unnecessary comments are arguably worse that no comments - because they actively waste time, and potentially create confusion
3. What does "tracking" mean? Not in the generic sense, but in this code? And why does it matter?
4. "We" are not doing anything with it, something specific in this code does, but what/where?
5. "Easily" is a bad fluff word - it doesn't contribute to the understanding or accuracy of the text; e.g. if efficiency or order are important (and they usually are) - why?
6. Typo: "correlated" --> "correlate"
7. "Correlate" just means that things are related to each other, which is obvious since this is a map, but why does it matter? There's still no explanation of the specific purpose

internal/backend/sessions/sessionworkflows/history.go
```diff
@@ -70,6 +74,10 @@ func (ws *workflows) GetWorkflowLog(ctx context.Context, filter sdkservices.Sess
 			}
 
 			riter.rs = append(riter.rs, event)
+
+			// On a descending query, we need to prepopulate the events so they'll be available when
```
daabr: I can guess that you meant "query in descending order". But bump where? Prepopulate where?

I'm looking at the raw code in this function and I don't understand what this comment is trying to explain, i.e. how `events[event.EventId] = event` relates to this text.

daabr (approved):

itayd:
I reworked the comments, I think they're much better now.

PR #1174, opened 2025-03-11
feat(sessions): speed up prints

itayd:
asynchronously save prints, don't do it directly in activity or workflow. this speeds up print behaviour significantly.

this also adds a "session prints" command that returns only prints, while removing the ability to do --prints-only on other commands as they no longer return any prints.

### DIFF
--- a/cmd/ak/cmd/sessions/prints.go
+++ b/cmd/ak/cmd/sessions/prints.go
@@ -35,10 +35,10 @@ var printsCmd = common.StandardCommand(&cobra.Command{
 			}
 
 			if !noTimestamps {
-				fmt.Printf("[%s] ", p.Timestamp.String())
+				fmt.Fprintf(cmd.OutOrStdout(), "[%s] ", p.Timestamp.String())
 			}
 
-			fmt.Println(text)
+			fmt.Fprintln(cmd.OutOrStdout(), text)
 		}
 
 		return nil
--- a/cmd/ak/cmd/sessions/prints.go
+++ b/cmd/ak/cmd/sessions/prints.go
@@ -1,0 +1,53 @@
+package sessions
+
+import (
+	"fmt"
+
+	"github.com/spf13/cobra"
+
+	"go.autokitteh.dev/autokitteh/cmd/ak/common"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var printsCmd = common.StandardCommand(&cobra.Command{
+	Use:   "prints [sessions ID | project] [--fail] [--no-timestamps]",
+	Short: "Get session prints",
+	Args:  cobra.ExactArgs(1),
+
+	RunE: func(cmd *cobra.Command, args []string) error {
+		sid, err := acquireSessionID(args[0])
+		if err = common.AddNotFoundErrIfCond(err, sid.IsValid()); err != nil {
+			return common.ToExitCodeWithSkipNotFoundFlag(cmd, err, "session")
+		}
+
+		ctx, done := common.LimitedContext()
+		defer done()
+
+		prints, err := sessions().GetPrints(ctx, sid, sdktypes.PaginationRequest{})
+		if err != nil {
+			return fmt.Errorf("get log: %w", err)
+		}
+
+		for _, p := range prints.Prints {
+			text, err := p.Value.ToString()
+			if err != nil {
+				text = fmt.Sprintf("error converting print to string: %v", err.Error())
+			}
+
+			if !noTimestamps {
+				fmt.Printf("[%s] ", p.Timestamp.String())
+			}
+
+			fmt.Println(text)
+		}
+
+		return nil
+	},
+})
+
+func init() {
+	// Command-specific flags.
+	printsCmd.Flags().BoolVarP(&noTimestamps, "no-timestamps", "n", false, "omit timestamps from watch output")
+
+	common.AddFailIfNotFoundFlag(printsCmd)
+}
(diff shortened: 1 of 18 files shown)


internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -584,14 +566,14 @@ func (w *sessionWorkflow) run(wctx workflow.Context, l *zap.Logger) (prints []sd
 		}
 
 		if retVal, err = run.Call(ctx, callValue, nil, inputs); err != nil {
-			return prints, sdktypes.InvalidValue, err
+			return printer.Finalize(), sdktypes.InvalidValue, err
 		}
 	}
 
-	return prints, retVal, w.updateState(
+	return printer.all, retVal, w.updateState(
```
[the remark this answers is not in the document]
itayd: nice catch oh magical ai in the cloud. fixed.

cmd/ak/cmd/sessions/prints.go
```diff
@@ -0,0 +1,53 @@
...
+
+		prints, err := sessions().GetPrints(ctx, sid, sdktypes.PaginationRequest{})
+		if err != nil {
+			return fmt.Errorf("get log: %w", err)
+		}
+
+		for _, p := range prints.Prints {
+			text, err := p.Value.ToString()
+			if err != nil {
+				text = fmt.Sprintf("error converting print to string: %v", err.Error())
```
tebeka, line 34: Print to `cmd.OutOrStdout()`

internal/backend/sessions/sessionworkflows/printer.go
```diff
@@ -0,0 +1,67 @@
...
+type print struct {
+	sdkservices.SessionPrint
+
+	// Correlates a print to an activity call.
+	activityCallSeq uint32
+}
+
+// Safe to call event if Finalize has been called.
+func (p *printer) Close() {
+	if p.ch != nil {
```
tebeka: There can be a race condition here and you'll close a `nil` channel - which panics.

internal/backend/sessions/sessionworkflows/printer.go
```diff
@@ -0,0 +1,67 @@
...
+}
+
+func (p *printer) Finalize() []sdkservices.SessionPrint {
+	p.Close()
+	<-p.done
+	return p.all
+}
+
+func (p *printer) Print(sp sdkservices.SessionPrint, activityCallSeq uint32, isReplay bool) {
+	p.all = append(p.all, sp)
```
tebeka, line 48: Limit size of `p.all`?

internal/backend/sessions/sessionworkflows/printer.go
```diff
@@ -0,0 +1,67 @@
...
+
+	if !isReplay {
+		p.ch <- &print{SessionPrint: sp, activityCallSeq: activityCallSeq}
```
tebeka, line 60: This blocks, maybe make `p.ch` buffered a bit?

internal/backend/sessions/sessionworkflows/printer.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+}
+
+func (w *sessionWorkflow) newPrinter() *printer {
+	pr := &printer{ch: make(chan *print, 32), done: make(chan struct{})}
+
+	go func() {
+		ctx := context.Background()
+
+		for p := range pr.ch {
```
tebeka, line 74: If we're doing that, why not write prints in bulks to the database?

internal/backend/sessions/sessionworkflows/printer.go
```diff
@@ -0,0 +1,67 @@
...
+			if err := w.ws.svcs.DB.AddSessionPrint(ctx, w.data.Session.ID(), p.Value, p.activityCallSeq); err != nil {
```
tebeka: Somewhat related question: If we only store prints to the database, why do we have issues with printing from multiple goroutines?

cmd/ak/cmd/sessions/prints.go
itayd: done

internal/backend/sessions/sessionworkflows/printer.go
itayd: this will solve this issue, you wouldn't have to do that from the same goroutine anymore since it's not going through temporal.

internal/backend/sessions/sessionworkflows/printer.go
itayd: it is - its cap is 32.

internal/backend/sessions/sessionworkflows/printer.go
itayd: this will never be called concurrently with any other method here. added comment re that.

itayd: changed the code a bit, killed Close.

internal/backend/sessions/sessionworkflows/printer.go
itayd: done

internal/backend/sessions/sessionworkflows/printer.go
itayd: this is over a channel. no one knows ahead of time how many prints are. we can implement buffering, but i don't think it's necessary right now.

internal/backend/sessions/sessionworkflows/printer.go
tebeka: OK

internal/backend/sessions/sessionworkflows/printer.go
tebeka: OK

tebeka (approved):

PR #1176, opened 2025-03-12
fix(python): Don't print in Start (race)

tebeka:
Fixes: ENG-2027 (again)


efiShtain (approved):
just make sure the frontend show the correct error before merging this one please

PR #1177, opened 2025-03-12
fix(python): 5 seconds timeout 

mario99logic:
- Deleted ak_print
- Replaced sys.stdout and sys.stderr with a tee like utility that will print to the stream and sticks the message to a queue.

refs: ENG-2021


runtimes/pythonrt/runner/tee.py
```diff
@@ -0,0 +1,51 @@
+from queue import Queue
+from threading import Thread
+from time import sleep
+import grpc
+import pb
```
tebeka, line 5:
```suggestion
from queue import Queue
from threading import Thread
from time import sleep

import grpc
import pb
```

Separate stdlib import from others.

runtimes/pythonrt/runner/tee.py
```diff
@@ -0,0 +1,51 @@
...
+
+
+class Tee:
```
tebeka, line 8: Add docstring on what it does.

runtimes/pythonrt/runner/tee.py
```diff
@@ -0,0 +1,51 @@
...
+        self.running = False
+        self.stream.close()
+
+    def write(self, data):
+        self.stream.write(data)
+        self.queue.put(data)
+
+    def flush(self):
+        self.stream.flush()
+        while not self.queue.empty():
```
tebeka, line 28: This might not finish if the thread doing gRPC calls is stuck. Limit this by time as well.

runtimes/pythonrt/runner/tee.py
```diff
@@ -0,0 +1,51 @@
...
+            sleep(0.001)
+
+    def print_thread(self):
+        while self.running:
+            data = self.queue.get()
+            self.rpc_call(data)
+
```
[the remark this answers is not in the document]
tebeka: :+1:

runtimes/pythonrt/runner/tee.py
```diff
@@ -0,0 +1,51 @@
...
+        req = pb.handler.PrintRequest(
+            runner_id=self.runner_id,
+            message=data,
+        )
+
+        try:
+            self.worker.Print(req)
+        except grpc.RpcError as err:
+            if err.code() in (grpc.StatusCode.UNAVAILABLE, grpc.StatusCode.CANCELLED):
+                print("gRPC connection unavailable, print failed", file=self.stream)
```
tebeka, line 49: Thinking if we should log this as well.

tebeka (changes requested):

itayd:
didn't we stop using the stderr and stdout because they were non-deterministic?

runtimes/pythonrt/runner/tee.py
```diff
@@ -0,0 +1,51 @@
...
+from threading import Thread
+from time import sleep
+import grpc
+import pb
+
+
+class Tee:
+    def __init__(self, stream, runner_id, worker):
+        self.stream = stream
+        self.queue = Queue()
```
daabr, line 11: I'm not an expert with this code, so this is more of a question that a comment:

Any queue is by definition a bottleneck, but (hopefully) a necessary one.

If the `Queue` here is destroyed half-way while a workflow is running because of a crash, what will happen to the prints in it? Will they be lost forever, or will we deterministically recreate them in the new `Queue` based on the workflow history from Temporal?

runtimes/pythonrt/runner/tee.py
daabr: Also unit tests? This is a general-purpose class that doesn't have any AK complexities that make testing harder.

itayd:
@tebeka this is fixed in https://github.com/autokitteh/autokitteh/pull/1180, right? if so - @mario99logic please close this one.

PR #1178, opened 2025-03-13
feat: add exponential backoff to `Subscribe` method in Salesforce gRPC

pashafateev:
Testing: verified that events work as expected. Tested exponential backoff by attempting to init a grpc connection with an expired token. Observed expected behavior.

Refs: INT-333

### DIFF
--- a/integrations/salesforce/grpc.go
+++ b/integrations/salesforce/grpc.go
@@ -26,11 +26,28 @@ const retryPolicy = `{
 	"methodConfig": [{
 		"name": [{"service": "eventbus.v1.PubSub", "method": "Subscribe"}],
 		"retryPolicy": {
-			"MaxAttempts": 4,
-			"InitialBackoff": "1s",
-			"MaxBackoff": "10s",
-			"BackoffMultiplier": 2.0,
-			"RetryableStatusCodes": [ ]
+			"maxAttempts": 4,
+			"initialBackoff": "1s",
+			"maxBackoff": "10s",
+			"backoffMultiplier": 2.0,
+			"retryableStatusCodes": [
+				"CANCELLED", 
+				"UNKNOWN", 
+				"INVALID_ARGUMENT",
+				"DEADLINE_EXCEEDED",
+				"NOT_FOUND",
+				"ALREADY_EXISTS",
+				"PERMISSION_DENIED",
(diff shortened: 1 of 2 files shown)


integrations/salesforce/grpc.go
```diff
@@ -19,6 +19,19 @@ type grpcAuth struct {
 	tenantID    string
 }
 
+const retryPolicy = `{
+	"methodConfig": [{
+		"name": [{"service": "eventbus.v1.PubSub", "method": "Subscribe"}],
+		"retryPolicy": {
+			"MaxAttempts": 4,
+			"InitialBackoff": "1s",
+			"MaxBackoff": "128s",
```
daabr: Based on `maxAttempts`, `initialBackoff` and `backoffMultiplier`, the max backoff should be 10s.

integrations/salesforce/grpc.go
```diff
@@ -19,6 +19,19 @@ type grpcAuth struct {
 	tenantID    string
 }
 
+const retryPolicy = `{
```
daabr, line 25:
```suggestion
// Based on:
// - https://grpc.io/docs/guides/retry/
// - https://github.com/grpc/grpc-proto/blob/master/grpc/service_config/service_config.proto
const retryPolicy = `{
```

integrations/salesforce/grpc.go
```diff
@@ -19,6 +19,19 @@ type grpcAuth struct {
...
+	"methodConfig": [{
+		"name": [{"service": "eventbus.v1.PubSub", "method": "Subscribe"}],
+		"retryPolicy": {
+			"MaxAttempts": 4,
```
daabr: Field names inside `retryPolicy`: the Go sample in GitHub does use `FooBar` casing, and I guess it works, but I think it should be `fooBar`.

daabr (approved):
Don't forget to remove the 2 TODO comments in `subscription.go`

integrations/salesforce/grpc.go
pashafateev: But we don't define these so wouldn't changing the casing break it?

integrations/salesforce/grpc.go
pashafateev: Wouldn't it be 8?

PR #1179, opened 2025-03-13
chore(python): Session test for import error

tebeka:


itayd:
@tebeka the test fails even though it returns the error correctly. i think the reason is that you don't *print* the error, you only return it. can you add a print?

tebeka:
> @tebeka the test fails even though it returns the error correctly. i think the reason is that you don't _print_ the error, you only return it. can you add a print?

OK

tebeka:
> > @tebeka the test fails even though it returns the error correctly. i think the reason is that you don't _print_ the error, you only return it. can you add a print?


@itayd: That took longer than expected. Please review.

itayd (approved):

PR #1187, opened 2025-03-16
fix: Fix build errors when git is not available

mario99logic:
This PR modifies the Makefile to gracefully handle environments where git is not installed. Instead of failing with errors during the build process, we now provide fallback values for version information when git commands cannot be executed.

Refs: INT-325

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -23,7 +23,7 @@ endif
 ARCH=$(shell uname -m)
 
 ifeq ($(COMMIT),)
-COMMIT=$(shell git rev-parse HEAD)$(shell git diff --quiet || echo '_dirty')
+COMMIT=$(shell git rev-parse HEAD 2>/dev/null || echo "unknown")$(shell git diff --quiet 2>/dev/null || : && echo '')
 endif
 
 ifeq ($(TIMESTAMP),)


Makefile
```diff
@@ -23,7 +23,7 @@ endif
 ARCH=$(shell uname -m)
 
 ifeq ($(COMMIT),)
-COMMIT=$(shell git rev-parse HEAD)$(shell git diff --quiet || echo '_dirty')
+COMMIT=$(shell git rev-parse HEAD 2>/dev/null || echo "unknown")$(shell git diff --quiet 2>/dev/null || : && echo '')
```
tebeka, line 26: - Use `true` instead of `:`
- Why did you replace `_dirty` with ''?
- I find this line cryptic, I prefer you'll use an `if`, maybe something like

```bash
if [ -x "$(command -v git)" ]; then
    COMMIT=$(shell git rev-parse HEAD)$(shell git diff --quiet || echo '_dirty')
elif
    COMMIT=unknown
fi
```

tebeka:
Why is this PR required? When are we building without git?

efiShtain:
+ @tebeka 

how would you even get the repo without git ?
why do we need this change ?

Makefile
daabr: I agree that it's too cryptic, but @tebeka's suggestion is in shell scripting syntax, but this is a Makefile

daabr (commented):
@tebeka / @efiShtain - the context for this PR is INT-352.

Suggestions of alternative approaches are welcome, because I also prefer not to over-generalize our Makefile. We *should* assume that git is installed when we run git.

@mario99logic - my idea (but not sure that it's good) is to add another `ifeq` above line 25 which checks if git is runnable, and if not set `COMMIT` to `"unknown"` - which will cause the condition in line 25 to be false.

mario99logic:
@efiShtain and I discussed this solution and agreed that this solution is not ideal and only hide the issue so I'm closing the PR

PR #1188, opened 2025-03-17
fix: don't run crons in test mode

efiShtain:


tebeka (approved):
Approving, can you say what prompted this change?

efiShtain:
I'm having problems building locally and thought this might be related

it's not
but it can't hurt to recode tests time

PR #1189, opened 2025-03-17
feat: switch to store jwt in cookie

efiShtain:
- This change makes the UI cookie store a standard jwt, unifying the "experience" of jwt and cookie

- it will also allow us simpler / easier usage in the gateway before the server

- it introduce an expiration time to the cookie so users are logged out automatically if it is not renewed (default for 14 days)

- auto renew cookies whenever the UI calls who-am-i endpoint, in case it will call it once a day, it will work fine without any probelms

### DIFF
--- a/internal/backend/auth/authsessions/config.go
+++ b/internal/backend/auth/authsessions/config.go
@@ -7,20 +7,23 @@ import (
 )
 
 type Config struct {
-	SameSite   http.SameSite
-	CookieKeys string `koanf:"cookie_keys"` // pairs of hash and block keys.
-	Domain     string `koanf:"ui_domain"`
-	Secure     bool
+	SameSite          http.SameSite
+	CookieKeys        string `koanf:"cookie_keys"` // pairs of hash and block keys.
+	Domain            string `koanf:"ui_domain"`
+	Secure            bool
+	ExpirationMinutes int `koanf:"expiration_minutes"`
 }
 
 var Configs = configset.Set[Config]{
 	Default: &Config{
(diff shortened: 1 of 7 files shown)


internal/backend/auth/authsessions/config.go
```diff
@@ -7,20 +7,23 @@ import (
...
+		Secure:            true,
+		ExpirationMinutes: 60 * 24 * 14, // 14 days
 	},
 	Dev: &Config{
-		Secure:     false,
-		SameSite:   http.SameSiteLaxMode,
-		CookieKeys: "0000000000000000000000000000000000000000000000000000000000000000,0000000000000000000000000000000000000000000000000000000000000000",
+		Secure:            false,
+		SameSite:          http.SameSiteLaxMode,
+		CookieKeys:        "0000000000000000000000000000000000000000000000000000000000000000,0000000000000000000000000000000000000000000000000000000000000000",
```
itayd, line 26: remove

internal/backend/auth/authsessions/store.go
```diff
@@ -1,158 +1,115 @@
...
 
-func (s store) newSessionWithData(data *sessionData) (*sessions.Session[[]byte], error) {
-	session := s.store.New(sessionName)
-
-	bs, err := json.Marshal(data)
-	if err != nil {
-		return nil, fmt.Errorf("failed to marshal user data: %w", err)
-	}
-
-	session.Set(sessionDataKeyName, bs)
```
itayd, line 92: check if we can get rid of sessionstore

itayd (approved):

PR #1192, opened 2025-03-18
feat: Reuse "accessTokenExpiration" from SFDC in "FreshOAuthToken"

mario99logic:
- Moved `AccessTokenExpiration` to a new package to be able to use it in both `common` and `salesforce` package.
- Call `AccessTokenExpiration` in `FreshOAuthToken`

### DIFF
--- a/integrations/common/vars.go
+++ b/integrations/common/vars.go
@@ -113,7 +113,7 @@ func FreshOAuthToken(ctx context.Context, l *zap.Logger, o sdkservices.OAuth, v 
 	if i.UniqueName().String() == "salesforce" && t2.Expiry.IsZero() {
 		err := salesforceauth.AccessTokenExpiration(ctx, cfg.Endpoint.TokenURL, t2, vsid, v)
 		if err != nil {
-			l.Warn("failed to get access token expiration for salesforce", zap.Error(err))
+			l.Error("failed to get access token expiration for salesforce", zap.Error(err))
 			t2.Expiry = time.Now().UTC().Add(2 * time.Hour)
 		}
 	}
--- a/integrations/common/vars.go
+++ b/integrations/common/vars.go
@@ -8,6 +8,7 @@ import (
 	"go.uber.org/zap"
 	"golang.org/x/oauth2"
 
+	"go.autokitteh.dev/autokitteh/integrations/salesforce/salesforceauth"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
@@ -105,15 +106,18 @@ func FreshOAuthToken(ctx context.Context, l *zap.Logger, o sdkservices.OAuth, v 
 	if err != nil {
 		return t1
 	}
+	vsid := vs.Get(OAuthAccessTokenVar).ScopeID()
 
 	// Special case: Salesforce access tokens are time-limited and yet
 	// they don't have an expiry timestamp - so we add it on our own.
 	if i.UniqueName().String() == "salesforce" && t2.Expiry.IsZero() {
-		// TODO(INT-322): Reuse "accessTokenExpiration" in SFDC's OAuth handler.
-		t2.Expiry = time.Now().UTC().Add(2 * time.Hour)
+		err := salesforceauth.AccessTokenExpiration(ctx, cfg.Endpoint.TokenURL, t2, vsid, v)
+		if err != nil {
+			l.Warn("failed to get access token expiration for salesforce", zap.Error(err))
+			t2.Expiry = time.Now().UTC().Add(2 * time.Hour)
+		}
 	}
 
-	vsid := vs.Get(OAuthAccessTokenVar).ScopeID()
 	l.Debug("refreshed OAuth token",
 		zap.String("integration", intg),
 		zap.String("connection_id", vsid.String()),
--- a/integrations/common/vars.go
+++ b/integrations/common/vars.go
@@ -113,7 +113,7 @@ func FreshOAuthToken(ctx context.Context, l *zap.Logger, o sdkservices.OAuth, v 
 	if i.UniqueName().String() == "salesforce" && t2.Expiry.IsZero() {
 		err := salesforceauth.AccessTokenExpiration(ctx, cfg.Endpoint.TokenURL, t2, vsid, v)
 		if err != nil {
-			l.Error("failed to get access token expiration for salesforce", zap.Error(err))
+			l.Warn("failed to get access token expiration for salesforce", zap.Error(err))
 			t2.Expiry = time.Now().UTC().Add(2 * time.Hour)
 		}
 	}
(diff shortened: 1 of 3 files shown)


integrations/common/vars.go
```diff
@@ -8,6 +8,7 @@ import (
 	"go.uber.org/zap"
 	"golang.org/x/oauth2"
 
+	"go.autokitteh.dev/autokitteh/integrations/salesforce/salesforceauth"
```
tebeka, line 11: This seems wrong. `common` should not import `salesforceauth`.

integrations/common/vars.go
```diff
@@ -105,15 +106,18 @@ func FreshOAuthToken(ctx context.Context, l *zap.Logger, o sdkservices.OAuth, v
...
+	vsid := vs.Get(OAuthAccessTokenVar).ScopeID()
 
 	// Special case: Salesforce access tokens are time-limited and yet
 	// they don't have an expiry timestamp - so we add it on our own.
 	if i.UniqueName().String() == "salesforce" && t2.Expiry.IsZero() {
-		// TODO(INT-322): Reuse "accessTokenExpiration" in SFDC's OAuth handler.
-		t2.Expiry = time.Now().UTC().Add(2 * time.Hour)
+		err := salesforceauth.AccessTokenExpiration(ctx, cfg.Endpoint.TokenURL, t2, vsid, v)
+		if err != nil {
+			l.Error("failed to get access token expiration for salesforce", zap.Error(err))
```
tebeka:
```suggestion
			l.Warn("failed to get access token expiration for salesforce", zap.Error(err))
```

integrations/common/vars.go
```diff
@@ -105,15 +106,18 @@ func FreshOAuthToken(ctx context.Context, l *zap.Logger, o sdkservices.OAuth, v
...
+			t2.Expiry = time.Now().UTC().Add(2 * time.Hour)
```
tebeka, line 117: Do we want to make this configurable?

integrations/salesforce/salesforceauth/salesforceauth.go
```diff
@@ -0,0 +1,66 @@
...
+	vs, err := vars.Get(ctx, vsid)
+	if err != nil {
+		return err
+	}
+
+	formData := url.Values{
+		"token":           {t.AccessToken},
+		"token_type_hint": {"access_token"},
+		"client_id":       {vs.GetValueByString("private_client_id")},
+		"client_secret":   {vs.GetValueByString("private_client_secret")},
```
tebeka, line 31: Check that `private_client_id` and `private_client_secret` are not empty? (unless they are optional in the request).

integrations/salesforce/salesforceauth/salesforceauth.go
```diff
@@ -0,0 +1,66 @@
...
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, u, strings.NewReader(formData.Encode()))
+	if err != nil {
+		return err
+	}
+
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	client := &http.Client{}
+	resp, err := client.Do(req)
```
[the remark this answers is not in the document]
tebeka: What Copilot said :)

integrations/salesforce/salesforceauth/salesforceauth.go
```diff
@@ -0,0 +1,66 @@
...
+	if err := json.NewDecoder(resp.Body).Decode(&tokenInfo); err != nil {
+		return errors.New("failed to parse token info")
+	}
+
+	// Extract the expiration timestamp.
+	expFloat, ok := tokenInfo["exp"].(float64)
+	if !ok {
+		return errors.New("missing or invalid expiration time in response")
+	}
+	t.Expiry = time.Unix(int64(expFloat), 0)
```
tebeka, line 63:
```suggestion
	var tokenInfo struct {
	    Exp float64
	}
	if err := json.NewDecoder(resp.Body).Decode(&tokenInfo); err != nil {
		return errors.New("failed to parse token info")
	}

	// Extract the expiration timestamp.
	if tokenInfo.Exp == 0 {
		return errors.New("missing or invalid expiration time in response")
	}
	t.Expiry = time.Unix(int64(tokenInfo.Exp), 0)
```

mario99logic:
The changes I made were from before the update to `accessTokenExpiration`, which is why there are conflicts. But my main question is: @daabr, we talked about moving the function to a new package so it could be used in the `common` package. Now, with the new changes, the function itself depends on the `common` package, so I can’t put it there anymore.
I could extract the parts that are used from the `common` package and create a separate package for them, but I think that would just overcomplicate things. What do you think?

integrations/salesforce/salesforceauth/salesforceauth.go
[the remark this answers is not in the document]
mario99logic: I did not implement this function, but could someone provide an explanation of what is incorrect in this section?

integrations/common/vars.go
mario99logic: Then how can we use the function `accessTokenExpiration` in the common package, don't we need to import it?

integrations/salesforce/salesforceauth/salesforceauth.go
mario99logic: there is a new implementation for this function in the new main

integrations/salesforce/salesforceauth/salesforceauth.go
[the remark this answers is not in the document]
pashafateev:
```
// Set the access token expiration time.
	if err := salesforceauth.AccessTokenExpiration(ctx, instanceURL, data.Token, vsid, h.vars); err != nil {
		l.Error("failed to get access token expiration", zap.Error(err))
		c.AbortServerError("failed to get access token expiration")
		return
```
If you click "show paths" and follow it towards the end you'll see the code above. `instanceURL` is a value provided by the user and we're using it as part of an http request. This poses a security risk, but we don't have any other choice. We're trusting Salesforce and in the worst case scenario, if Salesforce gets hacked, we don't reveal any sensitive information here.

It's annoying that GitHub is constantly bringing this up – I've already addressed this alert about 5 times.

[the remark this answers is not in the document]
pashafateev: i.e. you can dismiss alert and copy and paste my explanation.

integrations/common/vars.go
pashafateev: I agree with @tebeka. This function was originally intended to be local and is tightly coupled to Salesforce. If we want to reuse it _in_ `common`, I think it would be better to first generalize it, then import it _from_ `common` into SFDC.

pashafateev:
I'm confused about what the point of the PR is. Is it to be able to reuse `accessTokenExpiration` for other integrations?

mario99logic:
@pashafateev It addresses INT-322 , but its now old and based on an old branch I have to change it.

PR #1193, opened 2025-03-18
feat: Add linear connTest function 

mario99logic:
ref: INT-269

### DIFF
--- a/integrations/linear/checks.go
+++ b/integrations/linear/checks.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 
+	"go.uber.org/zap"
+
 	"go.autokitteh.dev/autokitteh/integrations"
 	"go.autokitteh.dev/autokitteh/integrations/common"
 	"go.autokitteh.dev/autokitteh/sdk/sdkintegrations"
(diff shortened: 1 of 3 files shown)


mario99logic:
I have tested the API key connection; however, I was unable to test the Linear OAuth connection as I do not have a Linear OAuth app. Could someone verify the OAuth connection?

integrations/linear/checks.go
```diff
@@ -37,26 +39,37 @@ func status(v sdkservices.Vars) sdkintegrations.OptFn {
 
 // test checks whether the connection is actually usable, i.e. the configured
 // authentication credentials are valid and can be used to make API calls.
-func test(v sdkservices.Vars) sdkintegrations.OptFn {
+func test(v sdkservices.Vars, o sdkservices.OAuth) sdkintegrations.OptFn {
```
tebeka, line 42: The name `test` is not accurate. Maybe `checkValid` or `isUsable`?

integrations/linear/checks.go
```diff
@@ -37,26 +39,37 @@ func status(v sdkservices.Vars) sdkintegrations.OptFn {
...
 	return sdkintegrations.WithConnectionTest(func(ctx context.Context, cid sdktypes.ConnectionID) (sdktypes.Status, error) {
 		vs, errStatus, err := common.ReadVarsWithStatus(ctx, v, cid)
 		if errStatus.IsValid() || err != nil {
 			return errStatus, err
 		}
 
+		body := `{ "query": "{ viewer { id  } }" }`
```
tebeka, line 49: It's risky to construct JSON by hand, but this is simple enough. If the query grows complex we'll need to use `json.Marshal`

integrations/linear/checks.go
```diff
@@ -37,26 +39,37 @@ func status(v sdkservices.Vars) sdkintegrations.OptFn {
...
 		case integrations.APIKey:
-			// TODO(INT-269): Implement.
+			apiKey := vs.GetValue(apiKeyVar)
+			_, err := common.HTTPPostJSON(ctx, linearAPIURL, apiKey, body)
+			if err != nil {
+				return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), err
+			}
+
 		default:
 			return sdktypes.NewStatus(sdktypes.StatusCodeError, "Bad auth type"), nil
```
tebeka, line 69: Add the auth type in the error

integrations/linear/vars.go
```diff
@@ -10,6 +10,8 @@ var (
 	orgIDVar  = sdktypes.NewSymbol("org_id")
 )
 
+var linearAPIURL = "https://api.linear.app/graphql"
```
tebeka: const?

tebeka (approved):

integrations/linear/checks.go
mario99logic: I agree the name `test` is relatively new, the original name for the function was `connTest` .
@daabr do we change the name to something more specific?

integrations/linear/checks.go
mario99logic: The query is only for verifying the connection, and I don’t expect it to become more complex.

integrations/linear/vars.go
mario99logic: changed it to const.

integrations/linear/checks.go
mario99logic: if we decide to change the name we will do this in another PR while changing the names of the other test functions

PR #1194, opened 2025-03-18
feat: refactor http requests with basic auth

mario99logic:
ref: INT-337


tebeka (changes requested):
I don't like this design. I prefer that we'll use [functional options](https://dave.cheney.net/2014/10/17/functional-options-for-friendly-apis):

So instead of 
```go
common.HTTPPostJSONWithAuth(ctx, url, creds, payload)
```

```go
common.HTTPRequest(ctx, url, WithMethod(http.MethodPost), WithCreds(user, passwd), WithJSON(payload))
```

The current design creates a lot of functions that encode that argument in the function name.

PR #1195, opened 2025-03-18
feat(python): Use dill for serialization

tebeka:
This solves issues with https://github.com/autokitteh/kittehub/pull/217 serialization of closures.

Ref: ENG-586


tebeka:
@itayd found another solution for https://github.com/autokitteh/kittehub/pull/217, closing.

PR #1196, opened 2025-03-19
chore(python): Add performance metrics

tebeka:
Ref: ENG-1985

Just a start, we'll probably add more later.


runtimes/pythonrt/pythonrt.go
```diff
@@ -169,9 +178,47 @@ func newSvc(cfg *Config, l *zap.Logger) (sdkservices.Runtime, error) {
...
+		return fmt.Errorf("create telemetry: %w", err)
+	}
+
+	mRun, err := tele.NewHistogram("python.run.duration", "Duration of Python runtime Run (ms)")
+	if err != nil {
+		return fmt.Errorf("create histogram: %w", err)
+	}
+	py.metrics.runDuration = mRun
+
+	mErr, err := tele.NewCounter("python.run.error", "Python Run error count")
```
efiShtain: this can probably set set in the session level with runtime dimension set to python so we can reuse it for node later

runtimes/pythonrt/pythonrt.go
```diff
@@ -169,9 +178,47 @@ func newSvc(cfg *Config, l *zap.Logger) (sdkservices.Runtime, error) {
...
+	tele, err := telemetry.New(py.log, &telemetry.Config{})
+	if err != nil {
+		return fmt.Errorf("create telemetry: %w", err)
+	}
+
+	mRun, err := tele.NewHistogram("python.run.duration", "Duration of Python runtime Run (ms)")
+	if err != nil {
+		return fmt.Errorf("create histogram: %w", err)
+	}
+	py.metrics.runDuration = mRun
```
efiShtain, line 208: I think we already track the session duration, do we need this specific metric as well ?

look at ```internal/backend/sessions/sessionworkflows/workflows.go```
```go
	sessionDurationHistogram.Record(metricsCtx, duration.Milliseconds(),
		metric.WithAttributes(attribute.Bool("replay", isReplaying), attribute.Bool("success", err == nil)))
```

runtimes/pythonrt/pythonrt.go
```diff
@@ -169,9 +178,47 @@ func newSvc(cfg *Config, l *zap.Logger) (sdkservices.Runtime, error) {
...
+
+	mErr, err := tele.NewCounter("python.run.error", "Python Run error count")
+	if err != nil {
+		return fmt.Errorf("create counter: %w", err)
+	}
+	py.metrics.runErrors = mErr
+
+	mStart, err := tele.NewHistogram("python.start.duration", "Duration of Python Start (ms)")
```
efiShtain: I think this is an important metric

can we rename it to runner.start_time with dimension python  so later on we can reuse it for node and have the possibility to filter ?

runtimes/pythonrt/pythonrt.go
```diff
@@ -169,9 +178,47 @@ func newSvc(cfg *Config, l *zap.Logger) (sdkservices.Runtime, error) {
...
+	if err != nil {
+		return fmt.Errorf("create histogram: %w", err)
+	}
+	py.metrics.startDuration = mStart
+
+	mErr, err = tele.NewCounter("python.start.error", "Python Start error count")
```
efiShtain: same as others, make it generic with dimension

runtimes/pythonrt/pythonrt.go
```diff
@@ -169,9 +178,47 @@ func newSvc(cfg *Config, l *zap.Logger) (sdkservices.Runtime, error) {
 		},
 	}
 
+	if err := svc.setupMetrics(); err != nil {
```
efiShtain, line 190: i think this should be at the global runtime level, not tied to specific pysvc so we don't initialize everything on each run

runtimes/pythonrt/pythonrt.go
tebeka: This is different than workflow failing. It's failing to start the runtime.

runtimes/pythonrt/pythonrt.go
tebeka: This is not workflow duration it's `Run` which sets up the python environment for running (start runner service).

efiShtain: so run is not a good name, maybe runner_setup or something like this

runtimes/pythonrt/pythonrt.go
efiShtain: so naming is important here, you're trying to add metrics on the setup runner part

tebeka:
Will move to `ak` in the `Run` method and others.

PR #1197, opened 2025-03-19
refactor(oauth): better design to support private OAuth apps seamlessly

daabr:
This PR has the primary goal of making it simple to support private OAuth apps in all integrations, by adding consideration for their customizations seamlessly when reading an OAuth config, instead of local patches in multiple disconnected places in the code.

The secondary goal (without which I couldn’t do the above): convert the `oauth` gRPC service into a non-service package: it was highly-coupled, violated the locality principle, and over-complicated.

Empirically, the new `oauth` package in this PR does much more then the old service, and includes new unit tests - but new size is only 50% of the old code!

Also added unit tests wherever it was reasonably possible.

On the way:

- Move `FreshOAuthToken` from the `integrations/common` package to the new `integrations/oauth` package
- Generalize an important GitHub-only optimization: handle gracefully OAuth flows starting outside the AutoKitteh server (i.e. they lack a `state` parameter with the connection ID)
- Get rid of many TODO comments related to the now-obsolete ENG-112
- Better starting point for private OAuth in Atlassian integrations (ENG-965) and GitHub (INT-194)
- Add TODO to support GHES in private GitHub apps (INT-349)

Testing:
- OAuth connection inits
- Running Kittehub projects: Slack sample, GitHub issue alert, Google Sheets sample

Refs: INT-320, ENG-112


internal/backend/cron/cron.go
```diff
@@ -74,18 +75,14 @@ type Cron struct {
...
-	oauth       sdkservices.OAuth
+	oauth       *oauth.OAuth
 }
 
 func New(c *Config, l *zap.Logger, t temporalclient.Client) *Cron {
 	return &Cron{cfg: c, logger: l, temporal: t}
 }
 
-func (cr *Cron) Start(ctx context.Context, c sdkservices.Connections, v sdkservices.Vars, o sdkservices.OAuth) error {
-	if !cr.cfg.Enabled {
```
daabr, line 85: Make sure this is intentional?

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -425,8 +425,10 @@ func (s *workerGRPCHandler) RefreshOAuthToken(ctx context.Context, req *userCode
 	}
 
 	// Get the integration's OAuth configuration.
-	// TODO(INT-46): pass vars service instead of nil
-	cfg, _, err := oauth.New(runner.log, nil).Get(ctx, req.Integration)
+	// TODO(INT-46): pass vars service instead of nil.
+	// TODO(INT-46): pass new OAuth service instead of constructing with nil params.
```
daabr: Fix comment, also cf. INT-322

integrations/oauth/token.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	t2, err := cfg.TokenSource(ctx, t1).Token()
+	if err != nil {
+		return t1
+	}
+
+	// Special case: Salesforce access tokens are time-limited and yet
+	// they don't have an expiry timestamp - so we add it on our own.
+	if integ == "salesforce" && t2.Expiry.IsZero() {
```
itayd: move to config

daabr: done

integrations/common/vars_test.go
```diff
@@ -35,6 +37,9 @@ func (s *fakeVarsService) Set(ctx context.Context, vs ...sdktypes.Var) error {
 
 func (s *fakeVarsService) Get(ctx context.Context, sid sdktypes.VarScopeID, names ...sdktypes.Symbol) (sdktypes.Vars, error) {
 	vs := sdktypes.NewVars()
+	if len(names) == 0 {
+		names = slices.Collect(maps.Keys(s.data[sid]))
+	}
```
mario99logic, line 42: what does this change do?

integrations/oauth/token.go
```diff
@@ -0,0 +1,64 @@
...
+	t2, err := cfg.TokenSource(ctx, t1).Token()
+	if err != nil {
+		return t1
+	}
+
+	// Special case: Salesforce access tokens are time-limited and yet
+	// they don't have an expiry timestamp - so we add it on our own.
+	if o.flags(integ).expiryMissingInToken && t2.Expiry.IsZero() {
+		// TODO(INT-322): Reuse "accessTokenExpiration" in SFDC's OAuth handler.
+		t2.Expiry = time.Now().UTC().Add(2 * time.Hour)
```
mario99logic, line 52: now that this function is not in the common package, I can do INT-322

integrations/oauth/fx.go
```diff
@@ -0,0 +1,89 @@
...
+
+func New(c *Config, l *zap.Logger, v sdkservices.Vars) *OAuth {
+	return &OAuth{cfg: c, logger: l, vars: v}
+}
+
+func (o *OAuth) Start(m *muxes.Muxes) error {
+	// Set the AutoKitteh server's base URL for webhooks.
+	o.BaseURL = o.cfg.Address
+	if o.BaseURL == "" {
+		o.BaseURL = os.Getenv("WEBHOOK_ADDRESS") // Legacy
```
mario99logic, line 43: What do you mean by Legacy?

integrations/oauth/fx.go
```diff
@@ -0,0 +1,89 @@
...
+	oauthConfigs map[string]oauthConfig
+}
+
+func New(c *Config, l *zap.Logger, v sdkservices.Vars) *OAuth {
+	return &OAuth{cfg: c, logger: l, vars: v}
+}
+
+func (o *OAuth) Start(m *muxes.Muxes) error {
+	// Set the AutoKitteh server's base URL for webhooks.
+	o.BaseURL = o.cfg.Address
```
mario99logic, line 41: before using `o.cfg.Address`, where do we initialize it?

integrations/common/vars_test.go
pashafateev: when no specific vars are specified, the vars service queries for all vars. This change introduces this behavior in the fake vars service.

integrations/salesforce/integration.go
```diff
@@ -60,27 +60,23 @@ func (h handler) reopenExistingPubSubConnections(ctx context.Context) {
...
-			cfg, _, err := h.oauth.Get(ctx, desc.UniqueName().String())
-			if err != nil {
-				h.logger.Error("failed to get Salesforce OAuth config", zap.Error(err))
-				continue
-			}
-			clientID = cfg.ClientID
+		cfg, _, err := h.oauth.GetConfig(ctx, desc.UniqueName().String(), cid)
+		if err != nil {
+			h.logger.Error("failed to get Salesforce OAuth config", zap.Error(err))
+			continue
```
pashafateev, line 74: Because the config is now privatized or default based on `cid`?

integrations/salesforce/oauth.go
```diff
@@ -118,36 +117,22 @@ func (h handler) saveConnection(ctx context.Context, vsid sdktypes.VarScopeID, t
 // Salesforce access tokens are time-limited, but they don't have an expiry timestamp
 // so we need to add it on our own. This function also returns the connection's client ID.
 // Based on: https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oidc_token_introspection_endpoint.htm
-func (h handler) accessTokenExpiration(ctx context.Context, instanceURL string, t *oauth2.Token, vsid sdktypes.VarScopeID) (string, error) {
-	var clientID, clientSecret string
-
-	vs, err := h.vars.Get(ctx, vsid, common.AuthTypeVar, clientIDVar, clientSecretVar)
+func (h handler) accessTokenExpiration(ctx context.Context, instanceURL string, t *oauth2.Token, cid sdktypes.ConnectionID) (string, error) {
+	u, err := url.JoinPath(instanceURL, "/services/oauth2/introspect")
```
pashafateev, line 122: what's the motivation behind the change from `vsid` to `cid`? Because `GetConfig` needs the `cid`?

internal/backend/svc/integrations.go
```diff
@@ -112,7 +113,7 @@ func integrationsFXOption() fx.Option {
 		integration("slack", configset.Empty, slack.New),
 		integration("twilio", configset.Empty, twilio.New),
 		integration("zoom", configset.Empty, zoom.New),
-		fx.Invoke(func(lc fx.Lifecycle, l *zap.Logger, muxes *muxes.Muxes, vars sdkservices.Vars, dispatch sdkservices.DispatchFunc, oauth sdkservices.OAuth) {
+		fx.Invoke(func(lc fx.Lifecycle, l *zap.Logger, muxes *muxes.Muxes, vars sdkservices.Vars, oauth *oauth.OAuth, dispatch sdkservices.DispatchFunc) {
```
pashafateev, line 116: This is question about injection. We went from injecting a service to injecting a package. 

1. Why inject a package, why not jump import it where needed?
2. As far as fx is concerned, it doesn't matter whether it's a service or a package?

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -425,8 +425,9 @@ func (s *workerGRPCHandler) RefreshOAuthToken(ctx context.Context, req *userCode
 	}
 
 	// Get the integration's OAuth configuration.
-	// TODO(INT-46): pass vars service instead of nil
-	cfg, _, err := oauth.New(runner.log, nil).Get(ctx, req.Integration)
+	// TODO(INT-46): pass new OAuth service instead of constructing with nil params.
+	o := oauth.New(nil, runner.log, nil)
+	cfg, _, err := o.GetConfig(ctx, req.Integration, sdktypes.InvalidConnectionID)
```
pashafateev, line 432: I think a comment to explain why we're passing `InvalidConnectionID` would be helpful.

internal/backend/svc/svc.go
```diff
@@ -218,6 +218,16 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 		Component("builds", configset.Empty, fx.Provide(builds.New)),
 		Component("connections", configset.Empty, fx.Provide(connections.New)),
 		Component("deployments", deployments.Configs, fx.Provide(deployments.New)),
+		Component(
+			"webhooks", // Koanf config prefix, reused by all integrations.
```
pashafateev, line 222: I'm not sure about this name. why not call it `oauth`?

integrations/oauth/configs.go
```diff
@@ -0,0 +1,675 @@
...
+		privatizeGitHub(vs, &cfg)
+	case "microsoft", "microsoft_teams":
+		privatizeMicrosoft(vs, &cfg)
+	}
+
+	return cfg.Config, cfg.Opts, nil
+}
+
+// deepCopy ensures that we don't modify the server's default OAuth 2.0 configurations.
+func deepCopy(c oauthConfig) oauthConfig {
```
pashafateev: nit: I would use `cfg` instead of `c` because everywhere I've seen a config we've called it that.

integrations/oauth/configs.go
```diff
@@ -0,0 +1,675 @@
...
+func (o *OAuth) initRedirectURLs() {
+	for k, v := range o.oauthConfigs {
+		if v.Config == nil {
+			continue
+		}
+
+		// Special case: Microsoft integrations all use "microsoft".
+		if strings.HasPrefix(k, "microsoft") {
+			k = "microsoft"
+		}
```
pashafateev: I think it's confusing that we have

```	// Optional integration-specific customizations.
	o.initAtlassianURLs()
	o.initGitHubURLs()
```

But Microsoft is handled inside the generic function.

integrations/oauth/configs.go
```diff
@@ -0,0 +1,675 @@
...
+	// Failed to read connection variables - return the default configuration.
+	vs, err := o.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
+	if err != nil {
+		o.logger.Error("failed to read connection variables",
+			zap.String("integration", integration),
+			zap.String("connection_id", cid.String()),
+			zap.Error(err),
+		)
+		return cfg.Config, cfg.Opts, nil
+	}
```
pashafateev, line 599: What if I am trying to authenticate an app that has both private and default OAuth modes? If there's an error, it will go to default and authenticate through the wrong app?

integrations/oauth/configs.go
```diff
@@ -0,0 +1,675 @@
...
+// fixAuth0 needs to run even when the connection is using the AutoKitteh server's
+// default OAuth app, not just a private one like the [privatize] function.
+func fixAuth0(vs sdktypes.Vars, c *oauthConfig) {
+	domain := vs.GetValue(auth0.DomainVar)
+	c.Config.Endpoint.AuthURL = fmt.Sprintf("https://%s/oauth/authorize", domain)
+	c.Config.Endpoint.DeviceAuthURL = fmt.Sprintf("https://%s/oauth/device/code", domain)
+	c.Config.Endpoint.TokenURL = fmt.Sprintf("https://%s/oauth/token", domain)
+
+	c.Opts["audience"] = fmt.Sprintf("https://%s/api/v2/", domain)
+}
```
pashafateev: To me the name is confusing. Other options: `configureAuth0`, `setupAuth0`

integrations/oauth/configs.go
```diff
@@ -0,0 +1,675 @@
...
+// TODO(INT-349): Support GHES in private OAuth mode too.
+func privatizeGitHub(vs sdktypes.Vars, c *oauthConfig) {
+	defaultAppName := os.Getenv("GITHUB_APP_NAME")
+	if defaultAppName == "" {
+		defaultAppName = defaultGitHubAppName
+	}
+	defaultAppName = fmt.Sprintf("/%s/", defaultAppName)
+
+	authURL := c.Config.Endpoint.AuthURL
+	// TODO: Make GitHub var symbols non-internal, and reuse here.
```
pashafateev: ticket?

integrations/oauth/configs_test.go
```diff
@@ -0,0 +1,261 @@
...
+			require.NoError(t, err)
+
+			assert.Equal(t, tt.wantID, cfg.ClientID)
+			assert.Equal(t, tt.wantSecret, cfg.ClientSecret)
+			assert.Equal(t, "https://example.com/oauth/redirect/slack", cfg.RedirectURL)
+		})
+	}
+}
+
+func TestFixAuth0(t *testing.T) {
```
pashafateev: change the name here if you change the name based on the other comment

integrations/oauth/configs_test.go
```diff
@@ -0,0 +1,261 @@
...
+
+	assert.Equal(t, "id", cfg.ClientID)
+	assert.Equal(t, "secret", cfg.ClientSecret)
+	assert.Equal(t, "https://example.com/oauth/redirect/auth0", cfg.RedirectURL)
+
+	assert.Contains(t, cfg.Endpoint.AuthURL, "https://domain/")
+	assert.Contains(t, cfg.Endpoint.DeviceAuthURL, "https://domain/")
+	assert.Contains(t, cfg.Endpoint.TokenURL, "https://domain/")
+	assert.Contains(t, opts, "audience")
+	assert.Contains(t, opts["audience"], "https://domain/")
```
pashafateev, line 233: I don't really get these checks for `https://domain/`

integrations/oauth/fx.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+	if strings.HasPrefix(addr, "http://") {
+		addr = strings.Replace(addr, "http://", "https://", 1)
+	}
+	if !strings.HasPrefix(addr, "https://") {
+		addr = "https://" + addr
+	}
+
+	// Parse and normalize the URL.
+	u, err := url.Parse(addr)
```
pashafateev, line 77: this is what does most of the error checking right? Meaning checking whether or not it's a valid url.

integrations/oauth/token.go
```diff
@@ -0,0 +1,64 @@
...
+
+	// Access token is still fresh - return it as-is.
+	if t1.Valid() {
+		return t1
+	}
+
+	// Otherwise, use the OAuth refresh flow.
+	integ := i.UniqueName().String()
+	vsid := vs.Get(common.OAuthAccessTokenVar).ScopeID()
+	cid := vsid.ToConnectionID()
```
pashafateev, line 31: just curious about this. When creating a new scope id it doesn't have to be a cid, so why can we convert it to one?

integrations/oauth/token.go
```diff
@@ -0,0 +1,64 @@
...
+	}
+
+	t2, err := cfg.TokenSource(ctx, t1).Token()
+	if err != nil {
+		return t1
+	}
+
+	// Special case: Salesforce access tokens are time-limited and yet
+	// they don't have an expiry timestamp - so we add it on our own.
+	if o.flags(integ).expiryMissingInToken && t2.Expiry.IsZero() {
```
pashafateev, line 50: isn't `IsZero()` equivalent to `expiryMissingInToken`? Or is this for additional safety?

integrations/oauth/vars_test.go
```diff
@@ -0,0 +1,58 @@
+package oauth
```
pashafateev, line 1: q: this is a file that is used for tests but doesn't actually have any tests in it right?

integrations/oauth/fx.go
daabr: Using AK config (`config.yaml` or the env var `AK_WEBHOOKS__ADDRESS`) is the new way, so the old way (`WEBHOOK_ADDRESS` env var) is a legacy way, which we will deprecate in the future.

integrations/salesforce/oauth.go
daabr: Yes, exactly. Technically I could extract `cid` from `vsid`, but it feels backwards to convert `vsid` before calling this function only to convert it back inside the function. Even though both approaches are the same, because we use both `cid` and `vsid`.

internal/backend/svc/integrations.go
daabr: 1. Because we're not injecting a package, we're injecting an **initialized instance** of it - that initialization depends on other components (component config from outside the process, vars services, etc.)
2. Yes, but again, service or "instance of a struct", not a package, because the package (the code itself) can't be referenced or replaced

runtimes/pythonrt/worker_grpc_handler.go
daabr: This relates to a previous discussion we had. I will add a comment here, but it's important to note that *here* it really is an invalid connection ID, because we *should* pass a real one.

internal/backend/svc/svc.go
daabr: Because it's (going to be) reused by all integrations, as the comment here says, not just OAuth. This name affects the namespace of the config struct's fields in `config.yaml` / `.env`.

integrations/oauth/configs.go
daabr: done, also in other functions below that one

integrations/oauth/configs.go
daabr: The functions `googleConfig`, `microsoftConfig`, and `offlineOpts` are static - they're just used instead of copy-pasting the same lines.

OTOH,  `initRedirectURLs, `initAtlassianURLs`, and `initGitHubURLs` are dynamic, they contain logic.

But it's a good point, I'll document this.

integrations/oauth/configs.go
daabr: Good point! One (partial) answer is that this failure shouldn't actually ever happen - if DB reads fail, the entire server will have an outage.

Another (partial) answer is that the user **is expected** to review an app installation: if the details (e.g. name) are incorrect, or the user doesn't like the requested permissions, they shouldn't approve it.

I'd prefer to add an error to this function's return value - Go doesn't like implicit errors based on output anyway - but that would be a major change across all integrations, so I preferred not to do it in this PR at least.

integrations/oauth/configs.go
daabr: changed to `setupAuth0`

integrations/oauth/configs.go
daabr: done

integrations/oauth/configs_test.go
daabr: done

integrations/oauth/configs_test.go
daabr: They check that the function under test does indeed modify `cfg.Endpoint`

integrations/oauth/fx.go
daabr: yes, but also an easy way to remove invalid/irrelevant affixes below.

integrations/oauth/token.go
daabr: VSIDs can also be project IDs - that's the meaning of vars being associated with either connections or projects - but in the context of OAuth, a VSID is **always** a connection ID.

integrations/oauth/token.go
daabr: No. `expiryMissingInToken` is a Boolean flag about handling a special case (where the expiry is missing even though it should be present), and `IsZero` is an actual check of the expiry.

You could argue that that first is enough, but I don't want to be in charge of Salesforce behavior, so if it gets fixed in the future it'll be seamless for us.

OTOH, the second condition is never enough on its own, because some integrations legitimately have non-expiring tokens, e.g. Slack.

integrations/oauth/vars_test.go
daabr: Correct

integrations/oauth/fx.go
daabr: AutoKitteh is in change of initializing its configs, based on the file `config.yaml` and other environment variables.

FX takes care of that.

pashafateev (approved):

PR #1199, opened 2025-03-20
fix(salesforce): reinitialize grpc client after errors

pashafateev:
testing: successfully received events after original token expiration

### DIFF
--- a/integrations/salesforce/subscription.go
+++ b/integrations/salesforce/subscription.go
@@ -94,6 +94,7 @@ func (h handler) eventLoop(ctx context.Context, clientID string, subscribeTopic 
 		for _, event := range msg.Events {
 			schema, err := client.GetSchema(ctx, &pb.SchemaRequest{SchemaId: event.Event.SchemaId})
 			if err != nil {
+				l.Error("failed to get Salesforce event schema", zap.String("schema_id", event.Event.SchemaId), zap.Error(err))
 				if gstatus.Code(err) != codes.Unauthenticated {
 					continue
 				}
@@ -216,7 +217,6 @@ func (h handler) initStream(ctx context.Context, l *zap.Logger, client pb.PubSub
 			cleanupClient(l, clientID)
 			h.initPubSubClient(cid, clientID, l, "failed to create gRPC stream for Salesforce events")
 			// TODO(INT-352): error handling
-			l.Error("failed to create gRPC stream for Salesforce events", zap.Error(err))
 			continue
 		}
 
--- a/integrations/salesforce/subscription.go
+++ b/integrations/salesforce/subscription.go
@@ -36,29 +36,10 @@ func (h handler) subscribe(clientID, orgID, instanceURL string, cid sdktypes.Con
 		return
 	}
 
-	ctx := authcontext.SetAuthnSystemUser(context.Background())
-	l := h.logger.With(zap.String("connection_id", cid.String()))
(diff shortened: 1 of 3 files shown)


integrations/salesforce/subscription.go
```diff
@@ -147,10 +151,71 @@ func (h handler) eventLoop(ctx context.Context, clientID string, subscribeTopic
...
+}
+
+// cleanupClient removes a client from the pubSubClients map.
+// This allows the underlying connection to be garbage collected.
+func cleanupClient(l *zap.Logger, clientID string) {
+	mu.Lock()
+	defer mu.Unlock()
+
+	if _, ok := pubSubClients[clientID]; ok {
+		delete(pubSubClients, clientID)
```
daabr, line 208: can you also close the client when you remove it?

integrations/salesforce/subscription.go
```diff
@@ -147,10 +151,71 @@ func (h handler) eventLoop(ctx context.Context, clientID string, subscribeTopic
...
+}
+
+func (h handler) initStream(ctx context.Context, l *zap.Logger, client pb.PubSubClient, cid sdktypes.ConnectionID, clientID string) pb.PubSub_SubscribeClient {
 	for {
 		stream, err := client.Subscribe(ctx)
 		if err != nil {
+			cleanupClient(l, clientID)
+			h.initPubSubClient(cid, clientID, l, "failed to create gRPC stream for Salesforce events")
+			// TODO(INT-352): error handling
 			l.Error("failed to create gRPC stream for Salesforce events", zap.Error(err))
```
daabr: Should you remove this log message here? Isn't it done inside `initPubSub`?

internal/backend/oauth/oauth.go
```diff
@@ -489,7 +489,8 @@ func New(l *zap.Logger, vars sdkservices.Vars) sdkservices.OAuth {
 				},
 				RedirectURL: redirectURL + "salesforce",
 				// https://help.salesforce.com/s/articleView?id=xcloud.remoteaccess_oauth_tokens_scopes.htm
-				Scopes: []string{"api", "refresh_token"},
+				//TODO: decrease scopes if possible
```
daabr: Go comments without a space after `//` are compiler directives. Add a space.

integrations/salesforce/subscription.go
```diff
@@ -106,9 +94,25 @@ func (h handler) eventLoop(ctx context.Context, clientID string, subscribeTopic
 		for _, event := range msg.Events {
 			schema, err := client.GetSchema(ctx, &pb.SchemaRequest{SchemaId: event.Event.SchemaId})
 			if err != nil {
-				l.Error("failed to get Salesforce event schema", zap.String("schema_id", event.Event.SchemaId), zap.Error(err))
-				continue
+				if gstatus.Code(err) != codes.Unauthenticated {
+					continue
```
daabr, line 99: Why remove the logging of this error if you `continue`?

integrations/salesforce/subscription.go
pashafateev: There is no way to close the client. Can only close the connection so either we need to also store the connection, or just store the connection but that might require some rearranging.

integrations/salesforce/subscription.go

daabr (approved):

integrations/salesforce/subscription.go
pashafateev: Done

PR #1202, opened 2025-03-23
fix(python): Ignore files/directories starting with .

tebeka:
Fixes: ENG-2051


cmd/ak/common/build.go
```diff
@@ -101,6 +101,10 @@ func walk(basePath string, uploads map[string][]byte) fs.WalkDirFunc {
 			return nil // Skip directory analysis, focus on files.
 		}
 
+		if path[0] == '.' {
+			return nil // Skip hidden files.
```
efiShtain, line 105: id might use strings.StartsWith instead of directly checking [0] due to runes / vs bytes storage
but i guess it should be fine for . char specifically

efiShtain (approved):

cmd/ak/common/build.go
tebeka: I'll keep the short version :)

PR #1205, opened 2025-03-25
fix: add thread_ts from app_mention event in slack

mario99logic:
As part of a client project, the requirement is to send all messages from a Slack thread to an AWS knowledge base whenever their app is mentioned within that thread.

To do this, we need the thread_ts to fetch the full thread. I noticed it wasn't included in the event payload, so I added it to ensure we can retrieve all messages correctly.


daabr (approved):
Nit: attach `ThreadTS` to be block above it, instead of the block below it.

PR #1206, opened 2025-03-25
chore: improve Jira event logs for parsing errors

pashafateev:


integrations/atlassian/jira/event.go
```diff
@@ -103,7 +103,11 @@ func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
 	for _, id := range ids {
 		u, err := transformIssueURL(jiraEvent, l)
 		if err != nil {
-			l.Error("Failed to transform Jira URL", zap.Error(err))
+			l.Error("Failed to transform Jira URL",
+				zap.Error(err),
+				zap.Any("webhookEvent", jiraEvent["webhookEvent"]),
+				zap.Any("issue", jiraEvent["issue"]),
```
daabr: Why not just `jiraEvent`? Also, while this is sufficient for short-term debugging, note the related Slack thread: these event payloads should be in a time-limited DB table.

daabr (approved):

integrations/atlassian/jira/event.go
pashafateev: That's what I originally did but then thought it would be too much

PR #1213, opened 2025-03-28
reproduce issue: pure python exception that is not picklable makes the runtime stuck

itayd:
not sure why, need to be investigated by @tebeka and @efiShtain 

based on https://github.com/autokitteh/autokitteh/pull/1212


cmd/ak/cmd/sessions/test.go
```diff
@@ -219,3 +245,16 @@ func init() {
...
+	expected = strings.TrimSpace(expected)
+	actual = strings.TrimSpace(actual)
+
+	if expected == actual {
+		return nil
+	}
+
+	edits := myers.ComputeEdits(span.URIFromPath("want"), expected, actual)
+	diff := gotextdiff.ToUnified("want", "got", expected, edits)
+	return errors.New(fmt.Sprint("%s:\n", what, diff))
```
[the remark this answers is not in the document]
itayd: fixed

cmd/ak/cmd/sessions/test.go
```diff
@@ -53,6 +53,28 @@ func normalizePath(p string) string {
...
+		buf.WriteRune('\n')
+	}
+
+	if err := s.Err(); err != nil {
+		return "", fmt.Errorf("normalize: %w", err)
+	}
+
+	return strings.TrimSpace(buf.String()), nil
+}
+
```
efiShtain, line 77: is this in use?

tests/sessions/python/pure_uncaught_custom_exception.txtar
```diff
@@ -0,0 +1,38 @@
...
+
+
+def main(_):
+    callstuff()
+
+-- error.txt --
+error: X(1, 2)
+
+Traceback (most recent call last):
+  File "/tmp/ak-runner-210142359/runner/main.py", line 442, in _call
```
efiShtain, line 22: this test is so tightly coupled that adding a space would break it
we can't have line number in expected output result

runtimes/pythonrt/runner/main.py
```diff
@@ -136,6 +135,19 @@ def is_pickleable(err):
...
+        # This is unexpected, but we can't not handle it.
+       # Logging so we can investigate.
+        log.exception("unexpected error: %r", pickle_err)
+        tb = "".join(format_exception(pickle_err))
+        log.error("traceback:\n%r", tb)
+        log.error('error we tried to pickle: %r', err)
+        try:
+            attrs = vars(err)
+            log.error('exception attributes: %r', attrs)
+        except TypeError:
```
efiShtain: if this will throw not TypeError we would get back to the same issue
and this time is for logging purposes of a different error
we have to except any exception everywhere it might throw

cmd/ak/cmd/sessions/test.go
tebeka: Yes. 3 places in the command `RunE`, see below.

runtimes/pythonrt/runner/deterministic.py
```diff
@@ -26,6 +26,18 @@ def is_deterministic(fn):
...
+    if (cls := getattr(fn, "__objclass__", None)):
+        # We don't check isinstance on built-in types since the user can subclass
+        # and then override a method. For example:
+        # class List(list):
+        #     def append(self, item):
+        #         ...
+        # And this append should run as an activity.
+        # We don't add Exception to builtin_types since then we'll need to add all
+        # built-in exceptions.
+        if cls in builtin_types or issubclass(cls, Exception):
```
tebeka, line 38: Not 100% sure about `issubclass`, I've opened ENG-2072 to think about it but we can commit this code.

runtimes/pythonrt/runner/main.py
tebeka: By `this` you mean `vars(err)` - it only raises `TypeError`. However the log might fail. Changed to Exception

tests/sessions/python/pure_uncaught_custom_exception.txtar
tebeka: `normalizeOutput` above takes care of this.

tebeka (approved):

runtimes/pythonrt/runner/deterministic.py
itayd: i can live with that

tests/sessions/python/import_error.txtar
```diff
@@ -1,4 +1,4 @@
-rpc error: code = InvalidArgument desc = can't load main from - error: No module named 'garfield_loves_pizza'
+rpc error: code = InvalidArgument desc = can't load main from - error: ModuleNotFoundError("No module named 'garfield_loves_pizza'")
```
itayd, line 1: we need to show where this is happening exactly without the ak-runner stuff

tests/sessions/python/pure_uncaught_unpicklable_exception.txtar
```diff
@@ -0,0 +1,34 @@
...
+@activity
+def callstuff():
+    raise X(1, 2)
+
+
+def main(_):
+    callstuff()
+
+-- error.txt --
+error: X(1, None)
```
itayd, line 19: this is because we don't override `__str__` right?

itayd: i can see this works ok in the previous test so ok

PR #1214, opened 2025-03-28
fix(systest): remove erronous exclusive

itayd:
this disabled most of the system test until now :(


daabr (commented):
It's not clear to me what this PR is/should be doing. Either way, it's looks like it's not ready for review, so I'm changing it's state to draft.

.github/workflows/go.yml
```diff
@@ -61,7 +61,7 @@ jobs:
         run: go build -trimpath -o bin/ak ./cmd/ak
 
       - name: Test
-        run: AK_SYSTEST_USE_PROC_SVC=1 gotestsum -f github-actions -- -timeout 0 -trimpath ./tests/system/...
+        run: AK_SYSTEST_USE_PROC_SVC=1 gotestsum -f github-actions -- -timeout 9m -trimpath ./tests/system/...
```
itayd: can you explain? 9m is a very specific number :)

tests/system/parse.go
```diff
@@ -30,7 +30,7 @@ var (
 	reCheck = regexp.MustCompile(`^capture_re\s+(\w+)\s+(.+)`)
 
 	// output <equals|equals_json|contains|regex> [file] *
-	akCheckOutput = regexp.MustCompile(`^output\s+(equals|equals_json|contains|regex|equals_jq)\s+(file\s+)?(.+|'.*')`)
+	akCheckOutput = regexp.MustCompile(`^output\s+(equals|equals_json|contains_json|contains|regex|equals_jq)\s+(file\s+)?(.+|'.*')`)
```
itayd: did i do this? what does `contains_json` means?

itayd: seems like it's not used anywhere?

tests/system/testdata/end2end/pyhttp.txtar
```diff
@@ -32,6 +32,7 @@ def on_http(event):
     "state": {
         "completed": {
             "prints": [
+                "ᓚᘏᗢ Python runs might take a while to start when running for the first time on a new runner, hang on!",
```
itayd: does this always take more than 10s? i think it's safer to filter this out somehow or set some env var in autokitteh to not show this message in testing mode.

tests/system/testdata/sessions/stop.txtar
```diff
@@ -79,11 +83,14 @@ def test():
...
 {"state":{"running":{"run_id":"run_0000000000000000000000000a"}}}
 {"state":{"running":{"run_id":"run_0000000000000000000000000a","call":{"function":{"executor_id":"run_0000000000000000000000000a","name":"test","desc":{},"data":"dGVzdCMx"}}}}}
-{"print":{"text":"started","value":{"string":{"v":"started"}}}}
 {"call_spec":{"function":{"function":{"executor_id":"int_3kth00testaf9d33c5697341f0","name":"freeze","desc":{"input":[{"name":"duration","optional":true},{"name":"allow_cancel","optional":true}]},"data":"Y29uXzAwMDAwMDAwMDAwMDAwMDAwMDAwMDAwMDA2"}},"args":[{"integer":{"v":"10"}},{"boolean":{"v":true}}],"seq":1}}
 {"stop_request":{}}
 {"state":{"stopped":{"reason":"test"}}}
+
+-- test-prints.txt --
+started
+
```
itayd: i think this empty line can be removed

.github/workflows/go.yml
mario99logic: 9m is just a random number, I'll change it to 10 minutes to make it clearer. The main point is that setting a timeout here helps us identify which test is stuck, unlike GitHub Actions' timeout, which doesn't show that.

tests/system/parse.go
mario99logic: I added this when I tried to implement a function to check if it contains the whole json, but I used `contains` instead, which is better.
I will remove it.

tests/system/testdata/end2end/pyhttp.txtar
mario99logic: I will use `equals_jq` for the fields, like I did for other tests.

.github/workflows/go.yml
itayd: got it. 9m is fine, was just curious.

itayd:
@mario99logic i cannot approve my own pr, so if you think it's ready - approve it. thanks.

mario99logic (approved):

PR #1215, opened 2025-03-29
feat: allow to start a session in another project

itayd:
due to customer demand.

adds another parameter to `start` - `project=`, where another project in the same org can be specified to start sessions in. will start a session on the latest deployment.


efiShtain:
can't the user just call the other project's webhook trigger instead of adding this ?

itayd:
> can't the user just call the other project's webhook trigger instead of adding this ?

This will expose the trigger to the outside world, which you don't always want to do. Also this returns a session id which you can wait on later.

cmd/ak/cmd/sessions/start.go
```diff
@@ -38,6 +40,20 @@ var startCmd = common.StandardCommand(&cobra.Command{
 			return err
 		}
 
+		if !bid.IsValid() && !did.IsValid() && pid.IsValid() {
```
tebeka, line 43: Does this mean they can't specify specific build/deployment on a given project?
IMO first get the project and then get builds & deployments for this specific project.

cmd/ak/cmd/sessions/start.go
```diff
@@ -60,7 +76,7 @@ func init() {
 	startCmd.Flags().StringVarP(&deploymentID, "deployment-id", "d", "", "deployment ID, mutually exclusive with --build-id and --env")
 	startCmd.Flags().StringVarP(&buildID, "build-id", "b", "", "build ID, mutually exclusive with --deployment-id")
 	startCmd.Flags().StringVarP(&project, "project", "o", "", "project name or ID, mutually exclusive with --deployment-id")
-	startCmd.MarkFlagsOneRequired("deployment-id", "build-id")
+	startCmd.MarkFlagsOneRequired("deployment-id", "project")
```
tebeka, line 79: They don't have to be mutually exclusive, only that the deployment-id is in the project. Same for build-in

internal/backend/sessions/sessionworkflows/callbacks_misc.go
```diff
@@ -6,32 +6,54 @@ import (
...
-		session := sdktypes.NewSession(w.data.Build.ID(), loc, inputs, memo).
-			WithParentSessionID(w.data.Session.ID()).
-			WithDeploymentID(w.data.Session.DeploymentID()).
-			WithProjectID(w.data.Session.ProjectID()).
-			WithNewID()
+		data := w.data
 
-		if err := workflow.ExecuteActivity(wctx, createSessionActivityName, session).Get(wctx, nil); err != nil {
-			return sdktypes.InvalidSessionID, err
+		if !project.IsValid() {
```
tebeka, line 22: I think `IsValid` is misleading name, should be `IsZero` (or `IsEmpty`)

runtimes/pythonrt/runner/syscalls.py
```diff
@@ -65,6 +71,7 @@ def ak_start(self, loc: str, data: dict = None, memo: dict = None) -> str:
             loc=loc,
             data=json_data.encode(),
             memo=json_memo.encode(),
+            project=project or "",
```
tebeka: Why isn't the define for project `""`? (instead of `None`). Then you don't need the `or`.

internal/backend/sessions/sessionworkflows/callbacks_misc.go
itayd: not sure, `reflect` is using `IsValid` so that's why i used it. in any case if we change this it's a huge search and replace - not for this pr.

runtimes/pythonrt/runner/syscalls.py
itayd: not sure what you mean here?

internal/backend/sessions/sessionworkflows/callbacks_misc.go
tebeka: OK

runtimes/pythonrt/runner/syscalls.py
tebeka:
```python
def start(
    loc: str,
    data: dict | None = None,
    memo: dict | None = None,
    project: str = "",   # <-- THIS
) -> str:
```

tebeka (approved):
I'd love to see the comments resolved, but OK with merging.

runtimes/pythonrt/runner/syscalls.py
itayd: fixed

cmd/ak/cmd/sessions/start.go
itayd: if you specify deployment id, it also has build id and project id in. thus it is mutual exclusive to the others.

cmd/ak/cmd/sessions/start.go
itayd: if a project is specified without a build id or a deployment id (which points to a build id), then we'll get the latest deployment which will give this info.

PR #1216, opened 2025-03-31
fix: add git to docker

mario99logic:
ref: INT-325

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -37,7 +37,7 @@ EOF
 # runtime dependencies for the application.
 FROM python:3.11-slim AS pydeps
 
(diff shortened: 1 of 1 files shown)


tebeka (approved):

Dockerfile
```diff
@@ -37,7 +37,7 @@ EOF
 # runtime dependencies for the application.
 FROM python:3.11-slim AS pydeps
 
-RUN apt-get update && rm -rf /var/lib/apt/lists/*
+RUN apt-get update && apt-get install -y git && rm -rf /var/lib/apt/lists/*
```
itayd, line 40: shouldn't this be in the golang build phase?

PR #1217, opened 2025-03-31
feat: convert jwt to rsa and expose jwks endpoint

efiShtain:
solves [ENG-2065] and [ENG-2066]

Merging this requires a change to the infrastructure code as well
[https://github.com/autokitteh/infrastructure/pull/12]

The change has no affect to dev/test mode

In standard mode, need to select the jwt signing algorithm
```bash
 # Can be rsa or hmac
AK_AUTHJWTTOKENS__ALGORITHM = rsa 

# If hmac is selected, a sign key needs to be provided
AK_AUTHJWTTOKENS__HMAC__SIGN_KEY = ""

# If rsa is selected, a private and public keys are required
AK_AUTHJWTTOKENS__RSA__PRIVATE_KEY = ""
AK_AUTHJWTTOKENS__RSA__PUBLIC_KEY = ""
```

This change also exposes a jwks endpoint over ```http://<api_domain>/.well-known/jwks.json```
The endpoint returns standard jwks if algorithm is rsa, otherwise it returns an error

### DIFF
--- a/internal/backend/auth/authtokens/authjwttokens/jwt.go
+++ b/internal/backend/auth/authtokens/authjwttokens/jwt.go
@@ -9,8 +9,6 @@ import (
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
-const issuer = "ak"
-
 type Algorithm string
 
 const (
--- a/internal/backend/auth/authtokens/authjwttokens/jwt.go
+++ b/internal/backend/auth/authtokens/authjwttokens/jwt.go
@@ -1,38 +1,31 @@
 package authjwttokens
 
 import (
-	"encoding/hex"
-	"encoding/json"
 	"errors"
-	"fmt"
 	"strings"
-	"time"
-
-	j "github.com/golang-jwt/jwt/v5"
-	"github.com/google/uuid"
 
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authtokens"
-	"go.autokitteh.dev/autokitteh/internal/backend/auth/authusers"
 	"go.autokitteh.dev/autokitteh/internal/backend/configset"
-	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
 const issuer = "ak"
 
-var (
-	method   = j.SigningMethodHS256
-	hashSize = method.Hash.Size()
+type Algorithm string
+
+const (
+	AlgorithmHMAC Algorithm = "hmac"
+	AlgorithmRSA  Algorithm = "rsa"
 )
 
+// Config represents the top-level JWT configuration
 type Config struct {
-	SignKey string `koanf:"sign_key"`
+	Algorithm Algorithm  `koanf:"algorithm"`
+	HMAC      HMACConfig `koanf:"hmac"`
+	RSA       RSAConfig  `koanf:"rsa"`
 }
 
-type tokens struct {
-	signKey []byte
-}
-
+// token is shared between HMAC and RSA implementations
 type token struct {
 	User sdktypes.User
 }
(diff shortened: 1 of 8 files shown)


internal/backend/auth/authtokens/authjwttokens/jwt.go
```diff
@@ -1,107 +1,57 @@
...
+// token is shared between HMAC and RSA implementations
 type token struct {
 	User sdktypes.User
 }
 
 var Configs = configset.Set[Config]{
 	Default: &Config{},
 	Dev: &Config{
-		SignKey: strings.Repeat("00", hashSize),
+		Algorithm: AlgorithmHMAC,
```
daabr, line 34: Technically speaking, this is "insecure by default" which is a bad practice, but it's good enough as a starting point.

Please add a ticket to replace this with a one-time generation of a random key.

daabr (approved):
Approved with one optional comment

internal/backend/auth/authtokens/authjwttokens/jwt.go
itayd: why is this insecure?

internal/backend/auth/authtokens/authjwttokens/jwtrsa.go
```diff
@@ -0,0 +1,178 @@
...
+type rsaTokens struct {
+	privateKey *rsa.PrivateKey
+	publicKey  *rsa.PublicKey
+}
+
+// Using same token structure as HMAC version
+
+var RSAConfigs = configset.Set[RSAConfig]{
+	Default: &RSAConfig{},
+	// For dev/test, you'll need to generate and provide RSA key pairs
```
itayd: Isn't this completely optional?

internal/backend/auth/authtokens/authjwttokens/jwtrsa.go
```diff
@@ -0,0 +1,178 @@
...
+
+type rsaTokens struct {
+	privateKey *rsa.PrivateKey
+	publicKey  *rsa.PublicKey
+}
+
+// Using same token structure as HMAC version
+
+var RSAConfigs = configset.Set[RSAConfig]{
+	Default: &RSAConfig{},
```
itayd: If you drop this, it'll always want you to specify them, which is what we want?

internal/backend/auth/authtokens/authjwttokens/jwt.go
efiShtain: I don't think we should do anything for dev mode, it works fine.
If you'll do random key, each restart of the service would break the previous one and user won't be able to talk to the backend. 

if someone wants to replace it, he can do that

in standard mode it is required and will fail if key is not provided

internal/backend/auth/authtokens/authjwttokens/jwtrsa.go
efiShtain: it is ignored if algorithm is hmac. for dev default is hmac, so I didn't provide any defaults here. it is required if algo is rsa. requirement is enforced here

does that answer your question ? 
I won't merge before you ack as well

```go
func newRSA(cfg *RSAConfig) (RSATokens, error) {
	if cfg.PrivateKey == "" || cfg.PublicKey == "" {
		return nil, errors.New("both private and public keys must be provided")
	}

	privateKey, err := parsePrivateKey(cfg.PrivateKey)
	if err != nil {
		return nil, fmt.Errorf("invalid private key: %w", err)
	}

	publicKey, err := parsePublicKey(cfg.PublicKey)
	if err != nil {
		return nil, fmt.Errorf("invalid public key: %w", err)
	}

	return &rsaTokens{
		privateKey: privateKey,
		publicKey:  publicKey,
	}, nil
}
```

internal/backend/auth/authtokens/authjwttokens/jwtrsa.go
efiShtain: it is actually un used, removed it

PR #1219, opened 2025-04-01
fix(salesforce): ignore self-triggered events based on `user_id` instead of `changeOrigin`

pashafateev:
testing: successfully received and filtered events

### DIFF
--- a/integrations/salesforce/subscription.go
+++ b/integrations/salesforce/subscription.go
@@ -56,6 +56,17 @@ func (h handler) eventLoop(ctx context.Context, clientID string, subscribeTopic 
 	client := pubSubClients[clientID]
 	stream := h.initStream(ctx, l, client, cid, clientID)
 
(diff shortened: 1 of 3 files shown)


integrations/salesforce/subscription.go
```diff
@@ -127,20 +127,26 @@ func (h handler) eventLoop(ctx context.Context, clientID string, subscribeTopic
...
 			if !ok {
-				l.Error("changeOrigin key is not present in Salesforce ChangeEventHeader")
+				l.Error("commitUser is not a string in ChangeEventHeader")
 				continue
 			}
-			changeOrigin, ok := changeOriginValue.(string)
-			if !ok {
-				l.Error("changeOrigin is not a string in Salesforce ChangeEventHeader")
+
+			vs, err := h.vars.Get(ctx, sdktypes.NewVarScopeID(cid))
```
daabr: This needs to happen only once, so do it before line 56, and also get only the user ID var, not all of the vars.

daabr (approved):

PR #1220, opened 2025-04-01
fix(slack): don't panic if interactive event doesn't contain enterprise ID

efiShtain:
unmarshalling payload has *Enterprise and *Team which can be nil
trying to reach Enterprise.ID or Team.ID throws a panic

I"m not sure this is the correct solution - @pashafateev  / @daabr. Feel free to discard the change.

**This does happen in production** 

```go
// Interaction payloads.
	var p BlockActionsPayload
	if err := json.Unmarshal([]byte(payload), &p); err != nil {
		l.Warn("failed to parse interaction payload for app/team IDs",
			zap.String("payload", payload),
			zap.Error(err),
		)
		return "", "", "", err
	}

	eID := ""
	if p.Enterprise != nil {
		eID = p.Enterprise.ID
	}

	tID := ""
	if p.Team != nil {
		tID = p.Team.ID
	}
	return p.APIAppID, eID, tID, nil
```

### DIFF
--- a/integrations/slack/webhooks/webhook.go
+++ b/integrations/slack/webhooks/webhook.go
@@ -232,8 +232,16 @@ func (h handler) extractIDs(l *zap.Logger, body []byte, wantContentType string) 
 		)
 		return "", "", "", err
 	}
+	eID := ""
+	if p.Enterprise != nil {
+		eID = p.Enterprise.ID
+	}
 
-	return p.APIAppID, p.Enterprise.ID, p.Team.ID, nil
+	tID := ""
+	if p.Team != nil {
+		tID = p.Team.ID
+	}
+	return p.APIAppID, eID, tID, nil
 }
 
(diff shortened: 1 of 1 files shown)


daabr (commented):
1. Var naming nit: `xID` should be `xid` ("blahID" is correct, but "bID" is not)
2. `Enterprise` is always empty (we don't have customers with enterprise Slack), and I never saw panics yet - so why/when is this change needed?
3. `Team` should never be empty - if it is, the received event can't be dispatched anyway, so why is this change needed?

itayd:
it panics in prod and local on line 236 in slack/webhook.go since the fields are null.

daabr (approved):

daabr (approved):
Team cannot be empty, and if it is, a panic is appropriate instead of returning an empty string.

But parsing a missing enterprise struct as `nil` is new behavior in `json.Unmarshal()` - keeping that fix.

PR #1224, opened 2025-04-01
fix(sessions): parent close policy to abandon

itayd:
this will prevent child workflows from being terminated when the parent is terminated. in the future, we might consider configuring this.

### DIFF
--- a/internal/backend/temporalclient/workflow.go
+++ b/internal/backend/temporalclient/workflow.go
@@ -38,7 +38,7 @@ func (wc WorkflowConfig) ToStartWorkflowOptions(qname, id, sum string, memo map[
 	}
 }
 
-func (wc WorkflowConfig) ToChildWorkflowOptions(qname, id, sum string, parentClosePolicy enumspb.ParentClosePolicy, memo map[string]string) workflow.ChildWorkflowOptions {
+func (wc WorkflowConfig) ToChildWorkflowOptions(qname, id, sum string, pcp enumspb.ParentClosePolicy, memo map[string]string) workflow.ChildWorkflowOptions {
 	wc = wc.With(defaultWorkflowConfig)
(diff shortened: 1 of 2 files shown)


internal/backend/temporalclient/workflow.go
```diff
@@ -37,12 +38,13 @@ func (wc WorkflowConfig) ToStartWorkflowOptions(qname, id, sum string, memo map[
 	}
 }
 
-func (wc WorkflowConfig) ToChildWorkflowOptions(qname, id, sum string, memo map[string]string) workflow.ChildWorkflowOptions {
+func (wc WorkflowConfig) ToChildWorkflowOptions(qname, id, sum string, parentClosePolicy enumspb.ParentClosePolicy, memo map[string]string) workflow.ChildWorkflowOptions {
```
daabr:
```suggestion
func (wc WorkflowConfig) ToChildWorkflowOptions(qname, id, sum string, pcp enumspb.ParentClosePolicy, memo map[string]string) workflow.ChildWorkflowOptions {
```

internal/backend/temporalclient/workflow.go
```diff
@@ -37,12 +38,13 @@ func (wc WorkflowConfig) ToStartWorkflowOptions(qname, id, sum string, memo map[
...
 	wc = wc.With(defaultWorkflowConfig)
 	return workflow.ChildWorkflowOptions{
 		WorkflowID:          id,
 		TaskQueue:           qname,
 		StaticSummary:       sum,
+		ParentClosePolicy:   parentClosePolicy,
```
daabr:
```suggestion
		ParentClosePolicy:   pcp,
```

daabr (approved):
Approved with 1 style comment

PR #1226, opened 2025-04-02
chore: upgrade webplatform

itayd:


efiShtain (approved):
just verify that the versions are compatible

PR #1228, opened 2025-04-02
feat(salesforce): track and close grpc connections

pashafateev:
Part of a multi phase approach to debug salesforce and close resources properly

Testing: received events – observed no behavior change.


daabr (approved):
Remove nested connection check

PR #1238, opened 2025-04-03
fix(build): only build directories with python

itayd:


runtimes/pythonrt/build.go
```diff
@@ -37,6 +37,13 @@ func isBuildFile(entry fs.DirEntry) bool {
 func (py *pySvc) Build(ctx context.Context, fsys fs.FS, path string, values []sdktypes.Symbol) (sdktypes.BuildArtifact, error) {
 	py.log.Info("build Python module", zap.String("path", path))
 
+	if path != "" {
```
tebeka, line 40: I don't see the value of sending a sub path. Is there a product requirement for that?

sdk/sdkruntimes/build.go
```diff
@@ -108,27 +109,51 @@ func Build(
...
-				data: &sdkbuildfile.RuntimeData{
-					Info: sdkbuildfile.RuntimeInfo{
-						Name: rt.Get().Name(),
-					},
+			data = &sdkbuildfile.RuntimeData{
+				Info: sdkbuildfile.RuntimeInfo{
+					Name: rt.Get().Name(),
 				},
 			}
+		} else {
```
tebeka, line 128: I think that instead of this logic, you should check for filewise around line 55 when you walk over the fs. If not - they send the fs to the runtime, otherwise do the loop. Probably in different functions.

The logic here is pretty complex.

tebeka (approved):
Approving, but would love to see my comments addressed.

runtimes/pythonrt/build.go
itayd: the param is there and i think it's better than ignoring it.

itayd: starlark is using it for what to build, so it's the same here now - just for directories.

sdk/sdkruntimes/build.go
itayd: You're right in principal - problem is that during the walk I when I encounter a dir I don't know if all the files in it are python or not.

PR #1241, opened 2025-04-05
feat(salesforce): add support for `replay_id`

pashafateev:
Without this PR, we will lose events if we're down for any reason or if there's an error processing an event.

refs: [replay docs](https://developer.salesforce.com/docs/platform/pub-sub-api/references/methods/subscribe-rpc.html?q=replay#replaying-an-event-stream)

testing: all previous functionality works as expected + I shut down server and received events on startup


integrations/salesforce/subscription.go
```diff
@@ -134,6 +136,16 @@ func (h handler) eventLoop(ctx context.Context, l *zap.Logger, clientID string,
 					continue
 				}
 
+				// Save the latest replay ID.
+				replayID := msg.LatestReplayId
+				encodedReplayID := base64.StdEncoding.EncodeToString(replayID)
```
tebeka: Why do we encode here and decode there? You can do `[]byte(msg.LatestReplayId)`

integrations/salesforce/vars.go
```diff
@@ -4,6 +4,7 @@ import "go.autokitteh.dev/autokitteh/sdk/sdktypes"
 
 var (
 	instanceURLVar = sdktypes.NewSymbol("instance_url")
+	replayIDVar    = sdktypes.NewSymbol("replay_id")
 	orgIDVar       = sdktypes.NewSymbol("organization_id")
```
daabr:
```suggestion
	orgIDVar       = sdktypes.NewSymbol("organization_id")
	replayIDVar    = sdktypes.NewSymbol("replay_id")
```

integrations/salesforce/subscription.go
daabr: @tebeka because connection vars (where we save this) can only be strings. I don't know what these IDs contain (i.e. if they can be naively converted to/from strings) so en/decoding seems reasonable.

daabr: I don't see `replayID` being used anywhere else in this function except in line 140, so this makes more sense to me:

```suggestion
				replayID := base64.StdEncoding.EncodeToString(msg.LatestReplayId)
```

integrations/salesforce/subscription.go
```diff
@@ -134,6 +136,16 @@ func (h handler) eventLoop(ctx context.Context, l *zap.Logger, clientID string,
...
+				l.Debug("saving Salesforce replay ID", zap.String("replay_id", encodedReplayID))
+
+				vsid := sdktypes.NewVarScopeID(cid)
+				if err := h.vars.Set(ctx, sdktypes.NewVar(replayIDVar).SetValue(encodedReplayID).WithScopeID(vsid)); err != nil {
```
daabr:
```suggestion
				v := sdktypes.NewVar(replayIDVar).SetValue(replayID)
				if err := h.vars.Set(ctx, v.WithScopeID(vsid)); err != nil {
```

integrations/salesforce/subscription.go
```diff
@@ -134,6 +136,16 @@ func (h handler) eventLoop(ctx context.Context, l *zap.Logger, clientID string,
 					continue
 				}
 
+				// Save the latest replay ID.
```
daabr: Let's discuss offline ASAP your question about the location of this block.

integrations/salesforce/subscription.go
```diff
@@ -223,19 +235,34 @@ func (h handler) initStream(ctx context.Context, l *zap.Logger, cid sdktypes.Con
...
+		decodedReplayID, err := base64.StdEncoding.DecodeString(replayID)
+		if err != nil {
+			return fmt.Errorf("failed to decode replay ID: %w", err)
+		}
+		fetchReq.ReplayId = decodedReplayID
+	}
 
-	err := stream.Send(fetchReq)
+	err = stream.Send(fetchReq)
 	if err != nil {
```
daabr:
```suggestion
	if err := stream.Send(fetchReq); err != nil {
```

daabr (approved):
Let's discuss offline your question about the location of this block, before merging this.

integrations/salesforce/subscription.go
pashafateev: @tebeka that wouldn't work because it's already a `[]byte`

```
	// Latest replay ID of a subscription. Enables clients with an updated replay value so that they can keep track
	// of their last consumed replay. Clients will not have to start a subscription at a very old replay in the case where a resubscribe is necessary.
	LatestReplayId []byte `protobuf:"bytes,2,opt,name=latest_replay_id,json=latestReplayId,proto3" json:"latest_replay_id,omitempty"`
	```

File: integrations/salesforce/subscription.go [hunk 6]
```diff
@@ -134,6 +136,16 @@ func (h handler) eventLoop(ctx context.Context, l *zap.Logger, clientID string,
 					continue
 				}
 
+				// Save the latest replay ID.
+				replayID := base64.StdEncoding.EncodeToString(msg.LatestReplayId)
+				l.Debug("saving Salesforce replay ID", zap.String("replay_id", replayID))
+
+				vsid := sdktypes.NewVarScopeID(cid)
```
itayd: wait - shouldn't we save the replayId after each event that we're processing?

pashafateev: easier to explain over zoom. 
Short answer is if we save for every event, we have to save before we parse the event. If there's an error parsing the event, we will lose it.
So the only events we're not saving are the keepalive messages.

itayd (approved):

PR #1245, opened 2025-04-09
feat(python): Switch to uv

tebeka:
Fixes: ENG-1984


itayd (commented):
how will migration be done? if we got someone running locally and suddenly it switches to uv - will that work? can we hide it for now behind a feature flag?

tebeka:
> how will migration be done? if we got someone running locally and suddenly it switches to uv - will that work? can we hide it for now behind a feature flag?

That's a good question. The simplest way is to tell people they need `uv` installed.
We can also install `uv` like we install the temporal dev server.

I don't want feature flag, we don't have any and setting them up and keep duplicate code is a pain.

See https://docs.google.com/document/d/163tlPso25zOaO8pyfaKT0oFmy81-O2qw5miVlr-FiP8/edit for a discussion.

efiShtain:
I don't think we should install anything for the user, especially not something sensitive like package managers

we should at most run a check at start time if uv is installed and if not instruct the user that it is required

efiShtain:
regarding migration, I think we can store in config something like venvUseUV with default to false, every start when we run ensureVenv we should do the following:
- if directory doesn't exists - install it
- if it exists but venvUseUV is false
    -  tell the user that we are migrating
    - remove it and install using UV
    - update config venvUseUV = true
- if it exists do nothing

tebeka:
> I don't think we should install anything for the user, especially not something sensitive like package managers
Why not? It's an internal tool for `ak` to use. We're not going to install it globally.

> we should at most run a check at start time if uv is installed and if not instruct the user that it is required
That's one option, we'll need to revisit it when NodeJS comes

itayd:
we'll make a special version only for this to test only this thing thoroughly before releasing

runtimes/pythonrt/local_runner_test.go
```diff
@@ -183,40 +183,42 @@ func Test_parsePyVersion(t *testing.T) {
...
-// 	}
-// 	ctx, cancel := testCtx(t)
-// 	defer cancel()
-
-// 	resp, err := client.Exports(ctx, &req)
-// 	require.NoError(t, err)
-
-// 	require.Equal(t, []string{"greet"}, resp.Exports)
-// }
+/*
```
itayd, line 188: ???

tebeka: Deleted commented out code.

itayd: it's still here?

itayd: @efiShtain can this be removed?

runtimes/pythonrt/dockerfilenodeps
```diff
@@ -1,21 +1,24 @@
...
 FROM gcr.io/distroless/python3-debian12
 USER 65532:65532
 
-COPY  --from=build /usr/local/lib/python3.11/site-packages /usr/lib/python3.11/site-packages
+COPY --from=build /usr/local/lib/python3.11/site-packages /usr/lib/python3.11/site-packages
+COPY --from=build /usr/local/bin/uv /usr/local/bin
 
-COPY  --chown=65532:65532  runner /runner
-COPY  --chown=65532:65532 workflow /workflow
+COPY --chown=65532:65532  runner /runner
```
itayd:
```suggestion
COPY --chown=65532:65532 runner /runner
```

runtimes/pythonrt/dockerfilewithdeps
```diff
@@ -1,25 +1,28 @@
...
+RUN uv pip install -r user_requirements.txt 
 
 # https://stackoverflow.com/questions/78599865/how-to-install-missing-python-modules-on-distroless-image
 FROM gcr.io/distroless/python3-debian12
 
 USER 65532:65532
 
-COPY  --from=build /usr/local/lib/python3.11/site-packages /usr/lib/python3.11/site-packages
+COPY --from=build /usr/local/lib/python3.11/site-packages /usr/lib/python3.11/site-packages
+COPY --from=build /usr/local/bin/uv /usr/local/bin
```
efiShtain: do we need uv here ?

runtimes/pythonrt/dockerfilenodeps
```diff
@@ -1,21 +1,24 @@
...
-RUN cd py-sdk && python -m pip install .[all]
+RUN cd py-sdk && uv pip install .[all]
 
 # https://stackoverflow.com/questions/78599865/how-to-install-missing-python-modules-on-distroless-image
 FROM gcr.io/distroless/python3-debian12
 USER 65532:65532
 
-COPY  --from=build /usr/local/lib/python3.11/site-packages /usr/lib/python3.11/site-packages
+COPY --from=build /usr/local/lib/python3.11/site-packages /usr/lib/python3.11/site-packages
+COPY --from=build /usr/local/bin/uv /usr/local/bin
```
efiShtain: do we need uv?

runtimes/pythonrt/local_runner.go
```diff
@@ -373,8 +373,21 @@ func overrideEnv(envMap map[string]string, runnerPath string) []string {
...
+	_, err := exec.LookPath("uv")
+	return err == nil
+}
+
+func createVEnv(log *zap.Logger, pyExe string, venvPath string) error {
+	var args []string
+	if hasUV() {
+		args = []string{"uv", "venv", "--python", pyExe, venvPath}
+	} else {
+		log.Warn("uv not found, using python venv with might be slower")
```
efiShtain:
```suggestion
		log.Warn("uv not found, using python venv which might be slower")
```

runtimes/pythonrt/local_runner_test.go
efiShtain: probably

runtimes/pythonrt/py-sdk/Makefile
```diff
@@ -14,8 +14,8 @@ upload: build
 		dist/autokitteh-*
 
 install-deps:
-	python -m pip install build twine
-	python -m pip install .[all,dev]
+	uv pip install build twine
```
efiShtain, line 17: need to ensure uv here as well?

runtimes/pythonrt/dockerfilewithdeps
tebeka: We gave uses way to install packages in the workflow (`autokitteh.packages`), it a hack but we need to support it. I did forget to update the py-sdk to try and use `uv` - will do.

runtimes/pythonrt/dockerfilenodeps
tebeka: See above :)

runtimes/pythonrt/py-sdk/Makefile
tebeka: I'm not a fan for check the environment in Makefile. This command will fail if there's no `uv` and `make` will fail, I'm OK with that.

runtimes/pythonrt/dockerfilewithdeps
efiShtain: this Dockerfile is for production use cases not open source
in this files we allow users installing stuff through requirements.txt, so no need to support installing extra stuff in other ways

itayd (approved):

PR #1244, opened 2025-04-09
feat: add tracing, refactor telemetry

itayd:

PR #1250, opened 2025-04-11
fix(pythonrt): reduce prints drain time to 10ms

itayd:
Per analysis from https://github.com/autokitteh/autokitteh/pull/1244, we got the `drainPrints` function hanging for a full second after each activity. There is no need to wait this long, if at all. Reduce grace time for print to 10ms, which will greatly speed up scripts run time.


itayd:
@tebeka do you think we could just remove the drainPrints function completely?

tebeka (approved):

PR #1252, opened 2025-04-14
bug(python): Cleaner test exit and restore print

tebeka:
Tests pass, but there are hanging threads the print error after the test is done due to the inactivity timer.

Fixes: ENG-2083


runtimes/pythonrt/runner/tests/main_test.py
```diff
@@ -25,22 +26,28 @@
...
 
 
+def new_test_runner(code_dir, worker=None, server=None):
+    runner = main.Runner(
+        id="runner1",
+        worker=worker,
+        code_dir=code_dir,
+        server=server,
+    )
+    runner._inactivty_timer.cancel()
```
itayd: is this the gist of this pr? not sure i understand - there is a lingering timer?

daabr: Same. But also FYI there's a typo in the timer name.

itayd (approved):

runtimes/pythonrt/runner/tests/main_test.py
tebeka: Fixed the typo, thanks!

PR #1253, opened 2025-04-14
feat(GitHub): support GHES in private OAuth

mario99logic:


runtimes/pythonrt/py-sdk/autokitteh/github.py
```diff
@@ -26,9 +26,11 @@ def github_client(connection: str, **kwargs) -> Github:
         ConnectionInitError: AutoKitteh connection was not initialized yet.
     """
     check_connection_name(connection)
-
     # Optional: GitHub Enterprise Server
-    base_url = os.getenv("GITHUB_ENTERPRISE_URL")
+    base_url = os.getenv(f"{connection}__enterprise_url")
+
+    if not base_url:
+        base_url = os.getenv("GITHUB_ENTERPRISE_URL")
```
daabr:
```suggestion
    base_url = os.getenv(f"{connection}__enterprise_url") or os.getenv("GITHUB_ENTERPRISE_URL")

```

runtimes/pythonrt/py-sdk/autokitteh/github.py
```diff
@@ -26,9 +26,11 @@ def github_client(connection: str, **kwargs) -> Github:
         ConnectionInitError: AutoKitteh connection was not initialized yet.
     """
     check_connection_name(connection)
-
```
daabr, line 29: Don't remove this empty line, reading the base URL should be in its own block.

integrations/oauth/configs.go
```diff
@@ -671,16 +671,20 @@ func setupAuth0(vs sdktypes.Vars, cfg *oauthConfig) {
 
 // TODO(INT-349): Support GHES in private OAuth mode too.
```
daabr: Remove this TODO

integrations/oauth/configs.go
```diff
@@ -671,16 +671,20 @@ func setupAuth0(vs sdktypes.Vars, cfg *oauthConfig) {
...
 func privatizeGitHub(vs sdktypes.Vars, cfg *oauthConfig) {
-	defaultAppName := os.Getenv("GITHUB_APP_NAME")
-	if defaultAppName == "" {
-		defaultAppName = defaultGitHubAppName
+	enterpriseURL := vs.GetValueByString("enterprise_url")
+	if enterpriseURL == "" {
+		enterpriseURL = defaultGitHubBaseURL
```
daabr, line 679: Not enough, you should also support the env var `GITHUB_ENTERPRISE_URL` in case `vs.GetValueByString("enterprise_url")` is empty, and use `defaultGitHubBaseURL` only if both of them are empty.

integrations/oauth/configs.go
```diff
@@ -671,16 +671,20 @@ func setupAuth0(vs sdktypes.Vars, cfg *oauthConfig) {
...
 	}
-	defaultAppName = fmt.Sprintf("/%s/", defaultAppName)
 
-	authURL := cfg.Config.Endpoint.AuthURL
-	// TODO(INT-353): Make GitHub var symbols non-internal, and reuse here.
-	privateAppName := fmt.Sprintf("/%s/", vs.GetValueByString("app_name"))
-	cfg.Config.Endpoint.AuthURL = strings.Replace(authURL, defaultAppName, privateAppName, 1)
+	appName := os.Getenv("GITHUB_APP_NAME")
```
daabr: What about privat app names?

integrations/oauth/configs.go
```diff
@@ -671,16 +671,20 @@ func setupAuth0(vs sdktypes.Vars, cfg *oauthConfig) {
...
+	if appName == "" {
+		appName = defaultGitHubAppName
+	}
+
+	cfg.Config.Endpoint.AuthURL = fmt.Sprintf("%s/apps/%s/installations/new", enterpriseURL, appName)
+	cfg.Config.Endpoint.TokenURL = enterpriseURL + "/login/oauth/access_token"
```
daabr: This is risky - you can't assume that `enterpriseURL` does or doesn't end with "/". If you construct the full URL, you need to join the path safely.

integrations/oauth/configs.go
```diff
@@ -671,16 +671,20 @@ func setupAuth0(vs sdktypes.Vars, cfg *oauthConfig) {
...
-	if defaultAppName == "" {
-		defaultAppName = defaultGitHubAppName
+	enterpriseURL := vs.GetValueByString("enterprise_url")
+	if enterpriseURL == "" {
+		enterpriseURL = defaultGitHubBaseURL
 	}
-	defaultAppName = fmt.Sprintf("/%s/", defaultAppName)
 
-	authURL := cfg.Config.Endpoint.AuthURL
-	// TODO(INT-353): Make GitHub var symbols non-internal, and reuse here.
```
daabr, line 681: It looks like you removed the TODO, but didn't actually do it

integrations/oauth/configs.go
```diff
@@ -671,16 +671,20 @@ func setupAuth0(vs sdktypes.Vars, cfg *oauthConfig) {
...
-	cfg.Config.Endpoint.AuthURL = strings.Replace(authURL, defaultAppName, privateAppName, 1)
+	appName := os.Getenv("GITHUB_APP_NAME")
+	if appName == "" {
+		appName = defaultGitHubAppName
+	}
+
+	cfg.Config.Endpoint.AuthURL = fmt.Sprintf("%s/apps/%s/installations/new", enterpriseURL, appName)
+	cfg.Config.Endpoint.TokenURL = enterpriseURL + "/login/oauth/access_token"
+	cfg.Config.Endpoint.DeviceAuthURL = enterpriseURL + "/login/device/code"
+	cfg.Config.Endpoint.AuthStyle = oauth2.AuthStyleInHeader
```
daabr: Why did you add this line?

integrations/oauth/configs.go
```diff
@@ -671,16 +671,20 @@ func setupAuth0(vs sdktypes.Vars, cfg *oauthConfig) {
...
-	privateAppName := fmt.Sprintf("/%s/", vs.GetValueByString("app_name"))
-	cfg.Config.Endpoint.AuthURL = strings.Replace(authURL, defaultAppName, privateAppName, 1)
+	appName := os.Getenv("GITHUB_APP_NAME")
+	if appName == "" {
+		appName = defaultGitHubAppName
+	}
+
+	cfg.Config.Endpoint.AuthURL = fmt.Sprintf("%s/apps/%s/installations/new", enterpriseURL, appName)
+	cfg.Config.Endpoint.TokenURL = enterpriseURL + "/login/oauth/access_token"
+	cfg.Config.Endpoint.DeviceAuthURL = enterpriseURL + "/login/device/code"
```
daabr: Same as the comment above: why are you constructing the URL manually instead of replacing the address in the existing string?

integrations/github/jwt.go
```diff
@@ -138,8 +139,14 @@ func getPrivateKey(privateKey string) []byte {
 	return pem.EncodeToMemory(b)
 }
 
-func enterpriseURL() (string, error) {
-	u := os.Getenv(enterpriseURLEnvVar)
+func enterpriseURL(vs sdktypes.Vars) (string, error) {
+	at := vs.Get(vars.AuthType)
```
daabr: Don't use `Get`, use `GetValue`

integrations/github/jwt.go
```diff
@@ -138,8 +139,14 @@ func getPrivateKey(privateKey string) []byte {
...
+	var u string
+	if at.IsValid() && at.Value() == integrations.OAuthPrivate {
+		u = vs.Get(vars.EnterpriseURL).Value()
```
daabr:
```suggestion
		u = vs.GetValue(vars.EnterpriseURL)
```

integrations/github/jwt.go
```diff
@@ -138,8 +139,14 @@ func getPrivateKey(privateKey string) []byte {
...
+	} else {
+		u = os.Getenv(enterpriseURLEnvVar)
```
daabr: Instead if `if-else`, use this:
```
u := ...
if ... {
  u = ...
}
```

integrations/oauth/configs.go
mario99logic: I wanted to check with you, since the app name isn’t one of the fields, should we add an app name field to the UI when creating the connection to support the cloud version, or is there a way to fetch it automatically?

integrations/oauth/configs.go
mario99logic: changing tom url.JoinPath since the baseURL is at the start of the URL

integrations/oauth/configs.go
mario99logic: Discussed offline.

integrations/oauth/configs.go
mario99logic: I misunderstood something, removing it.

daabr (approved):

PR #1256, opened 2025-04-16
feat(python): Support async handlers

tebeka:
This simplifies the usage of asyncio code, users can now write:

```python
import autokitteh


async def main(event):
    print('START')
    await act()
    print('END')


@autokitteh.activity
async def act():
    print('ACT')
```

We'll automatically run `main` in asyncio loop.

Ref: ENG-1976

### DIFF
--- a/runtimes/pythonrt/runner/loader.py
+++ b/runtimes/pythonrt/runner/loader.py
@@ -144,7 +144,7 @@ def exports(code_dir, file_name):
 
     tree = ast.parse(code, file_name, "exec")
     for node in tree.body:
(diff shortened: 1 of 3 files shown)


itayd:
how do we limit to only one parallel coroutine execution? because currently we do not support parallelism or concurrency.

tebeka:
> how do we limit to only one parallel coroutine execution? because currently we do not support parallelism or concurrency.

We currently don't limit. The change to the Python runtime to run in a single thread is a big one. IMO we should document it for now.

itayd (approved):

PR #1257, opened 2025-04-16
WIP: add jobs table and move start session to be async

efiShtain:


itayd:
@efiShtain i guess this can be closed?

PR #1263, opened 2025-04-22
feat: event deduping

itayd:
add deduplication_key to event. if set - event saving will return ErrAlreadyExists. if not exists - no deduping is done.


internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -206,14 +209,18 @@ func ParseEvent(e Event) (sdktypes.Event, error) {
 		return sdktypes.InvalidEvent, sdkerrors.NewInvalidArgumentError("event must have a connection or trigger")
 	}
 
+	// extract the actual deduped part.
+	_, dedupKey, _ := strings.Cut(e.DeduplicationKey, "/")
+
```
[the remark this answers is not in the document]
itayd: not relevant, added comment.

efiShtain:
not sure what you try to solve with this PR

we create events after receiving calls from external sources (webhooks, cron, etc)

most of this has some kind of id on them, we can use that id if we want to deduplicate events

I think it should be done on a higher level, let's talk about it

migrations/postgres/enterprise/20251104003836_dedup_events.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- modify "events" table
+ALTER TABLE "events" ADD COLUMN "deduplication_key" text NULL;
+-- create index "idx_ddk_not_empty" to table: "events"
+CREATE UNIQUE INDEX "idx_ddk_not_empty" ON "events" ("deduplication_key") WHERE (deduplication_key <> ''::text);
```
efiShtain: I think you want to check is not null, which is different than equal empty string
assuming you won't set the value to an empty string (the default is null if not set)

efiShtain: you also stated in the proto that you deduplicate based on key, destination and event type, so the unique check should reflect that
or you should update the comment in the proto

internal/backend/db/dbgorm/events.go
```diff
@@ -85,17 +96,23 @@ func (db *gormdb) SaveEvent(ctx context.Context, event sdktypes.Event) error {
...
+		ProjectID:        pid.UUIDValue(),
+		OrgID:            oid.UUIDValuePtr(),
+		EventID:          event.ID().UUIDValue(),
+		DestinationID:    event.DestinationID().UUIDValue(),
+		ConnectionID:     uuidPtrOrNil(connectionID),
+		TriggerID:        uuidPtrOrNil(event.DestinationID().ToTriggerID()),
+		EventType:        event.Type(),
+		Data:             kittehs.Must1(json.Marshal(event.Data())),
+		Memo:             kittehs.Must1(json.Marshal(event.Memo())),
+		DeduplicationKey: dedupKey,
```
efiShtain, line 108: better to have it as null instead of empty string
see my comment in the migration file

integrations/discord/websocket.go
```diff
@@ -73,20 +73,22 @@ func (h handler) dispatchAsyncEventsToConnections(cids []sdktypes.ConnectionID,
...
-		if err != nil {
-			if errors.Is(err, sdkerrors.ErrResourceExhausted) {
-				l.Info("Event dispatch failed due to resource exhaustion")
-			} else {
-				l.Error("Event dispatch failed", zap.Error(err))
-			}
-
-			return
+
+		switch {
```
efiShtain, line 82: replace this function with common.DispatchEvent
update the already exists error handling there if you want

efiShtain (commented):
Better to hide the discord sessions map behind a lock, it is currently in the global scope and not protected

efiShtain (commented):
see my comment

PR #1264, opened 2025-04-22
chore(python): Reduce server start time

tebeka:
Use gRPC to wait for Python server. This reduces start time (until first health check) from 1 second to about 270ms on my machine.

Adds some timing logs to runner, until we add telemetry.

Ref: ENG-1985

### DIFF
--- a/runtimes/pythonrt/runner_manager.go
+++ b/runtimes/pythonrt/runner_manager.go
@@ -56,6 +56,8 @@ func dialRunner(ctx context.Context, addr string) (*RunnerClient, error) {
 	span.SetAttributes(attribute.String("addr", addr))
 
 	creds := insecure.NewCredentials()
+	// Python takes it's time going up, this with grpc.WaitForReady(true) below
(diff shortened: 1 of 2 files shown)


runtimes/pythonrt/runner_manager.go
```diff
@@ -51,35 +49,20 @@ type Healthier interface {
...
-}
-
 func dialRunner(ctx context.Context, addr string) (*RunnerClient, error) {
 	ctx, span := telemetry.T().Start(ctx, "dialRunner")
 	defer span.End()
 
 	span.SetAttributes(attribute.String("addr", addr))
 
 	creds := insecure.NewCredentials()
+	params := grpc.ConnectParams{
```
itayd, line 61: please add comments here explainig why this is the way it is and how it waits to the connection to be open etc

tebeka: Done.

itayd (approved):

PR #1265, opened 2025-04-22
fix: refresh oauth token for custom oauth connnection

mario99logic:
Injected the OAuth service and mapped the connection ID to retrieve the corresponding variables.

This PR solves the issue of accessing the client ID and secret for private OAuth connections, which previously wasn't possible because these credentials were not available in environment variables.
Until now, only default OAuth connections worked correctly. With this update, private OAuth connections can also retrieve credentials using the connection variables.

Tested by directly calling the function from Python and debugging the flow — confirmed that the credentials are correctly retrieved for a private OAuth connection.

### DIFF
--- a/runtimes/pythonrt/worker_grpc_handler.go
+++ b/runtimes/pythonrt/worker_grpc_handler.go
@@ -433,12 +433,10 @@ func (s *workerGRPCHandler) RefreshOAuthToken(ctx context.Context, req *userCode
 	}
 
 	// Get the integration's OAuth configuration.
-	var cid sdktypes.ConnectionID = sdktypes.InvalidConnectionID
-	cid_str, ok := runner.envVars[req.Connection+"__connection_id"]
-	if ok && cid_str != "" {
+	var cid sdktypes.ConnectionID
+	if cid_str := runner.envVars[req.Connection+"__connection_id"]; cid_str != "" {
 		var err error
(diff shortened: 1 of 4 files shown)


runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -431,10 +433,20 @@ func (s *workerGRPCHandler) RefreshOAuthToken(ctx context.Context, req *userCode
...
+	if ok && cid_str != "" {
+		var err error
+		cid, err = sdktypes.ParseConnectionID(cid_str)
+		if err != nil {
+			runner.log.Warn("invalid connection ID",
+				zap.String("connection", req.Connection),
+				zap.String("connection_id", cid_str),
+				zap.Error(err))
+		}
+	}
```
itayd:
```suggestion
	var cid sdktypes.ConnectionID
	if cid_str := runner.envVars[req.Connection+"__connection_id"]; cid_str != "" {
		var err error
		if cid, err = sdktypes.ParseConnectionID(cid_str); err != nil {
			runner.log.Warn("invalid connection ID",
				zap.String("connection", req.Connection),
				zap.String("connection_id", cid_str),
				zap.Error(err))
		}
	}
```

itayd (approved):

PR #1269, opened 2025-04-24
fix(python): Append site-packages to sys.path

tebeka:
Previously we appended site-packages to sys.path and it caused issues with 3rd party packages overriding standard library ones.

Fixes: ENG-2096


efiShtain (approved):

efiShtain:
Do we need it in main ak Dockerfile as well ?

PR #1273, opened 2025-04-25
fix: separate temporal dev server download and start

itayd:
download now will not count for the start timeout, as well as download to a cache directory and not a temp dir. this will minimize the need to download the server every time and should make it more stable.

also add an "ak temporal download" command to warm up when needed.

this duplicates the temporal code in order to give us more control and tune what it does.

### DIFF
--- a/internal/backend/temporalclient/config.go
+++ b/internal/backend/temporalclient/config.go
@@ -4,11 +4,11 @@ import (
 	"path/filepath"
 	"time"
 
-	"go.temporal.io/sdk/testsuite"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zapcore"
 
 	"go.autokitteh.dev/autokitteh/internal/backend/configset"
+	"go.autokitteh.dev/autokitteh/internal/backend/temporaldevsrv"
 	"go.autokitteh.dev/autokitteh/internal/xdg"
 )
 
@@ -35,7 +35,7 @@ type Config struct {
 	Namespace             string `koanf:"namespace"`
 
 	// DevServer.ClientOptions is not used.
-	DevServer testsuite.DevServerOptions `koanf:"dev_server"`
+	DevServer temporaldevsrv.DevServerOptions `koanf:"dev_server"`
 
 	// Max number of attempts to start the dev server.
 	DevServerStartMaxAttempts int `koanf:"dev_server_start_max_attempts"`
@@ -76,7 +76,10 @@ var (
 		Dev: &Config{
 			Monitor:               defaultMonitorConfig,
 			StartDevServerIfNotUp: true,
-			DevServer: testsuite.DevServerOptions{
+			DevServer: temporaldevsrv.DevServerOptions{
+				CachedDownload: temporaldevsrv.CachedDownload{
+					DestDir: xdg.CacheHomeDir(),
+				},
 				LogLevel:   zapcore.WarnLevel.String(),
 				EnableUI:   true,
 				DBFilename: filepath.Join(xdg.DataHomeDir(), "temporal_dev.sqlite"),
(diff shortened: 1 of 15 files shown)


internal/backend/temporalclient/devsrv.go
```diff
@@ -0,0 +1,79 @@
...
+
+		startCtx := ctx
+
+		if c.cfg.DevServerStartTimeout != 0 {
+			var done func()
+			startCtx, done = context.WithTimeout(ctx, c.cfg.DevServerStartTimeout)
+			defer done()
+		}
+
+		c.srv, err = temporaldevsrv.StartDevServer(startCtx, devSrvCfg)
```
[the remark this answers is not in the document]
itayd: done, haha

internal/backend/temporaldevsrv/download.go
```diff
@@ -0,0 +1,202 @@
...
+func extractTarball(r io.Reader, toExtract string, w io.Writer) error {
+	r, err := gzip.NewReader(r)
+	if err != nil {
+		return err
+	}
+	tarRead := tar.NewReader(r)
+	for {
+		h, err := tarRead.Next()
+		if err != nil {
+			// This can be EOF which means we never found our file
```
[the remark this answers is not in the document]
itayd: done

internal/xdg/xdg.go
```diff
@@ -38,6 +39,8 @@ func ConfigHomeDir() string { return homeDir(xdg.ConfigHome) }
 // and guarantees that it exists, so callers can use it safely.
 func DataHomeDir() string { return homeDir(xdg.DataHome) }
 
+func CacheHomeDir() string { return homeDir(xdg.CacheHome) }
```
efiShtain, line 42: not sure what reload does, but maybe you should add this cache home dir function there too?

internal/backend/temporalclient/config.go
```diff
@@ -76,7 +76,10 @@ var (
 		Dev: &Config{
 			Monitor:               defaultMonitorConfig,
 			StartDevServerIfNotUp: true,
-			DevServer: testsuite.DevServerOptions{
+			DevServer: temporaldevsrv.DevServerOptions{
```
efiShtain, line 79: why do we need this option ?
it's a dev utility, shouldn't be used, can't we just download it to the config directory of AK and that's it ?

internal/backend/temporalclient/devsrv.go
```diff
@@ -0,0 +1,80 @@
...
+		return fmt.Errorf("open Temporal dev server log file: %w", err)
+	}
+
+	devSrvCfg := c.cfg.DevServer
+	devSrvCfg.ExistingPath = exePath
+	devSrvCfg.ClientOptions = &c.opts
+	devSrvCfg.Stderr = c.logFile
+	devSrvCfg.Stdout = c.logFile
+
+	for i := 0; i < c.cfg.DevServerStartMaxAttempts && c.srv == nil; i++ {
```
efiShtain: range c.cfg.DevServerStartMaxAttempts is more idiomatic

efiShtain (approved):
Overly engineered, I hope you copy pasted it from somewhere
but fine it it works

itayd:
> Overly engineered, I hope you copy pasted it from somewhere but fine it it works

yup, copied mostly from temporal and moved some stuff around.

internal/xdg/xdg.go
itayd: right!

internal/backend/temporalclient/devsrv.go
itayd: it doesn't have the c.srv condition in it though. but can change it, it'll be simpler.

internal/backend/temporalclient/config.go
itayd: there is more than just a dest dir there, i think it's useful.

PR #1277, opened 2025-04-27
bug(python): Global print hangs

tebeka:
This starts a print loop goroutine on `Run`.

Fixes: ENG-2099


runtimes/pythonrt/pythonrt.go
```diff
@@ -236,6 +236,16 @@ func entryPointFileName(entryPoint string) string {
 	return entryPoint
 }
 
+func (py *pySvc) printConsumer(ctx context.Context) {
+	for p := range py.channels.print {
+		py.log.Info("print", zap.String("message", p.message))
```
efiShtain: I think this should be at most a debug, 
better not even log it, that's user message not ours

runtimes/pythonrt/pythonrt.go
```diff
@@ -236,6 +236,16 @@ func entryPointFileName(entryPoint string) string {
...
+		if err := py.cbs.Print(ctx, py.runID, p.message); err != nil {
+			py.log.Error("print error", zap.Error(err))
+		}
+		close(p.doneChannel)
```
efiShtain: why do we close the done channel here ?

runtimes/pythonrt/pythonrt.go
tebeka: Done.

runtimes/pythonrt/pythonrt.go
tebeka: To notify the gRPC handler that the print is actually executed. See https://github.com/autokitteh/autokitteh/blob/main/runtimes/pythonrt/worker_grpc_handler.go#L139

runtimes/pythonrt/pythonrt.go
```diff
@@ -236,6 +234,32 @@ func entryPointFileName(entryPoint string) string {
...
+	py.log.Debug("print", zap.String("message", msg.message))
+	if err := py.cbs.Print(ctx, py.runID, msg.message); err != nil {
+		py.log.Error("print error", zap.Error(err))
+	}
+
+	close(msg.doneChannel)
+}
+
+func (py *pySvc) printConsumer(ctx context.Context) {
+	for {
```
efiShtain, line 246: any chance that this will never exit the loop ?

runtimes/pythonrt/pythonrt.go
```diff
@@ -236,6 +234,32 @@ func entryPointFileName(entryPoint string) string {
 	return entryPoint
 }
 
+func (py *pySvc) handlePrint(ctx context.Context, msg *logMessage) {
+	py.log.Debug("print", zap.String("message", msg.message))
```
efiShtain: already logging on worker_grpc_handler print request

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -129,7 +129,7 @@ func (s *workerGRPCHandler) Log(ctx context.Context, req *userCode.LogRequest) (
 }
 
 func (s *workerGRPCHandler) Print(ctx context.Context, req *userCode.PrintRequest) (*userCode.PrintResponse, error) {
-	s.log.Info("Print request", zap.String("message", req.Message))
+	s.log.Debug("Print request", zap.String("message", req.Message))
```
efiShtain: maybe print only after we find the runner ?
otherwise we can just ignore it

runtimes/pythonrt/pythonrt.go
```diff
@@ -579,7 +582,11 @@ func (py *pySvc) Call(ctx context.Context, v sdktypes.Value, args []sdktypes.Val
 		span.End()
 	}
 
-	defer py.drainPrints(ctx)
+	defer func() {
+		ctx, _ = context.WithTimeout(ctx, 10*time.Millisecond)
```
efiShtain: this is basically sleep ?
maybe time.after is better / more readable?

runtimes/pythonrt/pythonrt.go
```diff
@@ -236,6 +234,32 @@ func entryPointFileName(entryPoint string) string {
 	return entryPoint
 }
 
+func (py *pySvc) handlePrint(ctx context.Context, msg *logMessage) {
```
efiShtain, line 237: see that you remove the tracing stuff 
I think itay wants them

runtimes/pythonrt/pythonrt.go
tebeka: If the context is not cancelled, we don't close the `print` channel (tried but got panics)

runtimes/pythonrt/pythonrt.go
tebeka: Removed.

runtimes/pythonrt/worker_grpc_handler.go
tebeka: Done

runtimes/pythonrt/pythonrt.go
tebeka: But there's nothing to trace now :)

@itayd?

mario99logic (commented):
Just a general question, didn’t we decide to stop using goroutines because they’re not well-suited for Temporal?

runtimes/pythonrt/pythonrt.go
itayd: the tracing here will be done in the session level.

runtimes/pythonrt/pythonrt.go
tebeka: Changed to sleep

tebeka:
> Just a general question, didn’t we decide to stop using goroutines because they’re not well-suited for Temporal?

You're right in general, but @itayd made a change a while back that prints do not require a session - only write to database.

runtimes/pythonrt/pythonrt.go
```diff
@@ -559,9 +561,14 @@ func (py *pySvc) Call(ctx context.Context, v sdktypes.Value, args []sdktypes.Val
 			}
 			return sdktypes.InvalidValue, fmt.Errorf("start request: %w", err)
 		}
+
 		defer func() {
 			py.cleanup(context.Background())
+
+			time.Sleep(10 * time.Millisecond) // Give time to print consumer to finish
```
itayd, line 568: i really don't like it. but i know we'll refactor this soon.

itayd (approved):

PR #1278, opened 2025-04-27
add more versions to published images

efiShtain:


tebeka (approved):
I'm OK with it. Later we might want to switch to [ReadBuildInfo](https://pkg.go.dev/runtime/debug@go1.24.2#ReadBuildInfo) which stores `vcs.revision:`

PR #1279, opened 2025-04-27
fix(python): Handle non-pickleable return values

tebeka:
Fixes: ENG-2100


tebeka:
@itayd The session test still fails since the exception doesn't get to the print. Any idea how to fix it? (I do see it in the web UI)

itayd (approved):

PR #1294, opened 2025-05-08
feat: local pyhon multi env

itayd:
Support multiple venvs on local installation (meaning non-dockerized runners).
Each venv name is a hash of its requirements.


runtimes/pythonrt/build.go
```diff
@@ -74,7 +74,13 @@ func (py *pySvc) Build(ctx context.Context, fsys fs.FS, path string, values []sd
 	var art sdktypes.BuildArtifact
 	art = art.WithCompiledData(compiledData)
 
-	exports, err := findExports(py.log, fsys)
+	hash, err := getRequirementsHash(data)
```
tebeka: I'm not sure passing `hash` around is the right way. Maybe add a `pythonExe` field to `py` and populate it?

runtimes/pythonrt/build.go
```diff
@@ -93,10 +99,9 @@ func findExports(log *zap.Logger, fsys fs.FS) ([]sdktypes.BuildExport, error) {
 
 	var pyExe string
 	if !isUserPy {
-		if err := ensureVEnv(log, sysPyExe); err != nil {
+		if pyExe, err = ensureVEnv(log, envName, sysPyExe); err != nil {
```
tebeka: You need to pass the user requirements file (if they exist).

runtimes/pythonrt/requirements.go
```diff
@@ -0,0 +1,37 @@
...
+			continue
+		}
+
+		var b bytes.Buffer
+
+		if _, err := io.Copy(&b, tf); err != nil {
+			return "", err
+		}
+
+		return kittehs.SHA256Hash(b.Bytes())
```
tebeka: `kittehs.SHA256Hash` gobs encode the value first. IMO this can be not deterministic.
From the [gob documentation](https://go.dev/blog/gob#types-on-the-wire):

> The first time you send a given type, the gob package includes in the data stream a description of that type. In fact, what happens is that the encoder is used to encode, in the standard gob encoding format, an internal struct that describes the type and gives it a unique number.

Use sha256 (or just sha1) on the bytes.

runtimes/pythonrt/runner_manager_local.go
```diff
@@ -71,10 +72,9 @@ func configureLocalRunnerManager(log *zap.Logger, cfg LocalRunnerManagerConfig)
 	if !isUserPy {
 		if !cfg.LazyLoadVEnv {
 			log.Info("ensuring venv on start")
-			if err := ensureVEnv(log, pyExe); err != nil {
+			if lm.pyExe, err = ensureVEnv(log, "", pyExe); err != nil {
```
tebeka: Why the empty env name?

runtimes/pythonrt/runner_manager_local.go
```diff
@@ -262,11 +267,18 @@ func dirExists(path string) bool {
...
+func ensureVEnv(log *zap.Logger, name, pyExe string) (pyExePath string, err error) {
+	venv := venvPath(name)
+	pyExePath = path.Join(venv, "bin", "python")
+
+	if dirExists(venv) {
+		return
+	}
+
+	log.Info("creating venv", zap.String("path", venv))
+	if err := createVEnv(pyExe, venv); err != nil {
```
tebeka: Need to pass user requirements here.

runtimes/pythonrt/requirements.go
itayd: fixed

runtimes/pythonrt/runner_manager_local.go
itayd: no hash

itayd: (uses default)

runtimes/pythonrt/build.go
itayd: done

runtimes/pythonrt/runner_manager_local.go
itayd: now passing

runtimes/pythonrt/build.go
itayd: now passing the actual requirements

haimzlato (approved):

runtimes/pythonrt/build.go
```diff
@@ -84,7 +90,7 @@ func (py *pySvc) Build(ctx context.Context, fsys fs.FS, path string, values []sd
 	return art, nil
 }
 
-func findExports(log *zap.Logger, fsys fs.FS) ([]sdktypes.BuildExport, error) {
+func findExports(log *zap.Logger, fsys fs.FS, reqs string) ([]sdktypes.BuildExport, error) {
```
efiShtain: why do you pass reqs here? is is not used

runtimes/pythonrt/runner_manager_local.go
```diff
@@ -96,16 +99,36 @@ func (l *localRunnerManager) Start(ctx context.Context, sessionID sdktypes.Sessi
 		sessionID:     sessionID,
 	}
 
-	if l.cfg.LazyLoadVEnv {
-		_, venvSpan := telemetry.T().Start(ctx, "localRunnerManager.ehnsureVEnv")
+	err := func() error {
```
efiShtain: why is this an anonymous function which is immediatly invoked?

runtimes/pythonrt/runner_manager_local.go
```diff
@@ -262,11 +285,34 @@ func dirExists(path string) bool {
...
 
-func ensureVEnv(log *zap.Logger, pyExe string) error {
+func ensureVEnv(log *zap.Logger, reqs, pyExe string) (pyExePath string, err error) {
+	venvPath := venvPath(reqs)
+	pyExePath = path.Join(venvPath, "bin", "python")
+
+	// This locks the env creation so two concurrent runners don't try to
+	// create the same venv at the same time.
+	// Since this is a local runner manager, this is sufficient.
+	venvMutex.Lock()
```
efiShtain, line 286: you can optimize with readWrite lock, since it should almost never write

efiShtain (approved):
fix the issues we talked about before merging

runtimes/pythonrt/build.go
itayd: removed

runtimes/pythonrt/runner_manager_local.go
itayd: commented.

itayd: actually, not required anymore. removed.

PR #1297, opened 2025-05-15
feat: add downloadLog session service endpoint 

mario99logic:
Adds a new DownloadLogs endpoint to the session service, allowing client to get encoded text of the session logs.

Includes:
New DownloadLogs method in sessions service
Helper function for formatting multi-line log entries with timestamp alignment
proto definitions, service handlers, SDK client
Pagination support for large log sets

ref: ENG-2122


sdk/sdkservices/sessions.go
```diff
@@ -58,6 +59,7 @@ type Sessions interface {
 	List(ctx context.Context, filter ListSessionsFilter) (*ListSessionResult, error)
 	Get(ctx context.Context, sessionID sdktypes.SessionID) (sdktypes.Session, error)
 	GetLog(ctx context.Context, filter SessionLogRecordsFilter) (*GetLogResults, error)
+	DownloadLogs(ctx context.Context, sessionID sdktypes.SessionID) (io.ReadCloser, error)
```
efiShtain: why ReadCloser vs bytes ?

internal/backend/db/db.go
```diff
@@ -102,6 +102,7 @@ type DB interface {
 	CreateSession(ctx context.Context, session sdktypes.Session) error
 	GetSession(ctx context.Context, sessionID sdktypes.SessionID) (sdktypes.Session, error)
 	GetSessionLog(ctx context.Context, filter sdkservices.SessionLogRecordsFilter) (*sdkservices.GetLogResults, error)
+	GetSessionData(ctx context.Context, sessionID sdktypes.SessionID) ([]byte, error)
```
efiShtain: I'm not sure you need a new function here
you can probably reuse GetSessionLog, just need to add an argument to get all logs, basically ignoring the page limits

not sure, worth considering

converting to bytes can be done in the service layer

mario99logic: Yes I also considered it, wanted to ask you about it.

itayd: what efi said.

sdk/sdkservices/sessions.go
mario99logic: changed to bytes

internal/backend/db/db.go
mario99logic: used `GetSessionLog`

internal/backend/sessions/sessions.go
```diff
@@ -120,6 +123,55 @@ func (s *sessions) GetLog(ctx context.Context, filter sdkservices.SessionLogReco
...
+		allRecords = append(allRecords, logs.Records...)
+
+		if int32(len(logs.Records)) < pageSize { // if no more records.
+			break
+		}
+
+		skip += pageSize
+	}
+
+	data, err := json.Marshal(allRecords)
```
efiShtain: with this way, you are iterating over everything twice, once to marshal and then to gzip

try creating a json encoder and gzip writer at the beginning (before db access)

then write each record to the writer

at the end get the bytes

something like this (not sure this will work as is)
```go
var body bytes.Buffer
w := gzip.NewWriter(&body)

enc := json.NewEncoder(w)

for {

  //when you get the records
  //do for each record
  enc.Encode(row)
}

if err := w.Close(); err != nil {return err}

return body.Bytes()

```

mario99logic: I dont think this will work because enc.Encode create a separate json so it would produce something like:
```
{"field":"value1"}
{"field":"value2"}
{"field":"value3"}
```
but I can try this approach and add "," between records

internal/backend/sessions/sessions.go
```diff
@@ -120,6 +123,81 @@ func (s *sessions) GetLog(ctx context.Context, filter sdkservices.SessionLogReco
...
+func formatSessionLog(buf *bytes.Buffer, record sdktypes.SessionLogRecord) error {
+	value, ok := record.GetPrint()
+	if !ok {
+		return nil
+	}
+
+	printStr := value.String()
+	if strings.HasPrefix(printStr, "string:{v:\"") && strings.HasSuffix(printStr, "\"}") {
+		printStr = printStr[11 : len(printStr)-2]
+	}
```
efiShtain: don't try to parse stuff your self
convert the Value to StringValue and get the text from it

internal/backend/sessions/sessions.go
```diff
@@ -120,6 +123,81 @@ func (s *sessions) GetLog(ctx context.Context, filter sdkservices.SessionLogReco
...
+
+	// Decode escaped characters like \n and \".
+	if unquoted, err := strconv.Unquote(`"` + printStr + `"`); err == nil {
+		printStr = unquoted
+	}
+
+	lines := strings.Split(printStr, "\n")
```
efiShtain, line 181: why ?
if there is a new line, let it be

internal/backend/sessions/sessions.go
```diff
@@ -120,6 +123,81 @@ func (s *sessions) GetLog(ctx context.Context, filter sdkservices.SessionLogReco
...
+	}
+
+	// First line with timestamp.
+	firstLine := fmt.Sprintf("[%s]:  %s\n", record.Timestamp().Format("2006-01-02 15:04:05"), lines[0])
+	if _, err := buf.WriteString(firstLine); err != nil {
+		return fmt.Errorf("failed to write log line: %w", err)
+	}
+
+	// Indent subsequent lines.
+	for _, l := range lines[1:] {
```
efiShtain, line 193: not sure this is mandatory, 
are you sure it works ok ?

internal/backend/sessions/sessions.go
```diff
@@ -120,6 +123,81 @@ func (s *sessions) GetLog(ctx context.Context, filter sdkservices.SessionLogReco
...
+		if int32(len(logs.Records)) < pageSize {
+			break
+		}
+		skip += pageSize
+	}
+
+	return buf.Bytes(), nil
+}
+
+func formatSessionLog(buf *bytes.Buffer, record sdktypes.SessionLogRecord) error {
```
efiShtain: now that I see it like this, I think writeFormattedSessionLog(buf, record) sounds better

only format sounds like you are going to return the formatted output, but you don't

also, need to add tests to this function, 
record which is not print, print with multiple rows, etc...

efiShtain (commented):
Need to add CLI implementation as well

internal/backend/sessions/sessions.go
mario99logic: yes this is for if there are more than 1 line then the next line won't be directly under the timestamp, it will be under the previous line.

internal/backend/sessions/sessions.go
mario99logic: I split the lines to format them properly, without splitting, each new line would appear right under the timestamp instead of being indented like in the UI.

tests/system/checks.go
```diff
@@ -122,6 +124,45 @@ func checkAKOutput(t *testing.T, step string, ak *tests.AKResult) error {
...
+
+	if strings.HasPrefix(want, "file ") {
+		wantFilename := strings.TrimSpace(strings.TrimPrefix(want, "file "))
+		b, err := os.ReadFile(wantFilename)
+		if err != nil {
+			return fmt.Errorf("failed to read reference file %q: %w", wantFilename, err)
+		}
+		want = strings.TrimSpace(string(b))
+	}
+
```
efiShtain, line 149: can you extract this to some external function with a meaningful name ?
that's really hard to read

in addition, can you consider using named capture groups in regex instead of parsing stuff with lots of trimming ? will make stuff much less fragile

I'm ok with this current version of the code if you want to split it to a different PR, but if so, it has to be right next thing, since this is not very readable and kind of fragile

cmd/ak/cmd/sessions/prints.go
```diff
@@ -97,6 +102,42 @@ var printsCmd = common.StandardCommand(&cobra.Command{
 	},
 })
 
+var downloadLogsCmd = common.StandardCommand(&cobra.Command{
```
efiShtain: why is this in the same file as print ?
move to downloadlogs.go file

cmd/ak/cmd/sessions/prints.go
```diff
@@ -97,6 +102,42 @@ var printsCmd = common.StandardCommand(&cobra.Command{
...
+	Short: "Download logs for a session",
+	Args:  cobra.ExactArgs(1),
+
+	RunE: func(cmd *cobra.Command, args []string) error {
+		sid, err := acquireSessionID(args[0])
+		if err = common.AddNotFoundErrIfCond(err, sid.IsValid()); err != nil {
+			return common.ToExitCodeWithSkipNotFoundFlag(cmd, err, "session")
+		}
+
+		ctx, done := common.LimitedContext()
```
efiShtain: Did you test a large log file ? 
context might time out, need to thing about it

cmd/ak/cmd/sessions/prints.go
```diff
@@ -97,6 +102,42 @@ var printsCmd = common.StandardCommand(&cobra.Command{
...
+			return fmt.Errorf("failed to download logs: %w", err)
+		}
+
+		// Use default output filename if none provided.
+		out := outputPath
+		if out == "" {
+			timestamp := time.Now().Format("20060102_150405")
+			filename := fmt.Sprintf("%s_%s.txt", sid.String(), timestamp)
+			out = filepath.Join(".", filename)
+		}
```
efiShtain:
```suggestion
		
		if outputPath == "" {
			timestamp := time.Now().Format("20060102_150405")
			filename := fmt.Sprintf("%s_%s.txt", sid.String(), timestamp)
			outputPath = filepath.Join(".", filename)
		}
```

cmd/ak/cmd/sessions/prints.go
```diff
@@ -105,4 +146,7 @@ func init() {
 	printsCmd.Flags().DurationVarP(&pollInterval, "poll-interval", "i", defaultPollInterval, "poll interval")
 
 	common.AddFailIfNotFoundFlag(printsCmd)
+
+	downloadLogsCmd.Flags().StringVarP(&outputPath, "output", "o", "", "path to output file (default is ./<session_id>_<timestamp>.txt)")
+	common.AddFailIfNotFoundFlag(downloadLogsCmd)
```
efiShtain: again, move to downloadlogs file

internal/backend/sessions/sessiondownloadlogs_test.go
```diff
@@ -0,0 +1,127 @@
...
+
+	"github.com/stretchr/testify/assert"
+
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+// Common test time to use across all tests.
+var fixedTime = time.Date(2025, 5, 15, 10, 30, 0, 0, time.UTC)
+
+func TestWriteFormattedSessionLog_NormalPrint(t *testing.T) {
```
efiShtain: what is normal print ?
what is not normal print ?

internal/backend/sessions/sessiondownloadlogs_test.go
```diff
@@ -0,0 +1,127 @@
...
+	numValue := sdktypes.NewIntegerValue(42)
+	record := sdktypes.NewPrintSessionLogRecord(fixedTime, numValue, 1)
+
+	err := writeFormattedSessionLog(buf, record)
+
+	assert.NoError(t, err)
+	assert.Equal(t, "[2025-05-15 10:30:00]:  42\n", buf.String())
+}
+
+func TestWriteFormattedSessionLog_InvalidRecord(t *testing.T) {
```
efiShtain: why no error on invalid record ?

internal/backend/sessions/sessiondownloadlogs_test.go
```diff
@@ -0,0 +1,127 @@
...
+}
+
+// Mock buffer that returns an error on write.
+type mockBuffer struct {
+	bytes.Buffer
+}
+
+func (m *mockBuffer) WriteString(s string) (int, error) {
+	return 0, assert.AnError
+}
```
efiShtain, line 127: nicely done

internal/backend/sessions/sessions.go
```diff
@@ -120,6 +124,81 @@ func (s *sessions) GetLog(ctx context.Context, filter sdkservices.SessionLogReco
...
+		return nil
+	}
+
+	printStr, err := value.ToString()
+	if err != nil {
+		return fmt.Errorf("failed to convert value to string: %w", err)
+	}
+
+	// Decode escaped characters like \n and \".
+	if unquoted, err := strconv.Unquote(`"` + printStr + `"`); err == nil {
```
efiShtain, line 177: what's the purpose of unquote omething that you add quotes to it ?

mario99logic: the function `strconv.Unquote` converts escaped characters in a string to their meanings. 

For example:
\n becomes an actual line break
\" becomes just a quote character "
\\ becomes just a single backslash \

The function expects the input to be a properly quoted string, so I add the quotes.

internal/backend/sessions/sessiondownloadlogs_test.go
mario99logic: No error is returned because the function is designed to skip records it's not meant to process. It treats "not a print record" as an expected condition rather than an error condition, which I think its the right approach here.

internal/backend/sessions/sessiondownloadlogs_test.go
mario99logic: Normal prints successfully passes the record.GetPrint() check

Not normal is any record where record.GetPrint() returns ok = false which are tested in `TestWriteFormattedSessionLog_InvalidRecord`

tests/system/checks.go
mario99logic: I will open a ticket and do this on a separate PR, the only use of this func now is for the download-logs test.

cmd/ak/cmd/sessions/prints.go
mario99logic: tested a file with 20k logs and worked nicely.

internal/backend/sessions/sessiondownloadlogs_test.go
efiShtain: So make the test name more readable

It should indicate that you expect it to skip invalid log

so TestWriteFormattedSessionLog_SkipInvalidRecord

internal/backend/sessions/sessiondownloadlogs_test.go
efiShtain: Normal is not indicative of what you mean

Write something like TestPrintRecord

cmd/ak/cmd/sessions/prints.go
efiShtain: Not sure it will work the same with remote backend but ok, let's see

tests/system/checks.go
efiShtain: Add the PR ticket here please

internal/backend/sessions/sessions.go
efiShtain: weird naming for a function (not you, golang)

cmd/ak/cmd/sessions/downloadLogs.go
```diff
@@ -0,0 +1,54 @@
+package sessions
```
efiShtain: rename the file please
we don't use capital casing in files

download_logs

cmd/ak/cmd/sessions/downloadLogs.go
```diff
@@ -0,0 +1,54 @@
...
+		ctx, done := common.LimitedContext()
+		defer done()
+
+		data, err := sessions().DownloadLogs(ctx, sid)
+		if err != nil {
+			return fmt.Errorf("failed to download logs: %w", err)
+		}
+
+		// Use default output filename if none provided.
+		if outputPath == "" {
```
efiShtain: I'm thinking if output path should be output-dir instead, and you should make sure the provided path is a directory

or if you need to check if the path is a directory and if it exists



cause in current state use can provide a directory or a file, not sure what will happen

what do you think ?

tests/system/testdata/end2end/download_session_logs.txtar
```diff
@@ -0,0 +1,31 @@
...
+ak deploy --manifest project.yaml
+return code == 0
+
+# Send HTTP GET request to trigger deployment to start new session.
+http get /webhooks/00000000000000000000000003/
+resp code == 202
+
+wait 5s for session ses_00000000000000000000000007
+
+ak session download-logs ses_00000000000000000000000007 --output downloaded_logs.txt
```
efiShtain: please add another test that checks that no error is returned if you don't provide an output path
let's assume for now that it will create the correct file

efiShtain (approved):

cmd/ak/cmd/sessions/downloadLogs.go
mario99logic: I think its better to be a directory rather than a file, if we change it we need to make changes to the tests also lets address it with the other comment in a separate ticket if you agree.
Im creating one ticket for both issues

tests/system/checks.go
mario99logic: https://linear.app/autokitteh/issue/ENG-2129/refactor-file-check-parsing-in-checkfilecontent-logic-and-improve

PR #1298, opened 2025-05-15
feat: enterprise worker management and throttling

efiShtain:
working solution, throttling db based and agent based (not a general compute layer accounting)


internal/backend/workflowresourcemanager/workflowmanager.go
```diff
@@ -0,0 +1,23 @@
...
+	"go.uber.org/fx"
+)
+
+type Svcs struct {
+	fx.In
+	Temporal temporalclient.Client
+	DB       db.DB
+}
+
+type WorkflowResourcesManager interface {
```
itayd: so essentially this is the queuing strategy interface?

internal/backend/workflowresourcemanager/workflowmanager.go
```diff
@@ -0,0 +1,23 @@
...
+	Execute(ctx context.Context, options client.StartWorkflowOptions, name string, args any) error
```
itayd: this should replace temporal's ExecuteWorkflow?

internal/backend/workflowresourcemanager/enterprise_queue.go
```diff
@@ -0,0 +1,51 @@
...
+	}
+}
+
+func (q *q) push(job job) {
+	q.lock.Lock()
+	defer q.lock.Unlock()
+	q.jobs = append(q.jobs, job)
+}
+
+func (q *q) popX(n int) []job {
```
itayd: `popN` :P

internal/backend/workflowresourcemanager/manager.go
```diff
@@ -0,0 +1,56 @@
...
+		args,
+	)
+	if err != nil {
+		return err
+	}
+	e.l.Info("Started workflow", zap.String("workflow_id", r.GetID()), zap.String("workflow_name", name))
+	return nil
+}
+
+func (e *manager) NotifyDone(ctx context.Context, id string) error {
```
itayd: do you expect this being run from an activity? if not maybe it should receive `workflow.Context` - not sure.

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -155,9 +155,10 @@ func (ws *workflows) StartChildWorkflow(wctx workflow.Context, data sessiondata.
 }
 
 func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session, opts StartWorkflowOptions) error {
+
 	sessionID := session.ID()
 
-	l := ws.l.Sugar().With("session_id", sessionID)
+	// l := ws.l.Sugar().With("session_id", sessionID)
```
itayd, line 158: ?

internal/backend/sessions/sessionworkflows/workflows.go
```diff
@@ -166,7 +167,7 @@ func (ws *workflows) StartWorkflow(ctx context.Context, session sdktypes.Session
 
 	memo := memo(session, data.OrgID)
 
-	r, err := ws.svcs.Temporal.TemporalClient().ExecuteWorkflow(
+	err = ws.svcs.WorkflowResourceManager.Execute(
```
itayd: I think default impl for oss should be queueless - just relay to temporal.

efiShtain: That's exactly what we have in this pr

internal/backend/sessions/sessionworkflows/workflows.go
efiShtain: Cleanups
It's normally ready
Just needed your approval on the idea

internal/backend/workflowresourcemanager/manager.go
efiShtain: I prefer not passing workflow context. It couples to temporal for no reason. I thought to get temporal out of this code and just have the resources management here, passing an interface to actually create execute the workflow. This way it doesn't even have to be temporal.

internal/backend/workflowresourcemanager/workflowmanager.go
efiShtain: Yes
And it does 
See in workflows code, the pr works, it calls this interface instead of temporal directly

internal/backend/workflowresourcemanager/workflowmanager.go
efiShtain: Not exactly, queue is an implementation detail on how to achieve the resource management. This is a utility to manage the allowed workload on the machine.

A simple optimization is to immediate run if we have enough capacity instead of going through the queue

internal/backend/workflowexecutor/enterprise_executor.go
```diff
@@ -0,0 +1,188 @@
+//go:build enterprise
```
itayd, line 1: i think you should still be able to run enterprise with no queuing if we'd like. for example - some named customers. this should be gated not only by build tag but also by a config param.

internal/backend/sessions/sessionworkflows/activities.go
```diff
@@ -328,3 +334,7 @@ func (ws *workflows) terminateWorkflowActivity(ctx context.Context, sid sdktypes
 
 	return err
 }
+
+func (ws *workflows) notifyWorkflowEndedActivity(ctx context.Context, sid sdktypes.SessionID) error {
+	return ws.svcs.WorkflowExecutor.NotifyDone(ctx, workflowID(sid))
```
itayd, line 360: does this supposed to take a meaningful amount of time? if not - maybe no need to run it in an activity.

efiShtain: should be very quick, but why not run it in activity just to make sure if it fail it will retry in a meaningful way ?

internal/backend/workflowexecutor/enterprise_executor.go
efiShtain: I can add a config throttling_enabled or disable_async_execute
but we said we can have just a large number and it should be fine

internal/backend/db/dbinterface_enterprise.go
```diff
@@ -15,4 +18,13 @@ type WorkflowExecutionRequest struct {
 
 type DB interface {
 	Shared
+	// WorkerInfo
+	GetWorkerInfo(ctx context.Context, id string) (scheme.WorkerInfo, error)
+	IncActiveWorkflows(ctx context.Context, workerID string) (int, error)
```
itayd: check if you can replace this by just counting the execution requests

internal/backend/workflowexecutor/executor.go
```diff
@@ -0,0 +1,58 @@
+//go:build !enterprise
```
itayd: rename to executor_not_enterprise

internal/backend/workflowexecutor/enterprise_executor.go
```diff
@@ -0,0 +1,188 @@
...
+
+	return e.svcs.DB.CreateWorkflowExecutionRequest(ctx, db.WorkflowExecutionRequest{
+		SessionID: sessionID,
+		Args:      args,
+		Memo:      memo,
+	})
+}
+
+func (e *executor) startPoller(ctx context.Context) {
+	ticker := time.NewTicker(100 * time.Millisecond)
```
itayd: comment why this does not make a busy loop if load is high.

internal/backend/workflowexecutor/enterprise_executor.go
```diff
@@ -0,0 +1,188 @@
...
+	maxConcurrent int
+	l             *zap.Logger
+	sampledLogger *zap.Logger
+
+	stopChannel chan struct{}
+
+	workerInfo scheme.WorkerInfo
+	cfg        *Config
+
+	executeLock *sync.Mutex
```
itayd: no need to make this ptr, and no need to initialize is

internal/backend/workflowexecutor/executor.go
efiShtain: done

internal/backend/workflowexecutor/enterprise_executor.go
efiShtain: switched to timer that reset after runOnce

internal/backend/workflowexecutor/enterprise_executor.go
efiShtain: done

internal/backend/db/dbinterface_enterprise.go
```diff
@@ -4,15 +4,24 @@
...
 }
 
 type DB interface {
 	Shared
+
+	// WorkflowExecutionRequest
+	CreateWorkflowExecutionRequest(ctx context.Context, request WorkflowExecutionRequest) error
+	GetWorkflowExecutionRequests(ctx context.Context, workerID string, maxRequests int) ([]WorkflowExecutionRequest, error)
+	UpdateRequestStatus(ctx context.Context, workflowID string, status string) error
+	CountInProgressWorkflowExecutionRequests(ctx context.Context, workerID string) (int64, error)
```
itayd, line 26: not for this pr, but it would be nice if we could see for each state - pending or acquired or done

itayd:
tests!

internal/backend/db/dbinterface_enterprise.go
efiShtain: done

efiShtain:
> tests!

please re check @itayd

itayd (approved):

PR #1299, opened 2025-05-21
feat: suggest adding modules to requirements.txt message

mario99logic:
Suggest adding modules to requirements.txt when no module is found.

refs: ENG-2127

### DIFF
--- a/runtimes/pythonrt/runner/main.py
+++ b/runtimes/pythonrt/runner/main.py
@@ -101,8 +101,9 @@ for more details.
 
 suggest_add_package = """
 =======================================================================================================
-The below error means you need to add a package to the environment.
-It's done by adding it to the requirements.txt file.
+The below error means you need to add a package to the Python environment.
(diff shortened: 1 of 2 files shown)


runtimes/pythonrt/runner/main.py
```diff
@@ -99,6 +99,15 @@ def filter_traceback(stack, user_code):
 """
 
 
+suggest_add_package = """
+=======================================================================================================
+The below error means you need to add a package to the environment.
+It's done by adding it to the requirements.txt file.
```
efiShtain: @itayd 

we currently don't support requirements.txt files on non production workload

what should we do with this error ?

efiShtain (approved):

runtimes/pythonrt/runner/main.py
mario99logic: @haimzlato and I changed the message to support both version for now.

PR #1303, opened 2025-05-26
feat: add k8s integration

mario99logic:
Add the Kubernetes integration

-  client and server
-  saves the JSON config file used to initialize the Kubernetes connection
- py-sdk: used Kubernetes library and added it to `requirements.txt` and return the standard k8s client

### DIFF
--- a/integrations/kubernetes/client.go
+++ b/integrations/kubernetes/client.go
@@ -1,0 +1,44 @@
+package kubernetes
+
+import (
+	"context"
+
+	"go.autokitteh.dev/autokitteh/integrations/common"
+	"go.autokitteh.dev/autokitteh/sdk/sdkintegrations"
+	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var (
+	desc          = common.Descriptor("kubernetes", "Kubernetes", "/static/images/k8s.svg")
+	configFileVar = sdktypes.NewSymbol("configFile")
+	authTypeVar   = sdktypes.NewSymbol("authType")
+)
+
+type integration struct{ vars sdkservices.Vars }
+
+func New(vars sdkservices.Vars) sdkservices.Integration {
+	i := &integration{vars: vars}
+	return sdkintegrations.NewIntegration(
+		desc,
+		sdkmodule.New(),
+		connStatus(i),
+		connTest(i),
+		sdkintegrations.WithConnectionConfigFromVars(vars),
+	)
+}
+
+// TODO: Implement the connection status functions for Kubernetes integration.
+func connStatus(_ *integration) sdkintegrations.OptFn {
+	return sdkintegrations.WithConnectionStatus(func(ctx context.Context, cid sdktypes.ConnectionID) (sdktypes.Status, error) {
+		return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
+	})
+}
+
+// TODO: Implement the connection test function for Kubernetes integration.
+func connTest(_ *integration) sdkintegrations.OptFn {
+	return sdkintegrations.WithConnectionTest(func(ctx context.Context, cid sdktypes.ConnectionID) (sdktypes.Status, error) {
+		return sdktypes.NewStatus(sdktypes.StatusCodeUnspecified, "Not implemented"), nil
+	})
+}
--- a/integrations/kubernetes/client.go
+++ b/integrations/kubernetes/client.go
@@ -3,19 +3,20 @@ package kubernetes
 import (
 	"context"
 
+	"go.uber.org/zap"
+
 	"go.autokitteh.dev/autokitteh/integrations"
 	"go.autokitteh.dev/autokitteh/integrations/common"
 	"go.autokitteh.dev/autokitteh/sdk/sdkintegrations"
 	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
-	"go.uber.org/zap"
 )
 
 var (
 	desc          = common.Descriptor("kubernetes", "Kubernetes", "/static/images/k8s.svg")
-	configFileVar = sdktypes.NewSymbol("configFile")
-	authTypeVar   = sdktypes.NewSymbol("authType")
+	configFileVar = sdktypes.NewSymbol("config_file")
+	authTypeVar   = sdktypes.NewSymbol("auth_type")
 )
 
 type integration struct{ vars sdkservices.Vars }
--- a/integrations/kubernetes/client.go
+++ b/integrations/kubernetes/client.go
@@ -3,11 +3,13 @@ package kubernetes
 import (
 	"context"
 
+	"go.autokitteh.dev/autokitteh/integrations"
 	"go.autokitteh.dev/autokitteh/integrations/common"
 	"go.autokitteh.dev/autokitteh/sdk/sdkintegrations"
 	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+	"go.uber.org/zap"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


integrations/kubernetes/save.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	// Read and parse POST request body.
+	if err := r.ParseForm(); err != nil {
+		l.Warn("Failed to parse incoming HTTP request", zap.Error(err))
+		c.AbortBadRequest("form parsing error")
+		return
+	}
+
+	configFile := r.Form.Get("configFile")
```
efiShtain: what do you expect to get here ?
Not sure config file is the only valid option we want to support here

integrations/kubernetes/save.go
```diff
@@ -0,0 +1,48 @@
...
+
+	c.Finalize(sdktypes.NewVars().
+		Set(configFileVar, configFile, true).
```
efiShtain, line 46: will this set the config file as secret ? 
cause it has to be a secret

integrations/kubernetes/save.go
```diff
@@ -0,0 +1,48 @@
...
+		Set(authTypeVar, integrations.Init, false))
```
efiShtain, line 47: what is auth type var ? 
why do you need it ?

integrations/kubernetes/server.go
```diff
@@ -0,0 +1,21 @@
...
+	// connections, after users submit them via a web form.
+	savePath = "/kubernetes/save"
+)
+
+func Start(l *zap.Logger, m *muxes.Muxes) {
+	// Init webhooks save connection vars (via "c.Finalize" calls), so they need
+	// to have an authenticated user context, so the DB layer won't reject them.
+	// For this purpose, init webhooks are managed by the "auth" mux, which passes
+	// through AutoKitteh's auth middleware to extract the user ID from a cookie.
+	m.Auth.Handle("POST "+savePath, NewHTTPHandler(l))
```
efiShtain, line 20: why do you create a new handler each time ?
it's the exact same thing for all requests

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
```diff
@@ -0,0 +1,43 @@
...
+        raise ConnectionInitError(connection)
+
+    parsed = json.loads(config_file)
+    with tempfile.NamedTemporaryFile(delete=False, mode="w") as temp_file:
+        json.dump(parsed, temp_file, indent=2)
+        temp_path = temp_file.name
+
+    try:
+        config.load_kube_config(config_file=temp_path)
+    except config.ConfigException as e:
```
efiShtain, line 35: add an except to make sure you catch non specific error just in case and return internal server error (and log the error so we know it)

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
```diff
@@ -0,0 +1,43 @@
...
+        raise ConnectionInitError(connection) from e
+    finally:
+        os.unlink(temp_path)
+
+    return client
```
efiShtain, line 42: what's client ?
it's not defined

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
```diff
@@ -0,0 +1,43 @@
...
+
+    """
+    check_connection_name(connection)
+
+    config_file = os.getenv(connection + "__configFile")
+
+    if not config_file:
+        raise ConnectionInitError(connection)
+
+    parsed = json.loads(config_file)
```
efiShtain: this can fail, need to try except it

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
```diff
@@ -0,0 +1,43 @@
...
+    with tempfile.NamedTemporaryFile(delete=False, mode="w") as temp_file:
+        json.dump(parsed, temp_file, indent=2)
+        temp_path = temp_file.name
```
efiShtain: why ?
just write the config file to a file and try initiating the client
no need to json loads and the json dumps

integrations/kubernetes/save.go
mario99logic: yes the parameter `true` makes it a secret

integrations/kubernetes/save.go
mario99logic: this is for the connection status in the UI, in other integrations we specify the type of the integration (for example OAuth or PAT...).
when its a single connection type which is the case (config file) then we use this.

integrations/kubernetes/server.go
mario99logic: Its like this for all the integrations each one needs a new handler, but Im not sure why need to ask @itayd

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
mario99logic: client is a module provided by the [kubernetes Python client library](https://github.com/kubernetes-client/python) 
we return it after the initialization so we cn allow the user to do:

client = kubernetes_client("k8s_conn")
pods = client.CoreV1Api().list_namespaced_pod("default")
....

integrations/kubernetes/save.go
mario99logic: other than the configuration file method what other options we can support and is it needed?

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
mario99logic: removed json loads and dump

integrations/kubernetes/save.go
efiShtain: it's kind of complicated, lots of ways to get a valid config.
let's start with that and see

integrations/kubernetes/save.go
efiShtain: oh, fun stuff
bad readability (not your fault I know)

integrations/kubernetes/server.go
efiShtain: my bad, read it incorrectly, that's fine, initialized once.

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
```diff
@@ -28,14 +28,13 @@ def kubernetes_client(connection: str):  # TODO: add type hints
...
-        temp_path = temp_file.name
-
     try:
+        with tempfile.NamedTemporaryFile(delete=False, mode="w") as temp_file:
+            temp_file.write(config_file)
+            temp_path = temp_file.name
+
         config.load_kube_config(config_file=temp_path)
-    except config.ConfigException as e:
+    except Exception as e:
```
efiShtain: I wanted you to add another except, not replace the previous one
catch the connection init error like from
but just in case something we don't anticipate happen, add another except to catch unknown errors

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
```diff
@@ -0,0 +1,46 @@
...
+
+    if not config_file:
+        raise ConnectionInitError(connection)
+
+    try:
+        with tempfile.NamedTemporaryFile(delete=False, mode="w") as temp_file:
+            temp_file.write(config_file)
+            temp_path = temp_file.name
+
+        config.load_kube_config(config_file=temp_path)
```
efiShtain: If you have the content already you can use load from dict instead if you creating temporary file

Worth Checking

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
```diff
@@ -0,0 +1,46 @@
...
+    except config.ConfigException as e:
+        raise ConnectionInitError(connection) from e
+    except Exception as e:
+        raise ConnectionInitError(connection) from e
```
efiShtain: We said internal server error here
And you have to log it well
It should never happen

integrations/kubernetes/client.go
```diff
@@ -0,0 +1,64 @@
...
+			zap.L().Error("failed to read connection vars", zap.String("connection_id", cid.String()), zap.Error(err))
+			return sdktypes.InvalidStatus, err
+		}
+
+		at := vs.Get(authTypeVar)
+		if !at.IsValid() || at.Value() == "" {
+			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
+		}
+
+		if at.Value() == integrations.Init {
```
efiShtain, line 52: Init == ok?

runtimes/pythonrt/py-sdk/autokitteh/kubernetes.py
```diff
@@ -0,0 +1,42 @@
...
+        raise ConnectionInitError(connection)
+
+    try:
+        config_dict = json.loads(config_file)
+        config.load_kube_config_from_dict(config_dict)
+    except config.ConfigException as e:
+        raise ConnectionInitError(connection) from e
+    except Exception as e:
+        raise RuntimeError(
+            f"Internal error while initializing Kubernetes client for connection '{connection}''"
```
efiShtain, line 39: You still don't log the error

If you catch the error and hide it we can't know what happened 

Is it printed in another place ? Is it handled in another place ?
Does a user initiating kube client  is expecting errors ?

integrations/kubernetes/client.go
mario99logic: no, at.value() is a string that should be equal "Initialized".
checkout other connections that have one connection type like chatgpt, if there is one connection type then its called "Initialized".
for other connection that have multiple connections have multiple types: "using OAuth", "PAT" etc.

efiShtain: integrations.Init == initialized ?

efiShtain (approved):

integrations/kubernetes/client.go
mario99logic: yes

PR #1306, opened 2025-05-28
fix goreleaser brew part

efiShtain:
adds ids section to brew in goreleaser so it uses only ak and not ak enterprise build


mario99logic (approved):
can you explain this change thanks

PR #1310, opened 2025-06-02
bug(python): Activity marker doesn't work on async

tebeka:
Add support for async.
- In AST, if function is called after an `await` then use `async_call`
- When back from activity, if function is a coroutine function - wrap the return value is a coroutine

Fixes: ENG-2139


itayd:
hey @tebeka can you describe why this fails and how you fixed it in the desc? thanks

itayd:
I guess this is not a draft anymore?

itayd (approved):

PR #1312, opened 2025-06-08
feat: add deployment statistics table to db 

mario99logic:
Added `deployment_session_stats` table for the purpose of optimizing `listDeploymentsWithStats`, 
optimized the query by getting the count of the finished sessions from the new table, instead of counting all sessions from the sessions table.

This will help in the long run for the enterprise db when the session table is big.

ref: ENG-2142


migrations/postgres/20250608132530_deployment_stats_view.sql
```diff
@@ -0,0 +1,21 @@
+-- +goose Up
+CREATE VIEW deployment_stats AS
```
efiShtain: the original query had more filters on it, 
make sure you don't break the previous handling

migrations/postgres/20250610115543_add_deployment_stats_table.sql
```diff
@@ -0,0 +1,12 @@
+-- +goose Up
+-- create "deployment_session_stats" table
+CREATE TABLE "deployment_session_stats" (
+  "deployment_id" uuid NOT NULL,
+  "session_state" bigint NOT NULL,
+  "count" bigint NOT NULL DEFAULT 0,
+  PRIMARY KEY ("deployment_id", "session_state")
```
itayd, line 7: should you create an index on deployment_id?

itayd:
i think @efiShtain should make the final call on this pr

internal/backend/db/dbgorm/deployments.go
```diff
@@ -112,21 +112,21 @@ func (gdb *gormdb) listDeploymentsCommonQuery(ctx context.Context, filter sdkser
...
-	COUNT(case when sessions.current_state_type = ? then 1 end) AS stopped
-	`, int32(sdktypes.SessionStateTypeCreated.ToProto()), // Note:
+        deployments.*,
+        -- Active sessions from sessions table (subqueries)
+        (SELECT COUNT(*) FROM sessions WHERE deployment_id = deployments.deployment_id AND current_state_type = ?) AS created,
+        (SELECT COUNT(*) FROM sessions WHERE deployment_id = deployments.deployment_id AND current_state_type = ?) AS running,
+        -- Finished sessions from stats table (subqueries) 
+        (SELECT COALESCE(count, 0) FROM deployment_session_stats WHERE deployment_id = deployments.deployment_id AND session_state = ?) AS completed,
+		(SELECT COALESCE(count, 0) FROM deployment_session_stats WHERE deployment_id = deployments.deployment_id AND session_state = ?) AS error,
+		(SELECT COALESCE(count, 0) FROM deployment_session_stats WHERE deployment_id = deployments.deployment_id AND session_state = ?) AS stopped
```
efiShtain: Can we do it more efficiently?

This is technically 3 sub queries, I think you can do it one query like it was before

internal/backend/db/dbgorm/sessions.go
```diff
@@ -38,8 +38,24 @@ func (gdb *gormdb) createSession(ctx context.Context, session *scheme.Session) e
...
+	err := tx.writer.WithContext(ctx).Where("session_id = ?", sessionID).First(&session).Error
+	if err != nil {
+		return err
+	}
+
+	// If finished session, increment deleted_count in deployment_session_stats.
+	if session.CurrentStateType > 2 && session.DeploymentID != nil {
+		err = tx.writer.Model(&scheme.DeploymentSessionStats{}).
+			Where("deployment_id = ? AND session_state = ?", *session.DeploymentID, session.CurrentStateType).
+			Update("count", gorm.Expr("deployment_session_stats.count - 1")).Error
```
efiShtain: Should we do max of count-1 and 0?

internal/backend/db/dbgorm/sessions.go
```diff
@@ -50,13 +66,44 @@ func (gdb *gormdb) updateSessionState(ctx context.Context, sessionID uuid.UUID,
...
+
 		if err := tx.writer.Model(&scheme.Session{SessionID: sessionID}).Updates(sessionStateUpdate).Error; err != nil {
 			return err
 		}
+
+		oldStateType := session.CurrentStateType
+		newStateType := int(state.Type().ToProto())
+		runningState := int(sdktypes.SessionStateTypeRunning.ToProto())
+
+		if oldStateType < 3 && newStateType != runningState && session.DeploymentID != nil {
```
efiShtain: Use the enum type instead of 3, magic numbers aren't good

We have running state as numbered enum I think

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -419,6 +419,12 @@ type Deployment struct {
 
 func (Deployment) IDFieldName() string { return "deployment_id" }
 
+type DeploymentSessionStats struct {
+	DeploymentID uuid.UUID `gorm:"primaryKey;type:uuid;not null;"`
+	SessionState int       `gorm:"primaryKey"`
```
efiShtain, line 423: Did you think about storing the actual string here? Like completed or faild?
Not sure if it's better, might be

internal/backend/db/dbgorm/sessions.go
```diff
@@ -50,13 +66,44 @@ func (gdb *gormdb) updateSessionState(ctx context.Context, sessionID uuid.UUID,
...
+			if err := gdb.incrementSessionCount(tx, *session.DeploymentID, newStateType); err != nil {
+				return err
+			}
+		}
+
 		return createLogRecord(tx.writer, ctx, logr, stateSessionLogRecordType)
 	})
 }
 
+func (gdb *gormdb) incrementSessionCount(tx *gormdb, deploymentID uuid.UUID, stateType int) error {
```
efiShtain: SessionStateCounter
It's not really session counter

Alternatively incDeploymentStats(sessionType) and a matching decDeploymentStats

internal/backend/db/dbgorm/sessions_test.go
```diff
@@ -19,6 +19,18 @@ func (f *dbFixture) createSessionsAndAssert(t *testing.T, sessions ...scheme.Ses
 	for _, session := range sessions {
 		assert.NoError(t, f.gormdb.createSession(f.ctx, &session))
 		findAndAssertOne(t, f, session, "session_id = ?", session.SessionID)
+
+		// Manually increment count for sessions created directly in completed state.
+		// Normal sessions go through running->completed transition and auto-increment count.
+		// Test sessions skip the running state, so we need to increment count here.
+		if session.CurrentStateType > 2 && session.DeploymentID != nil {
```
efiShtain: No magic number please

migrations/postgres/20250610115543_add_deployment_stats_table.sql
efiShtain: There is a primary key on it so it's indexed

mario99logic: There is a primary key on it so it's indexed

internal/backend/db/dbgorm/sessions.go
mario99logic: is there a use case where we can go under 0, I don't think its possible

internal/backend/db/dbgorm/scheme/records.go
mario99logic: I think keep it an int like in session and deployment table. 
If you disagree I will change it no problem

internal/backend/db/dbgorm/deployments.go
mario99logic: I changed to one query

internal/backend/db/dbgorm/sessions.go
```diff
@@ -38,8 +38,20 @@ func (gdb *gormdb) createSession(ctx context.Context, session *scheme.Session) e
...
-func (gdb *gormdb) deleteSession(ctx context.Context, sessionID uuid.UUID) error {
-	return gdb.writer.WithContext(ctx).Delete(&scheme.Session{SessionID: sessionID}).Error
+func (tx *gormdb) deleteSession(ctx context.Context, sessionID uuid.UUID) error {
+	var session scheme.Session
+	err := tx.writer.WithContext(ctx).Where("session_id = ?", sessionID).First(&session).Error
+	if err != nil {
+		return err
+	}
+
+	// If finished session, increment deleted_count in deployment_session_stats.
```
efiShtain: the comment is irrelevant i think

internal/backend/db/dbgorm/sessions.go
```diff
@@ -50,9 +62,25 @@ func (gdb *gormdb) updateSessionState(ctx context.Context, sessionID uuid.UUID,
...
+
 		if err := tx.writer.Model(&scheme.Session{SessionID: sessionID}).Updates(sessionStateUpdate).Error; err != nil {
 			return err
 		}
+
+		oldStateType := session.CurrentStateType
+		newStateType := int(state.Type().ToProto())
+		runningState := int(sdktypes.SessionStateTypeRunning.ToProto())
+
+		if oldStateType <= runningState && newStateType != runningState && session.DeploymentID != nil {
```
efiShtain: to be on the safe size, maybe newState > runningState ?

internal/backend/db/dbgorm/sessions.go
```diff
@@ -200,7 +228,9 @@ func (db *gormdb) CreateSession(ctx context.Context, session sdktypes.Session) e
 }
 
 func (db *gormdb) DeleteSession(ctx context.Context, sessionID sdktypes.SessionID) error {
-	return translateError(db.deleteSession(ctx, sessionID.UUIDValue()))
+	return db.writeTransaction(ctx, func(tx *gormdb) error {
```
efiShtain: why do you wrap it here in transaction ?

migrations/postgres/enterprise/20250610151931_add_session_partial_index.sql
```diff
@@ -0,0 +1,15 @@
+-- +goose Up
+-- drop index "idx_sessions_current_state_type" from table: "sessions"
+DROP INDEX "idx_sessions_current_state_type";
+-- drop index "idx_sessions_deployment_id" from table: "sessions"
+DROP INDEX "idx_sessions_deployment_id";
+-- create index "idx_active_sessions" to table: "sessions"
+CREATE INDEX "idx_active_sessions" ON "sessions" ("deployment_id", "current_state_type") WHERE ((current_state_type = 1) OR (current_state_type = 2));
+
```
efiShtain: now sure here

no one uses the idx_sessions_current_state_type index ?

efiShtain (approved):

PR #1313, opened 2025-06-10
fix(ENG-2135): allow to export variables contents

itayd:
by default does not export vars. only on lint and when explicitly requested


cmd/ak/cmd/projects/export.go
```diff
@@ -20,10 +20,14 @@ var exportCmd = common.StandardCommand(&cobra.Command{
 	RunE: export,
 })
 
-var outputFileName string
+var (
+	outputFileName      string
+	IncludeVarsContents bool
```
efiShtain: why is this exported ?

internal/backend/projects/projects.go
```diff
@@ -317,6 +317,10 @@ func (ps *Projects) exportManifest(ctx context.Context, projectID sdktypes.Proje
 			Secret: v.IsSecret(),
 		}
 
+		if includeVarsContents {
```
efiShtain, line 320:
```suggestion
		if includeVarsContent {
```

efiShtain (approved):

cmd/ak/cmd/projects/export.go
itayd: fixed

internal/backend/projects/projects.go
itayd: diagree

PR #1318, opened 2025-06-15
ENG-2143 fix: add missing default org id on first login

efiShtain:
when user logs it initially to they system, a default org id is created for it, but we don't return it in create user method so the user object in the session is partial
this change fetches the created user from db after create making sure it has everything.
on subsequent calls the full user object is fetched anyway so the issue doesn't occur again

### DIFF
--- a/internal/backend/auth/authloginhttpsvc/svc.go
+++ b/internal/backend/auth/authloginhttpsvc/svc.go
@@ -273,7 +273,11 @@ func (a *svc) newSuccessLoginHandler(ctx context.Context, ld *loginData) http.Ha
 			return newErrHandler("internal server error", http.StatusInternalServerError)
 		}
 
-		u = u.WithID(uid)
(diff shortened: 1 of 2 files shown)


internal/backend/auth/authloginhttpsvc/svc.go
```diff
@@ -273,7 +273,11 @@ func (a *svc) newSuccessLoginHandler(ctx context.Context, ld *loginData) http.Ha
 			return newErrHandler("internal server error", http.StatusInternalServerError)
 		}
 
-		u = u.WithID(uid)
+		u, err = a.Users.Get(authcontext.SetAuthnSystemUser(ctx), uid, "")
+		if err != nil && !errors.Is(err, sdkerrors.ErrNotFound) {
```
itayd: remove check on not found

itayd (approved):

PR #1320, opened 2025-06-15
feat: values

itayd:
This adds a basic kv store to autokitteh scripts in starlark and python. 
A new service in backend: "store", which exposes a Mutate and List methods. Added to python sdk syscalls that call to the user_code service, which in turn uses new Callbacks by the session to access the store.

- Mutate(key, op, operands...): makes operation op on key using the operands. Example: k1 set 1 will set k1 to 1. 

Scope of store is project. Different projects cannot access each others values.

TODO: We need to discuss how to limit access to these things.


internal/backend/db/db.go
```diff
@@ -136,9 +136,9 @@ type Shared interface {
 	GetSecret(ctx context.Context, key string) (string, error)
 	DeleteSecret(ctx context.Context, key string) error
 
-	SetValue(ctx context.Context, pid sdktypes.ProjectID, key string, v sdktypes.Value) error
-	GetValue(ctx context.Context, pid sdktypes.ProjectID, key string) (sdktypes.Value, error)
-	ListValues(ctx context.Context, pid sdktypes.ProjectID) (map[string]sdktypes.Value, error)
+	SetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key string, v sdktypes.Value) error
+	GetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key string) (sdktypes.Value, error)
+	ListStoreValues(ctx context.Context, pid sdktypes.ProjectID, keys []string, getValues bool) (map[string]sdktypes.Value, error)
```
tebeka, line 144: What's the purpose of `getValues`?

internal/backend/db/dbgorm/store.go
```diff
@@ -13,22 +15,30 @@ import (
...
-func (db *gormdb) SetValue(ctx context.Context, pid sdktypes.ProjectID, key string, v sdktypes.Value) error {
+func (db *gormdb) SetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key string, v sdktypes.Value) error {
 	if !pid.IsValid() {
 		return sdkerrors.NewInvalidArgumentError("invalid project id")
 	}
 
+	q := db.writer.WithContext(ctx)
+
+	if !v.IsValid() {
+		return translateError(
```
tebeka, line 25: I don't like DB access inside `translateError`, call it outside.

internal/backend/db/dbgorm/store.go
```diff
@@ -57,13 +71,27 @@ func (db *gormdb) GetValue(ctx context.Context, pid sdktypes.ProjectID, key stri
...
 }
 
-func (db *gormdb) ListValues(ctx context.Context, pid sdktypes.ProjectID) (map[string]sdktypes.Value, error) {
-	var rs []*scheme.Value
-	if err := db.reader.WithContext(ctx).Where("project_id = ?", pid.UUIDValue()).Find(&rs).Error; err != nil {
+func (db *gormdb) ListStoreValues(ctx context.Context, pid sdktypes.ProjectID, keys []string, getValues bool) (map[string]sdktypes.Value, error) {
+	var rs []*scheme.StoreValue
+	q := db.reader.WithContext(ctx).Where("project_id = ?", pid.UUIDValue())
+
+	if len(keys) > 0 {
```
tebeka, line 75: I think empty keys means "everything" is not that intuitive, should be documented.

internal/backend/store/store.go
```diff
@@ -0,0 +1,91 @@
...
+type store struct {
+	db db.DB
+	l  *zap.Logger
+}
+
+func New(db db.DB, l *zap.Logger) sdkservices.Store {
+	return &store{db: db, l: l}
+}
+
+func (s *store) Do(ctx context.Context, pid sdktypes.ProjectID, key, op string, operands ...sdktypes.Value) (sdktypes.Value, error) {
```
tebeka: Rename to `Mutate`?

proto/autokitteh/store/v1/svc.proto
```diff
@@ -6,10 +6,22 @@ import "autokitteh/values/v1/values.proto";
 import "buf/validate/validate.proto";
 
 service StoreService {
+  rpc Do(DoRequest) returns (DoResponse);
   rpc Get(GetRequest) returns (GetResponse);
   rpc List(ListRequest) returns (ListResponse);
 }
 
+message DoRequest {
```
tebeka: I don't think `Do` should be in the API. There should be `Set`, `Update` & `Delete`

migrations/postgres/enterprise/20250616125426_values.sql
```diff
@@ -0,0 +1,32 @@
+-- +goose Up
+-- create "store_values" table
+CREATE TABLE "store_values" (
```
tebeka, line 3: Do all `ak` share the same database?

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
```diff
@@ -6,18 +6,24 @@
...
-    "Event",
-    "Signal",
     "activity",
+    "AttrDict",
+    "del_value",
     "errors",
+    "Event",
+    "get_value",
     "inhibit_activities",
+    "list_values",
```
tebeka: Create sub groups for each intent and sort in the sub group. e.g.

```python
    # Events
    "next_event",
    "subscribe",

     # Store
     "del_value",
     "get_value",
     "list_values",
     ...

File: runtimes/pythonrt/py-sdk/autokitteh/store.py [hunk 8]
```diff
@@ -0,0 +1,38 @@
...
+    }[op](key, *args)
+
+
+def get_value(key: str) -> any:
+    """Get a stored value."""
+    # Dummy implementation for local development.
+    return _local_dev_store.get(key)
+
+
+def set_value(key: str, value: any) -> None:
```
tebeka: Values should be `bytes`

runtimes/pythonrt/py-sdk/autokitteh/store.py
```diff
@@ -0,0 +1,38 @@
...
+    """Mutate a stored value."""
+    # Dummy implementation for local development.
+    return {
+        "set": set_value,
+        "get": get_value,
+        "del": del_value,
+    }[op](key, *args)
+
+
+def get_value(key: str) -> any:
```
tebeka: What's `any`?

runtimes/pythonrt/runner/main.py
```diff
@@ -305,8 +305,13 @@ def patch_ak_funcs(self):
         connections.encode_jwt = self.syscalls.ak_encode_jwt
         connections.refresh_oauth = self.syscalls.ak_refresh_oauth
 
+        autokitteh.del_value = self.syscalls.ak_del_value
```
tebeka, line 308: I'm starting to think we need some registry mechanism in `syscalls` to register function that need to be patched.

tebeka (changes requested):
I'd love to see design doc. Most importantly about limits - which I don't see any reference to them.

Also, why not https://github.com/temporalio/samples-go/blob/main/query/query_workflow.go ?

proto/autokitteh/store/v1/svc.proto
itayd: Disagree. It will lead to a lot of calls and a lot of boilerplate. Everything can be modelled in Do.

migrations/postgres/enterprise/20250616125426_values.sql
itayd: huh? well it's the ak db.

runtimes/pythonrt/py-sdk/autokitteh/store.py
itayd: why? you can put any value here. it'll be wrapped below.

runtimes/pythonrt/py-sdk/autokitteh/store.py
itayd: because it can return any value (unwrapped).

runtimes/pythonrt/runner/main.py
itayd: it used to be called "syscalls" :)

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
itayd: done

internal/backend/db/dbgorm/store.go
itayd: done

internal/backend/db/db.go
itayd: huh?

internal/backend/db/dbgorm/store.go
itayd: i kinda do it all over the place, i think it's ok

itayd: fixed anyway

internal/backend/store/store.go
itayd: I like it

itayd: done

internal/backend/db/db.go
```diff
@@ -136,9 +136,11 @@ type Shared interface {
...
 	DeleteSecret(ctx context.Context, key string) error
 
-	SetValue(ctx context.Context, pid sdktypes.ProjectID, key string, v sdktypes.Value) error
-	GetValue(ctx context.Context, pid sdktypes.ProjectID, key string) (sdktypes.Value, error)
-	ListValues(ctx context.Context, pid sdktypes.ProjectID) (map[string]sdktypes.Value, error)
+	SetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key string, v sdktypes.Value) error
+	GetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key string) (sdktypes.Value, error)
+
+	// If len(keys) == 0, it returns all keys.
+	ListStoreValues(ctx context.Context, pid sdktypes.ProjectID, keys []string, getValues bool) (map[string]sdktypes.Value, error)
```
efiShtain, line 144: What is getValues bool?
You either list values or list keys ?

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -528,10 +528,10 @@ func ParseSignal(r *Signal) (*types.Signal, error) {
 	}, nil
 }
 
-type Value struct {
+type StoreValue struct {
```
efiShtain, line 531: Why not make this whole feature enterprise only ?

internal/backend/db/dbgorm/store.go
```diff
@@ -0,0 +1,108 @@
...
+	"google.golang.org/protobuf/proto"
+	"gorm.io/gorm"
+
+	"go.autokitteh.dev/autokitteh/internal/backend/db/dbgorm/scheme"
+	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+const maxValueSize = 64 * 1024
```
efiShtain, line 15: Is this in kb?

internal/backend/db/dbgorm/store.go
```diff
@@ -0,0 +1,108 @@
...
+		return translateError(
+			q.Where("project_id = ? AND key = ?", pid.UUIDValue(), key).Delete(&scheme.StoreValue{}).Error,
+		)
+	}
+
+	if v.ProtoSize() > maxValueSize {
+		return sdkerrors.NewInvalidArgumentError("value too large > %d bytes", maxValueSize)
+	}
+
+	bs, err := proto.Marshal(v.ToProto())
```
efiShtain, line 34: Why the db is even aware of protos in this case?

internal/backend/db/dbgorm/store.go
```diff
@@ -0,0 +1,108 @@
...
+	if err != nil {
+		return err
+	}
+
+	err = q.Save(&scheme.StoreValue{
+		// This does not take into account if value was also previously set,
```
efiShtain, line 40: If you use model.updates you can set updated_at and updated_by in the same query

internal/backend/db/dbgorm/store.go
```diff
@@ -0,0 +1,108 @@
...
+		ProjectID: pid.UUIDValue(),
+		Key:       key,
+		Value:     bs,
+	}).Error
+
+	return translateError(err)
+}
+
+func (db *gormdb) GetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key string) (sdktypes.Value, error) {
+	r, err := getOne[scheme.StoreValue](db.reader.WithContext(ctx), "project_id = ? AND key = ?", pid.UUIDValue(), key)
```
efiShtain: Just call the list values to reduce duplicate code

sdk/sdkservices/runtimes.go
```diff
@@ -67,6 +67,10 @@ type RunCallbacks struct {
 	// Signals
 	Signal     func(ctx context.Context, rid sdktypes.RunID, sid sdktypes.SessionID, name string, payload sdktypes.Value) error
 	NextSignal func(ctx context.Context, rid sdktypes.RunID, names []string, timeout time.Duration) (*RunSignal, error)
+
+	// Store
+	ListStoreValues  func(ctx context.Context, rid sdktypes.RunID) ([]string, error)
+	MutateStoreValue func(ctx context.Context, rid sdktypes.RunID, key, op string, operands ...sdktypes.Value) (sdktypes.Value, error)
```
efiShtain, line 73: intentional diff between db layer ? (mutate vs set store value)

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -537,3 +537,48 @@ func (s *workerGRPCHandler) NextSignal(ctx context.Context, req *userCode.NextSi
 		},
 	}, nil
 }
+
+func (s *workerGRPCHandler) StoreList(ctx context.Context, req *userCode.StoreListRequest) (*userCode.StoreListResponse, error) {
```
efiShtain, line 541: any reason this does not match the ListStoreValues callback ?

runtimes/pythonrt/runner/syscalls.py
```diff
@@ -193,6 +193,32 @@ def ak_next_signal(
 
         return None
 
+    def ak_mutate_value(self, key: str, op: str, *args: list[str]) -> any:
+        self.log.info("ak_mutate_value: %r %r %r", key, op, args)
```
efiShtain: debug

efiShtain: also value can be secret, do we care?

runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -537,3 +537,48 @@ func (s *workerGRPCHandler) NextSignal(ctx context.Context, req *userCode.NextSi
...
+		return &userCode.StoreListResponse{Error: err.Error()}, nil
+	}
+
+	return &userCode.StoreListResponse{
+		Keys: resp.value.([]string),
+	}, nil
+}
+
+func (s *workerGRPCHandler) StoreMutate(ctx context.Context, req *userCode.StoreMutateRequest) (*userCode.StoreMutateResponse, error) {
+	operands, err := kittehs.TransformError(req.Operands, sdktypes.ValueFromProto)
```
efiShtain, line 562: maybe we should limit set/update size here ? instead of the db layer ? 
we can use grpc max size like in resources

proto/autokitteh/store/v1/svc.proto
```diff
@@ -6,10 +6,22 @@ import "autokitteh/values/v1/values.proto";
...
+  rpc Mutate(MutateRequest) returns (MutateResponse);
   rpc Get(GetRequest) returns (GetResponse);
   rpc List(ListRequest) returns (ListResponse);
 }
 
+message MutateRequest {
+  string project_id = 1;
+  string key = 2 [(buf.validate.field).string.min_len = 1];
+  string operation = 3 [(buf.validate.field).string.min_len = 1];
+  repeated values.v1.Value operands = 4 [(buf.validate.field).repeated.items.required = true];
```
efiShtain, line 18: what is operands ? 
wouldn't it be better to be concise like DelKey, SetKey, GetKey, ListKeys - that's it ?

migrations/postgres/enterprise/20250616125426_values.sql
```diff
@@ -0,0 +1,32 @@
...
+  "created_by" uuid NULL,
+  "created_at" timestamptz NULL,
+  "project_id" uuid NOT NULL,
+  "key" text NOT NULL,
+  "value" bytea NULL,
+  "updated_by" uuid NULL,
+  "updated_at" timestamptz NULL,
+  PRIMARY KEY ("project_id", "key"),
+  CONSTRAINT "fk_store_values_project" FOREIGN KEY ("project_id") REFERENCES "projects" ("project_id") ON UPDATE NO ACTION ON DELETE NO ACTION
```
efiShtain, line 12: maybe on delete cascade it better here

migrations/postgres/enterprise/20250616125426_values.sql
```diff
@@ -0,0 +1,32 @@
+-- +goose Up
+-- create "store_values" table
+CREATE TABLE "store_values" (
+  "created_by" uuid NULL,
```
efiShtain, line 4: technically how do you know who creates it ?
workflows that starts by webhooks are not "user" oriented

efiShtain (commented):
left some comments, 

can you explain why we don't just use the current "vars" table with "project_id" scope and that's it ? 

also, not sure why we have "mutate" function vs just set / get function

internal/backend/store/values.go
```diff
@@ -0,0 +1,42 @@
...
+var ops = map[string]func(sdktypes.Value, []sdktypes.Value) (set sdktypes.Value, ret sdktypes.Value, _ error){
+	"get": func(v sdktypes.Value, vs []sdktypes.Value) (sdktypes.Value, sdktypes.Value, error) {
+		if len(vs) > 0 {
+			return sdktypes.InvalidValue, sdktypes.InvalidValue, errTooManyOperands
+		}
+		return sdktypes.InvalidValue, v, nil
+	},
+	"set": func(_ sdktypes.Value, vs []sdktypes.Value) (sdktypes.Value, sdktypes.Value, error) {
+		if len(vs) != 1 {
+			return sdktypes.InvalidValue, sdktypes.InvalidValue, errTooManyOperands
```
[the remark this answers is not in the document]
tebeka: What Copilot said :)

internal/backend/db/db.go
tebeka: The `getValues` parameter to `ListStoreValues`

internal/backend/db/dbgorm/scheme/tables.go
```diff
@@ -16,6 +16,6 @@ var Tables = []any{
 	&Signal{},
 	&Trigger{},
 	&User{},
-	&Value{},
+	&StoreValue{},
```
tebeka, line 19: Do we want to store customer values with our own database? Maybe a dedicatd database for values? Need to discuss.

internal/backend/store/store_test.go
```diff
@@ -0,0 +1,160 @@
...
+	ivs = []sdktypes.Value{
+		sdktypes.NewIntegerValue(0),
+		sdktypes.NewIntegerValue(1),
+		sdktypes.NewIntegerValue(2),
+	}
+)
+
+func TestMain(m *testing.M) {
+	authz.DisableCheckForTesting()
+	m.Run()
```
tebeka: You need to os.Exit the result of m.Run

migrations/postgres/enterprise/20250616125426_values.sql
tebeka: Still, we might decide run several `ak` maybe each with its own db. @efiShtain ?

itayd:
> can you explain why we don't just use the current "vars" table with "project_id" scope and that's it ?

Vars are immutable and do not store state between projects.

> 
> also, not sure why we have "mutate" function vs just set / get function

So we can implement "inc", "push", etc in the future

runtimes/pythonrt/py-sdk/autokitteh/store.py
tebeka: I mean `any` is a [built-in function](https://docs.python.org/3/library/functions.html#any) not the "any" type. There's `typing.Any`

internal/backend/db/db.go
itayd: it specifies if to fill in the values or just return a set of keys

internal/backend/db/dbgorm/scheme/records.go
itayd: i don't think this should be oss only, i'd defer to @haimzlato on this

runtimes/pythonrt/py-sdk/autokitteh/store.py
tebeka: OK.

Again misus of `any` :)

internal/backend/db/dbgorm/store.go
itayd: yes

itayd: i mean 64kb

internal/backend/db/dbgorm/store.go
itayd: its just a way to store the value. proto is more efficient than json.

internal/backend/db/dbgorm/store.go
itayd: not sure i follow? we don't use gorm.Model.

sdk/sdkservices/runtimes.go
itayd: i don't follow?

runtimes/pythonrt/py-sdk/autokitteh/store.py
```diff
@@ -0,0 +1,38 @@
...
+def del_value(key: str) -> None:
+    """Delete a stored value."""
+    # Dummy implementation for local development.
+    try:
+        del _local_dev_store[key]
+    except KeyError:
+        pass
+
+
+def list_values() -> list[str]:
```
tebeka: This should return a map of key->value of a list of tuples (key, value)

sdk/sdkservices/runtimes.go
itayd: oh, the mutate does a transaction underneath that includes get and set. this makes it more generic for other options as well.

proto/autokitteh/store/v1/svc.proto
itayd: nope - see pr description. i want to support more operations in the future.

runtimes/pythonrt/py-sdk/autokitteh/store.py
```diff
@@ -0,0 +1,38 @@
+_local_dev_store = {}
+
+
+def mutate_value(key: str, op: str, *args: list[any]) -> any:
```
tebeka: Thinking about this. IMO the right thing will be to have a `Store` class that implement `MutableMapping`. This way the user will have a familiar dict like object to work with.

runtimes/pythonrt/runner/syscalls.py
```diff
@@ -193,6 +193,32 @@ def ak_next_signal(
...
+
+    def ak_get_value(self, key: str) -> any:
+        return self.ak_mutate_value(key, "get")
+
+    def ak_del_value(self, key: str) -> any:
+        return self.ak_mutate_value(key, "del")
+
+    def ak_list_values(self) -> list[str]:
+        self.log.info("ak_list_values")
+        req = pb.StoreListRequest(runner_id=self.runner_id)
```
tebeka, line 218: I read `StoreListRequest` and a request to store a list. Maybe we need better naming.

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
```diff
@@ -6,19 +6,29 @@
...
     "unsubscribe",
+    # Signals
+    "next_signal",
+    "signal",
+    "Signal",
+    # Values
+    "del_value",
+    "get_value",
+    "list_values",
+    "mutate_value",
```
tebeka, line 32: `mutate_value` should not be part of the public API

internal/backend/db/dbgorm/store.go
itayd: done

runtimes/pythonrt/worker_grpc_handler.go
itayd: ? this is the grpc handler.

runtimes/pythonrt/runner/syscalls.py
itayd: correct - changed to debug.

tests/sessions/python/store.txtar
tebeka, line 1: This doesn't really tests the store. You need to have two activities each using the store at least. The better solution is to re-run and see you get the values.

tebeka (changes requested):

runtimes/pythonrt/runner/syscalls.py
itayd: not sure about secrets - should we remove the log?

runtimes/pythonrt/worker_grpc_handler.go
itayd: agree, will do in future pr

migrations/postgres/enterprise/20250616125426_values.sql
itayd: this is for when someone does it via rpc.

itayd: (for instance, in the future users will be able to see this vars in ui)

migrations/postgres/enterprise/20250616125426_values.sql
itayd: not sure i fllow

internal/backend/db/db.go
itayd: oh - replied below to efi. will add comment.

migrations/postgres/enterprise/20250616125426_values.sql
itayd: i don't follow - each ak will have its own db per config.

runtimes/pythonrt/py-sdk/autokitteh/store.py
itayd: i don't think it's a misuse - it can return anything?

runtimes/pythonrt/py-sdk/autokitteh/store.py
itayd: ooooooohhhhhhhh ok fixed

internal/backend/db/dbgorm/scheme/tables.go
itayd: i don't think we need to go overboard here, at least not now

internal/backend/store/store_test.go
itayd: done

runtimes/pythonrt/py-sdk/autokitteh/store.py
itayd: this returns only keys. i will rename the function.

runtimes/pythonrt/py-sdk/autokitteh/store.py
itayd: interesting - but i'd rather do it in a new pr

runtimes/pythonrt/runner/syscalls.py
itayd: ack

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
itayd: it is?

tests/sessions/python/store.txtar
itayd: not sure i follow, in any case i have a unit test as well

proto/autokitteh/store/v1/svc.proto
efiShtain: I agree with @tebeka

migrations/postgres/enterprise/20250616125426_values.sql
efiShtain: @tebeka what do you mean ? 
SaaS has one db, all instances share same db (with replicas and such, but one)

internal/backend/db/db.go
efiShtain: why do we need it ? 
just return everything and keep the API simpler

internal/backend/db/dbgorm/store.go
efiShtain: can you add KB to the name ? 
maxValueSizeKB

internal/backend/db/dbgorm/store.go
efiShtain: let the db store jsonb and handle how to store it
if you want to can switch to bytes all the way
I don't think the db should be aware of the transport

internal/backend/db/dbgorm/store.go
efiShtain: see this as an example (the gorm.Expr part)
```go
db.Model(&scheme.StoreValue{}).Update("updated_at", gorm.Expr("NOW()"))
```

or this
```go
if session.CurrentStateType > runningState && session.DeploymentID != nil {
		return tx.writer.WithContext(ctx).Model(&scheme.DeploymentSessionStats{}).
			Where("deployment_id = ? AND session_state = ?", *session.DeploymentID, session.CurrentStateType).
			Update("count", gorm.Expr("deployment_session_stats.count - 1")).Error
	}
```

runtimes/pythonrt/runner/syscalls.py
efiShtain: not sure we really need the log, we can just log in AK

runtimes/pythonrt/worker_grpc_handler.go
efiShtain: open ticket and add todo please

proto/autokitteh/store/v1/svc.proto
efiShtain: I think its future proofing something not sure we need and the api is now less readable
if you insist then fine

migrations/postgres/enterprise/20250616125426_values.sql
efiShtain: change the delete to cascade so if project is removed the values are removed as well automatically

migrations/postgres/enterprise/20250616125426_values.sql
efiShtain: again, future proofing, something we don't support in the UI and won't ever be supported in the workflows, I think you should remove it

internal/backend/db/db.go
itayd: might be too big. i want to enable the user to get just the list to see what's there.

runtimes/pythonrt/runner/syscalls.py
itayd: @tebeka i just copied your stuff - can we remove the log?

proto/autokitteh/store/v1/svc.proto
itayd: strong disagree. we should discuss.

efiShtain (approved):

migrations/postgres/enterprise/20250616125426_values.sql
tebeka: OK then

runtimes/pythonrt/runner/syscalls.py
tebeka: Aye

runtimes/pythonrt/py-sdk/autokitteh/__init__.py
tebeka: "should not", my bad

PR #1321, opened 2025-06-15
fix: make linter work again

itayd:


tebeka:
@itayd Did you try `golangci-lint migrate` and continue from there? I'd do that and then tell claude to fix the errors.

tebeka (changes requested):
Lint still failing

tebeka (approved):

PR #1322, opened 2025-06-16
feat: add Anthropic integration 

mario99logic:
Add Anthropic connection with API key


itayd:
lgtm, why draft?

mario99logic:
> lgtm, why draft?

Wanted to test it first

itayd (approved):

integrations/anthropic/save.go
```diff
@@ -0,0 +1,67 @@
...
+		l.Warn("save connection: invalid connection ID", zap.Error(err))
+		c.AbortBadRequest("invalid connection ID")
+		return
+	}
+
+	vsid := sdktypes.NewVarScopeID(cid)
+	common.SaveAuthType(r, h.vars, vsid)
+
+	apiKey := r.FormValue("api_key")
+	if apiKey == "" {
```
efiShtain, line 56: Maybe first thing check the request is valid ?
If there is no api key no real reason for the rest of the checks

integrations/anthropic/server.go
```diff
@@ -0,0 +1,22 @@
...
+	// connections, after users submit them via a web form.
+	savePath = "/anthropic/save"
+)
+
+func Start(l *zap.Logger, muxes *muxes.Muxes, v sdkservices.Vars) {
+	// Init webhooks save connection vars (via "c.Finalize" calls), so they need
+	// to have an authenticated user context, so the DB layer won't reject them.
+	// For this purpose, init webhooks are managed by the "auth" mux, which passes
+	// through AutoKitteh's auth middleware to extract the user ID from a cookie.
+	muxes.Auth.Handle("POST "+savePath, NewHTTPHandler(l, v))
```
efiShtain, line 21: We always put routes in start method? Cause we don't really start anything here

runtimes/pythonrt/py-sdk/docs/requirements.txt
```diff
@@ -1,6 +1,7 @@
 # Generated by gen-reqs.py, DO NOT EDIT
 # This is an export of ../pyproject.toml for builds in readthedocs.
 
+anthropic
```
efiShtain: Pin a version please

efiShtain (approved):

integrations/anthropic/save.go
mario99logic: in the beginning of the function I check the content type of the request, read and parse the POST request body and do a sanity check, do I need to do another check to see if the request is valid?

integrations/anthropic/server.go
mario99logic: yes its like this in every integration.

PR #1325, opened 2025-06-19
feat: support oss and enterprise system tests

efiShtain:


.github/workflows/go.yml
```diff
@@ -5,84 +5,94 @@
...
+            - name: Check out code
+              uses: actions/checkout@v4
+            - name: Set up Go
+              uses: actions/setup-go@v5
+              with:
+                  go-version-file: go.mod
+            - name: Download Go modules
+              run: go mod download
+            - name: Install gotestsum
+              run: go install gotest.tools/gotestsum@latest
```
tebeka, line 29: This is a tool in go.mod already. Run `go tool gotestssum` instead of just `gotestsum` and you should be good to go.

tests/system/ak_ee_test.go
```diff
@@ -0,0 +1,67 @@
+//go:build enterprise
+// +build enterprise
```
tebeka, line 2: Remove this, we don't support old Go versions that use `+build`

tests/system/ak_ee_test.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	t.Cleanup(func() {
+		if err := pgContainer.Terminate(context.Background()); err != nil {
+			t.Fatalf("failed to terminate pgContainer: %s", err)
+		}
+	})
+	connStr, err := pgContainer.ConnectionString(t.Context(), "sslmode=disable")
+	cfg := map[string]any{}
+	// cfg = maps.Clone(test.config.Server)
```
tebeka: Delete commented out code.

tests/system/ak_oss_test.go
```diff
@@ -0,0 +1,36 @@
+//go:build !enterprise
+// +build !enterprise
```
tebeka, line 2: Delete

tests/system/ak_oss_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+/*
```
tebeka, line 4: I don't find this comment that useful (AI generated?)

tests/system/ak_oss_test.go
efiShtain: that's not comment, it's build tags

tests/system/ak_ee_test.go
efiShtain: I didn't add it, the formatter added it automatically

.github/workflows/go.yml
efiShtain: didn't change what we had before
we run it using make
doesn't really matter

tests/system/ak_oss_test.go
efiShtain: it's not a new comment, was there before, I just renamed the file so git blames me

I tend to agree but I don't mind leaving it there

itayd (approved):

PR #1326, opened 2025-06-21
feat(values): limit number of values per project and max value size refactoring

itayd:


internal/backend/db/dbgorm/store.go
```diff
@@ -68,6 +62,34 @@ func (db *gormdb) GetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key
...
 }
 
+func (db *gormdb) HasStoreKey(ctx context.Context, pid sdktypes.ProjectID, key string) (bool, error) {
+	q := db.reader.WithContext(ctx).
+		Model(&scheme.StoreValue{}).
+		Where("project_id = ? AND key == ?", pid.UUIDValue(), key).
+		Select("count(*)")
+
+	var count int
+	if err := q.Find(&count).Error; err != nil {
```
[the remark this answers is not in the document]
itayd: fixed

internal/backend/store/store.go
```diff
@@ -65,6 +80,35 @@ func (s *store) Mutate(ctx context.Context, pid sdktypes.ProjectID, key, op stri
 		}
 
 		if r.write {
+			if s.cfg.MaxValueSize != 0 && next.ProtoSize() > s.cfg.MaxValueSize {
+				return sdkerrors.NewInvalidArgumentError("value size %d exceeds maximum allowed size %d for a single value", next.ProtoSize(), s.cfg.MaxValueSize)
+			}
+
+			if maxCount := int64(s.cfg.MaxStoreKeysPerProject); maxCount > 0 {
```
efiShtain: why would this ever be zero ? 
and if it can be zero, why do we check it here ? 
you can't read / write anything if the max is zero, better fail fast at the top of the function

internal/backend/store/store.go
```diff
@@ -65,6 +80,35 @@ func (s *store) Mutate(ctx context.Context, pid sdktypes.ProjectID, key, op stri
...
+					has, err := tx.HasStoreKey(ctx, pid, key)
+					if err != nil {
+						return fmt.Errorf("has: %w", err)
+					}
+
+					isNewKey = !has
+				}
+
+				if isNewKey {
+					count, err := tx.CountStoreKeys(ctx, pid)
```
efiShtain: this won't hold in concurrent executions

internal/backend/store/store.go
itayd: if it's 0 - no limit. see comment in config.

internal/backend/store/store.go
itayd: yes it will - it is in a transaction.

itayd: unless i misunderstand how transactions work.

internal/backend/db/dbgorm/store.go
```diff
@@ -68,6 +62,33 @@ func (db *gormdb) GetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key
 	return v, nil
 }
 
+func (db *gormdb) HasStoreKey(ctx context.Context, pid sdktypes.ProjectID, key string) (bool, error) {
+	q := db.reader.WithContext(ctx).
+		Model(&scheme.StoreValue{}).
+		Where("project_id = ? AND key == ?", pid.UUIDValue(), key).
+		Select("count(*)")
```
tebeka: Why not use [Count](https://gorm.io/docs/advanced_query.html#Getting-the-Count-of-Matched-Records)?

internal/backend/store/store.go
```diff
@@ -3,25 +3,40 @@ package store
...
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authz"
+	"go.autokitteh.dev/autokitteh/internal/backend/configset"
 	"go.autokitteh.dev/autokitteh/internal/backend/db"
 	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
 
+type Config struct {
+	MaxValueSize           int `koanf:"max_value_size"`             // 0 to disable size limit.
```
tebeka, line 21: IMO -1 should be no limit, 0 can be use to disallow users from using this feature.
Also, there always should be a limit. We don't want people storing 100GB value by accident.

internal/backend/store/store.go
```diff
@@ -65,6 +80,35 @@ func (s *store) Mutate(ctx context.Context, pid sdktypes.ProjectID, key, op stri
...
+				if r.read {
+					isNewKey = curr.IsNothing()
+				} else {
+					has, err := tx.HasStoreKey(ctx, pid, key)
+					if err != nil {
+						return fmt.Errorf("has: %w", err)
+					}
+
+					isNewKey = !has
+				}
```
tebeka: All this validation code (up to line 111 - couldn't make the mouse get there) should be in a different method.

tebeka (changes requested):

itayd:
@efiShtain added locking

internal/backend/db/dbgorm/store.go
itayd: done

internal/backend/store/store.go
itayd: done

internal/backend/store/store.go
itayd: not sure i agree re -1 vs 0. in other places i use the same scheme. maybe we should overhaul.

efiShtain:
@itayd 

Why do we need locks table with counting instead of just locking the project row ?
it should be very short anyway, and you can lock only if you have something to update or delete

itayd:
> @itayd
> 
> Why do we need locks table with counting instead of just locking the project row ? it should be very short anyway, and you can lock only if you have something to update or delete

now locking the project directly.

internal/backend/db/dbgorm/store.go
efiShtain: What do you think about exists?
More expressive (and better performance on large data sets but irrelevant here)

internal/backend/db/dbgorm/tx.go
```diff
@@ -8,10 +8,25 @@ import (
...
 )
 
-func (db *gormdb) Transaction(ctx context.Context, f func(db db.DB) error) error {
-	return db.writeTransaction(ctx, func(tx *gormdb) error { return f(tx) })
+type txImpl struct {
+	*gormdb
+}
+
+func (tx txImpl) LockProject(ctx context.Context, pid sdktypes.ProjectID) error {
+	return translateError(
```
efiShtain, line 21: What does this function do?

efiShtain (commented):
Not sure I understood the locking you added

internal/backend/db/dbgorm/tx.go
itayd: now it locks...

internal/backend/db/dbgorm/store.go
itayd: done

efiShtain (approved):

PR #1327, opened 2025-06-22
bug(python): Add asyncio function to deterministic

tebeka:
Since asyncio runs in the same thread, it is safe not to run it as an activity.

Fixes: ENG-2147

### DIFF
--- a/runtimes/pythonrt/runner/deterministic.py
+++ b/runtimes/pythonrt/runner/deterministic.py
@@ -5,6 +5,7 @@
 # - Common enough to be included
 # So, `re` is OK, `random` is not
 
+import asyncio
 import datetime
 import functools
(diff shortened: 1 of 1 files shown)


mario99logic (approved):
adding these functions here means they are considered deterministic therefore they do not run as an activity?

itayd:
> adding these functions here means they are considered deterministic therefore they do not run as an activity?

iiuc, these has no side effects by themselves, they are determinstic, so no need to run them in an activity. BUT: the things they do run (as supplied by an external caller), might be non-deterministic and will be wrapped in an activity if needed.

itayd:
@tebeka i really want a session test for these. will approve and merge, but please create session tests in another pr.

PR #1328, opened 2025-06-22
feat: add AirTable integration

mario99logic:
- Added Airtable integration with `default OAuth` and `PAT` connectors
- Introduced PKCE handling in the `oauth` package

Note: @itayd — I'm not fully confident about how I handled the PKCE logic in `getConfig`. The issue is that `getConfig` is called not only during `startOAuthFlow`, but also in `exchangeCodeToToken` and in `refreshOAuthToken`. This leads to a warning when `refreshOAuthToken` tries to save a variable, which ideally shouldn't happen in that context.
Open to suggestions or a better approach if you have one.

ref: ENG-434


itayd (commented):
don't forget to add a pysdk function as well

runtimes/pythonrt/py-sdk/autokitteh/airtable.py
```diff
@@ -0,0 +1,44 @@
...
+    pat = os.getenv(f"{connection}__pat")
+    if pat:
+        return Api(pat)
+
+    refresh_token = os.getenv(connection + "__oauth_refresh_token")
+    if not refresh_token:
+        raise ConnectionInitError(connection)
+
+    try:
+        access_token, _ = refresh_oauth("airtable", connection)
```
itayd, line 40: what's the second return value here?

integrations/oauth/configs.go
```diff
@@ -49,6 +53,30 @@ type oauthConfig struct {
 // for all AutoKitteh integrations. This map must not be modified during runtime.
 func (o *OAuth) initConfigs() {
 	o.oauthConfigs = map[string]oauthConfig{
+		"airtable": {
+			Config: &oauth2.Config{
+				ClientID:     os.Getenv("AIRTABLE_CLIENT_ID"),
```
itayd, line 58: did you create our own app?

integrations/oauth/configs.go
```diff
@@ -729,3 +763,72 @@ func privatizeGitHub(vs sdktypes.Vars, cfg *oauthConfig) {
...
+
+	if cfg.Opts == nil {
+		cfg.Opts = make(map[string]string)
+	}
+	cfg.Opts["code_challenge"] = challenge
+	cfg.Opts["code_challenge_method"] = "S256"
+}
+
+func generatePKCEPair() (verifier, challenge string, err error) {
+	// 64 bytes gives ~86-character base64 string
```
itayd: why the ~?

runtimes/pythonrt/py-sdk/autokitteh/airtable.py
mario99logic: Expiry, there is no use for it in this case I think.

integrations/oauth/configs.go
mario99logic: yes

mario99logic:
added connection test function

integrations/airtable/integration.go
```diff
@@ -0,0 +1,45 @@
...
+
+	"go.autokitteh.dev/autokitteh/integrations/common"
+	"go.autokitteh.dev/autokitteh/integrations/oauth"
+	"go.autokitteh.dev/autokitteh/internal/backend/muxes"
+	"go.autokitteh.dev/autokitteh/sdk/sdkintegrations"
+	"go.autokitteh.dev/autokitteh/sdk/sdkmodule"
+	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
+)
+
+var desc = common.Descriptor("airtable", "Airtable", "/static/images/airtable.svg")
```
itayd, line 14: @RonenMars any change is needed in frontend?

integrations/oauth/configs.go
```diff
@@ -618,6 +646,12 @@ func (o *OAuth) GetConfig(ctx context.Context, integration string, cid sdktypes.
 		return cfg.Config, cfg.Opts, nil
 	}
 
+	// If the integration uses PKCE, inject a dynamic code_challenge/code_verifier
+	// pair into the config before we apply any private OAuth app overrides.
+	if o.flags(integration).usePKCE {
```
itayd: move to after deepCopy

integrations/oauth/configs.go
```diff
@@ -729,3 +763,72 @@ func privatizeGitHub(vs sdktypes.Vars, cfg *oauthConfig) {
...
+	if err != nil {
+		o.logger.Error("failed to generate PKCE pair", zap.Error(err))
+		return
+	}
+
+	pkceVerifier := sdktypes.NewSymbol("pkce_verifier")
+	v := sdktypes.NewVar(pkceVerifier).SetValue(verifier).SetSecret(true)
+
+	// Try to save the new verifier (this might fail during token refresh).
+	if err := o.vars.Set(ctx, v.WithScopeID(sdktypes.NewVarScopeID(cid))); err != nil {
```
itayd, line 795: probably add system user ctx here

integrations/oauth/configs.go
mario99logic: done

integrations/oauth/configs.go
mario99logic: done

itayd: why isn't it there?

mario99logic: When `RefreshOAuthToken` calls `getConfig`, it doesn't reach this part of the code because it returns early if the verifier exists, so it's not needed here.

I added the system user context in var.get since it's needed there to check if it exists.

itayd: but what happens if the verifier doesn't exist?

itayd: if called from refresh_token, verifier must exist

itayd (approved):

PR #1330, opened 2025-06-23
feat(python): More Pythoninc store API

tebeka:
Adds a dict-like `store` to the Python SDK. Users can use familiar API:

```python
import autokitteh

def on_event(event):
    autokitteh.store['api-key'] = 's3cr3t'
    # ...
    api_key = autokitteh.store['api-key']
```

Fixes: ENG-2148


itayd:
@tebeka  i am not sure - in case of adding new value we might raise - if the size is too big or number of elements too much. is it expected in this abstraction?

itayd (approved):

PR #1335, opened 2025-06-26
fix: allow disable workflow executor poller

efiShtain:
Allow disabling poller so we can have servers that only push requests and others that consume the requests

### DIFF
--- a/internal/backend/workflowexecutor/enterprise_executor.go
+++ b/internal/backend/workflowexecutor/enterprise_executor.go
@@ -28,6 +28,7 @@ var (
 	Configs = configset.Set[Config]{
 		Default: &Config{
 			MaxConcurrentWorkflows: 1,
+			WorkerID:               "default-worker",
 		},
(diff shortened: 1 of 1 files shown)


tebeka (approved):

internal/backend/workflowexecutor/enterprise_executor.go
```diff
@@ -58,6 +65,14 @@ func (e *executor) WorkflowQueue() string {
...
 // Start implements WorkflowResourcesManager.
 func (e *executor) Start(ctx context.Context) error {
+	if !e.cfg.EnablePoller {
+		e.l.Info("Workflow executor polling is disabled, skipping start")
+		return nil
+	}
+
+	if e.cfg.WorkerID == "" {
+		return errors.New("worker_id is required")
+	}
```
itayd, line 76:
```suggestion
	}

```

itayd (approved):

PR #1338, opened 2025-06-26
feat(store): add

itayd:


runtimes/pythonrt/py-sdk/autokitteh/store.py
```diff
@@ -94,6 +97,28 @@ def set_value(key: str, value: Any) -> None:
     _local_dev_store[key] = value
 
 
+def add_values(key: str, value: Any) -> Any:
```
tebeka:
```suggestion
def add_values(key: str, value: int|float) -> int|float:
```

tebeka: IMO counters should be only integers, don't see a use case for floats.

runtimes/pythonrt/py-sdk/autokitteh/store.py
```diff
@@ -94,6 +97,28 @@ def set_value(key: str, value: Any) -> None:
...
+
+    Returns:
+        New result value. Always the same type as the value stored under the key.
+    """
+
+    # Dummy implementation for local development.
+    if key not in _local_dev_store:
+        _local_dev_store[key] = value
+    else:
+        _local_dev_store[key] += value
```
tebeka:
```suggestion
_local_dev_store[key] = _local_dev_store.get(key, 0) + value
```

sdk/sdktypes/value.go
```diff
@@ -112,75 +113,152 @@ func (v Value) Concrete() concreteValue {
...
+	switch vv := v.Concrete().(type) {
+	case CustomValue:
+		return vv.Value().ToInt64()
+	case DurationValue:
+		return int64(vv.Value()), nil
+	case IntegerValue:
+		return vv.Value(), nil
+	case StringValue:
+		return strconv.ParseInt(vv.Value(), 0, 64)
+	default:
```
tebeka, line 171: What about FloatValue?

sdk/sdktypes/value.go
```diff
@@ -112,75 +113,152 @@ func (v Value) Concrete() concreteValue {
...
+
+func (v Value) ToInt64() (int64, error) {
+	switch vv := v.Concrete().(type) {
+	case CustomValue:
+		return vv.Value().ToInt64()
+	case DurationValue:
+		return int64(vv.Value()), nil
+	case IntegerValue:
+		return vv.Value(), nil
+	case StringValue:
```
tebeka: I don't like this implicit conversion.

sdk/sdktypes/value.go
```diff
@@ -112,75 +113,152 @@ func (v Value) Concrete() concreteValue {
...
+	switch vv := v.Concrete().(type) {
+	case CustomValue:
+		return vv.Value().ToFloat64()
+	case DurationValue:
+		return float64(vv.Value()), nil
+	case IntegerValue:
+		return float64(vv.Value()), nil
+	case FloatValue:
+		return vv.Value(), nil
+	case StringValue:
```
tebeka: I don't like this implicit conversion.

sdk/sdktypes/value.go
```diff
@@ -112,75 +113,152 @@ func (v Value) Concrete() concreteValue {
...
 	case DurationValue:
-		return v.Value(), nil
+		return vv.Value(), nil
 	case IntegerValue:
-		return time.Second * time.Duration(v.Value()), nil
+		return time.Second * time.Duration(vv.Value()), nil
 	case FloatValue:
-		return time.Duration(float64(time.Second) * v.Value()), nil
+		return time.Duration(float64(time.Second) * vv.Value()), nil
 	case StringValue:
```
tebeka, line 201: I don't like this implicit conversion.

sdk/sdktypes/value.go
```diff
@@ -112,75 +113,152 @@ func (v Value) Concrete() concreteValue {
...
 func (v Value) ToTime() (time.Time, error) {
-	switch v := v.Concrete().(type) {
+	switch vv := v.Concrete().(type) {
 	case CustomValue:
-		return v.Value().ToTime()
+		return vv.Value().ToTime()
 	case TimeValue:
-		return v.Value(), nil
+		return vv.Value(), nil
 	case StringValue:
```
tebeka, line 214: I don't like this implicit conversion.

sdk/sdktypes/value.go
```diff
@@ -112,75 +113,152 @@ func (v Value) Concrete() concreteValue {
...
-		return dateparse.ParseAny(v.Value())
+		return dateparse.ParseAny(vv.Value())
 	case IntegerValue:
-		return time.Unix(v.Value(), 0), nil
+		return time.Unix(vv.Value(), 0), nil
 	default:
```
tebeka, line 222: Float?

sdk/sdktypes/value_ops.go
```diff
@@ -0,0 +1,36 @@
+package sdktypes
+
+import (
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
+)
+
+func AddValues(a, b Value) (Value, error) {
```
tebeka, line 7: Another approach is to have `Add` method on specific types and then something like:

```go
type Adder interface {
    Add(v sdktypes.Value) (sdktypes.Value, error)
}

adder, ok := v.(Adder)
if !ok {
    return sdktypes.InvalidValue, "no support for addition")
}
return adder.Add(other)
```

tebeka (approved):

runtimes/pythonrt/py-sdk/autokitteh/store.py
itayd: da

itayd: wait, this is an "add", not "inc". i agree re inc, not re add.

runtimes/pythonrt/py-sdk/autokitteh/store.py
itayd: done

sdk/sdktypes/value.go
itayd: good catch

sdk/sdktypes/value.go
itayd: yeah you're right. removed.

sdk/sdktypes/value.go
itayd: i think it is warranted here.

sdk/sdktypes/value.go
itayd: removed.

sdk/sdktypes/value.go
itayd: done

sdk/sdktypes/value_ops.go
itayd: interesting - will do it if needs further extending in the future.

PR #1341, opened 2025-06-29
feat: add anthropic connection test func

mario99logic:
Add connection test func that send an HTTP request to anthropic API to check if the connection is valid.

ref: INT-435

### DIFF
--- a/integrations/anthropic/client.go
+++ b/integrations/anthropic/client.go
@@ -93,7 +93,7 @@ func makeAnthropicAPIRequest(ctx context.Context, apiKey string) error {
 		return fmt.Errorf("failed to marshal request: %w", err)
 	}
 
-	req, err := http.NewRequestWithContext(ctx, http.MethodPost, AnthropicURL, bytes.NewBuffer(jsonData))
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, AnthropicURL, bytes.NewReader(jsonData))
 	if err != nil {
 		return fmt.Errorf("failed to construct request: %w", err)
 	}
--- a/integrations/anthropic/client.go
+++ b/integrations/anthropic/client.go
@@ -1,7 +1,12 @@
 package anthropic
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
 
 	"go.autokitteh.dev/autokitteh/integrations"
 	"go.autokitteh.dev/autokitteh/integrations/common"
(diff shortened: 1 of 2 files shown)


integrations/anthropic/client.go
```diff
@@ -50,9 +55,73 @@ func connStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
...
+		return fmt.Errorf("failed to construct request: %w", err)
+	}
+
+	// Set Anthropic-specific headers.
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("x-api-key", apiKey)
+	req.Header.Set("anthropic-version", "2023-06-01")
+
+	// Send the request.
+	resp, err := http.DefaultClient.Do(req)
```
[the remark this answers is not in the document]
mario99logic: @tebeka do we need this if we send the http request with context?

integrations/anthropic/client.go
```diff
@@ -50,9 +55,73 @@ func connStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
...
+		}
+
+		return sdktypes.NewStatus(sdktypes.StatusCodeOK, "OK"), nil
 	})
 }
+
+// makeAnthropicAPIRequest makes a minimal request to validate the Anthropic API key.
+func makeAnthropicAPIRequest(ctx context.Context, apiKey string) error {
+	request := ChatRequest{
+		Model:     "claude-3-5-haiku-20241022", // Use the most cost-effective model.
```
tebeka, line 81: Do we want this configurable? Not sure it's worth the effort.

integrations/anthropic/client.go
```diff
@@ -50,9 +55,73 @@ func connStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
...
+			return sdktypes.NewStatus(sdktypes.StatusCodeError, err.Error()), nil
+		}
+
+		return sdktypes.NewStatus(sdktypes.StatusCodeOK, "OK"), nil
 	})
 }
+
+// makeAnthropicAPIRequest makes a minimal request to validate the Anthropic API key.
+func makeAnthropicAPIRequest(ctx context.Context, apiKey string) error {
+	request := ChatRequest{
```
tebeka, line 80: I'm not sure you need the types just to create a request. You can use a map:

```go
	request := map[string]any{
		"model":      "claude-3-5-haiku-20241022",
		"max_tokens": 1,
		"messages": []map[string]any{
			{"role": "user", "content": "hi"},
		},
	}
```

integrations/anthropic/client.go
```diff
@@ -50,9 +55,73 @@ func connStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
...
+			},
+		},
+	}
+
+	jsonData, err := json.Marshal(request)
+	if err != nil {
+		return fmt.Errorf("failed to marshal request: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, AnthropicURL, bytes.NewBuffer(jsonData))
```
tebeka: IMO `bytes.NewReader` is better.

integrations/anthropic/client.go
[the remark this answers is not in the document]
tebeka: IMO no

integrations/anthropic/client.go
```diff
@@ -50,9 +55,73 @@ func connStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
...
+	req.Header.Set("anthropic-version", "2023-06-01")
+
+	// Send the request.
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to send request: %w", err)
+	}
+	defer resp.Body.Close()
+
+	payload, err := io.ReadAll(resp.Body)
```
tebeka, line 113: Please an `io.LimitReader` over `resp.Body` to limit how much you read.
This can be done in another PR after a security review.

tebeka (approved):

integrations/anthropic/client.go
mario99logic: will do this in another PR, thanks.

integrations/anthropic/client.go
mario99logic: I think its safer if we want to make changes in the api call in the future.

integrations/anthropic/client.go
mario99logic: nice, changed it to newReader

integrations/anthropic/client.go
mario99logic: not worth it, we are going to use this model as a long as its available.

PR #1345, opened 2025-07-02
feat(ENG-2141): initial support for async builds

efiShtain:
this lays the ground for async builds
it currently doesn't handle the building process for async builds
it defaults to sync builds like before


itayd (approved):

itayd:
@efiShtain i rebased this, this is backward compatible, so i'm merging this.

PR #1350, opened 2025-07-14
ENG-2171: add external sessions triggering so limiter would work

efiShtain:
use external triggering for sessions in prod
stop activity retry on resource exhausted

### DIFF
--- a/internal/backend/sessions/sessions.go
+++ b/internal/backend/sessions/sessions.go
@@ -19,8 +19,6 @@ import (
 	"go.autokitteh.dev/autokitteh/internal/backend/sessions/sessionsvcs"
 	"go.autokitteh.dev/autokitteh/internal/backend/sessions/sessionworkflows"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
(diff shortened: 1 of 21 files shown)


internal/backend/sessions/sessions.go
```diff
@@ -308,3 +311,29 @@ func (s *sessions) Start(ctx context.Context, session sdktypes.Session) (sdktype
...
+	})
+
+	if err != nil {
+		return sdktypes.InvalidSessionID, fmt.Errorf("create internal token: %w", err)
+	}
+
+	cli := sdkclients.New(sdkclient.Params{
+		URL:       s.config.ExternalStart.URL,
+		AuthToken: internalToken,
+	}.Safe())
```
itayd: This is really the same thing as you do with dispatcher. Maybe you can make a helper function to create such clients?

itayd (approved):

internal/backend/sessions/sessions.go
efiShtain: refactored

PR #1357, opened 2025-07-23
feat: add descriptions to vars

itayd:


cmd/ak/cmd/vars/get.go
```diff
@@ -52,7 +52,12 @@ func (v V) ToString() string {
 		vv = strconv.Quote(v.Value())
 	}
 
-	return fmt.Sprintf("%v=%s", v.Name(), vv)
+	var desc string
+	if v.Description() != "" {
```
efiShtain, line 56: I'm not sure about this format, 

I think we should always display an empty description if there is none
makes it more consistent

efiShtain (approved):
LGTM

see my comment about formatting

cmd/ak/cmd/vars/get.go
itayd: the idea is to return kvs, if there is a description i add # ...
i don't want to change the format right now, but will rehaul in the future

PR #1362, opened 2025-08-06
try force close runner to clean up

efiShtain:


runtimes/pythonrt/runner/main.py
```diff
@@ -222,6 +223,11 @@ def short_name(func_name: str):
     return func_name[i + 1 :]
 
 
+def force_close(server):
+    server.stop(SERVER_GRACE_TIMEOUT)
```
itayd, line 227: does this block?

runtimes/pythonrt/runner/main.py
```diff
@@ -731,4 +737,9 @@ def dir_type(value):
         Thread(target=runner.should_keep_running, daemon=True).start()
         log.info("started 'should_keep_running' thread")
 
-    server.wait_for_termination()
+    try:
+        server.wait_for_termination()
```
itayd, line 741: why waiting?

tebeka (approved):

runtimes/pythonrt/runner/main.py
efiShtain: suppose to block for grace period

runtimes/pythonrt/runner/main.py
efiShtain: that's was here before
otherwise the process will stop without waiting for connections

PR #1365, opened 2025-08-12
feat: add hubspot events handling

mario99logic:
- added support for hubspot events using webhook 
- added scopes to the default hubspot OAuth app 
- signature check

### DIFF
--- a/integrations/hubspot/events.go
+++ b/integrations/hubspot/events.go
@@ -25,6 +25,8 @@ var (
 )
 
 // handleEvent receives and dispatches asynchronous HubSpot events.
+// We must respond with 200 OK within 5 seconds, otherwise HubSpot
+// will consider the event failed and may retry.
 func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
 	defer r.Body.Close()
 
--- a/integrations/hubspot/events.go
+++ b/integrations/hubspot/events.go
@@ -136,8 +136,6 @@ func (h handler) checkRequest(w http.ResponseWriter, r *http.Request) map[string
 		return nil
 	}
 
-	l.Warn("incoming event: missing HubSpot verification headers")
-
 	return payload[0]
 }
 
--- a/integrations/hubspot/events.go
+++ b/integrations/hubspot/events.go
@@ -1,0 +1,172 @@
+package hubspot
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	"go.uber.org/zap"
+
+	"go.autokitteh.dev/autokitteh/integrations/common"
+	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+)
+
+var portalIDVar = sdktypes.NewSymbol("portal_id")
+
+// handleEvent receives and dispatches asynchronous HubSpot events.
+func (h handler) handleEvent(w http.ResponseWriter, r *http.Request) {
+	defer r.Body.Close()
+
+	// Check the request's headers and parse its body.
+	hubspotEvents := h.checkRequest(w, r)
+	if hubspotEvents == nil {
+		return
+	}
+
+	eventType, ok := hubspotEvents["subscriptionType"].(string)
+	if !ok {
+		h.logger.Warn("received HubSpot event without subscription type",
+			zap.Any("event", hubspotEvents),
+		)
+		common.HTTPError(w, http.StatusBadRequest)
+		return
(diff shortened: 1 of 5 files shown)


integrations/hubspot/events.go
```diff
@@ -0,0 +1,172 @@
...
+
+	var payload []map[string]any
+	if err := json.Unmarshal(body, &payload); err != nil {
+		l.Warn("incoming event: failed to parse JSON body", zap.Error(err))
+		common.HTTPError(w, http.StatusBadRequest)
+		return nil
+	}
+
+	l.Warn("incoming event: missing HubSpot verification headers")
+
```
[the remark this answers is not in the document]
itayd: ^

integrations/hubspot/oauth.go
```diff
@@ -58,5 +63,37 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	vsid := sdktypes.NewVarScopeID(cid)
+	common.SaveAuthType(r, h.vars, vsid)
+
+	// Get portal ID
+	accountURL := "https://api.hubapi.com/integrations/v1/me"
+	accountResp, err := common.HTTPGet(r.Context(), accountURL, auth)
+	if err != nil {
+		l.Warn("failed to get HubSpot account details", zap.Error(err))
```
itayd, line 81: you need to abort here, otherwise connection will be finalized.

integrations/hubspot/oauth.go
```diff
@@ -58,5 +63,37 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	} else {
+		// Parse response and extract portal ID.
+		var accountData map[string]any
+		if json.Unmarshal(accountResp, &accountData) == nil {
+			if id, ok := accountData["portalId"].(float64); ok {
```
itayd: why not just unmarshal to a struct with the desired types?

itayd:
> note: in hubspot/oauth.go I save a new var called portal ID so I can use it to get the connection id before dispatching the event, is there a cleaner way to save the var than the current method? can we add it to the data struct somehow?

I think this is the right way.

integrations/hubspot/events.go
```diff
@@ -0,0 +1,170 @@
...
+func (h handler) signingSecret() (string, error) {
+	secret := os.Getenv("HUBSPOT_CLIENT_SECRET")
+	if secret == "" {
+		return "", errors.New("missing HubSpot secret token")
+	}
+	return secret, nil
+}
+
+func checkSignature(signature, timestamp, secret string, body []byte, method, uri string) (bool, error) {
+	allowed_Duration := 5 * time.Minute
```
itayd: this should be configurable

integrations/hubspot/events.go
```diff
@@ -0,0 +1,170 @@
...
+	currentTime := time.Now()
+	tsInt, err := strconv.ParseInt(timestamp, 10, 64)
+	if err != nil {
+		return false, err
+	}
+	tsTime := time.UnixMilli(tsInt)
+
+	// Check if the timestamp is within the allowed duration according to hubspot's guidelines.
+	if currentTime.Sub(tsTime) > allowed_Duration {
+		return false, errors.New("timestamp is too old")
```
itayd: put the ts in the error

integrations/hubspot/events.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	return secret, nil
+}
+
+func checkSignature(signature, timestamp, secret string, body []byte, method, uri string) (bool, error) {
+	allowed_Duration := 5 * time.Minute
+	currentTime := time.Now()
+	tsInt, err := strconv.ParseInt(timestamp, 10, 64)
+	if err != nil {
+		return false, err
```
itayd: `fmt.Errorf("parse int(%s): %w", timestamp, err)`

integrations/hubspot/oauth.go
```diff
@@ -58,5 +63,42 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+	vsid := sdktypes.NewVarScopeID(cid)
+	common.SaveAuthType(r, h.vars, vsid)
+
+	// Get portal ID
+	accountURL := "https://api.hubapi.com/integrations/v1/me"
+	accountResp, err := common.HTTPGet(r.Context(), accountURL, auth)
+	if err != nil {
+		l.Warn("failed to get HubSpot account details", zap.Error(err))
+		c.AbortServerError("failed to get account details")
```
itayd, line 82: i don't see a return here? this will still progress to finalize. 
also - when you put the return there is no need for the else afterwards, you can deindent that code.

itayd (approved):

PR #1368, opened 2025-08-14
refactor: maintain proper requirements and client names between pythonrt and linter

itayd:


tebeka (approved):
I'm approving due to time constraints. However:

I think a better approach is to have Python code that runs during build of `ak` and generates a Go file with all the definitions. It does require Python as part of the build, but moves errors to build time and not runtime and also less error prone.

PR #1379, opened 2025-08-20
feat: implement microsoft events

mario99logic:
- fix daemon app connection error ref: INT-468
- add microsoft events ref: INT-469


mario99logic:
this PR is still a draft due to two reasons:
- Event subscriptions are too slow due to Microsoft API calls. Used goroutines to fix this, need to discuss the solution.
- When saving a connection, integration starts as microsoft but becomes microsoft_teams after saving . Made a change in integration.go in `resources()` func so it go through the subscription process, I need to discuss it offline

PR #1382, opened 2025-08-21
refactor(ENG-2245):  dockerfile in pythonrt in a simpler way

efiShtain:
make base image on start
don't build if no custom req files
reuse previous images if exists


runtimes/pythonrt/Dockerfile
```diff
@@ -1,27 +1,31 @@
...
 RUN uv pip install .[all] --system
 
 COPY py-sdk py-sdk
 RUN cd py-sdk && uv pip install .[all] --system
 
+FROM base_requirements AS custom_requirements
+
 COPY workflow/user_requirements.txt .
-RUN uv pip install -r user_requirements.txt --system    
+RUN uv pip install -r user_requirements.txt --system
```
itayd, line 19: I thought we're just adding it inside the container, not in the image?

itayd: i think i might have misunderstood what @haimzlato told me, we'll need to talk.

efiShtain: we're mounting the user code at runtime, not the installed packages which has to be installed in the container

I changed the Dockerfile to be one, and I always add a default empty requirement file so no conditionals are required when building

itayd (approved):

PR #1387, opened 2025-08-25
feat(ENG-1770): add container default limits

efiShtain:


runtimes/pythonrt/dockerclient.go
```diff
@@ -30,31 +30,35 @@ const (
...
-	allRunnerIDs    map[string]struct{}
-	mu              *sync.Mutex
-	runnerLabels    map[string]string
-	logBuildProcess bool
-	logRunner       bool
-	logger          *zap.Logger
+	client                     *client.Client
+	activeRunnerIDs            map[string]struct{}
+	allRunnerIDs               map[string]struct{}
+	mu                         *sync.Mutex
```
itayd: this shouldn't be a ptr

runtimes/pythonrt/dockerclient.go
```diff
@@ -30,31 +30,35 @@ const (
...
+	runnerLabels               map[string]string
+	logBuildProcess            bool
+	logRunner                  bool
+	logger                     *zap.Logger
+	maxMemoryBytesPerContainer int64
+	maxNanoCPUPerContainer     int64
```
itayd, line 42: you can just store the config here, and then you don't need to copy all the stuff. it's dry

runtimes/pythonrt/dockerclient.go
```diff
@@ -112,6 +116,10 @@ func (d *dockerClient) StartRunner(ctx context.Context, runnerImage string, sess
 			NetworkMode:  container.NetworkMode(networkName),
 			PortBindings: nat.PortMap{internalRunnerPort: []nat.PortBinding{{HostIP: "127.0.0.1"}}},
 			Tmpfs:        map[string]string{"/tmp": "size=64m"},
+			Resources: container.Resources{
+				Memory:   d.maxMemoryBytesPerContainer, // 512MB memory limit
```
itayd: kill comment

itayd (commented):
.

runtimes/pythonrt/config.go
```diff
@@ -19,12 +19,17 @@ type Config struct {
...
+
+	MaxMemoryPerWorkflowMB int64   `koanf:"max_memory_per_workflow_mb"`
+	MaxCPUsPerWorkflow     float32 `koanf:"max_cpus_per_workflow"`
 }
 
 var Configs = configset.Set[Config]{
 	Default: &Config{
 		RunnerType:               "local",
 		DelayedStartPrintTimeout: 10 * time.Second,
+		MaxMemoryPerWorkflowMB:   512, // 512MB
```
itayd, line 31: can we go below 512? how many can fit in a machine?

runtimes/pythonrt/dockerclient.go
```diff
@@ -112,6 +116,10 @@ func (d *dockerClient) StartRunner(ctx context.Context, runnerImage string, sess
...
+				NanoCPUs: d.maxNanoCPUPerContainer,
```
itayd, line 121: nice of docker to give these options. what happens if they are exceeded?

runtimes/pythonrt/dockerclient.go
efiShtain: it's a different config type with less keys and different values, and I don't really need to original value
if it will grow more I'll probably refactor it to a wrapper type

runtimes/pythonrt/dockerclient.go
efiShtain: done

runtimes/pythonrt/config.go
efiShtain: it's max memory not assigned memory, so it should be ok to have multiple
I played with some values, seems like 512 is ok for our current state
same for 1 cpu 
below that python starts to choke

I need to check how many we can fit on current runners, but we can always increase their size

runtimes/pythonrt/dockerclient.go
efiShtain: memory should OOM
CPU should throttle

itayd (approved):

PR #1391, opened 2025-08-30
feat(ENG-2257): no durability

itayd:
- Triggers has "is_durable" field.
- For v1 manifests, it is true by default.
- For v2 manifests, it is false by default.
- Can be set explicitly for both versions.

To be followed by https://github.com/autokitteh/autokitteh/pull/1396.


cmd/ak/cmd/runtimes/run.go
```diff
@@ -133,7 +133,7 @@ func run(ctx context.Context, b *sdkbuildfile.BuildFile, path string) (map[strin
 			return nil, sdkerrors.ErrNotFound
 		},
 		Call: func(context.Context, sdktypes.RunID, sdktypes.Value, []sdktypes.Value, map[string]sdktypes.Value) (sdktypes.Value, error) {
-			return sdktypes.InvalidValue, sdkerrors.ErrNotImplemented
+			panic("since running as nondurable, must never be called")
```
efiShtain: Is it customary in GO to panic intentionally instead of exiting in an orderly manner ?

cmd/ak/cmd/sessions/start.go
```diff
@@ -88,6 +89,7 @@ func init() {
 	startCmd.Flags().DurationVarP(&pollInterval, "poll-interval", "i", defaultPollInterval, "watch poll interval")
 	startCmd.Flags().BoolVarP(&noTimestamps, "no-timestamps", "n", false, "omit timestamps from watch output")
 	startCmd.Flags().BoolVarP(&quiet, "quiet", "q", false, "don't print anything, just wait to finish")
+	startCmd.Flags().BoolVarP(&durable, "durable", "D", false, "durable run")
```
efiShtain, line 92: just making sure we notice, this is a breaking change
so next version should might require changing the major key (not a must since this is a low key change)

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -233,6 +233,7 @@ type Trigger struct {
 	EventType    string
 	Filter       string
 	CodeLocation string
+	IsDurable    bool
```
efiShtain: We want to allow durability on the trigger? 
isn't it too low level ? 
I though it would be on the project level

what happen if a workflow starts and then we change the field value and the workflow restarts ?

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -348,6 +350,7 @@ type Session struct {
 	Entrypoint       string
 	Inputs           datatypes.JSON
 	Memo             datatypes.JSON
+	IsDurable        bool
```
efiShtain: why is this required on the session ?

internal/backend/projects/projects.go
```diff
@@ -290,9 +290,12 @@ func (ps *Projects) exportManifest(ctx context.Context, projectID sdktypes.Proje
...
 
 	for _, t := range triggers {
+		isDurable := t.IsDurable()
+
 		mt := manifest.Trigger{
-			Name: t.Name().String(),
-			Call: t.CodeLocation().CanonicalString(),
+			Name:      t.Name().String(),
+			Call:      t.CodeLocation().CanonicalString(),
+			IsDurable: &isDurable,
```
efiShtain, line 298: why is this a reference and not a default bool value false ?

internal/backend/projects/projects.go
```diff
@@ -301,6 +304,8 @@ func (ps *Projects) exportManifest(ctx context.Context, projectID sdktypes.Proje
 			mt.EventType = etype
 		}
 
+		mt.IsDurable = &isDurable
```
efiShtain: isn't this line duplicate of line 298 ?

internal/backend/sessions/sessionworkflows/workflow.go
```diff
@@ -462,6 +461,11 @@ func (w *sessionWorkflow) run(wctx workflow.Context, l *zap.Logger) (_ []sdkserv
 				return sdktypes.InvalidValue, errors.New("cannot call self during initial evaluation")
 			}
 
+			if !session.IsDurable() {
```
efiShtain, line 464: what's the problem with call in non durable mode?
or is it not required? 
maybe i need to see the documented design for this

migrations/postgres/enterprise/20250903024555_trigger_durability.sql
```diff
@@ -0,0 +1,7 @@
+-- +goose Up
+-- modify "triggers" table
+ALTER TABLE "triggers" ADD COLUMN "is_durable" boolean NULL;
```
efiShtain: you can make the field default to false and non nullable
just make it a 3 step migration

```sql
ALTER TABLE "triggers" ADD COLUMN "is_durable" boolean NULL DEFAULT false;
UPDATE "triggers" SET is_durable = false;
ALTER TABLE "triggers" ALTER COLUMN "is_durable" SET NOT NULL;
```

proto/autokitteh/sessions/v1/session.proto
```diff
@@ -148,6 +148,8 @@ message Session {
 
   SessionStateType state = 12;
 
+  bool is_durable = 13;
```
efiShtain, line 151: consistency please
either is_durable like runtimes/v1 or durable like here

runtimes/configrt/runtime.go
```diff
@@ -32,6 +32,7 @@ func (svc) Run(
 	mainPath string,
 	compiled map[string][]byte,
 	_ map[string]sdktypes.Value,
+	_ bool,
```
efiShtain, line 35: why is this not named?

runtimes/pythonrt/py-sdk/embed.go
```diff
@@ -21,7 +21,7 @@ var pyProject string
 var clientDefRegex = regexp.MustCompile(`^def (\w+_client)\(`)
 
 func ClientNames() (names []string) {
-	fs.WalkDir(codeFS, "autokitteh", func(path string, d fs.DirEntry, err error) error {
+	kittehs.Must0(fs.WalkDir(codeFS, "autokitteh", func(path string, d fs.DirEntry, err error) error {
```
efiShtain: if you want to must this, i think you should read at at init instead of on demand
we don't want panics here

cmd/ak/cmd/runtimes/run.go
itayd: you know what, i'll revert this.

cmd/ak/cmd/sessions/start.go
itayd: oh yeah we're going to be 0.20.0. this is intentional.

internal/backend/db/dbgorm/scheme/records.go
itayd: product wise this is ok - there might some triggers that start in a durable way and other that starts in nondurable.

re trigger changing - it is cached by temporal anyway in an activity so we're fine.

internal/backend/db/dbgorm/scheme/records.go
itayd: a session can be durable or not. i mean this is the crux of the change.

internal/backend/projects/projects.go
itayd: we need to know if it was explicitly specified in the manifest or not. if it is specified, we take the value. if it is not - it depends on the manifest version. v1 will assume true. v2 will assume false. this is for backward compatability.

internal/backend/projects/projects.go
itayd: right, removed.

internal/backend/sessions/sessionworkflows/workflow.go
itayd: there are no calls in nondurable mode - everything is self contained in the python runtime except "syscalls" (callbacks).

runtimes/pythonrt/py-sdk/embed.go
itayd: right, fixed

runtimes/configrt/runtime.go
itayd: it's unused, so why write the name?

migrations/postgres/enterprise/20250903024555_trigger_durability.sql
itayd: not sure what is better. lets discuss at sync.

proto/autokitteh/sessions/v1/session.proto
itayd: what do you preffer?

efiShtain (approved):

proto/autokitteh/sessions/v1/session.proto
itayd:
```
$ ag --proto is_durable
user_code/v1/runner_svc.proto
27:  bool is_durable = 3;

triggers/v1/trigger.proto
22:  bool is_durable = 8;

sessions/v1/session.proto
151:  bool is_durable = 13;
```

in proto it's always `is_durable`...

migrations/postgres/enterprise/20250903024555_trigger_durability.sql
itayd: done

itayd:
merged as part of https://github.com/autokitteh/autokitteh/pull/1396.

PR #1402, opened 2025-09-11
feat: synchronous http responses

itayd:
allow sync http responses. see new triggers fields. looks like this in manifest:

```
  triggers:
    - name: webhook
      webhook:
        sync: true
      call: program.py:wh
```

currently works by polling the session logs for a new record type of "HTTP Response", which is intentionally specific for HTTP. Can be generalized in the future for "outcomes", which might be used for slack, etc.


runtimes/pythonrt/worker_grpc_handler.go
```diff
@@ -582,3 +582,26 @@ func (s *workerGRPCHandler) StoreMutate(ctx context.Context, req *userCode.Store
...
+
+	resp, err := s.callback(ctx, req.RunnerId, "http_respond", fn)
+	if err != nil {
+		return &userCode.HTTPResponseResponse{Error: err.Error()}, nil
+	}
+
+	if resp.err != nil {
+		err = status.Errorf(codes.Internal, "http_respond(%v, %v, %v, %v) -> %v", req.Status, req.Body, req.Headers, req.More, err)
+		return &userCode.HTTPResponseResponse{Error: err.Error()}, nil
+	}
```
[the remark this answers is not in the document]
itayd: fixed

sdk/sdkclients/sdkdispatcherclient/client.go
```diff
@@ -34,46 +35,64 @@ func (c *client) Redispatch(ctx context.Context, eventID sdktypes.EventID, opts
...
-func (c *client) Dispatch(ctx context.Context, event sdktypes.Event, opts *sdkservices.DispatchOptions) (sdktypes.EventID, error) {
+func (c *client) Dispatch(ctx context.Context, event sdktypes.Event, opts *sdkservices.DispatchOptions) (*sdkservices.DispatchResponse, error) {
 	if opts == nil {
 		opts = &sdkservices.DispatchOptions{}
 	}
 
 	resp, err := c.client.Dispatch(ctx, connect.NewRequest(&dispatcherv1.DispatchRequest{
 		Event:        event.ToProto(),
 		DeploymentId: opts.DeploymentID.String(),
+		Wait:         opts.Wait,
```
efiShtain, line 74: what is wait? wait for what ? 
is this a breaking change ?

sdk/sdkclients/sdkdispatcherclient/client.go
```diff
@@ -34,46 +35,64 @@ func (c *client) Redispatch(ctx context.Context, eventID sdktypes.EventID, opts
...
+	}
+
+	return &sdkservices.DispatchResponse{
+		EventID:    eventId,
+		SessionIDs: sids,
+	}, nil
 }
 
-func (c *client) Dispatch(ctx context.Context, event sdktypes.Event, opts *sdkservices.DispatchOptions) (sdktypes.EventID, error) {
+func (c *client) Dispatch(ctx context.Context, event sdktypes.Event, opts *sdkservices.DispatchOptions) (*sdkservices.DispatchResponse, error) {
```
efiShtain, line 66: why is this change required in this PR ?

sdk/sdkservices/dispatcher.go
```diff
@@ -9,11 +9,20 @@ import (
 type DispatchOptions struct {
 	// If set, dispatch only to this specific deployment.
 	DeploymentID sdktypes.DeploymentID
+
+	// If true, the call will block until the dispatch is done.
+	Wait bool
```
efiShtain, line 14: why? what's the purpose of it? 
and no one sohuld actually call dispatch directly

sdk/sdkservices/runtimes.go
```diff
@@ -71,6 +71,9 @@ type RunCallbacks struct {
 	// Store
 	ListStoreValues  func(ctx context.Context, rid sdktypes.RunID) ([]string, error)
 	MutateStoreValue func(ctx context.Context, rid sdktypes.RunID, key, op string, operands ...sdktypes.Value) (sdktypes.Value, error)
+
+	// HTTP Response
+	HTTPResponse func(ctx context.Context, rid sdktypes.RunID, resp sdktypes.SessionHTTPResponse) error
```
efiShtain: why is this a call back?

runtimes/pythonrt/runner/main.py
```diff
@@ -345,6 +345,7 @@ def patch_ak_funcs(self):
         autokitteh.start = self.syscalls.ak_start
         autokitteh.subscribe = self.syscalls.ak_subscribe
         autokitteh.unsubscribe = self.syscalls.ak_unsubscribe
+        autokitteh.http_respond = self.syscalls.ak_http_respond
```
efiShtain: http_response?

efiShtain:
I really don't like this entire PR (as you already know)

you need to consider non durable workflows and their affect on this, 
probably only non durable workflow should support sync responses
you should also consider timeouts - a triggered workflow might take time to start since we queue up workflows, then you have timeouts on the load balancers and reverse proxies and such, you also don't want people just having open connections for long times for no good reason and probably don't want lots of polling on the database searching for this sync values

but if you are moving forward with it, 
I think you should refactor it to SessionOutputs instead of specific http response log record (just like you said in the pr description), probably a dedicated table we can query fast and ditch fast since stuff might be large and wasteful

allow the session to store outputs as key value pairs (like the store, or use vars if you want but that wouldn't be a very good design)

session would then could store any number of outputs and the user can poll for them (or wait on the webhook for specific output(s))

sdk/sdkclients/sdkdispatcherclient/client.go
itayd: wait for a session to be created. in sync mode we need to wait for it to be created to know what to poll on.

this is backward compatible. if sync is not specified - this wont wait.

sdk/sdkclients/sdkdispatcherclient/client.go
itayd: we want to get the session id that was created in order to know what to track.

itayd: we could do without, but then we'll have to make another list call to find all sessions for event, which is not really needed if we do this.

sdk/sdkservices/dispatcher.go
itayd: see above

sdk/sdkservices/runtimes.go
itayd: it's a "syscall". we can discuss.

runtimes/pythonrt/runner/main.py
itayd: uuuuggghh...

itayd:
> but if you are moving forward with it, I think you should refactor it to SessionOutputs instead of specific http response log record (just like you said in the pr description), probably a dedicated table we can query fast and ditch fast since stuff might be large and wasteful

Hahaha. I did it explicitly "http response" and not "session outcomes" becuase i thought you'll be against premature generalization here :)

> allow the session to store outputs as key value pairs (like the store, or use vars if you want but that wouldn't be a very good design)
> 
> session would then could store any number of outputs and the user can poll for them (or wait on the webhook for specific output(s))

we should discuss - not sure we want this as a first step. but i agree long term this can be useful (regardless of sync).

sdk/sdkclients/sdkdispatcherclient/client.go
efiShtain: wait in sync context sounds like wait for response, 
also, the dispatcher creates an event workflow and only later on there is a session workflow
so we want to wait for session ?

efiShtain (approved):
see my last 2 comments

Im making sure this is set in git's history - I'm against this change at it is

but let's try it, glad to be wrong

itayd:
> Im making sure this is set in git's history - I'm against this change at it isZ

it wont - only the PR description will be 🙉

sdk/sdkclients/sdkdispatcherclient/client.go
itayd: wait is for waiting for dispatch done. will think of another name.

PR #1404, opened 2025-09-11
feat: get events for connections only with active deployments

mario99logic:
Temporary fix: Updated findConnectionID (now findActiveConnectionID) to handle connection events for deployed projects only.

### DIFF
--- a/internal/backend/db/dbgorm/triggers.go
+++ b/internal/backend/db/dbgorm/triggers.go
@@ -133,7 +133,7 @@ func (db *gormdb) hasActiveDeployment(ctx context.Context, projectID uuid.UUID) 
 	var deploymentCount int64
 	err := db.reader.WithContext(ctx).
 		Model(&scheme.Deployment{}).
-		Where("project_id = ? AND state = ? AND deleted_at IS NULL", projectID, sdktypes.DeploymentStateActive.ToProto()).
+		Where("project_id = ? AND state = ? AND deleted_at IS NULL", projectID, int32(sdktypes.DeploymentStateActive.ToProto())).
 		Count(&deploymentCount).Error
 	if err != nil {
 		return translateError(err)
--- a/internal/backend/db/dbgorm/triggers.go
+++ b/internal/backend/db/dbgorm/triggers.go
@@ -130,7 +130,7 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
 	return scheme.ParseTrigger(*r)
 }
 
-func (db *gormdb) GetTriggerWithActiveDeploymentByID(ctx context.Context, triggerID sdktypes.TriggerID) (sdktypes.Trigger, bool, error) {
+func (db *gormdb) GetTriggerWithActiveDeploymentByID(ctx context.Context, triggerID uuid.UUID) (sdktypes.Trigger, bool, error) {
 	var triggerAndDeployment struct {
 		scheme.Trigger
 		HasActiveDeployment bool `gorm:"column:has_active_deployment"`
@@ -141,7 +141,7 @@ func (db *gormdb) GetTriggerWithActiveDeploymentByID(ctx context.Context, trigge
 		Select("triggers.*, CASE WHEN deployments.deployment_id IS NOT NULL THEN true ELSE false END as has_active_deployment").
 		Joins("LEFT JOIN deployments ON triggers.project_id = deployments.project_id AND deployments.state = ? AND deployments.deleted_at IS NULL",
 			int32(sdktypes.DeploymentStateActive.ToProto())).
-		Where("triggers.trigger_id = ?", triggerID.UUIDValue()).
+		Where("triggers.trigger_id = ?", triggerID).
 		First(&triggerAndDeployment).Error
 	if err != nil {
 		if errors.Is(err, gorm.ErrRecordNotFound) {
--- a/internal/backend/db/dbgorm/triggers.go
+++ b/internal/backend/db/dbgorm/triggers.go
@@ -150,12 +150,8 @@ func (db *gormdb) GetTriggerWithActiveDeploymentByID(ctx context.Context, trigge
 		return sdktypes.InvalidTrigger, false, translateError(err)
 	}
 
-	if !triggerAndDeployment.HasActiveDeployment {
-		return sdktypes.InvalidTrigger, false, nil // Trigger exists but no active deployment.
-	}
-
 	trigger, err := scheme.ParseTrigger(triggerAndDeployment.Trigger)
-	return trigger, true, err
+	return trigger, triggerAndDeployment.HasActiveDeployment, err
 }
 
 func (db *gormdb) ListTriggers(ctx context.Context, filter sdkservices.ListTriggersFilter) ([]sdktypes.Trigger, error) {
--- a/internal/backend/db/dbgorm/triggers.go
+++ b/internal/backend/db/dbgorm/triggers.go
@@ -10,6 +10,7 @@ import (
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authcontext"
 	"go.autokitteh.dev/autokitteh/internal/backend/db/dbgorm/scheme"
 	"go.autokitteh.dev/autokitteh/internal/kittehs"
+	"go.autokitteh.dev/autokitteh/sdk/sdkerrors"
 	"go.autokitteh.dev/autokitteh/sdk/sdkservices"
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
 )
@@ -127,6 +128,39 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
 	return scheme.ParseTrigger(*r)
 }
 
+// hasActiveDeployment checks if a project has active deployments
+func (db *gormdb) hasActiveDeployment(ctx context.Context, projectID uuid.UUID) error {
+	var deploymentCount int64
+	err := db.reader.WithContext(ctx).
+		Model(&scheme.Deployment{}).
+		Where("project_id = ? AND state = ? AND deleted_at IS NULL", projectID, sdktypes.DeploymentStateActive.ToProto()).
+		Count(&deploymentCount).Error
+	if err != nil {
+		return translateError(err)
+	}
+
+	if deploymentCount == 0 {
+		return sdkerrors.ErrFailedPrecondition
+	}
+
+	return nil
+}
+
+func (db *gormdb) GetTriggerWithActiveDeploymentByID(ctx context.Context, triggerID sdktypes.TriggerID) (sdktypes.Trigger, error) {
+	// Check if the trigger exists.
+	trigger, err := db.getTriggerByID(ctx, triggerID.UUIDValue())
+	if trigger == nil || err != nil {
+		return sdktypes.InvalidTrigger, translateError(err)
+	}
+
+	// Check if there's an active deployment for this trigger's project.
+	if err := db.hasActiveDeployment(ctx, trigger.ProjectID); err != nil {
+		return sdktypes.InvalidTrigger, err
+	}
+
+	return scheme.ParseTrigger(*trigger)
+}
+
 func (db *gormdb) ListTriggers(ctx context.Context, filter sdkservices.ListTriggersFilter) ([]sdktypes.Trigger, error) {
 	ts, err := db.listTriggers(ctx, filter)
 	if ts == nil || err != nil {
@@ -135,11 +169,21 @@ func (db *gormdb) ListTriggers(ctx context.Context, filter sdkservices.ListTrigg
 	return kittehs.TransformError(ts, scheme.ParseTrigger)
 }
 
-func (db *gormdb) GetTriggerByWebhookSlug(ctx context.Context, slug string) (sdktypes.Trigger, error) {
-	r, err := getOne[scheme.Trigger](db.reader.WithContext(ctx), "webhook_slug = ?", slug)
+func (db *gormdb) GetTriggerWithActiveDeploymentByWebhookSlug(ctx context.Context, slug string) (sdktypes.Trigger, error) {
+	// Check if the trigger exists.
+	var trigger scheme.Trigger
+	err := db.reader.WithContext(ctx).
+		Model(&scheme.Trigger{}).
+		Where("webhook_slug = ?", slug).
+		First(&trigger).Error
 	if err != nil {
 		return sdktypes.InvalidTrigger, translateError(err)
 	}
 
-	return scheme.ParseTrigger(*r)
+	// Check if there's an active deployment for this trigger's project.
+	if err := db.hasActiveDeployment(ctx, trigger.ProjectID); err != nil {
+		return sdktypes.InvalidTrigger, err
+	}
+
+	return scheme.ParseTrigger(trigger)
 }
--- a/internal/backend/db/dbgorm/triggers.go
+++ b/internal/backend/db/dbgorm/triggers.go
@@ -2,10 +2,12 @@ package dbgorm
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"time"
 
 	"github.com/google/uuid"
+	"gorm.io/gorm"
 
 	"go.autokitteh.dev/autokitteh/internal/backend/auth/authcontext"
 	"go.autokitteh.dev/autokitteh/internal/backend/db/dbgorm/scheme"
@@ -128,22 +130,21 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
 	return scheme.ParseTrigger(*r)
 }
 
-// hasActiveDeployment checks if a project has active deployments
-func (db *gormdb) hasActiveDeployment(ctx context.Context, projectID uuid.UUID) error {
-	var deploymentCount int64
+// hasActiveDeployment checks if a project has active deployments.
+func (db *gormdb) hasActiveDeployment(ctx context.Context, projectID uuid.UUID) (bool, error) {
+	var deployment scheme.Deployment
 	err := db.reader.WithContext(ctx).
-		Model(&scheme.Deployment{}).
 		Where("project_id = ? AND state = ? AND deleted_at IS NULL", projectID, int32(sdktypes.DeploymentStateActive.ToProto())).
-		Count(&deploymentCount).Error
+		Take(&deployment).Error
+
 	if err != nil {
-		return translateError(err)
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return false, nil
+		}
+		return false, translateError(err)
 	}
 
-	if deploymentCount == 0 {
-		return sdkerrors.ErrFailedPrecondition
-	}
-
-	return nil
+	return true, nil
 }
 
 func (db *gormdb) GetTriggerWithActiveDeploymentByID(ctx context.Context, triggerID sdktypes.TriggerID) (sdktypes.Trigger, error) {
@@ -154,8 +155,12 @@ func (db *gormdb) GetTriggerWithActiveDeploymentByID(ctx context.Context, trigge
 	}
 
 	// Check if there's an active deployment for this trigger's project.
-	if err := db.hasActiveDeployment(ctx, trigger.ProjectID); err != nil {
+	hasActive, err := db.hasActiveDeployment(ctx, trigger.ProjectID)
+	if err != nil {
 		return sdktypes.InvalidTrigger, err
+	}
+	if !hasActive {
+		return sdktypes.InvalidTrigger, sdkerrors.ErrFailedPrecondition
 	}
 
 	return scheme.ParseTrigger(*trigger)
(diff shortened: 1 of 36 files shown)


internal/backend/vars/vars.go
```diff
@@ -189,7 +189,7 @@ func (v *Vars) Get(ctx context.Context, sid sdktypes.VarScopeID, names ...sdktyp
 	})
 }
 
-func (v *Vars) FindConnectionIDs(ctx context.Context, iid sdktypes.IntegrationID, name sdktypes.Symbol, value string) ([]sdktypes.ConnectionID, error) {
+func (v *Vars) FindActiveConnectionIDs(ctx context.Context, iid sdktypes.IntegrationID, name sdktypes.Symbol, value string) ([]sdktypes.ConnectionID, error) {
```
itayd, line 192: I don't see any actual change to the function other than the name?

internal/backend/db/dbgorm/vars.go
```diff
@@ -74,16 +74,23 @@ func (gdb *gormdb) listVars(ctx context.Context, scopeID uuid.UUID, names ...str
...
 	if v != "" {
-		db = db.Where("value = ? AND is_secret is false", v)
+		db = db.Where("vars.value = ? AND vars.is_secret is false", v)
 	}
 
+	// Join with connections and deployments to filter by active projects only.
+	db = db.Model(&scheme.Var{}).
+		Joins("JOIN connections ON vars.var_id = connections.connection_id").
+		Joins("JOIN deployments ON connections.project_id = deployments.project_id").
+		Where("deployments.state = ? AND deployments.deleted_at IS NULL", 1)
```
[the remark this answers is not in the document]
itayd: what it said

internal/backend/db/dbgorm/vars.go
```diff
@@ -74,16 +74,23 @@ func (gdb *gormdb) listVars(ctx context.Context, scopeID uuid.UUID, names ...str
 }
 
 func (gdb *gormdb) findConnectionIDsByVar(ctx context.Context, integrationID uuid.UUID, name string, v string) ([]uuid.UUID, error) {
```
itayd: findConnectionIDsWithActiveDeploymentByVar

internal/backend/vars/vars.go
```diff
@@ -189,7 +189,7 @@ func (v *Vars) Get(ctx context.Context, sid sdktypes.VarScopeID, names ...sdktyp
...
 	if err := authz.CheckContext(ctx, sdktypes.InvalidIntegrationID, "read:find-var-connections-ids", authz.WithData("integration_id", iid), authz.WithData("name", name)); err != nil {
```
itayd, line 193: rename function name used here

internal/backend/db/dbgorm/triggers.go
```diff
@@ -127,6 +128,39 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
 	return scheme.ParseTrigger(*r)
 }
 
+// hasActiveDeployment checks if a project has active deployments
+func (db *gormdb) hasActiveDeployment(ctx context.Context, projectID uuid.UUID) error {
```
efiShtain: I think since the name is hasActive... the function should return (bool,error) where false means no and true means yes

internal/backend/db/dbgorm/triggers.go
```diff
@@ -127,6 +128,39 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
...
+	var deploymentCount int64
+	err := db.reader.WithContext(ctx).
```
efiShtain: Not a must, worth checking if exists would give the same results instead of count
It is negligible performance wise since there is always 1 or 0 rows matching, but it's more readable and express what you try to achieve

internal/backend/db/dbgorm/triggers.go
```diff
@@ -127,6 +128,39 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
...
+		Model(&scheme.Deployment{}).
+		Where("project_id = ? AND state = ? AND deleted_at IS NULL", projectID, int32(sdktypes.DeploymentStateActive.ToProto())).
+		Count(&deploymentCount).Error
+	if err != nil {
+		return translateError(err)
+	}
+
+	if deploymentCount == 0 {
+		return sdkerrors.ErrFailedPrecondition
```
efiShtain: See my other comment on the name of the function and return signature

Anyhow, the error doesn't seems like a good fit here 

If you keep it as error, not found is much better

internal/backend/db/dbgorm/triggers.go
```diff
@@ -127,6 +128,39 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
...
+	}
+
+	return nil
+}
+
+func (db *gormdb) GetTriggerWithActiveDeploymentByID(ctx context.Context, triggerID sdktypes.TriggerID) (sdktypes.Trigger, error) {
+	// Check if the trigger exists.
+	trigger, err := db.getTriggerByID(ctx, triggerID.UUIDValue())
```
efiShtain: Why not exists with a join instead of two queries?

internal/backend/db/dbgorm/triggers.go
```diff
@@ -135,11 +169,21 @@ func (db *gormdb) ListTriggers(ctx context.Context, filter sdkservices.ListTrigg
 	return kittehs.TransformError(ts, scheme.ParseTrigger)
 }
 
-func (db *gormdb) GetTriggerByWebhookSlug(ctx context.Context, slug string) (sdktypes.Trigger, error) {
-	r, err := getOne[scheme.Trigger](db.reader.WithContext(ctx), "webhook_slug = ?", slug)
+func (db *gormdb) GetTriggerWithActiveDeploymentByWebhookSlug(ctx context.Context, slug string) (sdktypes.Trigger, error) {
+	// Check if the trigger exists.
+	var trigger scheme.Trigger
+	err := db.reader.WithContext(ctx).
```
efiShtain, line 167: Same, try using one simple joined query with exists

internal/backend/db/dbgorm/vars.go
```diff
@@ -73,17 +73,24 @@ func (gdb *gormdb) listVars(ctx context.Context, scopeID uuid.UUID, names ...str
...
+		Joins("JOIN deployments ON connections.project_id = deployments.project_id").
+		Where("deployments.state = ? AND deployments.deleted_at IS NULL", int32(sdktypes.DeploymentStateActive.ToProto()))
+
 	// Note(s):
 	// - will skip not user owned vars
 	// - not checking if scope is deleted, since scope deletion will cascade deletion of relevant vars
+	// - only returns connections for projects with active deployments
 	var ids []uuid.UUID
-	if err := db.Model(&scheme.Var{}).Distinct("var_id").Find(&ids).Error; err != nil {
+	if err := db.Distinct("vars.var_id").Find(&ids).Error; err != nil {
```
efiShtain: What's the purpose of distinct here?
Var_id is not unique?

internal/backend/scheduler/scheduler.go
```diff
@@ -133,15 +133,28 @@ func (sch *Scheduler) activity(ctx context.Context, tid sdktypes.TriggerID) erro
...
+
+			if err := sch.Delete(ctx, tid); err != nil {
+				return temporalclient.TranslateError(err, "delete schedule for %v", tid)
+			}
+
+			return nil
 		}
+
+		if errors.Is(err, sdkerrors.ErrFailedPrecondition) {
+			sl.Warnf("trigger %v found but project not deployed, skipping execution", tid)
```
efiShtain: This should be info

There is error from our side

The user won't see this logs anyway

internal/backend/scheduler/scheduler.go
```diff
@@ -133,15 +133,28 @@ func (sch *Scheduler) activity(ctx context.Context, tid sdktypes.TriggerID) erro
...
 	if err != nil {
-		if !errors.Is(err, sdkerrors.ErrNotFound) {
-			return temporalclient.TranslateError(err, "get trigger %v", tid)
+		if errors.Is(err, sdkerrors.ErrNotFound) {
+			sl.Warnf("trigger %v not found, removing schedule", tid)
+
+			if err := sch.Delete(ctx, tid); err != nil {
+				return temporalclient.TranslateError(err, "delete schedule for %v", tid)
+			}
+
```
efiShtain: Add info log removed with the details

sdk/sdkclients/sdktriggersclient/client.go
```diff
@@ -85,6 +86,12 @@ func (c *client) Get(ctx context.Context, triggerID sdktypes.TriggerID) (sdktype
 	return sdktypes.TriggerFromProto(resp.Msg.Trigger)
 }
 
+func (c *client) GetWithActiveDeployment(ctx context.Context, triggerID sdktypes.TriggerID) (sdktypes.Trigger, error) {
```
efiShtain: Not sure I understand why it is not supported for clients
And if so, why is it in the sdk?

sdk/sdkservices/triggers.go
```diff
@@ -22,5 +22,6 @@ type Triggers interface {
 	Update(ctx context.Context, trigger sdktypes.Trigger) error
 	Delete(ctx context.Context, triggerID sdktypes.TriggerID) error
 	Get(ctx context.Context, triggerID sdktypes.TriggerID) (sdktypes.Trigger, error)
+	GetWithActiveDeployment(ctx context.Context, triggerID sdktypes.TriggerID) (sdktypes.Trigger, error)
```
efiShtain: Why not use the list triggers and extend it with active deployments capabilities?
Not sure it works but did you consider it?

internal/backend/db/dbgorm/triggers.go
itayd: +1

internal/backend/db/dbgorm/triggers.go
mario99logic: I used Take() better in this case than count or find

internal/backend/scheduler/scheduler.go
mario99logic: changed it to Info

internal/backend/db/dbgorm/triggers.go
mario99logic: I used `ErrFailedPrecondition` instead of `ErrNotFound` because the trigger does exist meaning we successfully retrieved it from the database. 
ErrNotFound would incorrectly suggest the trigger doesn't exist, when the actual issue is that the trigger's project lacks active deployments. 
I think its better in terms of readability and overall purpose to keep `ErrFailedPrecondition`

internal/backend/db/dbgorm/triggers.go
mario99logic: I used two queries to distinguish between two different error scenarios:
  1. Trigger doesn't exist 
  2. Trigger exists but no active deployment 

isn't it better to keep these two scenarios separate for readability and debugging? thats why I used `ErrFailedPrecondition` like I mentioned in the previous comment.

internal/backend/db/dbgorm/triggers.go
mario99logic: same reason I used to queries to distinguish between scenarios
for webhooks: We need to to return the correct HTTP response to clients:
  - Trigger doesn't exist → HTTP 404  "Not Found"
  - Trigger exists but no active deployment → HTTP 412 "Project Not Deployed"

internal/backend/db/dbgorm/vars.go
mario99logic: removed it

sdk/sdkservices/triggers.go
mario99logic: mainly for semantic mismatch: List() suggests "get multiple", but we want exactly one.

sdk/sdkclients/sdktriggersclient/client.go
mario99logic: it can be supported for clients but there is no point for a client to use it.
do I remove it from the sdk and use it as an internal func? if so how can I give the scheduler access to it

internal/backend/db/dbgorm/triggers.go
efiShtain: you're looking for something which you don't find, looking for active connections and there are no active connections, literally not found

not sure which precondition failed here

precondition error might be relevant if the user tries to run the workflow assuming there are active deployments and then there are none, then the DB should say not found, the service might say failed precondition

in the DB layer, there is no notion of per condition on active deployments, its application logic

internal/backend/db/dbgorm/triggers.go
efiShtain: it really depends, 

as I see it (I might be wrong), this specific function is used mostly for asynchronous tasks, like webhooks/events/etc... not user generated. so it doesn't matter if the trigger isn't found or the deployment isn't active, there is just no trigger with active deployment which is what you try to find

if it was user generated, it also means it is of low call rate (no user would call this 1K times per second, but a webhook might, which requires a different handling) - so for user generated, where you want to explain something to the user, it might be better to have 2 different calls


another interesting option (and this is for learning purposes here), you can return a left join, which will return the trigger if it exists even if there is no deployment, but if there is a deployment, you'll get it as well

something like (don't catch me on specific details here)

```go
type triggerAndDeployment struct{
   TriggerID string,
   DeploymentID string
}
```
```sql
SELECT t.trigger_id, d.deployment_id FROM triggers t JOIN projects p USING(project_id) LEFT JOIN deployments d USING(project_id) where t.trigger_id = '' and d.state = 'active'
```

internal/backend/db/dbgorm/triggers.go
efiShtain: webhooks are for machines, any 4xx has the same effect

sdk/sdkclients/sdktriggersclient/client.go
efiShtain: which object the scheduler uses ? 
use type assertion maybe to get the underlying object instead of the sdk client

sdk/sdkservices/triggers.go
efiShtain: I'm fine with this logic

internal/backend/db/dbgorm/triggers.go
```diff
@@ -128,22 +130,21 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
 	return scheme.ParseTrigger(*r)
 }
 
-// hasActiveDeployment checks if a project has active deployments
-func (db *gormdb) hasActiveDeployment(ctx context.Context, projectID uuid.UUID) error {
-	var deploymentCount int64
+// hasActiveDeployment checks if a project has active deployments.
```
efiShtain: that's kind of redundant since the function signature is good and self explanatory

internal/backend/db/dbgorm/triggers.go
```diff
@@ -128,22 +130,21 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
...
+func (db *gormdb) hasActiveDeployment(ctx context.Context, projectID uuid.UUID) (bool, error) {
+	var deployment scheme.Deployment
```
efiShtain: you can do exists to just get a boolean answer from the db directly, try it

internal/backend/db/dbgorm/triggers.go
mario99logic: since we are doing single query for both the webhook and the scheduler then this function is no longer needed

internal/backend/db/dbgorm/triggers.go
mario99logic: my comment above

sdk/sdkclients/sdktriggersclient/client.go
mario99logic: I injected the db in the scheduler service and used the function

internal/backend/svc/svc.go
```diff
@@ -268,7 +268,7 @@ func makeFxOpts(cfg *Config, opts RunOptions) []fx.Option {
 			scheduler.Configs,
 			fx.Provide(scheduler.New),
 			fx.Invoke(
-				func(lc fx.Lifecycle, sch *scheduler.Scheduler, d sdkservices.Dispatcher, ts sdkservices.Triggers) {
+				func(lc fx.Lifecycle, sch *scheduler.Scheduler, db db.DB, d sdkservices.Dispatcher, ts sdkservices.Triggers) {
```
efiShtain: did you check it? 
cause you get the db, but you don't pass it download anywhere

internal/backend/webhookssvc/svc.go
```diff
@@ -60,17 +60,11 @@ func (s *Service) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	t, err := s.db.GetTriggerWithActiveDeploymentByWebhookSlug(ctx, slug)
 	if err != nil {
 		if errors.Is(err, sdkerrors.ErrNotFound) {
-			sl.Infof("slug %q not found", slug)
+			sl.Infof("slug %q not found or project not deployed", slug)
```
efiShtain: I know it was slug not found before, 
I think a better test would be "could not find an active deployment for trigger by slug %q"

internal/backend/svc/svc.go
mario99logic: I removed it there is no need for it here we only need it in the constructor, I misunderstood. 
So trigger and dispatcher are passed because they might need each other therefore they are created separately in the constructor and then they are connected together in the runtime.
`db` is static so its okay to inject it in the construction time only.

internal/backend/db/dbgorm/triggers.go
```diff
@@ -128,19 +131,30 @@ func (db *gormdb) GetTriggerByID(ctx context.Context, triggerID sdktypes.Trigger
...
 	if err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return sdktypes.InvalidTrigger, sdkerrors.ErrNotFound // Trigger doesn't exist.
+		}
 		return sdktypes.InvalidTrigger, translateError(err)
 	}
 
-	return scheme.ParseTrigger(trigger)
+	if !triggerAndDeployment.HasActiveDeployment {
+		return sdktypes.InvalidTrigger, sdkerrors.ErrFailedPrecondition // Trigger exists but no active deployment.
```
efiShtain: I still think it should be not found
pre condition is irrelevant

internal/backend/scheduler/scheduler.go
```diff
@@ -137,21 +137,20 @@ func (sch *Scheduler) activity(ctx context.Context, tid sdktypes.TriggerID) erro
 
 	trigger, err := sch.db.GetTriggerWithActiveDeploymentByID(ctx, tid)
 	if err != nil {
-		if errors.Is(err, sdkerrors.ErrNotFound) {
-			sl.Infof("skipping execution for trigger %v: not found or project not deployed", tid)
+		if errors.Is(err, sdkerrors.ErrNotFound) && !trigger.IsValid() {
```
efiShtain: why do you need to check trigger.isvalid if there is an error ? 
if err not found, just remove the schedule

mario99logic: I agree, in the main branch there is check for errNotFound and then another check if the trigger is invalid, maybe there is a specific case for it Im not sure.

efiShtain: this checks ?
it has a different semantic then your code
but if it works the same as before I'm fine with not fixing it in this PR
if you changed the handling, need to make sure it doesn't introduce a new bug

```go
if err != nil {
		if !errors.Is(err, sdkerrors.ErrNotFound) {
			return temporalclient.TranslateError(err, "get trigger %v", tid)
		}
	}

	if !t.IsValid() {
		sl.Warnf("trigger %v not found, removing schedule", tid)

		if err := sch.Delete(ctx, tid); err != nil {
			return temporalclient.TranslateError(err, "delete schedule for %v", tid)
		}

		return nil
	}

```

internal/backend/db/dbgorm/triggers.go
```diff
@@ -145,16 +145,17 @@ func (db *gormdb) GetTriggerWithActiveDeploymentByID(ctx context.Context, trigge
...
-			return sdktypes.InvalidTrigger, sdkerrors.ErrNotFound // Trigger doesn't exist.
+			return sdktypes.InvalidTrigger, false, sdkerrors.ErrNotFound // Trigger doesn't exist.
 		}
-		return sdktypes.InvalidTrigger, translateError(err)
+		return sdktypes.InvalidTrigger, false, translateError(err)
 	}
 
 	if !triggerAndDeployment.HasActiveDeployment {
-		return sdktypes.InvalidTrigger, sdkerrors.ErrFailedPrecondition // Trigger exists but no active deployment.
+		return sdktypes.InvalidTrigger, false, nil // Trigger exists but no active deployment.
```
efiShtain: trigger is valid in this case, just doesn't have a deployment

internal/backend/scheduler/scheduler.go
```diff
@@ -135,24 +135,24 @@ func (sch *Scheduler) activity(ctx context.Context, tid sdktypes.TriggerID) erro
 
 	ctx = authcontext.SetAuthnSystemUser(ctx)
 
-	trigger, err := sch.db.GetTriggerWithActiveDeploymentByID(ctx, tid)
+	_, hasTrigger, err := sch.db.GetTriggerWithActiveDeploymentByID(ctx, tid)
```
efiShtain: trigger is the first variable, you're looking for hadActiveDeployment

efiShtain (approved):

internal/backend/db/dbgorm/triggers.go
mario99logic: yes but there is no need to parse it and return a valid trigger because it won't be used.

efiShtain: You don't know what the caller want
You have a function that returns a trigger with its active deployment

If you want a function that only returns true/false if there is an active deployment for a trigger than rename the function and return a bool

Otherwise it's either error or the trigger and true/false

Returning invalid trigger is not the correct state since there is a valid trigger

PR #1406, opened 2025-09-17
feat(ENG-2272): organization level connections support

efiShtain:
Design decisions can be found here

https://linear.app/autokitteh/issue/ENG-2272/allow-org-level-connections


proto/autokitteh/connections/v1/connection.proto
```diff
@@ -19,6 +19,9 @@ message Connection {
   common.v1.Status status = 5;
   Capabilities capabilities = 6;
   map<string, string> links = 7;
+
+  string scope = 8;
```
itayd: enum

internal/backend/connections/connections.go
```diff
@@ -27,13 +27,24 @@ type Connections struct {
 func New(c Connections) sdkservices.Connections { return &c }
 
 func (c *Connections) Create(ctx context.Context, conn sdktypes.Connection) (sdktypes.ConnectionID, error) {
+	// TODO: This is for backwards compatibility.
+	// We should remove it once the UI is updated to always pass org_id
+	if conn.OrgID() == sdktypes.InvalidOrgID {
```
itayd: !conn.OrgID().IsValid()

PR #1416, opened 2025-09-27
feat: mark python sdk as typed

itayd:


efiShtain (approved):

tebeka (commented):
Late to the game :) Don't really know about this and why do we need it.
Saw the uvx error, not sure where it's coming from.

itayd:
> Late to the game :) Don't really know about this and why do we need it. Saw the uvx error, not sure where it's coming from.

when using mypy on scripts, mypy complained autokitteh is not typed

PR #1424, opened 2025-09-30
bug(python): Support UUID in wrap/unwrap

tebeka:
Note that unwrapping will return string, not UUID. I'm OK with that :)

The underlying reason is that Python saves UUID as an int (unlimited in size in Python), and we convert it to a Proto int. 

Fixes: ENG-2285


itayd:
@tebeka doesn't this break the behavior of uuid4?

efiShtain (approved):

PR #1436, opened 2025-10-08
fix: better error message for the user in telegram

mario99logic:
- better error message for the user in telegram
- connection status: only after successful toke authentication and setting the webhook url we set the auth type

### DIFF
--- a/integrations/telegram/client.go
+++ b/integrations/telegram/client.go
@@ -5,7 +5,6 @@ import (
 
 	"go.uber.org/zap"
 
-	"go.autokitteh.dev/autokitteh/integrations"
 	"go.autokitteh.dev/autokitteh/integrations/common"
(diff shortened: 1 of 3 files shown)


integrations/telegram/client.go
```diff
@@ -40,10 +39,7 @@ func connStatus(i *integration) sdkintegrations.OptFn {
 			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
 		}
 
-		if at.Value() == integrations.Init {
```
efiShtain, line 43: Are you sure we should remove this?

mario99logic: Yes
1. we only have one auth type for telegram 
2. I only save the auth type with the token after the authentication process, so if there is an auth_type and its not empty this means the status is initialized

efiShtain (approved):

PR #1440, opened 2025-10-13
feat: save anthropic api key only if key is valid

mario99logic:


integrations/anthropic/save.go
```diff
@@ -59,6 +59,14 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Validate the API key by making a test request.
+	err = makeAnthropicAPIRequest(r.Context(), apiKey)
+	if err != nil {
+		l.Debug("save connection: API key validation failed", zap.Error(err))
```
efiShtain: do we have any context to this message ?

efiShtain (approved):

PR #1441, opened 2025-10-13
feat: add project display name

itayd:
optional, backward compatible


migrations/postgres/20251013231828_project_display_name.sql
```diff
@@ -0,0 +1,11 @@
+-- +goose Up
+-- modify "projects" table
+ALTER TABLE "projects" ADD COLUMN "display_name" text NULL;
+-- create index "idx_projects_display_name" to table: "projects"
+CREATE INDEX "idx_projects_display_name" ON "projects" ("display_name");
```
efiShtain: why do we want to index this?

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -124,11 +124,12 @@ func (v *Var) BeforeCreate(tx *gorm.DB) (err error) {
...
 
-	ProjectID uuid.UUID `gorm:"primaryKey;type:uuid;not null"`
-	OrgID     uuid.UUID `gorm:"index;type:uuid"` // TODO(authz-migration): not null.
-	Name      string    `gorm:"index;not null"`
-	RootURL   string
-	Resources []byte
+	ProjectID   uuid.UUID `gorm:"primaryKey;type:uuid;not null"`
+	OrgID       uuid.UUID `gorm:"index;type:uuid"` // TODO(authz-migration): not null.
+	Name        string    `gorm:"index;not null"`
+	DisplayName string    `gorm:"index"`
```
efiShtain: I think this shouldn't be null as well

itayd: kill index

efiShtain (approved):

PR #1446, opened 2025-10-16
fix: panic in dispatchAsyncEventsToConnections

efiShtain:
there was a use of resp.EventID.String() for response which was error causing a panic due to nil exception in code


mario99logic (approved):
Nice

PR #1448, opened 2025-10-17
feat: allow opa to read policy from local filesystem

itayd:


internal/backend/policy/opapolicy/opa.go
```diff
@@ -28,6 +30,8 @@ type Config struct {
 	MinLogLevel        string `koanf:"log_level"`         // log level threshold to emit. if empty: "warn".
 	MinConsoleLogLevel string `koanf:"console_log_level"` // console log level threshold to emit. if empty: "warn".
 
+	LocalPolicyContentBase64 string `koanf:"local_policy_content_base64"` // If ConfigPath is empty, and this is set, load local policies from this path.
```
efiShtain: I'd set the name to PolicyContentBase64, not sure why local is helpful

efiShtain (approved):

internal/backend/policy/opapolicy/opa.go
itayd: fixed

PR #1456, opened 2025-10-28
feat: store check_and_set

itayd:


internal/backend/store/values.go
```diff
@@ -37,6 +37,28 @@ var ops = map[string]op{
 		},
 		write: true,
 	},
+	"check_and_set": {
+		fn: func(curr sdktypes.Value, vs []sdktypes.Value) (sdktypes.Value, sdktypes.Value, error) {
+			if len(vs) < 2 {
+				return sdktypes.InvalidValue, sdktypes.InvalidValue, sdkerrors.NewInvalidArgumentError("expecting two operands: next_value, check_value")
+			} else if len(vs) > 2 {
```
efiShtain, line 44: for readability can you change this else if to 
```golang
if len(vs) < 2 {
   return
}

if len(vs) > 2 {
   return 
}

// rest of code
```

internal/backend/store/values.go
```diff
@@ -37,6 +37,28 @@ var ops = map[string]op{
...
+			if !curr.IsValid() {
+				curr = sdktypes.Nothing
+			}
+
+			if !curr.Equal(vs[1]) {
+				// Failed check, no change.
+				return curr, sdktypes.FalseValue, nil
+			}
+
+			return vs[0], sdktypes.TrueValue, nil
```
efiShtain, line 57: race conditions between workflows ?

efiShtain (approved):

internal/backend/store/values.go
itayd: this is done inside a transaction

internal/backend/store/values.go
itayd: i don't see much difference...

PR #1457, opened 2025-10-28
feat: add timezone to trigger

mario99logic:
- Added timezone field to the trigger
- pass it to temporal cli to set the timezone in the scheduler
- Added it in the CLI


cmd/ak/cmd/triggers/create.go
```diff
@@ -13,8 +13,8 @@ import (
 )
 
 var (
-	call, event, filter, name, project, schedule string
-	webhook                                      bool
+	call, event, filter, name, project, schedule, timezone string
+	webhook                                                bool
 )
 
 var createCmd = common.StandardCommand(&cobra.Command{
```
itayd, line 20: add doc for tz in Use

itayd (approved):

PR #1463, opened 2025-11-02
fix: Don't save Gemini connection if key is not valid

mario99logic:
- Validate gemini api key before saving connection
- Add Sanity check: the connection ID is valid
- Added logs to contest func

### DIFF
--- a/integrations/google/gemini/client.go
+++ b/integrations/google/gemini/client.go
@@ -17,19 +17,12 @@ const IntegrationName = "googlegemini"
 
 var desc = common.Descriptor(IntegrationName, "Google Gemini", "/static/images/google_gemini.svg")
 
-type integration struct {
-	cvars sdkservices.Vars
-	l     *zap.Logger
-}
-
 func New(cvars sdkservices.Vars, l *zap.Logger) sdkservices.Integration {
-	i := &integration{cvars: cvars, l: l}
-
 	return sdkintegrations.NewIntegration(
 		desc,
 		sdkmodule.New(),
-		i.connStatus(),
-		i.connTest(),
+		connStatus(cvars),
+		connTest(cvars),
 		sdkintegrations.WithConnectionConfigFromVars(cvars),
 	)
 }
@@ -37,13 +30,13 @@ func New(cvars sdkservices.Vars, l *zap.Logger) sdkservices.Integration {
 // connStatus is an optional connection status check provided by
 // the integration to AutoKitteh. The possible results are "Init
 // required" (the connection is not usable yet) and "Initialized".
-func (i integration) connStatus() sdkintegrations.OptFn {
+func connStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
 	return sdkintegrations.WithConnectionStatus(func(ctx context.Context, cid sdktypes.ConnectionID) (sdktypes.Status, error) {
 		if !cid.IsValid() {
 			return sdktypes.NewStatus(sdktypes.StatusCodeWarning, "Init required"), nil
 		}
 
-		vs, err := i.cvars.Get(ctx, sdktypes.NewVarScopeID(cid))
+		vs, err := cvars.Get(ctx, sdktypes.NewVarScopeID(cid))
 		if err != nil {
 			return sdktypes.InvalidStatus, err
 		}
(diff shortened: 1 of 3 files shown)


integrations/google/gemini/client.go
```diff
@@ -3,43 +3,47 @@ package gemini
...
+		i.connTest(),
 		sdkintegrations.WithConnectionConfigFromVars(cvars),
 	)
 }
 
 // connStatus is an optional connection status check provided by
 // the integration to AutoKitteh. The possible results are "Init
 // required" (the connection is not usable yet) and "Initialized".
-func connStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
+func (i integration) connStatus() sdkintegrations.OptFn {
```
efiShtain: Are you sure this works?

You use i as a reference but this is not a pointer receiver

integrations/google/gemini/save.go
```diff
@@ -41,5 +45,48 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	// Sanity check: the connection ID is valid.
+	cid, err := sdktypes.StrictParseConnectionID(c.ConnectionID)
+	if err != nil {
+		l.Warn("save connection: invalid connection ID", zap.Error(err))
```
efiShtain: Info or debug, not our system problem

integrations/google/gemini/save.go
```diff
@@ -41,5 +45,48 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		c.AbortBadRequest("invalid connection ID")
+		return
+	}
+
+	api_key := r.Form.Get("key")
+	if api_key == "" {
+		l.Debug("API key is missing")
```
efiShtain: Some context about what you're trying to do

API key is missing can mean a lot of things without context

integrations/google/gemini/save.go
```diff
@@ -41,5 +45,48 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	client, err := genai.NewClient(ctx, option.WithAPIKey(apiKey))
+	if err != nil {
+		return err
+	}
+	defer client.Close()
+
+	iter := client.ListModels(ctx)
+	for {
+		_, err := iter.Next()
+		if err == iterator.Done {
```
efiShtain: Do you need to iterate all models?

integrations/google/gemini/client.go
```diff
@@ -3,43 +3,47 @@ package gemini
...
 	"go.autokitteh.dev/autokitteh/sdk/sdktypes"
+	"go.uber.org/zap"
 )
 
 const IntegrationName = "googlegemini"
 
 var desc = common.Descriptor(IntegrationName, "Google Gemini", "/static/images/google_gemini.svg")
 
-func New(cvars sdkservices.Vars) sdkservices.Integration {
+type integration struct {
```
efiShtain: What's the purpose of this struct?
What was wrong with pure functions?

mario99logic: to pass the logs

integrations/google/gemini/client.go
mario99logic: missed the pointer, but I used the previous approach now need for this approach

integrations/google/gemini/save.go
mario99logic: no Need I changed it

integrations/google/gemini/client.go
```diff
@@ -3,47 +3,41 @@ package gemini
...
+		connTest(cvars),
 		sdkintegrations.WithConnectionConfigFromVars(cvars),
 	)
 }
 
 // connStatus is an optional connection status check provided by
 // the integration to AutoKitteh. The possible results are "Init
 // required" (the connection is not usable yet) and "Initialized".
-func (i integration) connStatus() sdkintegrations.OptFn {
+func connStatus(cvars sdkservices.Vars) sdkintegrations.OptFn {
```
efiShtain, line 34: The "go" convention for such optionals, is calling it withX
so withConnStatus, withConnText

efiShtain (approved):

PR #1471, opened 2025-11-08
fix: properly render service base url

itayd:


cmd/ak/cmd/manifest/deploy.go
```diff
@@ -157,7 +157,7 @@ func PrintSuggestions(cmd *cobra.Command, effects manifest.Effects) {
 			}
 
 			if slug := t.WebhookSlug(); slug != "" {
-				urlPath := webhookssvc.WebhooksPathPrefix + slug
+				urlPath := "/" + webhookssvc.WebhooksPathPrefix + slug
```
efiShtain, line 160: you should probably check you don't get double / somehow
isn't there a url parse or url join utility?

efiShtain (commented):
see my comment

cmd/ak/cmd/manifest/deploy.go
itayd: WebhookPathPrefix is a const, it will always not have the leading slash.

efiShtain (approved):

PR #1473, opened 2025-11-10
fix: try and set error to session log if container failed

efiShtain:


runtimes/pythonrt/runner_manager_docker.go
```diff
@@ -202,7 +202,24 @@ func (rm *dockerRunnerManager) RunnerHealth(ctx context.Context, runnerID string
 		return err
 	}
 	if !isRunning {
-		return errors.New("runner not running")
+		exitCode, err := rm.client.getContainerExitCode(ctx, cid)
```
itayd, line 205: should we remove the container here?

itayd (approved):

PR #1474, opened 2025-11-11
feat: allow to publish store values

itayd:
stored values that are published are world readable by http, prefixed by the project id and their key.


internal/backend/db/dbgorm/store.go
```diff
@@ -68,6 +68,33 @@ func (db *gormdb) GetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key
 	return v, nil
 }
 
+func (db *gormdb) PublishStoreValue(ctx context.Context, pid sdktypes.ProjectID, key string) error {
+	if !pid.IsValid() {
+		return sdkerrors.NewInvalidArgumentError("invalid project id")
+	}
+
+	q := db.writer.WithContext(ctx).Update("public", "true").Where("project_id = ? AND key = ?", pid.UUIDValue(), key)
```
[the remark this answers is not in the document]
efiShtain: @itayd

internal/backend/db/dbgorm/store.go
```diff
@@ -68,6 +68,33 @@ func (db *gormdb) GetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key
...
+	if q.RowsAffected == 0 {
+		return sdkerrors.ErrNotFound
+	}
+
+	return nil
+}
+
+func (db *gormdb) IsStoreValuePublished(ctx context.Context, pid sdktypes.ProjectID, key string) (bool, error) {
+	var sv scheme.StoreValue
+	err := db.reader.WithContext(ctx).Select("public").Where("project_id = ? AND key = ?", pid.UUIDValue(), key).First(&sv).Error
```
efiShtain: technically you don't need the value so you can just check for exists where public is true

optimization
you can ignore if you want

internal/backend/db/dbgorm/store.go
```diff
@@ -68,6 +68,33 @@ func (db *gormdb) GetStoreValue(ctx context.Context, pid sdktypes.ProjectID, key
 	return v, nil
 }
 
+func (db *gormdb) PublishStoreValue(ctx context.Context, pid sdktypes.ProjectID, key string) error {
```
efiShtain, line 80: do we need unpublish?

internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -559,6 +559,7 @@ type StoreValue struct {
 	ProjectID uuid.UUID `gorm:"primaryKey;type:uuid;not null"`
 	Key       string    `gorm:"primaryKey;not null"`
 	Value     []byte
+	Public    bool
```
efiShtain: it's kind of nit, but consistent,
you call the field public and call the function publish, 
I think it's better be published bool 
so Publish would make it published instead of public

internal/backend/sessions/sessionworkflows/activities.go
```diff
@@ -32,6 +32,7 @@ const (
 	mutateStoreValueActivityName             = "mutate_store_value"
 	notifyWorkflowEndedActivity              = "notify_workflow_ended"
 	outcomeActivityName                      = "outcome"
+	publishStoreValueActivityName            = "publish_store_value"
```
efiShtain, line 35: why is this an activity? 
making a value public should happen only once
or I'm missing something?

efiShtain (commented):
not sure I understand why you need to publish from the user code
just make it a decision in the UI

internal/backend/storehttpsvc/svc.go
```diff
@@ -0,0 +1,79 @@
...
+	db db.DB
+}
+
+func Init(muxes *muxes.Muxes, db db.DB, l *zap.Logger) {
+	s := &svc{db: db, l: l}
+	muxes.NoAuth.Handle("/"+storePathPrefix+"{pid}/{key}", s)
+}
+
+func (s *svc) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	pid, err := sdktypes.StrictParseProjectID(r.PathValue("slug"))
```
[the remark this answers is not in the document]
itayd: fixed

internal/backend/db/dbgorm/store.go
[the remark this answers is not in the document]
itayd: fixed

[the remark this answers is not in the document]
itayd: fixed

internal/backend/db/dbgorm/store.go
itayd: i currently don't see a reason for that.

internal/backend/db/dbgorm/store.go
itayd: nah, i think maybe in the future if we want to store false and stuff.

itayd:
> not sure I understand why you need to publish from the user code just make it a decision in the UI

we don't have a ui for these. also i was thinking users my want to publish something with high cardinality, like maybe a uuid and send this uuid to somewhere.

internal/backend/sessions/sessionworkflows/activities.go
itayd: yeah, but if we do it from a workflow, having it in activity will prevent it from happening again unneccesarily on replays. i know it's idempotent, but it requires context, and we might want to retry on db issues, etc.

internal/backend/db/dbgorm/scheme/records.go
itayd: changed

internal/backend/db/dbgorm/store.go
efiShtain: If you unpublish by mistake for testing you're stuck

internal/backend/sessions/sessionworkflows/activities.go
efiShtain: I understand why it is an activity if it's in a workflow
I don't understand why it is part of a workflow vs project configuration 

I guess there are no real product requirements

itayd: you can't put it in configuration since these might be dynamically named.

efiShtain (approved):

PR #1479, opened 2025-11-19
fix(ENG-2312): block webhooks with too large a body 

itayd:
also prevent converting to a string if not valid UTF8, which might panic on JSON serialization later on. Discovered this by pushing random data as http body.

### DIFF
--- a/internal/backend/webhookssvc/svc.go
+++ b/internal/backend/webhookssvc/svc.go
@@ -85,10 +85,14 @@ func (s *Service) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	sl.Infof("webhook request: %s %s", r.Method, r.URL.Path)
 
-	if s.cfg.MaxBodySize > 0 && uint64(r.ContentLength) > s.cfg.MaxBodySize {
-		sl.Warnf("request body too large: %d bytes (max: %d bytes)", r.ContentLength, s.cfg.MaxBodySize)
-		http.Error(w, fmt.Sprintf("Request Entity Too Large (%d>%d bytes)", r.ContentLength, s.cfg.MaxBodySize), http.StatusRequestEntityTooLarge)
-		return
+	if s.cfg.MaxBodySize > 0 {
+		if r.ContentLength > 0 && uint64(r.ContentLength) > s.cfg.MaxBodySize {
+			sl.Warnf("request body too large: %d bytes (max: %d bytes)", r.ContentLength, s.cfg.MaxBodySize)
+			http.Error(w, fmt.Sprintf("Request Entity Too Large (%d>%d bytes)", r.ContentLength, s.cfg.MaxBodySize), http.StatusRequestEntityTooLarge)
+			return
+		}
+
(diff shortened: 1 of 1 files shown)


internal/backend/webhookssvc/svc.go
```diff
@@ -406,7 +417,10 @@ func bodyData(body []byte, form url.Values) sdktypes.Value {
 
 	if len(body) > 0 {
 		bytes = sdktypes.NewBytesValue(body)
-		text = sdktypes.NewStringValue(string(body))
+
+		if utf8.Valid(body) {
```
efiShtain, line 424: if not valid?

efiShtain, line 424: if not valid?

efiShtain (approved):
not sure this will work if encoding is transfer-chunk
so you have to enforce content-type application/json

another thing is that it might be better do it in envoy instead of here (limit and content type verification)

internal/backend/webhookssvc/svc.go
itayd: no translation to string, because it's not a string.

internal/backend/webhookssvc/svc.go
```diff
@@ -78,10 +81,17 @@ func (s *Service) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		zap.String("method", r.Method),
 		zap.String("url", r.URL.String()),
 		zap.String("slug", slug),
+		zap.Int64("content_length", r.ContentLength),
 	).Sugar()
 
 	sl.Infof("webhook request: %s %s", r.Method, r.URL.Path)
 
+	if s.cfg.MaxBodySize > 0 && uint64(r.ContentLength) > s.cfg.MaxBodySize {
```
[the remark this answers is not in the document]
itayd: done

internal/backend/webhookssvc/svc.go
```diff
@@ -78,10 +81,17 @@ func (s *Service) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		sl.Warnf("request body too large: %d bytes (max: %d bytes)", r.ContentLength, s.cfg.MaxBodySize)
+		http.Error(w, fmt.Sprintf("Request Entity Too Large (%d>%d bytes)", r.ContentLength, s.cfg.MaxBodySize), http.StatusRequestEntityTooLarge)
+		return
+	}
+
```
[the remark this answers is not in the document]
itayd: done

internal/backend/webhookssvc/svc.go
itayd: no this is ok - we set the text if the body is valid utf8.

itayd:
> not sure this will work if encoding is transfer-chunk so you have to enforce content-type application/json

if content-length is not defined, we go with limiting the actual body read now.

> 
> another thing is that it might be better do it in envoy instead of here (limit and content type verification)

maybe? but then it wont work on oss. worth hanging a lantern on this though.

PR #1485, opened 2025-11-20
feat: unpublish store value

itayd:


internal/backend/db/dbgorm/store.go
```diff
@@ -94,6 +94,23 @@ func (db *gormdb) PublishStoreValue(ctx context.Context, pid sdktypes.ProjectID,
...
 }
 
+func (db *gormdb) UnpublishStoreValue(ctx context.Context, pid sdktypes.ProjectID, key string) error {
+	if !pid.IsValid() {
+		return sdkerrors.NewInvalidArgumentError("invalid project id")
+	}
+
+	q := db.writer.WithContext(ctx).Model(&scheme.StoreValue{}).Where("project_id = ? AND key = ?", pid.UUIDValue(), key).Update("published", false)
+	if err := q.Error; err != nil {
+		return translateError(err)
```
efiShtain, line 104: i think we should log db errors, they shouldn't happen

internal/backend/store/store.go
```diff
@@ -117,3 +117,16 @@ func (s *store) Publish(ctx context.Context, pid sdktypes.ProjectID, key string)
...
 }
+
+func (s *store) Unpublish(ctx context.Context, pid sdktypes.ProjectID, key string) error {
+	if err := authz.CheckContext(
+		ctx,
+		pid,
+		"write:unpublish",
+		authz.WithData("key", key),
+	); err != nil {
+		return err
```
efiShtain, line 128: not sure relaed only to this PR, but maybe CheckContext should return bool,error? so we know (and log) operational error then unauthorized?

runtimes/pythonrt/runner/syscalls.py
```diff
@@ -231,6 +231,13 @@ def ak_publish_value(self, key: str) -> None:
         )
         call_grpc("store_publish", self.worker.StorePublish, req)
 
+    def ak_unpublish_value(self, key: str) -> None:
+        req = pb.StoreUnpublishRequest(
+            runner_id=self.runner_id,
+            key=key,
+        )
+        call_grpc("store_unpublish", self.worker.StoreUnpublish, req)
```
efiShtain, line 239: logs? something?

efiShtain (approved):
please add some logs where needed

internal/backend/store/store.go
itayd: maaaybe, but the thing is that the error is containing the right error to return to the user. maybe i can split this function in two to do that. not for this pr anyway.

runtimes/pythonrt/runner/syscalls.py
itayd: i don't see we do this in other places? maybe call_grpc should do the logging?

PR #1486, opened 2025-11-20
fix: properly limit stored value size and number

itayd:


internal/backend/store/store.go
```diff
@@ -48,6 +53,16 @@ func (s *store) Mutate(ctx context.Context, pid sdktypes.ProjectID, key, op stri
 		)
 
 		if r.read {
+			if err := authz.CheckContext(
+				ctx,
+				pid,
+				"read:get-store-value",
```
efiShtain: are you sure we are going to need this fine grained rbac?
read:store is not enough?
or even just read?

internal/backend/store/store.go
```diff
@@ -65,6 +80,32 @@ func (s *store) Mutate(ctx context.Context, pid sdktypes.ProjectID, key, op stri
...
+				return sdkerrors.NewInvalidArgumentError("value size (%d bytes) exceeds maximum allowed (%d bytes)", next.ProtoSize(), s.cfg.MaxValueSizeBytes)
+			}
+
+			if (!curr.IsValid() || curr.IsNothing()) && next.IsValid() {
+				count, err := tx.CountStoreValues(ctx, pid)
+				if err != nil {
+					return err
+				}
+
+				if int(count)+1 > s.cfg.MaxValuesPerProject {
```
efiShtain, line 93: technically we do limits using envoy externally
same like project limits
you can leave it here if you want
but its not consistent with other limits

internal/backend/store/store.go
```diff
@@ -65,6 +80,32 @@ func (s *store) Mutate(ctx context.Context, pid sdktypes.ProjectID, key, op stri
...
+					return sdkerrors.NewInvalidArgumentError("number of stored values (%d) exceeds maximum allowed (%d)", count+1, s.cfg.MaxValuesPerProject)
+				}
+			}
+
+			if err := authz.CheckContext(
```
efiShtain, line 98: i think you should better do read and write checks before any db access, espcially counts which might be expensive operations (even if user has small number of items, it can affect cache for other users)

internal/backend/store/store.go
```diff
@@ -82,7 +123,7 @@ func (s *store) Get(ctx context.Context, pid sdktypes.ProjectID, keys []string)
 	if err := authz.CheckContext(
 		ctx,
 		pid,
-		"read:get",
+		"read:get-store-value",
```
efiShtain: if you are going to reuse this strings set them as const at the top
or better have one rbac file with all permissions in one place

efiShtain (commented):
see my comments

internal/backend/store/store.go
itayd: i mean, the policy can choose to ignore the key, it's just the name of the operation.

internal/backend/store/store.go
itayd: this is not a per-user limit or billable limit, it's lower level. i think it's more important to get this out. but will consider.

internal/backend/store/store.go
itayd: problem is that i don't know the size and if it's a new var ahead of time.

internal/backend/store/store.go
itayd: i like the rbac idea, but i'll do it in a different pr

efiShtain (approved):

PR #1490, opened 2025-11-26
refactor: move all authz ops to constants

itayd:


efiShtain (approved):

internal/backend/auth/authz/ops.go
```diff
@@ -0,0 +1,107 @@
...
+	// Store operations
+	OpStoreWriteDo        = "write:do"
+	OpStoreReadGet        = "read:get"
+	OpStoreReadList       = "read:list"
+	OpStoreWritePublish   = "write:publish"
+	OpStoreWriteUnpublish = "write:unpublish"
+
+	// Organization operations
+	OpOrgCreateCreate       = "create:create"
+	OpOrgReadGet            = "read:get"
```
efiShtain: You have multiple read:get
You probably want resource prefix or something like orgs:read

PR #1491, opened 2025-11-26
feat: log all unhandled failures in db

itayd:


internal/backend/db/dbgorm/dbgorm.go
```diff
@@ -112,7 +112,7 @@ func (db *gormdb) Connect(ctx context.Context) (err error) {
 	return
 }
 
-func translateError(err error) error {
+func translateError(l *zap.Logger, operation string, err error) error {
```
efiShtain, line 115: I understand this is one place to log it
but I think it is mixing stuff and this function should not log anything

efiShtain (approved):
I don't like this implementation but I did approve it

PR #1495, opened 2025-11-27
feat: add "fix action" to status

itayd:


proto/autokitteh/common/v1/status.proto
itayd, line 11: extends codes

PR #1498, opened 2025-12-01
fix: pass cookie configuration to OAuth redirect cookies

mario99logic:
Configure the OAuth redirect cookie with proper domain, secure, and SameSite settings from the auth login config to ensure cookies survive OAuth flows across different environments.

Without proper SameSite settings, the redirect cookie could be lost during third-party OAuth flows, causing users to be redirected to `/` instead of their intended destination after login.

### DIFF
--- a/internal/backend/auth/authloginhttpsvc/config.go
+++ b/internal/backend/auth/authloginhttpsvc/config.go
@@ -39,9 +39,9 @@ type Config struct {
 	RejectNewUsers bool `koanf:"reject_new_users"`
 
 	// Session cookie settings (should match authsessions config)
-	CookieDomain   string            `koanf:"cookie_domain"`
-	SecureCookie   bool              `koanf:"secure_cookie"`
-	CookieSameSite http.SameSite     `koanf:"cookie_samesite"`
+	CookieDomain   string        `koanf:"ui_domain"`
+	SecureCookie   bool          `koanf:"secure_cookie"`
+	CookieSameSite http.SameSite `koanf:"cookie_samesite"`
 }
 
 var Configs = configset.Set[Config]{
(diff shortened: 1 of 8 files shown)


internal/backend/auth/authloginhttpsvc/svc.go
```diff
@@ -28,18 +28,37 @@ import (
...
-	Sessions authsessions.Store
-	Tokens   authtokens.Tokens
-	Users    sdkservices.Users
+	Muxes      *muxes.Muxes
+	L          *zap.Logger
+	Cfg        *Config
+	Sessions   authsessions.Store
+	Tokens     authtokens.Tokens
+	Users      sdkservices.Users
+	ServiceURL string `optional:"true"`
```
itayd, line 37: why is this optional?

internal/backend/auth/authloginhttpsvc/redir.go
```diff
@@ -5,14 +5,28 @@ import (
 	"net/url"
 )
 
-const redirCookieName = "redir"
+const (
+	redirCookieName = "redir"
+	cookieMaxAge    = 5 * 60 // Short-lived cookie for OAuth flow
+)
+
+type redirectConfig struct {
```
itayd: why not make this a config struct?

internal/backend/auth/authloginhttpsvc/config.go
```diff
@@ -35,10 +37,18 @@ type Config struct {
 	// If set, reject logins from new users, meaning users that are
 	// not already in the database are rejected.
 	RejectNewUsers bool `koanf:"reject_new_users"`
+
+	// Session cookie settings (should match authsessions config)
+	CookieDomain   string        `koanf:"ui_domain"`
```
itayd, line 42:
```
Cookie CookieConfig `koanf:"cookie"`
```

itayd (approved):
see what's up with the service url

internal/backend/auth/authloginhttpsvc/svc.go
mario99logic: in localhost the serviceURL us optional

PR #1500, opened 2025-12-02
feat: report peak rss in python rt

itayd:


runtimes/pythonrt/runner/prof.py
```diff
@@ -0,0 +1,21 @@
...
+_process = psutil.Process(os.getpid())
+_peak_rss = 0
+
+
+def sample() -> psutil.pmem:
+    global _peak_rss
+    curr = _process.memory_info()
+    if curr.rss > _peak_rss:
+        _peak_rss = curr.rss
+    log.debug("memory sample: rss=%d, peak=%d", curr.rss, _peak_rss)
```
efiShtain, line 16: why not reuse the report function you defined below?

efiShtain (approved):
- not sure this is standard to publish the metrics in a production code
- need to check that it publishes correct results when running in containers, on a Mac and on a linux
- probably better to get the metrics from containerd instead of printing stuff internally
- not sure how it behaves on windows 
- I think If you do go with it, print it in a loop once a second once reaching a threshold or some thresholds
- also need to probably globally catch an unknown exception before the process dies and print it, not sure if possible, you can try, maybe better the the container wrapper instead of directly in python

efiShtain:
what about this one?
can we close?

PR #1502, opened 2025-12-05
feat: github slash commands recognition

itayd:
this will add to all slash command capable events `slash_commands` key at the top level. this will allow for filtering for them.

also send slash_command events per command.

### DIFF
--- a/integrations/github/webhooks/slash_commands.go
+++ b/integrations/github/webhooks/slash_commands.go
@@ -2,9 +2,8 @@ package webhooks
 
 import (
 	"strings"
+	"unicode"
 
-	"github.com/gomarkdown/markdown/ast"
-	"github.com/gomarkdown/markdown/parser"
 	"github.com/google/go-github/v52/github"
 )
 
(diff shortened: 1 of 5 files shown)


integrations/github/webhooks/slash_commands.go
```diff
@@ -0,0 +1,80 @@
...
+			inCodeBlock = !inCodeBlock
+			continue
+		}
+
+		// Skip lines inside code blocks
+		if inCodeBlock {
+			continue
+		}
+
+		if !strings.HasPrefix(line, "/") {
```
efiShtain, line 62: You assume slash command is on its  own line always?

integrations/github/webhooks/slash_commands.go
```diff
@@ -0,0 +1,80 @@
...
+		return nil
+	}
+
+	return extractSlashCommandsFromMD(md)
+}
+
+func extractSlashCommandsFromMD(md string) (commands []slashCommand) {
+	inCodeBlock := false
+
+	for line := range strings.SplitSeq(md, "\n") {
```
efiShtain, line 49: Do we need to care about really long texts here?
Can it cause some kind of ddos on the server?

efiShtain:
The idea seems fine
But do we want to limit the number of commands in one message?
Or in an event?
What about retries if some command is failing?
What about order?
What about edit of the text? Will it re fire the command?

integrations/github/webhooks/slash_commands.go
itayd: yes, it doesn't really make sense otherwise

integrations/github/webhooks/slash_commands.go
itayd: it's coming verified from github, so i think it's fine. will add a limit anyway though

itayd:
> The idea seems fine But do we want to limit the number of commands in one message? 

makes sense, will do.

> Or in an event?

same thing?

> What about retries if some command is failing? 

not sure i follow you here - retries on what?

> What about order? 

should not matter. need to document.

> What about edit of the text? 
> Will it re fire the command?

user should check the underlying message to see the action.

---

when this is in, i will add examples in kittehub and docs.

efiShtain (approved):

PR #1503, opened 2025-12-07
fix: migration auth_type primary key conflict

mario99logic:
The original migration renamed authType → auth_type using UPDATE, which caused issues:
  - Some var_ids had BOTH authType AND auth_type (from partial migration runs or code changes during migration)
  - Migration failed with duplicate key errors due to composite primary key (var_id, name)
  - Deleting duplicates risked data loss and broke backward compatibility

Changed migration strategy from UPDATE to INSERT:
  - Inserts auth_type copy for var_ids that only have authType
  - Keeps both authType and auth_type for backward compatibility
  - Skips var_ids that already have auth_type


itayd:
@efiShtain i'll need you to approve it

itayd (approved):

PR #1504, opened 2025-12-08
feat: add polling go routine for discord WebSocket connections 

mario99logic:
- Removed WebSocket opening from the save.go
- Added background polling mechanism configurable via env var with default fallback value of 5 minutes 
- Polling queries active connections and ensures WebSockets are open for all bot tokens


integrations/discord/server.go
```diff
@@ -48,4 +56,73 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.Disp
...
+		// OpenWebSocketConnection is idempotent - it checks the discordSessions
+		// map and won't create a duplicate if one already exists for this bot token.
+		wsh.OpenWebSocketConnection(botToken)
+	}
+}
+
+func getPollingInterval(l *zap.Logger) time.Duration {
+	interval := pollInterval
+
+	if envInterval := os.Getenv("DISCORD_POLL_INTERVAL_MINUTES"); envInterval != "" {
```
efiShtain: This is not how we handle configuration, nor does it confirm to our variable structure

integrations/discord/server.go
```diff
@@ -48,4 +56,73 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.Disp
...
+		if minutes, err := strconv.Atoi(envInterval); err == nil && minutes > 0 {
+			interval = time.Duration(minutes) * time.Minute
+			l.Info("Using custom Discord WebSocket poll interval", zap.Duration("interval", interval))
+		} else {
+			l.Warn("Invalid DISCORD_POLL_INTERVAL_MINUTES, using default", zap.String("value", envInterval))
```
efiShtain: if someone provides an invalid value, it should fail
you'll get it automatically if you will use the configration library we use

integrations/discord/server.go
```diff
@@ -48,4 +56,73 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.Disp
 
 		wsh.OpenWebSocketConnection(data.GetValue(vars.BotToken))
 	}
+
+	// Start background polling to ensure WebSocket connections exist.
+	go pollAndSyncWebSockets(context.Background(), l, v, wsh)
+}
```
efiShtain: Move this to a separate function that can be started from out side

it should be behind a flag as we discussed - enable / disable the web sockets handling

you can invoke it probably in ```internal/backend/integrations/registry.go```

note the difference between starting and invoking, it should be invoked once the rest of the server is up and running (you can see other components started and invoked using fx)

integrations/discord/server.go
```diff
@@ -48,4 +56,73 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.Disp
...
+		case <-ticker.C:
+			syncWebSocketConnections(ctx, l, v, wsh)
+		}
+	}
+}
+
+// syncWebSocketConnections queries the database for all active connections
+// and ensures WebSocket connections are open for them.
+func syncWebSocketConnections(ctx context.Context, l *zap.Logger, v sdkservices.Vars, wsh handler) {
+	cids, err := v.FindActiveConnectionIDs(ctx, integrationID, vars.BotToken, "")
```
efiShtain: technically, this won't do anything smart 99.9999% percent of the time

assuming we have X connections, it will fetch them for nothing

can you find a way to get only updated/new connections?

also, how do you handle deleted connections ? 
need to know to close the websocket

integrations/discord/server.go
efiShtain: it can also be its own struct with configuration or what ever you need

integrations/discord/server.go
```diff
@@ -17,6 +21,10 @@ const (
 	// savePath is the URL path for our handler to save new
 	// connections, after users submit them via a web form.
 	savePath = "/discord/save"
+
+	// pollInterval is the interval at which we poll the database
+	// for active connections and ensure WebSockets are open.
+	pollInterval = 5 * time.Minute
```
efiShtain: that's way too long assuming someone create a connection now and he won't be able to use it for 5 minutes

i think you should do something better
maybe poll faster
or subscribe to changes somehow

integrations/discord/server.go
mario99logic: whats the difference from these examples?

https://github.com/autokitteh/autokitteh/blob/86438b964db81fc8720dae05e9c6399f79779bc9/integrations/atlassian/jira/oauth.go#L126

https://github.com/autokitteh/autokitteh/blob/980e98620f9f98869194c0c603a6b43b5aaf9811/integrations/github/oauth.go#L103

https://github.com/autokitteh/autokitteh/blob/f0b13ecbc3f6da37b9fc830a55bf085d605078c9/integrations/google/calendar/client.go#L101

efiShtain: no difference
problematic handling in the other places as well

I think you should try and not add more wrong to the current case

integrations/discord/server.go
itayd: this needs to be from config, order of 5s

integrations/discord/server.go
```diff
@@ -48,4 +56,73 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.Disp
...
 		wsh.OpenWebSocketConnection(data.GetValue(vars.BotToken))
 	}
+
+	// Start background polling to ensure WebSocket connections exist.
+	go pollAndSyncWebSockets(context.Background(), l, v, wsh)
+}
+
+// pollAndSyncWebSockets runs a periodic check to ensure all active connections
+// have WebSocket connections open.
+func pollAndSyncWebSockets(ctx context.Context, l *zap.Logger, v sdkservices.Vars, wsh handler) {
```
itayd: remove context

integrations/discord/save.go
```diff
@@ -42,12 +42,13 @@ func (h handler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
 
-	h.OpenWebSocketConnection(token)
-
 	c.Finalize(sdktypes.NewVars().
 		Set(vars.BotID, bot.ID, false).
 		Set(vars.BotToken, token, true).
 		Set(vars.AuthType, integrations.Init, false))
+
+	// The background polling mechanism (in server.go) will detect this new connection
+	// and open the WebSocket within the next poll interval.
```
itayd: explain why we do this. we need to document the fact that the server must run as a singleton.

integrations/discord/server.go
itayd: kill this function

integrations/discord/server.go
```diff
@@ -19,9 +21,19 @@ const (
...
+	PollInterval     int  `koanf:"poll_interval"` // in seconds
+}
+
+// StartWithConfig initializes all the HTTP handlers of the Discord integration.
 // This includes connection UIs, initialization webhooks, and event webhooks.
-func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.DispatchFunc) {
+func StartWithConfig(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.DispatchFunc, cfg *Config) {
+	// Extract config values from integrations config
+	enableWebsockets := cfg.EnableWebsockets
+	pollInterval := time.Duration(cfg.PollInterval) * time.Second
```
efiShtain: please move this below where it's needed
no real reason to use it here

integrations/discord/server.go
```diff
@@ -32,7 +44,12 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.Disp
 
 	m.Auth.Handle("POST "+savePath, wsh)
 
-	// Initialize WebSocket pool.
+	// Initialize WebSocket pool if enabled.
+	if !enableWebsockets {
+		l.Info("Discord WebSockets are disabled")
```
efiShtain: add a log indicating when it is enabled
I don't really mind when it's disabled
since we assume the default is disabled

integrations/discord/server.go
```diff
@@ -48,4 +65,52 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.Disp
 
 		wsh.OpenWebSocketConnection(data.GetValue(vars.BotToken))
 	}
```
efiShtain: this is exactly what the pollAndSyncWebsockets function is doing
why is it here as well?

integrations/discord/server.go
```diff
@@ -48,4 +65,52 @@ func Start(l *zap.Logger, m *muxes.Muxes, v sdkservices.Vars, d sdkservices.Disp
...
+// have WebSocket connections open.
+func pollAndSyncWebSockets(l *zap.Logger, v sdkservices.Vars, wsh handler, interval time.Duration) {
+	l.Info("Starting Discord WebSocket polling", zap.Duration("interval", interval))
+
+	ticker := time.NewTicker(interval)
+	defer ticker.Stop()
+
+	for {
+		<-ticker.C
+		syncWebSocketConnections(l, v, wsh)
```
efiShtain: can you make sure that if syncWebSocketConnections takes a long time, then it won't fire many times in a row next time?
meaning you don't have a backlog of ticks in this channel 

(I'm not sure what's the behavior, this might be fine, just need to verify

mario99logic: I will use timer instead this way there is no buffering

efiShtain (approved):

PR #1512, opened 2025-12-14
refactor: events, next_event and callbacks

itayd:
- formalize event type passed to sessions, next_event, etc.
- take session_id out of event.
- allow to get all event's details from next_event when `full=True` is specified.


runtimes/pythonrt/runner/syscalls.py
```diff
@@ -119,11 +119,24 @@ def ak_next_event(self, subscription_id, *, timeout=None):
...
             raise AutoKittehError(f"next_event: invalid event: {err}")
 
-        return AttrDict(data) if isinstance(data, dict) else data
+        if not event:
+            return None
+
+        data = event.get("data")
+        only_data = AttrDict(data) if isinstance(data, dict) else data
+
+        if not full:
```
efiShtain, line 132: why do we need this full option?

runtimes/pythonrt/runner/main.py
```diff
@@ -654,11 +662,11 @@ def _call(self, fn, args, kw):
 
         return Result(value, error, stack)
 
-    def on_event(self, fn, event):
+    def on_event(self, fn, event, session_id):
         func_name = full_func_name(fn)
         log.info("start event: %s", func_name)
 
-        result = self._call(fn, [event], {})
+        result = self._call(fn, [event], {}, opts={"session_id": session_id})
```
efiShtain, line 669: why?

efiShtain (approved):

runtimes/pythonrt/runner/syscalls.py
itayd: backward compatability. otherwise it would have returned full event and current scripts would fail.

runtimes/pythonrt/runner/main.py
itayd: session_id is no longer part of event. i've checked and nobody is using it. now if the entrypoint function specifies "session_id" arg, it would be supplied.

PR #1516, opened 2025-12-17
feat: allow to write outcomes for events coming from next_event

itayd:


internal/backend/db/dbgorm/scheme/records.go
```diff
@@ -318,6 +317,9 @@ type SessionLogRecord struct {
 	Data      datatypes.JSON
 	Type      string `gorm:"index"`
 
+	// For outcomes, what event id they refer to.
+	OutcomeEventID *uuid.UUID `gorm:"index;type:uuid"`
```
efiShtain: I assume 99.9 percent of this will be null
so update the index please
```
 WHERE outcome_event_id  IS NOT NULL;
```

internal/backend/dispatcher/workflow.go
```diff
@@ -198,8 +198,11 @@ func (d *Dispatcher) signalWorkflow(wctx workflow.Context, wid string, sigid uui
 		return
 	}
 
-	var nferr *serviceerror.NotFound
-	if errors.As(err, &nferr) {
+	var (
+		nferr      *serviceerror.NotFound
+		unknownErr *temporal.UnknownExternalWorkflowExecutionError
+	)
+	if errors.As(err, &nferr) || errors.As(err, &unknownErr) {
```
efiShtain, line 205: didn't you change this in another PR as well?

efiShtain (approved):
fix the index please before merging

internal/backend/dispatcher/workflow.go
itayd: yeah this should have been based on it.

internal/backend/db/dbgorm/scheme/records.go
itayd: done

PR #1526, opened 2025-12-27
fix cli dispatch using incorrect connection name

efiShtain:


cmd/ak/cmd/events/dispatch.go
itayd: s/args/_/

itayd (approved):

cmd/ak/cmd/events/dispatch.go
itayd:
```suggestion
	RunE: func(cmd *cobra.Command, _ []string) error {
```

